PR #16, opened 2021-08-18
feat: add ability to configure SDK

levibostian:
Changes: 

1. We were using "config" previously to set the siteId and apiKey on the SDK but this has been renamed to "credentials" to separate these from other configuration options of the SDK. 
2. Create a DSL on `CustomerIO` class to configure the SDK in a Swifty way. You can now change the URL used for the track API.


Sources/SDK/CustomerIO.swift
```diff
@@ -50,26 +62,83 @@ public class CustomerIO {
...
+        credentialsStore.save(siteId: siteId, credentials: credentials)
+
+        // Some default values of the SDK configuration may depend on credentials. Reset default values.
+        sdkConfig = setDefaultValuesSdkConfig(config: sdkConfig)
+    }
+
+    /**
+     Configure the Customer.io SDK.
+
+     This will configure the singleton shared instance of the CustomerIO class. It will also be the defafult
```
hownowstephen, line 99: typo defafult

hownowstephen (approved):

PR #21, opened 2021-08-24
feat: add feature to identify a customer

levibostian:


Sources/Common/Util/KeyValueStorageKey.swift
```diff
@@ -3,8 +3,10 @@ import Foundation
 /**
  The keys for `KeyValueStorage`. Less error-prone then hard-coded strings.
  */
-internal enum KeyValueStorageKey: String {
+public enum KeyValueStorageKey: String {
     case sharedInstanceSiteId
     case apiKey
     case regionCode
+    case identifiedProfileId
```
hownowstephen, line 10: list of identifiers

Sources/Tracking/Repository/IdentifyRepository.swift
```diff
@@ -0,0 +1,61 @@
...
+        let httpRequestParameters = HttpRequestParams(endpoint: .identifyCustomer(identifier: identifier), headers: nil,
+                                                      body: bodyData)
+
+        httpClient
+            .request(httpRequestParameters) { [weak self] result in
+                guard let self = self else { return }
+
+                switch result {
+                case .success:
+                    self.keyValueStorage.setString(siteId: self.siteId, value: identifier, forKey: .identifiedProfileId)
```
hownowstephen, line 47: this then obviously has ripple effects down to how we store them, whether or not we go the struct-identity route, IMO we should be storing any supplied identities as a list, and then we can just take the first of them when performing attribute update calls

Sources/Tracking/Tracking.swift
```diff
@@ -0,0 +1,146 @@
...
+}
+
+/**
+ The automatically generated dependency injection graph project
+ complains when there isn't at least 1 `case` in the enum.
+
+ This solves that problem by placing a placeholder into the graph until we have a class that makes sense.
+ */
+// sourcery: InjectRegister = "Placeholder"
+internal class Placeholder {}
```
hownowstephen, line 146: garbage now?

Sources/Tracking/Tracking.swift
```diff
@@ -0,0 +1,146 @@
...
+    /**
+     Stop identifying the currently persisted customer. All future calls to the SDK will no longer
+     be associated with the previously identified customer.
+
+     Note: If you simply want to identify a *new* customer, this function call is optional. Simply
+     call `identify()` again to identify the new customer profile over the existing.
+
+     If no profile has been identified yet, this function will ignore your request.
+     */
+    public func identifyStop() {
```
hownowstephen, line 130: wonder if we could come up with a better func name here, `identifyStop` feels not super intuitive

Tests/Tracking/Repository/IdentifyRepositoryTest.swift
```diff
@@ -0,0 +1,128 @@
...
+    override func setUp() {
+        super.setUp()
+
+        httpClientMock = HttpClientMock()
+        keyValueStorageMock = KeyValueStorageMock()
+        siteId = String.random
+
+        repository = CIOIdentifyRepository(httpClient: httpClientMock, keyValueStorage: keyValueStorageMock,
+                                           siteId: siteId)
+        integrtionRepository = CIOIdentifyRepository(httpClient: httpClientMock, keyValueStorage: keyValueStorage,
```
hownowstephen, line 24: integrtion

Tests/Tracking/Repository/IdentifyRepositoryTest.swift
```diff
@@ -0,0 +1,128 @@
...
+    func test_addOrUpdateCustomer_givenHttpFailure_expectDoNotSaveData_expectGetError() {
+        httpClientMock.requestClosure = { params, onComplete in
+            onComplete(Result.failure(HttpRequestError.unsuccessfulStatusCode(500, message: "")))
+        }
+
+        let expect = expectation(description: "Expect to complete")
+        repository.addOrUpdateCustomer(identifier: String.random, email: nil) { result in
+            guard case .failure(let actualError) = result else { return XCTFail() }
+            guard case .httpError(let httpError) = actualError else { return XCTFail() }
+            guard case .unsuccessfulStatusCode = httpError else { return XCTFail() }
```
hownowstephen, line 61: should we check that it's actually the 500 status code?

Sources/Tracking/Service/Request/AddUpdateCustomerRequestBody.swift
```diff
@@ -0,0 +1,7 @@
+import Foundation
```
hownowstephen, line 1: maybe I missed it, but are we not actually using this?

also this isn't correct, the body is just a json object but there are specific well-known fields:
- email
- _update
- anonymous_id
- created_at

but they will get serialized into the main object body
```json
{
  "email": "abc@example.com",
  "_update": true,
  "anonymous_id": "RIOEJOIMLW",
  "created_at": 1629854504,
  "t_shirt_size": "M",
  "premium": true,
  "ltv": 123400
}
```
It wouldn't be so bad to work with a struct and then do some custom serialization on it to make this happen, makes the fields more explicit. But we should be careful here to match the semantics of the API

Sources/Tracking/Repository/IdentifyRepository.swift
```diff
@@ -0,0 +1,67 @@
...
+
+                    onComplete(Result.success(()))
+                case .failure(let error):
+                    onComplete(Result.failure(.http(error)))
+                }
+            }
+    }
+
+    func removeCustomer() {
+        keyValueStorage.setString(siteId: siteId, value: nil, forKey: .identifiedProfileId)
```
hownowstephen, line 65: one upshot of storing relative to the site ID is that if we don't clear it out, then there could be old data if someone moves between two site IDs.

Not something to worry about at this stage, 99% of integrations will set & forget a single site ID, but we should have a quick discussion with the squad about what we expect a customer would expect in a scenario where the site ID changes - both in terms of the new state on the device and the existing state

hownowstephen (approved):
LGT

Sources/Tracking/Repository/IdentifyRepository.swift
levibostian: Good idea. 

How I handle that in apps that I build is set an expiration on data. If there is data that is considered "too old" in my standards I delete it. When the app starts up I run a quick background task that checks for old data and deletes it. 

Without much thinking about it I would lean towards that. We get to decide what "too old" is. 

The nice thing is that key/value storage is small in size. If it does stick around it will not hurt the device storage much. It's not like we are caching MB of data in a database on the device.

Issue #19, opened 2021-08-18
Register APN device token for push notifications


levibostian:
In order to send a push notification to a mobile device from CIO, you need to: 
1. Setup the push notification backend service (APN/FCM) with your mobile app. FCM and APN has steps that they require to setup your app. 
2. FCM and APN then give the mobile app a device token. CIO needs this device token. A customer can use Segment or our API.

This task here has to do with item 2 above and customers who need to use our API. 

# Why are we doing this for our customers?

Today, our customers send push notification device tokens to CIO via (1) 3rd party services like Segment or (2) they write custom code in their apps to send the token to CIO via our API. This task is to do item (2) for our customers. 

Sending a push notification device token to CIO will now just be 1 line of code:

```swift
Customers.deviceToken = "XXXXXXXXXXXXX"
```

...and that's it! 

Along with sending a device token, this task will also set the `platform` and `last_used` properties automatically. 

# How do we test that this change is successful? 

This feature will be tested with automated tests + manual testing with a sample app. 

# How are we building this? 

- [x] Create a separate swift package just for APN push messaging. 

- [ ] Create a class `MessagingPush`

```swift
public class MessagingPush {
  // convenient shared singleton instance 
  public static private(set) var instance: MessagingPushAPN = MessagingPushAPN(CustomerIO.instance)

  // needs customerIO so MessagingPushAPN SDK features can get credentials and config
  init(customerIO: CustomerIO) {}
}
````

The idea is that this 1 class is where all the functionality of Push messaging exists. When you add a new SDK, it will extend `MessagingPush`. 

- [ ] Create extensions to `MessagingPush` to add APN specific functionality:

```swift
public extension MessagingPush {
  public func application(_ application: UIApplication, didRegisterForRemoteNotificationsWithDeviceToken deviceToken: Data, onComplete: @escaping (Result<String, Error>) -> Void) {
    // perform API call to send device token up to API. 
    // if successful, save device token in key/value storage 
    // if successful or error, send result to onComplete()
  }

  func application(_ application: UIApplication, didFailToRegisterForRemoteNotificationsWithError error: Error) {
    // Print to XCode logs the error 
  }
}
```

- [ ] The next features will be added once [identifying a customer is complete](https://github.com/customerio/customerio-ios/issues/18). 

* When a customer is identified, after performing a HTTP request to add/update a customer, check if a device token exists in key/value storage. If one does, perform a HTTP request to register the device token with the customer's profile. 

[...]

PR #26, opened 2021-08-31
feat: register/delete device tokens

hownowstephen:
https://github.com/customerio/customerio-ios/issues/19

Adds support for registering device tokens with the `MessagingPush` base class, adds the `MessagingPushAPNS` class as a thin wrapper, for usage like:
```swift
func application(_ application: UIApplication, didRegisterForRemoteNotificationsWithDeviceToken deviceToken: Data) {
   self.messagingPush.registerDeviceToken(deviceToken: deviceToken)
}
```

we will only keep the device token in-memory, [per the apple docs](https://developer.apple.com/documentation/usernotifications/registering_your_app_with_apns) to avoid weird issues with backups etc
> Important
>
> Never cache device tokens in local storage. APNs issues a new token when the user restores a device from a backup, when the user installs your app on a new device, and when the user reinstalls the operating system. If you ask the system to provide the token each time, you’re guaranteed to get an up-to-date token.

multiple calls to `registerDeviceToken` currently only overwrite the active token in memory, I need to research further if there's a reasonable way of determining when a device token is revoked


Sources/Tracking/CustomerIO.swift
```diff
@@ -230,6 +234,17 @@ public class CustomerIO {
             }
         }
     }
+    
+    /**
+     Set the active identifier without calling the customer.io API
+    **/
```
levibostian, line 240: Why?

Sources/Tracking/CustomerIO.swift
```diff
@@ -230,6 +234,17 @@ public class CustomerIO {
...
+    public func setIdentifier(identifier: String,  onComplete: @escaping (Result<Void, CustomerIOError>) -> Void){
```
levibostian, line 241: If this function is a synchronous function we don't need to add `onComplete()`

Sources/Tracking/Util/KeyValueStorageKey.swift
```diff
@@ -9,4 +9,5 @@ public enum KeyValueStorageKey: String {
     case regionCode
     case identifiedProfileId
     case identifiedProfileEmail
+    case deviceToken
```
levibostian, line 12: Can we delete this since we aren't saving it?

Tests/MessagingPush/MessagingPushTest.swift
```diff
@@ -0,0 +1,156 @@
...
+        mockCustomerIO.setIdentifier(identifier: String.random, onComplete: { result in
+            guard case .success = result else { return XCTFail() }
+            XCTAssertNotNil(self.mockCustomerIO.identifier)
+        })
+        
+        httpClientMock.requestClosure = { params, onComplete in
+            onComplete(Result.success(Data()))
+        }
+        
+        let actualToken = Data(String.random.utf8)
```
levibostian, line 58:
```suggestion
        let actualToken = String.random.data!
```

some nice extension sugar.

Sources/MessagingPush/MessagingPush.swift
```diff
@@ -0,0 +1,122 @@
...
+
+    private var credentials: SdkCredentials? {
+        customerIO.credentials
+    }
+
+    private var sdkConfig: SdkConfig {
+        customerIO.sdkConfig
+    }
+    
+    public var deviceToken: Data?
```
levibostian, line 23:
```suggestion
    @Atomic public var deviceToken: Data?
```

make it thread safe since we may be accessing it in many places.

Sources/MessagingPush/MessagingPush.swift
```diff
@@ -0,0 +1,122 @@
+import CioTracking
+import Foundation
+
+/**
+ Swift code goes into this module that are common to *all* of the Messaging Push modules (APN, FCM, etc).
+ So, performing an HTTP request to the API with a device token goes here.
+  */
+open class MessagingPush {
```
levibostian, line 8: I know that you linked to apple docs stating they recommend you do not persist the device token. 

However, since we are storing it in memory, do we know for sure that APN will call `didRegisterForRemoteNotificationsWithDeviceToken` for us the next time that the app is restarted? 

I am nervous that there will be a scenario where a device token is valid for a device but `deviceToken` is still nil in this class.

Tests/MessagingPush/MessagingPushTest.swift
```diff
@@ -0,0 +1,156 @@
...
+
+        mockCustomerIO = CustomerIO(credentialsStore: nil, sdkConfig: SdkConfig(), identifyRepository: identifyRepositoryMock,
+                                keyValueStorage: nil)
+        
+        mockCustomerIO.setCredentials(siteId: String.random, apiKey: String.random, region: Region.EU)
+        
+        messagingPush = MessagingPush(customerIO: mockCustomerIO, httpClient: httpClientMock, keyValueStorage: DITracking.shared.keyValueStorage)
+    }
+
+    // MARK: registerDeviceToken
```
levibostian, line 29: I would write a test for the scenario of: http error happens - assert device token not saved by the SDK.

Tests/MessagingPush/MessagingPushTest.swift
```diff
@@ -0,0 +1,156 @@
...
+            expect.fulfill()
+        }
+        
+        waitForExpectations()
+    }
+    
+    func test_registerDeviceToken_givenHttpSuccess_expectSaveExpectedData() {
+        
+        identifyRepositoryMock.setIdentifierClosure = { identifier in
+            self.identifyRepositoryMock.identifier = identifier
```
levibostian, line 46: I would avoid setting properties on mocks since each time that this is done, values like `mock.identifierCalled: Bool` is changed. it may give us false positive/negatives with our tests.

Tests/MessagingPush/MessagingPushTest.swift
```diff
@@ -0,0 +1,156 @@
...
+        }
+        
+        let expect = expectation(description: "Expect to fail to delete device token")
+        self.messagingPush.deleteDeviceToken() { result in
+            guard case .failure(let error) = result else { return XCTFail() }
+            guard case .deviceNotRegistered = error else { return XCTFail() }
+            expect.fulfill()
+        }
+
+        waitForExpectations()
```
levibostian, line 116: Good to also make sure the http mock is never called. Something like: `AssertFalse(httpclientmock.requestcalled)` to make sure our SDK when it errors does not try to perform the action of the function.

Tests/MessagingPush/MessagingPushTest.swift
```diff
@@ -0,0 +1,156 @@
...
+        guard let storedToken = self.messagingPush.deviceToken else {
+            return XCTFail()
+        }
+        
+        XCTAssertEqual(storedToken, actualToken)
+
+        waitForExpectations()
+    }
+    
+    // MARK: deleteDeviceToken
```
levibostian, line 75: I would add a test case for when http fail the SDK does not try to delete the device token persisted.

levibostian (changes requested):

Sources/MessagingPushAPN/MessagingPushAPN.swift
```diff
@@ -0,0 +1,21 @@
...
+    
+    func application(_ application: Any, didRegisterForRemoteNotificationsWithDeviceToken deviceToken: Data, onComplete: @escaping (Result<Void, CustomerIOError>) -> Void) {
+        self.registerDeviceToken(deviceToken: deviceToken, onComplete:onComplete)
+      }
+
+    func application(_ application: Any, didFailToRegisterForRemoteNotificationsWithError error: Error, onComplete: @escaping (Result<Void, CustomerIOError>) -> Void) {
+        // Print to XCode logs the error
+        if self.deviceToken != nil {
+            self.deleteDeviceToken(onComplete: onComplete)
+        }
```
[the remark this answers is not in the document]
hownowstephen: Good point!

Sources/MessagingPush/MessagingPush.swift
```diff
@@ -0,0 +1,122 @@
...
+     Delete the currently registered device token
+     */
+    public func deleteDeviceToken(onComplete: @escaping (Result<Void, CustomerIOError>) -> Void) {
+        guard let bodyData = JsonAdapter.toJson(DeleteDeviceRequest()) else {
+            return onComplete(Result.failure(.httpError(.noResponse)))
+        }
+        
+        guard let identifier = self.customerIO.identifier else {
+            return onComplete(Result.failure(.noCustomerIdentified))
+        }
```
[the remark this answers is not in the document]
hownowstephen: yeah, I'll leave it off for now. on some level what makes the most sense to do here is potentially to hold the token until the identify happens and then push it through then. But for now just not failing seems like enough

Sources/MessagingPush/MessagingPush.swift
hownowstephen: from my reading, as long as the app calls `registerForRemoteNotificationTypes`, the `didRegisterForRemoteNotificationsWithDeviceToken` delegate will get called every time it is opened, also gets called on settings change as well. I vote we stick with the Apple best practice for the time being and then do a few tests. Actually persisting the device token is also pretty trivial, so if we have any concern after testing that the token doesn't show up or something then we can raise it at that point.

https://stackoverflow.com/questions/10983536/when-is-didregisterforremotenotificationswithdevicetoken-called

Sources/MessagingPush/Service/Request/Device.swift
```diff
@@ -0,0 +1,14 @@
+import Foundation
+
+// Standard device structure
+internal struct Device: Codable {
+    let token: Data
```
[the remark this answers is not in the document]
hownowstephen: ah yeah, Data does the byte-array-to-base64 thing. going to just make it a String

Tests/MessagingPush/MessagingPushTest.swift
hownowstephen: changed up the way the test works to avoid it. also removed `setIdentifier`

Sources/Tracking/CustomerIO.swift
hownowstephen: was to get around some test stuff I was trying to work with. found an alt approach & removed it

Tests/MessagingPush/MessagingPushTest.swift
hownowstephen: test_registerDeviceToken_givenHttpFailure_expectNilDeviceToken

Tests/MessagingPush/MessagingPushTest.swift
hownowstephen: test_deleteDeviceToken_givenHttpFailure_expectTokenNotCleared

Sources/Tracking/CustomerIO.swift
hownowstephen: did one better. it is a 🔪 function now

Tests/MessagingPush/MessagingPushTest.swift
```diff
@@ -18,12 +18,36 @@ class MessagingPushTest: UnitTest {
...
+        
+        let identifier: String? = String.random
+
+    
+        cio.identify(identifier: identifier!) { result in
+            guard case .success = result else {  return XCTFail() }
+            XCTAssertEqual(cio.identifier, identifier)
+        }
+        
+        return MessagingPush(customerIO: cio, httpClient: httpClientMock, jsonAdapter: jsonAdapter)
```
levibostian, line 50: This will be improved when I perform the refactors that I have mentioned a couple of times. Dependency injection is not not really setup the best at the moment which makes creating mocks and instances harder. This is a good workaround for now.

levibostian, line 50: This will be improved when I perform the refactors that I have mentioned a couple of times. Dependency injection is not not really setup the best at the moment which makes creating mocks and instances harder. This is a good workaround for now.

Sources/MessagingPush/MessagingPush.swift
levibostian: I agree to follow what they mention. I have never tested this myself before (I have always used a background queue in an app that persisted this data for me) so I am not sure on the behavior. 

QA will help assert this is true. 

I am working with QA right now to help them test the behavior of the SDK better. I am thinking we create log statements throughout the SDK (this is a feature of the SDK for customers, not just QA so helping everyone!) and so QA can check the logs of the RH app and see "device token registered" when app starts up. Then we can confirm that this fact is true.

levibostian (approved):

PR #37, opened 2021-09-13
chore: create eventbus for events between modules

levibostian:
Functions of the SDK need to react when certain events happen. Example: When an APN device token is registered with the device and a new profile is identified. The existing device token needs to be registered to the device. 

This is the building blocks of that work. However, because we use async handlers we cannot do this work at this time. ﻿


Sources/MessagingPush/MessagingPush.swift
```diff
@@ -31,10 +35,22 @@ open class MessagingPush {
         self.customerIO = customerIO
         self.httpClient = CIOHttpClient(credentials: customerIO.credentials!, config: customerIO.sdkConfig)
         self.jsonAdapter = DITracking.shared.jsonAdapter
+        self.eventBus = DITracking.shared.eventBus
+
+        self.identifyCustomerEventBusCallback = eventBus.register(event: .identifiedCustomer) { event in
```
hownowstephen, line 40: why would we need to `switch` on the event below if it's always of type `.identifiedCustomer`?

hownowstephen (approved):

Issue #36, opened 2021-09-13
Create background queue 


levibostian:

levibostian:
Merged into `alpha`.

PR #42, opened 2021-09-15
feat: event tracking

hownowstephen:
Adds basic event tracking to the SDK with data payload and support for setting `timestamp` in the `identifyRepository`


Sources/Tracking/CustomerIO.swift
```diff
@@ -10,6 +10,17 @@ public protocol CustomerIOInstance: AutoMockable {
         onComplete: @escaping (Result<Void, CustomerIOError>) -> Void,
         jsonEncoder: JSONEncoder?
     )
+
+    // sourcery:Name=track
+    func track<RequestBody: Encodable>(
```
levibostian, line 15: Should we move `track()` into the `Tracking` class? We talked about `identify` being in `CustomerIO` since all features depend on it but this function shouldn't have that same problem, yeah?

Sources/Tracking/Service/Request/TrackRequestBody.swift
```diff
@@ -0,0 +1,25 @@
...
+internal struct TrackRequestBody<T: Encodable>: Encodable {
+    let name: String
+    let data: T?
+    let timestamp: Date?
+
+    enum CodingKeys: String, CodingKey {
+        case name
+        case data
+        case timestamp
+    }
```
levibostian, line 13: I believe you can remove this. The compiler should generate it for you.

Tests/Tracking/CustomerIOTest.swift
```diff
@@ -262,4 +262,89 @@ class CustomerIOTest: UnitTest {
...
+
+            onComplete(Result.success(()))
+        }
+
+        let expect = expectation(description: "Expect to complete track")
+        customerIO.track(name: givenEventName, data: givenEventData) { result in
+            expect.fulfill()
+        }
+
+        waitForExpectations()
```
levibostian, line 301:
```suggestion
        waitForExpectations()
        
        XCTAssertEqual(identifyRepositoryMock.trackEventCalledCount, 1)
```

Since as of now, the test function doesn't test that the repository did in fact get called.

Tests/Tracking/Repository/IdentifyRepositoryTest.swift
```diff
@@ -147,4 +147,176 @@ class IdentifyRepositoryTest: UnitTest {
...
+        }
+
+        integrationRepository.trackEvent(name: givenEventName, data: EmptyRequestBody(), timestamp: nil,
+                                         jsonEncoder: nil) { _ in
+            expect.fulfill()
+        }
+
+        waitForExpectations()
+
+        integrationRepository.removeCustomer()
```
levibostian, line 187: Is this for test cleanup? 

If so, you can put it in the test class function `tearDown()`.

Tests/Tracking/Repository/IdentifyRepositoryTest.swift
```diff
@@ -147,4 +147,176 @@ class IdentifyRepositoryTest: UnitTest {
...
+
+            onComplete(Result.success(params.body!))
+        }
+
+        integrationRepository.trackEvent(name: givenEventName, data: EmptyRequestBody(), timestamp: nil,
+                                         jsonEncoder: nil) { _ in
+            expect.fulfill()
+        }
+
+        waitForExpectations()
```
levibostian, line 185: The test function currently tests that `expect` is fulfilled which tests that `repo.trackEvent()`'s `onComplete` gets called however it doesn't check that the http client does get called. Adding assertions against `httpClientMock.requestCalledCount` and other properties would take care of that. 

Same comment for a couple of other test functions below added to this class.

levibostian (approved):

Sources/Tracking/CustomerIO.swift
```diff
@@ -321,4 +340,42 @@ public class CustomerIO: CustomerIOInstance {
...
+        timestamp: Date? = nil,
+        jsonEncoder: JSONEncoder? = nil,
+        onComplete: @escaping (Result<Void, CustomerIOError>) -> Void
+    ) {
+        guard let identifyRepository = self.identifyRepository else {
+            return onComplete(Result.failure(.notInitialized))
+        }
+
+        identifyRepository
+            .trackEvent(name: name, data: data, timestamp: timestamp, jsonEncoder: jsonEncoder) { [weak self] result in
```
levibostian, line 368: What is the main use case for `timestamp` for a customer? Is it simply tracking the time that an event happened? 

If so, I propose that the default timestamp instead of `nil` (so the API records a time when the API request succeeded) is `Date()` to record when the event happened in the app.

levibostian (approved):

Sources/Tracking/CustomerIO.swift
```diff
@@ -64,6 +75,14 @@ public extension CustomerIOInstance {
     ) {
         identify(identifier: identifier, body: body, onComplete: onComplete, jsonEncoder: jsonEncoder)
     }
+
+    func track(
+        name: String,
+        onComplete: @escaping (Result<Void, CustomerIOError>) -> Void,
+        jsonEncoder: JSONEncoder? = nil
+    ) {
```
levibostian, line 83: When using the SDK in Remote Habits, I noticed all 4 of the options that we give customers for using this function:

2021-09-16 at 9 52 AM

We want to try and keep `onComplete` always as the last parameter because it's a lambda and that's common practice to have lambda be last. 

Notice in the screenshot 1 of the 4 options doesn't have `onComplete` as the last option. I believe this is the function that makes that so.

levibostian (approved):

Sources/Tracking/CustomerIO.swift
hownowstephen: yeah I waffled a little on this one. In most of our SDKs it is treated as optional, because normally we don't really recommend using it. It also has the possibility of causing some [unexpected behaviour](https://github.com/customerio/services/blob/c630505b789742c6a0c05234f68acdb456605e9b/campaigns/match.go#L81) since we only allow events from the last 72h  to trigger a campaign.

That said, as a happy path & even with a bg queue, I think it's probably reasonable to default to `Date()` but still make it optional in case someone were to want it explicitly to be dated based on the receive time

Sources/Tracking/CustomerIO.swift
hownowstephen: should fix that for `identify` then as well I think it doesn't fit that pattern
Screen Shot 2021-09-16 at 4 10 02 PM

Sources/Tracking/CustomerIO.swift
hownowstephen: that said, I like _even more_ @clabland's idea that we follow the approach in our other SDKs and hide timestamp from the customer, but that we set the value if the request gets sent to the bg queue

levibostian: I see you [made a comment](https://github.com/customerio/customerio-ios/blob/6ffa57404c7cba73d6c4e4e01d129667bd0d0302/Sources/Tracking/CustomerIO.swift#L365) in the source code for doing this in the future. 

I also [made a note in the bg queue ticket](https://github.com/customerio/customerio-ios/issues/36) to do this. 

This is resolved!

Sources/Tracking/CustomerIO.swift
hownowstephen: I spent some time tooling around with doing the stuff in tracking and I think without a reorg of the `identifyRepository` stuff it ends up being overly complicated. Again referring to our other libraries, what we tend to see as the core / in the box functionality is always identifies & events, so IMO it makes sense for those to all be in the same place, instead of having to wrap a whole other object

PR #45, opened 2021-09-20
feat: rich push deep links

levibostian:


Sources/MessagingPush/PushContent.swift
```diff
@@ -0,0 +1,63 @@
...
+/**
+ The content of a push notification. Single source of truth for getting and setting properties of a push notification.
+
+ Meant to be used by the SDK to set/get the push content but also designed for SDK user to further modify the push
+ if they wish.
+ */
+public class PushContent {
+    public var title: String {
+        didSet {
+            modifyNotificationContent()
```
hownowstephen, line 14: why do this over setting the individual pieces in each of these?

hownowstephen (approved):

Sources/MessagingPushAPN/rich push/RichPushRequest.swift
```diff
@@ -0,0 +1,33 @@
...
+    init(
+        pushContent: PushContent,
+        request: UNNotificationRequest,
+        completionHandler: @escaping (UNNotificationContent) -> Void
+    ) {
+        self.completionHandler = completionHandler
+        self.pushContent = pushContent
+    }
+
+    func start() {
```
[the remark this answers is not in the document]
levibostian: `start` is meant to mean to start the rich push request since the class is named `RichPushRequest`. 

The function doesn't do much at this time but it's plumbing for me to do images next. We start async operations when you call this function.

Sources/MessagingPush/PushContent.swift
```diff
@@ -0,0 +1,63 @@
...
+        }
+    }
+
+    public let mutableNotificationContent: UNMutableNotificationContent?
+
+    public static func parse(notificationContent: UNNotificationContent) -> PushContent? {
+        let raw = notificationContent.userInfo
+
+        guard let cio = raw["CIO"] as? [AnyHashable: Any], let cioPush = cio["push"] as? [AnyHashable: Any] else {
```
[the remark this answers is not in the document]
levibostian: The reason to be so generic is for safe type casting. You can assume that if Customer.io sent the push to the SDK, all of this code will execute correctly so we can do something like:

```swift
(raw["CIO"] as! [String: SomeCodable])["push"].link
```

But if the JSON payload from APN is not the format that we expect, a runtime crash could happen. To avoid that we be extra careful by using `Any` and ` if ... as? ...` to prevent the crashes. 

Is that the question you were asking? 

Oh, and `[AnyHashable: Any]` is what `notificationContent.userInfo` gives you so I continued to use that because it's a safe casting type.

Sources/MessagingPush/PushContent.swift
levibostian: I don't understand the question

Sources/MessagingPush/MessagingPush+NotificationCenter.swift
```diff
@@ -0,0 +1,57 @@
...
+    ) -> Bool {
+        guard let pushContent = PushContent.parse(notificationContent: response.notification.request.content) else {
+            return false
+        }
+
+        switch response.actionIdentifier {
+        case UNNotificationDefaultActionIdentifier: // push notification was touched.
+            if let deepLinkurl = pushContent.deepLink {
+                UIApplication.shared.open(url: deepLinkurl)
+
```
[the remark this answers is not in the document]
levibostian: Yeah, I meant to delete the code below. Done. 

Originally `onOpenURL` and `application(open:options)` was to track that a deep link got clicked but we can allow customers to do that themselves if they wish on their own.

Sources/MessagingPushAPN/rich push/RichPushRequestHandler.swift
```diff
@@ -0,0 +1,40 @@
...
+
+        let newRequest = RichPushRequest(pushContent: content, request: request, completionHandler: completionHandler)
+        requests[requestId] = newRequest
+
+        newRequest.start()
+    }
+
+    func stopAll() {
+        requests.forEach { item in
+            let request = item.value
```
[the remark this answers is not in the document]
levibostian: I created the `let` out of habit. Since `item` is a tuple key/value pair I usually name the `item.value` to a variable name that makes more sense in the context. In this scenario, value == request. 

But for this very simple loop, no need.

Sources/MessagingPush/PushContent.swift
hownowstephen: like
```
public var title: String {
        didSet {
            mutableNotificationContent?.title = title
        }
}
```

Sources/MessagingPush/PushContent.swift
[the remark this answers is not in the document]
hownowstephen: right, the approach I'd take if this were done in Golang would've been something like

```go
type CIOPushData struct {
  Link string `json:"link"`
}

cio, ok := raw["CIO"]
if !ok {
  return
}

var pushData CIOPushData
err := json.Unmarshal(&pushData, cio["push"])
if err != nil {
  return false
}

...
```
or something similar that basically checks whether it unmarshals into a well-defined struct and then passes that along, instead of continuing to work directly with arbitrary json

hownowstephen (approved):

levibostian:
merge https://github.com/customerio/customerio-ios/pull/51 after this is merged

Sources/MessagingPush/PushContent.swift
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    private let jsonAdapter: JsonAdapter
+    public let mutableNotificationContent: UNMutableNotificationContent
+
+    public static func parse(notificationContent: UNNotificationContent, jsonAdapter: JsonAdapter) -> PushContent? {
+        let raw = notificationContent.userInfo
+
+        guard let cioUserInfo = raw["CIO"] as? [AnyHashable: Any],
+              let _: CioPushPayload = jsonAdapter.fromDictionary(cioUserInfo),
```
hownowstephen, line 69: if someone sends in garbage here, like a link that is an integer for some reason, will this crash?

hownowstephen (approved):

Sources/MessagingPush/PushContent.swift
levibostian: [No, that should not happen](https://github.com/customerio/customerio-ios/blob/rich-push-links/Tests/Tracking/Util/JsonAdapterTest.swift#L61-L71)

`fromDictionary`'s logic is wrapped in a try/catch so if the payload doesn't exactly match the format, you get `nil` back.

hownowstephen: 🆒

PR #52, opened 2021-09-22
chore: register apn device token Data

hownowstephen:
Adds an overload to the `registerDeviceToken` func on the `MessagingPushAPN` class for those who want to call it directly instead of using our wrapper


levibostian (approved):
Nice naming of the functions 😄!

PR #59, opened 2021-09-23
feat: rich push images

levibostian:


Sources/Tracking/Service/HttpRequestRunner.swift
```diff
@@ -57,6 +58,33 @@ internal class UrlRequestHttpRequestRunner: HttpRequestRunner {
         }.resume()
     }
 
+    public func downloadFile(url: URL, onComplete: @escaping (URL?) -> Void) {
```
hownowstephen, line 61: is there a way for us to set a lifespan for cache items explicitly? or is that all managed by iOS under the hood

Sources/Tracking/Service/HttpRequestRunner.swift
```diff
@@ -57,6 +58,33 @@ internal class UrlRequestHttpRequestRunner: HttpRequestRunner {
...
+                return onComplete(nil)
+            }
+
+            // note make sure that the file name below is unique to not overwrite previous file.
+            var uniqueFileName = String.random
+            if let fileExtension = url.absoluteString.fileExtension {
+                uniqueFileName += ".\(fileExtension)"
+            }
+
+            let destinationURL = documentsDirectoryURL.appendingPathComponent(uniqueFileName)
```
hownowstephen, line 76: is there any value in putting these in a `cio` subdirectory in the cache or something?

Sources/Tracking/Service/HttpRequestRunner.swift
```diff
@@ -57,6 +58,33 @@ internal class UrlRequestHttpRequestRunner: HttpRequestRunner {
...
+        else { return }
+
+        session.downloadTask(with: url) { tempLocation, _, _ in
+            guard let tempLocation = tempLocation else {
+                return onComplete(nil)
+            }
+
+            // note make sure that the file name below is unique to not overwrite previous file.
+            var uniqueFileName = String.random
+            if let fileExtension = url.absoluteString.fileExtension {
```
hownowstephen, line 72: file metadata is also available as a `Content-Type` header if we wanted to be particularly robust, but extension is probably fine and saves us having to parse the MIME Type

Sources/MessagingPushAPN/RichPush/RichPushRequestHandler.swift
```diff
@@ -14,14 +14,22 @@ internal class RichPushRequestHandler {
...
+        customerIO: CustomerIO,
         completionHandler: @escaping (UNNotificationContent) -> Void
     ) {
         let requestId = request.identifier
 
         let existingRequest = requests[requestId]
         if existingRequest != nil { return }
 
-        let newRequest = RichPushRequest(pushContent: content, request: request, completionHandler: completionHandler)
+        // XXX: After we are able to inject HttpClient using DI graph, this is no longer needed.
```
hownowstephen, line 25: cool, that explains it then. definitely seemed weird to have that dep added

hownowstephen (approved):

Sources/Tracking/Service/HttpRequestRunner.swift
hownowstephen: additionally, in a bunch of our Go code we use a `LimitReader` to avoid loading an unbounded-size file, does it make sense to do that in the SDK as well?

levibostian:
merge https://github.com/customerio/customerio-ios/pull/60 after this

Sources/Tracking/Service/HttpRequestRunner.swift
levibostian: > is there any value in putting these in a cio subdirectory in the cache or something?
Yes, I believe so. Good idea. 

> additionally, in a bunch of our Go code we use a LimitReader to avoid loading an unbounded-size file, does it make sense to do that in the SDK as well?
I created [a ticket](https://github.com/customerio/customerio-ios/issues/61) for this. I suggest we do this another day because (1) it will take a bit more work to get this working and (2) we talked about for `alpha` we suggest to customers in the docs to limit their file size of the image they want to show.

Sources/Tracking/Service/HttpRequestRunner.swift
levibostian: I changed from using the system cache directory to system tmp directory. Better place to store these files since `cache` must be managed by us I believe.

levibostian: I also added a `cleanup` function when a push notification is touched or dismissed to delete images that we saved.

hownowstephen (approved):

PR #63, opened 2021-09-24
feat: automatic push events

hownowstephen:
Introduces a new configuration option `autoTrackPushEvents` to send events to Customer.io when a push is handled to indicate it was delivered, or in the case of a non-dismiss interaction, opened

### DIFF
--- a/Sources/MessagingPush/MessagingPush+NotificationCenter.swift
+++ b/Sources/MessagingPush/MessagingPush+NotificationCenter.swift
@@ -61,6 +61,7 @@ public extension MessagingPush {
         }
         
         trackMetric(deliveryID: deliveryID, event: event, deviceToken: deviceToken) { result in
+            // XXX: pending background queue so that this can get retried instead of discarding the result
         }
     }
 
--- a/Sources/MessagingPush/MessagingPush+NotificationCenter.swift
+++ b/Sources/MessagingPush/MessagingPush+NotificationCenter.swift
@@ -21,8 +21,7 @@ public extension MessagingPush {
         switch response.actionIdentifier {
         case UNNotificationDismissActionIdentifier, UNNotificationDefaultActionIdentifier:
             if customerIO.sdkConfig.autoTrackPushEvents {
-                trackMetric(notificationContent: response.notification.request.content, event: .delivered,
-                            jsonAdapter: DITracking.shared.jsonAdapter) { _ in
+                trackMetric(notificationContent: response.notification.request.content, event: .delivered) { _ in
                     // XXX: pending background queue so that this can get retried instead of discarding the result
                 }
             }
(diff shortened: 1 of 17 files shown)


Sources/MessagingPush/MessagingPush+NotificationCenter.swift
```diff
@@ -42,6 +59,24 @@ public extension MessagingPush {
         return false
     }
 
+    func trackMetric(
+        notificationContent: UNNotificationContent,
+        event: Metric,
+        onComplete: @escaping (Result<Void, CustomerIOError>) -> Void
+    ) {
```
levibostian, line 66: Should this function go into `MessagingPush` instead of `MessagingPush+NotificationCenter`?

Sources/MessagingPush/MessagingPush+NotificationCenter.swift
```diff
@@ -42,6 +59,24 @@ public extension MessagingPush {
...
+        guard let deliveryID: String = notificationContent.userInfo["CIO-Delivery-ID"] as? String else {
+            onComplete(Result.success(()))
+            return
+        }
+
+        guard let deviceToken: String = notificationContent.userInfo["CIO-Delivery-Token"] as? String else {
+            onComplete(Result.success(()))
+            return
+        }
```
levibostian, line 75:
```suggestion
        guard let deliveryID: String = notificationContent.userInfo["CIO-Delivery-ID"] as? String, 
                  let deviceToken: String = notificationContent.userInfo["CIO-Delivery-Token"] as? String else {
            return onComplete(Result.success(()))
        }
```

Sources/MessagingPush/MessagingPush+NotificationCenter.swift
levibostian: Ah, nevermind. I see this is more of a convenient extension. Got it.

Sources/MessagingPush/MessagingPush+NotificationCenter.swift
```diff
@@ -42,6 +59,24 @@ public extension MessagingPush {
...
+
+        trackMetric(deliveryID: deliveryID, event: event, deviceToken: deviceToken, onComplete: onComplete)
```
levibostian, line 77: I find `deviceToken` named parameter confusing because it is the same naming we give FCM and APN device tokens. I am wondering if customers will find it confusing? 

The [api body](https://customer.io/docs/api/#operation/pushMetrics) names it `deliver_id` while the push JSON payload's name is `CIO-Delivery-Token` as we see here. 

Any suggestions?

levibostian (approved):

Sources/MessagingPush/MessagingPush+NotificationCenter.swift
hownowstephen: `CIO-Delivery-ID` is a delivery ID, and `CIO-Delivery-Token` _is_ the deviceToken of the app instance that received the push. I assume we did it this way in case the customer wasn't storing a token locally in their integratino. IMO using the one on the headers is probs the right move, but it wouldn't matter either way.

hownowstephen: in the api we call it `delivery_id` and `device_id` so we could I guess use `deviceID` but since it is actually a deviceToken, felt like it would be weird to introduce a new term into the SDK

levibostian (approved):

Issue #58, opened 2021-09-23
Better way of managing singleton classes 


levibostian:
Today in top-level classes like `CustomerIO` we have to have [logic in place](https://github.com/customerio/customerio-ios/blob/5cd4f2d12a9d606c06cc477fe190b99593ac2f7c/Sources/Tracking/CustomerIO.swift#L307-L309) to determine if the shared singleton instance of `CustomerIO` has been initialized or not. Today it's done with conditionals in each public facing SDK function but that can get complex to handle and error prone as the SDK grows in functionality. 

This issue is to come up with a better solution to this problem so it's simple to handle using a singleton or not a singleton instance of the top-level classes.

levibostian:
Idea:

```swift
class CustomerIOImplementation {
  init(config, credentials) {}

  fun identifyCustomer() {
    repository.identify() 
   }
}

class CustomerIO {
  init(siteid, apikey) {
     self.implementation = CustomerIOImplementation(...)
  }
  fun identifyCustomer() {
    self.implentation?.identifyCustomer()
  }
}
```

We have 1 class where all of the logic of the `CustomerIO` class exists. It never has to handle the different use scenarios if the SDK has been initialized or not. It must be constructed only after the SDK has been initialized. Then, in the public facing API of the `CustomerIO` class, we simply handle the "have you initialized the SDK?" scenario there in each function. 

pros:
* scalable
* `XImplementation` class more easily testable
* simple

cons:
* 2 classes required to perform the work of the public facing function. The nice thing is we only need 2 classes for the top-level classes of the SDK like `CustomerIO`, `MessagingPush`.

Issue #57, opened 2021-09-23
Improve dependency injection graph to allow integration tests and easier use


levibostian:
Today, the dependency injection graph is:
* One graph for the whole SDK. All instances of the SDK objects no matter what unique siteid and apikey you want to use. This means that if you want to use different site ids in your app, the SDK will not work. 
* Because of the use case defined above with supporting multiple different site ids in the SDK, the dependency injection graph does not truly work as intended today. 

Here is an example of what we want:
```swift
class HttpClient {
  init(httpRunner: HttpRunner) {
    // httpRunner dependency automatically gets resolved by the DI graph
  }
}
```

But, what we have today:
```swift
class HttpClient {
  init(siteid: String) {
    self.httprunner = DI.shared.httpRunner
  }
}
```

As you can see, today we are not really using the DI graph as it should be used. 

Proposal:
- [ ] Create 1 DI graph per site id.  

Some ideas of how to do this:
* create a new DI graph instance in each top-level class (such as `CustomerIO`) where all dependencies get resolved from that top-level. You can inject in the constructor a DI graph instance into these top-level classes for integration test purposes. 
*  create a singleton manager where you must provide a site id in order to get a shared instance of a DI graph. 
- [ ] We can now write integration tests! They are difficult to create today because some classes are not injected into the SDK by use of the DI graph (such as HttpClient). Now that *all* dependencies are in the graph, we can override them in tests.

PR #69, opened 2021-10-06
refactor: separate di graphs by siteid

levibostian:
> Note: Not ready to merge until implemented and tested into Remote Habits app. 

Big refactor. The DI graph touches *a lot* of classes! 

Here is an overview of what was changed:
* There is now 1 DI graph per `siteId`. There used to be 1 DI graph shared with *all* instances and singletons of the SDK no matter the `siteId`. [learn more](https://github.com/customerio/customerio-ios/issues/57)
* Convert dependencies to be injected via the DI graph. Some dependencies (such as HttpClient) were not originally being injected with the DI graph. I refactored each class to now be part of the graph. 
* Fixed [this issue](https://github.com/customerio/customerio-ios/issues/58)` as it fit well into this work.

### DIFF
--- a/Tests/Tracking/Service/HttpClientTest.swift
+++ b/Tests/Tracking/Service/HttpClientTest.swift
@@ -109,10 +109,10 @@ class HttpClientTest: UnitTest {
 
     func test_request_givenUnsuccessfulStatusCode_expectError() {
         let expectedCode = 500
-        let expectedError = HttpRequestError.unsuccessfulStatusCode(expectedCode, apiMessage: "")
+        let expectedError = HttpRequestError.unsuccessfulStatusCode(expectedCode, apiMessage: "invalid id")
 
         requestRunnerMock.requestClosure = { _, _, _, onComplete in
-            onComplete(nil,
+            onComplete(#"{"meta": { "error": "invalid id" }}"#.data,
                        HTTPURLResponse(url: self.url, statusCode: expectedCode, httpVersion: nil, headerFields: nil),
                        nil)
         }


Sources/MessagingPush/MessagingPush+NotificationCenter.swift
```diff
@@ -18,9 +18,18 @@ public extension MessagingPush {
         didReceive response: UNNotificationResponse,
         withCompletionHandler completionHandler: @escaping () -> Void
     ) -> Bool {
+        guard let siteId = customerIO.siteId else {
```
hownowstephen, line 21: been thinking a little about all local state being held relative the `siteId` - I think the most likely case for a `siteId` ever changing is going to be in a scenario where it needs to be rotated for some reason, and doing so will lose any local context / pending work. Don't know that there's a great solution for that, except that we should at least have some recommendations prepared for a future customer as to how to safely rotate their key.

no action items for this, just food for thought

Sources/MessagingPush/MessagingPush.swift
```diff
@@ -20,124 +20,44 @@ public protocol MessagingPushInstance: AutoMockable {
...
         self.customerIO = customerIO
-        self.httpClient = CIOHttpClient(credentials: customerIO.credentials!, config: customerIO.sdkConfig)
-        self.jsonAdapter = DITracking.shared.jsonAdapter
-        self.eventBus = DITracking.shared.eventBus
-
-        self.identifyCustomerEventBusCallback = eventBus.register(event: .identifiedCustomer) {
-//            if let deviceToken = self.deviceToken {
-            // register device token with customer.
-//            }
+        if let siteId = customerIO.siteId {
```
hownowstephen, line 34: mostly consistent with how our other API clients work - they only error once they're actually used - but would it be beneficial to developers to log or emit something at this point if the siteId doesn't exist?

Sources/Tracking/CustomerIOImplementation.swift
```diff
@@ -0,0 +1,110 @@
...
+     Constructor for singleton, only.
+
+     Try loading the credentials previously saved for the singleton instance.
+     */
+    internal init(siteId: String) {
+        self._siteId = siteId
+
+        self.diGraph = DITracking.getInstance(siteId: siteId)
+
+        self.identifyRepository = diGraph.identifyRepository
```
hownowstephen, line 38: is there still value in having the concept of the `identifyRepository` - since we can get access to the `profileStore` directly? feels like it would simplify development by removing one more layer (similar to how the `MessagingPush` module just does the http calls directly within its methods)

levibostian: Good question. 

The idea behind a repository is re-usability and keeping this class more simple by delegating tasks. However, as you have stated, the logic is not re-used anywhere at the moment but this 1 place. 

I imagine that when a background queue comes, all of the logic can be moved out of this class and in another class that runs background queue operations. Let's consider removing repositories then?

Sources/MessagingPush/MessagingPushImplementation.swift
```diff
@@ -0,0 +1,148 @@
...
+            return onComplete(Result.failure(.noCustomerIdentified))
+        }
+
+        let httpRequestParameters = HttpRequestParams(endpoint: .registerDevice(identifier: identifier), headers: nil,
+                                                      body: bodyData)
+
+        httpClient
+            .request(httpRequestParameters) { [weak self] result in
+                DispatchQueue.main.async { [weak self] in
+                    guard let self = self else { return }
```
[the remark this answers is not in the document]
levibostian: Can you think of a use case where calling a completion handler with an error would be good? 

`self` is nil when the SDK objects are being garbage collected and therefore, means the app is probably exiting. That sounds like the expected behavior that a customer would expect to happen, no?

Sources/Tracking/Service/HttpClient.swift
```diff
@@ -14,74 +14,75 @@ public protocol HttpClient: AutoMockable {
...
-                        errorBodyString = errorMessageBody.meta.error
+                let statusCode = response.statusCode
+                guard statusCode < 300 else {
+                    switch statusCode {
+                    case 401:
+                        onComplete(.failure(.unauthorized))
+                    default:
+                        var errorBodyString: String = data?.string ?? ""
+                        if let data = data,
+                           let errorMessageBody: ErrorMessageResponse = self.jsonAdapter.fromJson(data) {
```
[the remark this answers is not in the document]
levibostian: Good catch. I made a change for that https://github.com/customerio/customerio-ios/pull/69/commits/68fe875f8cf0693a2dce686ae254d05ec5604e77

PR #71, opened 2021-10-09
feat: pre-login tracking

hownowstephen:
Add support for tracking customer activity before they have been identified.

### Logged out tracking flow:
- If configured, identify a profile with a randomized ID generated on-device
- Until an `identifier` is set, all tracks / identifies / device tokens go to the temporary customer
- At any time you can manually merge into another customer, which will set the `identifier`
- Or calling `identify` will do so automatically

WIP: still pending some test coverage and addressing the `XXX:` comments


hownowstephen:
closing for now while we work out some implementation on the CIO side

PR #74, opened 2021-11-01
fix: improve user-agent with more detail

ami-aman:
[Issue 6245](https://github.com/customerio/issues/issues/6245) 

This PR contains : 

### Adding user-agent value in request header ### 

- Add value of user-agent in header which can be easily parsed by the backend
- Format followed : `Customer.io iOS Client/1.0.0-alpha.16 (x86_64; iOS 14.5) User App/1.0`
- In case UIKit is unavailable : `Customer.io iOS Client/1.0.0-alpha.16`

Manual testing done on :
- [x] Simulator
- [ ] Physical Device

### DIFF
--- a/Sources/Tracking/Extensions/DeviceExtension.swift
+++ b/Sources/Tracking/Extensions/DeviceExtension.swift
@@ -1,0 +1,18 @@
+import Foundation
+import UIKit
+
+public extension UIDevice {
+
+    static let deviceModelCode : String = {
+            
+        var systemInfo = utsname()
+        uname(&systemInfo)
+        let machineMirror = Mirror(reflecting: systemInfo.machine)
+        let identifier = machineMirror.children.reduce("") { identifier, element in
+            guard let value = element.value as? Int8, value != 0 else { return identifier }
+            return identifier + String(UnicodeScalar(UInt8(value)))
+        }
+        return identifier
+    }()
+    
+}
--- a/Sources/Tracking/Extensions/DeviceExtension.swift
+++ b/Sources/Tracking/Extensions/DeviceExtension.swift
@@ -2,6 +2,18 @@ import Foundation
 #if canImport(UIKit)
 import UIKit
 
+
+/**
+ Extend `UIDevice` to get user's device information such as
+ device's model. If running on Simulator `deviceModelCode`
+ will return values like`x86_64` but when running on a device, this function
+ returns exact device model for example, `iPhone12,3`
+ 
+ Use case :
+ To get model detail, simply use
+ 
+ let deviceModelInfo = UIDevice.deviceModelCode
+ */
 public extension UIDevice {
 
     static let deviceModelCode : String = {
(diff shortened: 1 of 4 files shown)


Sources/Tracking/Service/HttpClient.swift
```diff
@@ -127,4 +128,15 @@ extension CIOHttpClient {
 
         return encodedRawHeader.base64EncodedString(options: NSData.Base64EncodingOptions(rawValue: 0))
     }
+    
+    static func getUserAgent() -> String {
```
levibostian, line 132: Can you add some unit tests against this function? 

I suggest a test making sure that the returned User-Agent returns a User-Agent string that meets the requirement of the API. That way if the code ever gets modified we can feel confident our code always meets the requirement of the API.

Sources/Tracking/Service/HttpClient.swift
```diff
@@ -127,4 +128,15 @@ extension CIOHttpClient {
...
+        var userAgent = "Customer.io iOS Client/"
+        userAgent += SdkVersion.version
+        #if canImport(UIKit)
+        userAgent += " (\(DeviceInfo.deviceInfo); \(DeviceInfo.osInfo))"
+        userAgent += " \(DeviceInfo.customerAppName)/"
+        userAgent += DeviceInfo.customerAppVersion
+        #endif
+        return userAgent
```
levibostian, line 140: I think it would be helpful to add a code comment for this function with the User-Agent format the API requires. It would be good documentation for the iOS dev team to make sure we are consistent with the requirement.

Sources/Tracking/Extensions/DeviceExtension.swift
```diff
@@ -0,0 +1,19 @@
+import Foundation
+#if canImport(UIKit)
+import UIKit
+
+public extension UIDevice {
+
+    static let deviceModelCode : String = {
```
levibostian, line 7: Mind adding helpful code comments to this function to help developers understand what this function is trying to do and how it works? Reading this, I have no idea what's going on.

levibostian (changes requested):
After adding tests and reviewing comments, this is ready to go 🎊

Sources/Tracking/Service/HttpClient.swift
ami-aman: Done. Have added a unit test case for this function. Can be found in latest commit in this PR

Sources/Tracking/Service/HttpClient.swift
ami-aman: Added detailed comment on this function along with the expected results when UIKit is found and when not found

Sources/Tracking/Extensions/DeviceExtension.swift
ami-aman: Have added comments to this function as well. The comment also documents a use case to use this extension.

levibostian (approved):

PR #80, opened 2021-11-19
fix: logs now show up in mac console app

levibostian:
To better improve the QA experience of the team, log messages now show up in the macOS Console app instead of just the Xcode console. This allows the team to be able to view logs for pre-built versions of the app and not just builds built in Xcode.

### DIFF
--- a/Sources/Tracking/Util/Log.swift
+++ b/Sources/Tracking/Util/Log.swift
@@ -1,5 +1,7 @@
 import Foundation
+#if canImport(os)
 import os.log
+#endif
 import OSLog
 
 /// mockable logger + abstract that allows you to log to multiple places if you wish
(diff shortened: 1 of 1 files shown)


Sources/Tracking/Util/Log.swift
```diff
@@ -1,34 +1,49 @@
...
+public protocol Logger: AutoMockable {
+    /// the noisey log level. Feel free to spam this log level with any
+    /// information about the SDK that would be useful for debugging the SDK.
     func debug(_ message: String)
+    /// Not noisy log messages. Good for general information such as
+    /// when the background queue begins and ends running but use `debug`
+    /// for the status of each background queue task running.
     func info(_ message: String)
-    func warning(_ message: String)
+    /// the SDK is in an unstable state that you want to notify
```
hownowstephen, line 14: decided to cut the warn level?

hownowstephen (approved):

Sources/Tracking/Util/Log.swift
levibostian: Yes. The original list of log levels (info, warning, verbose, etc) mostly came [Android](https://developer.android.com/reference/android/util/Log) TBH. 

[The Swift Logger class](https://developer.apple.com/documentation/os/oslogtype) does not include a `warning` case so I removed it.

PR #81, opened 2021-11-22
feat: run queue tasks in other SDKs via "hooks"

levibostian:
- Implement "hooks" into background queue. This allows each separate SDK at runtime to communicate between each other. If an SDK is not installed by a customer, hooks are never registered for that SDK. 

- I did begin to convert over some of the SDK functions to use hooks. You will notice this in `CustomerIOImplementation` where the `onComplete` callback is removed. I did these couple of conversions to make sure my hooks implementation would solve the use cases that we need it to in the SDK.


Sources/MessagingPush/Repository/PushDeviceTokenRepository.swift
```diff
@@ -0,0 +1,69 @@
+import CioTracking
+import Foundation
+
+internal protocol PushDeviceTokenRepository: AutoMockable {
```
levibostian, line 4: I am now realizing that having this repository class doesn't add any value. I think I created it for a reason that no longer is necessary? 

Either way, in my work I am doing now for the background queue, I am removing this class and moving the register/delete device token work into `MessagingPushImplementation`

Sources/Tracking/CustomerIOImplementation.swift
```diff
@@ -70,28 +70,35 @@ public class CustomerIOImplementation: CustomerIOInstance {
...
         jsonEncoder: JSONEncoder? = nil
     ) {
-        let isChangingIdentifiedProfile = profileStore.identifier != nil && profileStore.identifier == identifier
-        if isChangingIdentifiedProfile {
-            // TODO: add to background queue delete device token from currently registered profile.
+        let currentlyIdentifiedProfileIdentifier = profileStore.identifier
+        let isChangingIdentifiedProfile = currentlyIdentifiedProfileIdentifier != nil &&
+            currentlyIdentifiedProfileIdentifier != identifier
+
+        if let currentlyIdentifiedProfileIdentifier = currentlyIdentifiedProfileIdentifier,
```
hownowstephen, line 77: this some sort of garbage collection thing? why are we re-`let`ing the same var here?

Sources/Tracking/CustomerIOImplementation.swift
```diff
@@ -70,28 +70,35 @@ public class CustomerIOImplementation: CustomerIOInstance {
...
+           isChangingIdentifiedProfile {
+            hooks.profileIdentifyHooks.forEach { hook in
+                hook.beforeNewProfileIdentified(oldIdentifier: currentlyIdentifiedProfileIdentifier,
```
hownowstephen, line 80: not sure the name is super expressive. `beforeIdentifiedProfileChange` maybe?

hownowstephen (approved):

Sources/Tracking/CustomerIOImplementation.swift
levibostian: It's a common swift pattern to safely check a variable for null. 

Since `currentlyIdentifiedProfileIdentifier` is `String?`, if I ever want to use it I must do `currentlyIdentifiedProfileIdentifier!` which is dangerous. Instead, we can safely check by using `if let ...`

```swift
if let id = currentlyIdentifiedProfileIdentifier {
```

but, I don't like to introduce 2 variable names for the same 1 variable. That's why we use the same name:

```swift
if let currentlyIdentifiedProfileIdentifier = currentlyIdentifiedProfileIdentifier {
  // now, when you use `currentlyIdentifiedProfileIdentifier` in this scope, it's String not String?
```

Sources/Tracking/CustomerIOImplementation.swift
levibostian: Yeah, I like that. As long as the name is suggesting that we are *changing to a new profile* when there was a previous profile identified, that will work.

PR #82, opened 2021-11-22
feat: add screen view tracking

hownowstephen:
Adds support for sending `screen` type events to Customer.io.

For customers using `UIKit` this also introduces _automatic screen view tracking_ as an opt-in feature using the `autoTrackScreenViews` parameter.

To set a custom default body handler when tracking screen views, you can additionally set the `autoScreenViewBody: (() -> [String:AnyEncodable])?` method on the config as well, and it will be called prior to tracking a screen view


Sources/Tracking/CustomerIO.swift
```diff
@@ -22,7 +24,18 @@ public protocol CustomerIOInstance: AutoMockable {
         jsonEncoder: JSONEncoder?,
         onComplete: @escaping (Result<Void, CustomerIOError>) -> Void
     )
-    func clearIdentify()
+
+    // sourcery:Name=screenView
```
levibostian, line 28: Why not use `screen` as the name of the mocked function to match the name of the function here?

Sources/Tracking/CustomerIO.swift
```diff
@@ -22,7 +24,18 @@ public protocol CustomerIOInstance: AutoMockable {
...
+    func screen<RequestBody: Encodable>(
+        name: String,
+        // sourcery:Type=AnyEncodable
+        // sourcery:TypeCast="AnyEncodable(data)"
+        data: RequestBody?,
+        jsonEncoder: JSONEncoder?,
+        onComplete: @escaping (Result<Void, CustomerIOError>) -> Void
+    )
+
+    func enableAutoScreenviewTracking()
```
levibostian, line 38: I imagined this being in the `CustomerIO.config {}` code, only. Is there a reason for there being a function here? 

Also, if there is a function to *enable* auto screen view tracking, should there also be a way to *disable* it, too?

Sources/Tracking/CustomerIO.swift
```diff
@@ -84,6 +97,14 @@ public extension CustomerIOInstance {
     ) {
         track(name: name, data: EmptyRequestBody(), jsonEncoder: jsonEncoder, onComplete: onComplete)
     }
+
+    func screen(
+        name: String,
+        jsonEncoder: JSONEncoder? = nil,
```
levibostian, line 103: No need to include the `jsonEncoder` param since there is not `body` param.

Sources/Tracking/CustomerIO.swift
```diff
@@ -115,6 +136,15 @@ public class CustomerIO: CustomerIOInstance {
     internal var implementation: CustomerIOImplementation?
 
     internal var globalData: GlobalDataStore = CioGlobalDataStore()
+    
+    public var autoScreenViewBody: () -> ScreenViewData {
```
levibostian, line 140: I imagined this being in the `CustomerIO.config {}` code, only. Is there a reason for there being a function here?

Sources/Tracking/CustomerIO.swift
```diff
@@ -287,4 +317,37 @@ public class CustomerIO: CustomerIOInstance {
...
         implementation.track(name: name, data: data, jsonEncoder: jsonEncoder, onComplete: onComplete)
     }
+
+    /**
+     Track a a screen view
+
+     [Learn more](https://customer.io/docs/events/) about events in Customer.io
+
+     - Parameters:
+     - name: Name of the currently active screen. When automatically tracked, we use the name of the controller without any `ViewController` substrings
```
levibostian, line 327: > When automatically tracked, we use the name of the controller without any `ViewController` substrings

When I read these docs, I wonder, "When I am not using automatic tracking, does the SDK also remove `ViewController` substring?"

Sources/Tracking/CustomerIOImplementation+ScreenViews.swift
```diff
@@ -0,0 +1,58 @@
...
+        var viewController = window!.rootViewController
+        if viewController is UINavigationController {
+            viewController = (viewController as! UINavigationController).visibleViewController
+        }
+
+        var name = String(describing: type(of: viewController)).replacingOccurrences(of: "ViewController", with: "")
+
+        if name.isEmpty || name == "" {
+            if viewController?.title != nil {
+                name = viewController?.title ?? "Unknown"
```
ami-aman, line 38: default value will never be executed since the check is for `viewController?.title != nil` which means the code block will be executed only when value is not nil. Also, it is always better to use if let in such cases

Sources/Tracking/CustomerIOImplementation+ScreenViews.swift
```diff
@@ -0,0 +1,58 @@
...
+        screen(name: name, data: autoScreenViewBody()) { _ in
+            // TODO: global error handling of result here
+        }
+    }
+}
+
+#else
+public extension CustomerIOImplementation {
+    func enableAutoScreenviewTracking() {
+        // TODO: log warning that tracking is not available
```
levibostian, line 55: Our Swift linting tool marks `TODO` as a warning. I prefer we stick with `XXX: ` to make the linter happy. I like this lint behavior because you can still use TODO as personal reminders to yourself before your work is done but it's not the best for longer term future work like in this case, logging.

Sources/Tracking/CustomerIOImplementation+ScreenViews.swift
```diff
@@ -0,0 +1,58 @@
+import Foundation
+#if canImport(UIKit)
+import UIKit
+
+public extension CustomerIOImplementation {
+    func enableAutoScreenviewTracking() {
+        let selector1 = #selector(UIViewController.viewDidAppear(_:))
```
levibostian, line 7: There are many `UIViewController.viewXXX` functions. Mind sharing more about why you chose `viewDidAppear`? Adding a code comment for that would be helpful for documentation.

Sources/Tracking/CustomerIOImplementation+ScreenViews.swift
```diff
@@ -0,0 +1,58 @@
...
+
+        guard let delegate = UIApplication.shared.delegate else {
+            return
+        }
+
+        guard let window = delegate.window else {
+            return
+        }
+
+        var viewController = window!.rootViewController
```
levibostian, line 29:
```suggestion
        var viewController = window.rootViewController
```

Shouldn't need `!` because `guard let window =`  above?

Sources/Tracking/CustomerIOImplementation+ScreenViews.swift
```diff
@@ -0,0 +1,58 @@
...
+
+        if name.isEmpty || name == "" {
+            if viewController?.title != nil {
+                name = viewController?.title ?? "Unknown"
+            }
+            if name.isEmpty || name == "" {
+                // Could not infer screen name
+                name = "Unknown"
+            }
+        }
```
ami-aman, line 44: Can be reduced to

`if (name.isEmpty || name == "") { 
if let vcTitle = viewController?.title {
name = vcTitle
}else {
name = "unknown"
}
}`

Sorry for bad indentation here. 😅

Sources/Tracking/CustomerIOImplementation+ScreenViews.swift
```diff
@@ -0,0 +1,58 @@
...
+
+        guard let window = delegate.window else {
+            return
+        }
+
+        var viewController = window!.rootViewController
+        if viewController is UINavigationController {
+            viewController = (viewController as! UINavigationController).visibleViewController
+        }
```
levibostian, line 32:
```suggestion
        if let navigationController = viewController as? UINavigationController {
            viewController = navigationController.visibleViewController
        }
```

To avoid force cast.

Sources/Tracking/CustomerIOImplementation+ScreenViews.swift
```diff
@@ -0,0 +1,58 @@
...
+
+        var name = String(describing: type(of: viewController)).replacingOccurrences(of: "ViewController", with: "")
+
+        if name.isEmpty || name == "" {
+            if viewController?.title != nil {
+                name = viewController?.title ?? "Unknown"
+            }
```
levibostian, line 39:
```suggestion
            if let title = viewController?.title {
                name = title
            }
```

Sources/Tracking/CustomerIOImplementation+ScreenViews.swift
```diff
@@ -0,0 +1,58 @@
...
+            if name.isEmpty || name == "" {
+                // Could not infer screen name
+                name = "Unknown"
+            }
```
levibostian, line 43: Do we want to track a screen with the text "Unknown"? 

I would imagine a customer using the web app might find it not valuable to see a screen tracked "Unknown" in their campaign?

Sources/Tracking/CustomerIOImplementation.swift
```diff
@@ -59,6 +63,10 @@ public class CustomerIOImplementation: CustomerIOInstance {
         handler(&configToModify)
 
         sdkConfigStore.config = configToModify
+
+        if sdkConfigStore.config.autoTrackScreenViews {
+            enableAutoScreenviewTracking()
+        }
```
levibostian, line 69: Because this code exists, I don't see a value in adding `enableAutoScreenviewTracking()` to `CustomerIOInstance` protocol that is public facing that a customer could use. I find it confusing to have multiple places where a customer could configure the SDK for automatic screen view tracking.

Sources/Tracking/CustomerIOImplementation.swift
```diff
@@ -107,4 +115,30 @@ public class CustomerIOImplementation: CustomerIOInstance {
...
+                    case .failure(let error):
+                        return onComplete(Result.failure(error))
+                    }
+                }
+            }
+    }
+
+    public static func defaultScreenViewBody() -> ScreenViewData {
+        ScreenViewData(data: ScreenViewDefaultData())
+    }
```
levibostian, line 143: A more Swift-y way of doing this is:
```swift
public static var defaultScreenViewBody: ScreenViewData {
  ScreenViewData(data: ScreenViewDefaultData())
}
```

Sources/Tracking/Service/Request/TrackRequestBody.swift
```diff
@@ -2,7 +2,13 @@ import Foundation
 
 /// https://customer.io/docs/api/#operation/track
 internal struct TrackRequestBody<T: Encodable>: Encodable {
+    let type: EventType
     let name: String
     let data: T?
     let timestamp: Date?
 }
+
+public enum EventType: String, Codable {
```
levibostian, line 11: Since this is `Codable` and we allow customers to pass in a customer `jsonEncoder` for tracking screen views, we will need to explicitly define the `CodingKeys` for these Strings to make sure the API accepts the request.

Sources/Tracking/Store/SdkConfig.swift
```diff
@@ -18,6 +18,12 @@ public struct SdkConfig {
      */
     public var autoTrackPushEvents: Bool = true
 
+    /**
+     Automatic tracking of screen views will generate `screen`-type events on every screen transition within
+        your application. If you would prefer to handle these yourself, disable this feature
+     */
+    public var autoTrackScreenViews: Bool = true
```
levibostian, line 25: Should we make this an opt-in feature?

Sources/Tracking/Service/Request/ScreenViewDefaultData.swift
```diff
@@ -0,0 +1,13 @@
+import Foundation
+
+// ScreenViewData wraps an encoder to allow returning of custom data for automatic screenview tracking
+public struct ScreenViewData: Encodable {
```
levibostian, line 4: Is this for doing custom attributes? 

Wondering what the need for this struct is since `trackEvent()` doesn't use something like this?

Sources/Tracking/autogenerated/AutoMockable.generated.swift
```diff
@@ -166,6 +170,27 @@ public class CustomerIOInstanceMock: CustomerIOInstance, TrackingMock {
         identifyClosure?(identifier, AnyEncodable(body), onComplete, jsonEncoder)
     }
 
+    // MARK: - clearIdentify
+
+    /// Number of times the function was called.
+    public private(set) var clearIdentifyCallsCount = 0
+    /// `true` if the function was ever called.
+    public var clearIdentifyCalled: Bool {
+        clearIdentifyCallsCount > 0
```
ami-aman, line 179: true if function never called OR true if function called atleast once ?

Sources/Tracking/CustomerIO.swift
```diff
@@ -287,4 +317,37 @@ public class CustomerIO: CustomerIOInstance {
...
+     - data: Optional event body data
+     - onComplete: Asynchronous callback with `Result` of tracking an event.
+     Check result to see if error or success. Callback called on main thread.
+     - jsonEncoder: Provide custom JSONEncoder to have more control over the JSON request body
+     */
+    public func screen<RequestBody: Encodable>(
+        name: String,
+        data: RequestBody,
```
ami-aman, line 335: is data optional ?

levibostian: Yes. We do this by using [protocol extensions](https://github.com/customerio/customerio-ios/blob/e376bf46b33128d6265893be8a80cf0b0e7a45bb/Sources/Tracking/CustomerIO.swift#L101-L108)

Sources/Tracking/autogenerated/AutoMockable.generated.swift
levibostian: true if function called atleast once

Sources/Tracking/CustomerIO.swift
hownowstephen: haha whoops, forgot to update the sourcery part when I changed

Sources/Tracking/CustomerIOImplementation+ScreenViews.swift
hownowstephen: 👍

Sources/Tracking/CustomerIOImplementation+ScreenViews.swift
hownowstephen: based this on how Segment handles theirs. Is there a lifecycle event that you think is more suited to tracking? https://github.com/segmentio/analytics-ios/blob/e35f1c984687f3216810145c8f1a8360fad20a72/Segment/Internal/NSViewController%2BSEGScreen.m#L77

Sources/Tracking/CustomerIOImplementation+ScreenViews.swift
hownowstephen: this is a good question, again this was how it is handled in the Segment SDK 🤔

Sources/Tracking/CustomerIO.swift
hownowstephen: yeah this is more of a `setup` than an `enable` func, renamed

Sources/Tracking/CustomerIO.swift
hownowstephen: 👍

Sources/Tracking/CustomerIO.swift
hownowstephen: yeah so this is to allow people to hook in and gather a custom body instead of our default (which today is nothing). Moving around a bit to help clarify

Sources/Tracking/CustomerIO.swift
hownowstephen: tidied

Sources/Tracking/CustomerIOImplementation+ScreenViews.swift
hownowstephen: https://customerio.slack.com/archives/C01QYDKD0SU/p1637786068084900

Sources/Tracking/Service/Request/ScreenViewDefaultData.swift
hownowstephen: yeah I was fighting with some interface things. came up with an alternative

Sources/Tracking/CustomerIOImplementation.swift
hownowstephen: 👍

Sources/Tracking/Store/SdkConfig.swift
hownowstephen: probably yeah. I was writing it with the context of the JS snippet in my head, where we do auto-track
> To capture page view events, add our JavaScript snippet to your site. We will track page views on any page where that snippet is present. The event name of a page view event is the URL of the page that a person visited.

but I think the usage pattern is likely not analogous

Sources/Tracking/CustomerIOImplementation+ScreenViews.swift
hownowstephen: yeah, dropping it gives
```
Value of optional type 'UIWindow?' must be unwrapped to refer to member 'rootViewController' of wrapped base type 'UIWindow'
```
so I assume either I'm doing something wrong here or the field is somehow doubly nullable? would appreciate some expert help here 😂

Sources/Tracking/CustomerIOImplementation+ScreenViews.swift
hownowstephen: thanks! switched to using a slightly more swift-y approach that tidies it

Sources/Tracking/Service/Request/ScreenViewDefaultData.swift
hownowstephen: that alternative is `[String:AnyEncodable]` and another protocol signature extension. Do you think we also need to support `[String:Any]` or are people likely to be able to conform to `[String:AnyEncodable]` (calling back to the convo yesterday with Lasting, trying to see whether this'll help them as well)

levibostian: I looked at what other SDKs use as a type signature and `[String: Any]` is preferred. [Here is an idea](https://github.com/segmentio/analytics-swift/blob/c5a7f4ffc2635fd834ae029f130744ca87c30a94/Sources/Segment/Events.swift#L139-L149) for how to get started on that. 

To get `[String: Any]` working with the SDK, it might require more work since `Codable` may or may not be compatible with the `JSON` data type. Should we ignore `[String: Any]` in this PR and move that into a new PR where we make *all* custom attributes in the SDK compatible with `[String: Any]`?

levibostian (approved):

Tests/Tracking/Util/StringAnyEncodable.swift
```diff
@@ -19,9 +19,73 @@ struct DummyData: Codable, Equatable {
...
+        let data = ["foo": Unencodable(data: 12345)] as [String:Any]
+        
+        let json = StringAnyEncodable(data)
+        
+        guard let actual = jsonAdapter.toJson(json, encoder: nil) else {
+            XCTFail("couldn't encode to JSON")
+            return
+        }
+        
+        XCTAssertEqual(expect, String(decoding:actual, as:UTF8.self))
```
levibostian, line 40: To make this a little less copy/paste, the SDK does have [an extension](https://github.com/customerio/customerio-ios/blob/664819caf48d23168ce97e027e51f8040b7922d1/Sources/Tracking/Extensions/DataExtensions.swift#L4) you can use here. 

```swift
actual.string
```

PR #85, opened 2021-11-24
feat: push badge support

ami-aman:
[Issue #77](https://github.com/customerio/customerio-ios/issues/77) - Add support for badges

This PR contains :

- Adding support for badges using Customer.io SDK. 
- Steps to follow can be found [here](Adding Badges to App icon using CIO SDK) . 
- This document also explains in detail about this task, open questions and troubleshooting while implementing this feature from a customer point of view.

Manual testing done on :
- [ ] Simulator
- [x] Physical Device

### DIFF
--- a/Sources/MessagingPush/PushContent.swift
+++ b/Sources/MessagingPush/PushContent.swift
@@ -45,6 +45,15 @@ public class PushContent {
         }
     }
 
+    public var cioHandleBadge: Bool? {
+        get {
+            cio.push.cioHandleBadge
+        }
+        set {
+            cioPush = cioPush.cioHandleBadgeSet(newValue)
+        }
+    }
+    
     public var cioAttachments: [UNNotificationAttachment] {
         mutableNotificationContent.attachments.filter { $0.identifier.starts(with: Self.cioAttachmentsPrefix) }
     }
@@ -119,5 +128,6 @@ struct CioPushPayload: Codable {
     struct Push: Codable, AutoLenses {
         let link: String?
         let image: String?
+        let cioHandleBadge : Bool?
     }
 }
--- a/Sources/MessagingPushAPN/RichPush/MessagingPush+NotificationServiceExtension.swift
+++ b/Sources/MessagingPushAPN/RichPush/MessagingPush+NotificationServiceExtension.swift
@@ -43,9 +43,11 @@ public extension MessagingPush {
             return false
         }
 
+        // If cioHandleBadge exists and value is true
         if let cioHandleBadge = pushContent.cioHandleBadge, cioHandleBadge  {
             
             let badgeCount = store.badgeCount ?? 0
+            // Update badge count in payload
             if let newRequest = request.content.mutableCopy() as? UNMutableNotificationContent {
                 newRequest.badge = NSNumber(value: badgeCount + 1)
                 store.badgeCount = badgeCount + 1
--- a/Sources/MessagingPushAPN/RichPush/MessagingPush+NotificationServiceExtension.swift
+++ b/Sources/MessagingPushAPN/RichPush/MessagingPush+NotificationServiceExtension.swift
@@ -1,7 +1,6 @@
 import CioMessagingPush
 import CioTracking
 import Foundation
-import UIKit
 #if canImport(UserNotifications)
 import UserNotifications
 
--- a/Sources/MessagingPushAPN/RichPush/MessagingPush+NotificationServiceExtension.swift
+++ b/Sources/MessagingPushAPN/RichPush/MessagingPush+NotificationServiceExtension.swift
@@ -1,6 +1,7 @@
 import CioMessagingPush
 import CioTracking
 import Foundation
+import UIKit
 #if canImport(UserNotifications)
 import UserNotifications
 
@@ -18,12 +19,14 @@ public extension MessagingPush {
         _ request: UNNotificationRequest,
         withContentHandler contentHandler: @escaping (UNNotificationContent) -> Void
     ) -> Bool {
+        
         guard let siteId = customerIO.siteId else {
             contentHandler(request.content)
             return false
         }
 
-        let diGraph = DITracking.getInstance(siteId: siteId)
+        let diGraph = DI.getInstance(siteId: siteId)
+        var store  = diGraph.globalDataStore
         let sdkConfig = diGraph.sdkConfigStore.config
         let jsonAdapter = diGraph.jsonAdapter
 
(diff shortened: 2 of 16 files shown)


Sources/MessagingPushAPN/RichPush/MessagingPush+NotificationServiceExtension.swift
```diff
@@ -40,6 +42,18 @@ public extension MessagingPush {
...
 
+        // If cioHandleBadge exists and value is true
+        if let cioHandleBadge = pushContent.cioHandleBadge, cioHandleBadge  {
+            
+            let badgeCount = store.badgeCount ?? 0
+            // Update badge count in payload
+            if let newRequest = request.content.mutableCopy() as? UNMutableNotificationContent {
+                newRequest.badge = NSNumber(value: badgeCount + 1)
+                store.badgeCount = badgeCount + 1
+                contentHandler(newRequest)
```
hownowstephen, line 53: this means `contentHandler` will be called twice I think? once here and once by the `RichPushRequestHandler`. should we instead be mutating the request that gets passed along, or handling this within the `RichPushRequestHandler`

Sources/MessagingPushAPN/RichPush/MessagingPush+NotificationServiceExtension.swift
```diff
@@ -40,6 +42,18 @@ public extension MessagingPush {
...
         }
 
+        // If cioHandleBadge exists and value is true
+        if let cioHandleBadge = pushContent.cioHandleBadge, cioHandleBadge  {
+            
+            let badgeCount = store.badgeCount ?? 0
+            // Update badge count in payload
+            if let newRequest = request.content.mutableCopy() as? UNMutableNotificationContent {
+                newRequest.badge = NSNumber(value: badgeCount + 1)
+                store.badgeCount = badgeCount + 1
```
hownowstephen, line 52: this only supports incrementing the badge, planning a follow-up to add decrementing, clearing and setting a discrete value?

Sources/MessagingPush/PushContent.swift
```diff
@@ -119,5 +128,6 @@ struct CioPushPayload: Codable {
     struct Push: Codable, AutoLenses {
         let link: String?
         let image: String?
+        let cioHandleBadge : Bool?
```
hownowstephen, line 131: seems a little restrictive, I imagined we'd want some representation of 
```
action -> one of `incr`, `decr`, `set` or `clear` // to increment, decrement, or set discrete values
value -> an integer
```
(note: this is just off the top of my head, not canon - feel free to play with a better API than this - for example, we may not need a decr-type action if we instead had `incr` (or a better named thing) support positive and negative numbers, and depending on what `set to zero` is interpreted as by the os, a `clear` may not be necessary either?)

also I've seen on my device number-less badges - how does that work?

Sources/MessagingPushAPN/RichPush/MessagingPush+NotificationServiceExtension.swift
```diff
@@ -40,6 +42,18 @@ public extension MessagingPush {
             return false
         }
 
+        // If cioHandleBadge exists and value is true
+        if let cioHandleBadge = pushContent.cioHandleBadge, cioHandleBadge  {
+            
+            let badgeCount = store.badgeCount ?? 0
+            // Update badge count in payload
+            if let newRequest = request.content.mutableCopy() as? UNMutableNotificationContent {
```
levibostian, line 50: Can you move this logic of creating `UNMutableNotificationContent` and modifying it into `PushContent` class? That class exists for that reason already.

Sources/MessagingPushAPN/RichPush/MessagingPush+NotificationServiceExtension.swift
levibostian: I agree in that the logic that modifies the badge count here should be moved into `RichPushRequestHandler`/`RichPushRequest`. 

This extension function built on top of `NotificationService`'s job is to delegate work to other functions, not perform logic in itself. 

I am realizing now that we don't document that very well. How we use extension functions for convenience and not as a way to perform logic. Code gets re-used better and is easier to test when the logic is in a SDK class instead of in an extension function.

Sources/MessagingPushAPN/RichPush/MessagingPush+NotificationServiceExtension.swift
```diff
@@ -40,6 +42,18 @@ public extension MessagingPush {
             return false
         }
 
+        // If cioHandleBadge exists and value is true
+        if let cioHandleBadge = pushContent.cioHandleBadge, cioHandleBadge  {
+            
+            let badgeCount = store.badgeCount ?? 0
```
levibostian, line 48: What is the use case for getting the stored badge count from global storage? What is the app has a badge count already or the global storage value is out-of-date?

Sources/Tracking/autogenerated/AutoDependencyInjection.generated.swift
```diff
@@ -70,170 +71,165 @@ public enum DependencyTracking: CaseIterable {
...
- We must use 1+ different graphs because of the hierarchy of modules in this SDK.
- Example: You can't add classes from `Tracking` module in `Common`'s DI graph. However, classes
- in `Common` module can be in the `Tracking` module.
+ We must use 1+ different graphs because of the hierarchy of modules in this SDK. 
+ Example: You can't add classes from `Tracking` module in `Common`'s DI graph. However, classes 
+ in `Common` module can be in the `Tracking` module. 
  */
-public class DITracking {
-    private var overrides: [DependencyTracking: Any] = [:]
+ public class DI {
```
levibostian, line 82: I don't understand why the class name changed here from `DITracking`?

Sources/Tracking/autogenerated/AutoDependencyInjection.generated.swift
```diff
@@ -8,59 +8,60 @@ import Foundation
...
 
-/**
- enum that contains list of all dependencies in our app.
- This allows automated unit testing against our dependency graph + ability to override nodes in graph.
+/** 
+ enum that contains list of all dependencies in our app. 
+ This allows automated unit testing against our dependency graph + ability to override nodes in graph. 
  */
-public enum DependencyTracking: CaseIterable {
+ enum Dependency: CaseIterable {
```
levibostian, line 60: I don't understand why the class name changed here from `DependencyTracking`?

Sources/Tracking/autogenerated/AutoMockable.generated.swift
```diff
@@ -8,389 +8,376 @@ import FoundationNetworking
...
     func resetAll() {
-        mocks.forEach {
+        self.mocks.forEach {
             $0.reset()
         }
     }
 }
 
-public protocol TrackingMock {
+public protocol Mock {
```
levibostian, line 96: I don't understand why the class name changed here from `TrackingMock`?

levibostian:
Closing as it's old and out-dated.

PR #86, opened 2021-12-03
docs: trim readme to basic instructions

glosier:
I _think_ this is what we want 🤞: I trimmed the instructions down to _just_ the installation path and linked to customer.io/docs/sdk/ios in a few places. We have all that information in /docs now.


README.md
```diff
@@ -44,6 +46,17 @@ https://github.com/customerio/customerio-ios.git
 
 ![](docs/img/xcode_spm_select_sdks.jpeg)
 
+### Available SDK packages
+
+Here are the list of the SDKs that you can install. You can find more details on both in [our SDK documentation](/docs/sdk/ios/).
```
levibostian, line 51: > You can find more details on both in [our SDK documentation](/docs/sdk/ios/)

I am wondering if this provides any value to the docs here? When I read docs for products, I find it annoying when someone says, "Learn more" and sends me to the landing page where I am expected to browse my way to where I need to be. 

One idea could be that instead of having a generic "learn more about all the products" here, we instead make `Tracking`, `MessagingPush`, etc text in the table below link to each page in the docs talking about that product?

README.md
```diff
@@ -8,13 +8,15 @@
 ![min ios version is 13](https://img.shields.io/badge/min%20iOS%20version-13-blue)
 [![Contributor Covenant](https://img.shields.io/badge/Contributor%20Covenant-2.0-4baaaa.svg)](code_of_conduct.md) 
 
-# Customer.io iOS SDK
+# Summary
```
levibostian, line 11: I prefer the previoius "Customer.io iOS SDK" as it is a good header introducing the project. This is a good header for introducing the text below. 

One idea is change the markdown to:

```markdown
# Customer.io iOS SDK

### Summary 
...
```

I like the way we do it in our [ruby SDK](https://github.com/customerio/customerio-ruby). Maybe something similar to that?

README.md
```diff
@@ -44,6 +46,17 @@ https://github.com/customerio/customerio-ios.git
 
 ![](docs/img/xcode_spm_select_sdks.jpeg)
 
+### Available SDK packages
```
levibostian, line 49: 🤔 What is the value of this section? 

Instead of us introducing each SDK from a technical standpoint, would it be better to instead introduce each *feature* of the SDK? In the docs for each feature, example FCM messaging, our docs can tell you "Make sure to install the FCM messaging SDK to use this feature...".

README.md
glosier: That's fair. I should probably also alias some /docs/ pages to to the name of the package.

README.md
glosier: Yeah, when I wrote this, I was questioning how deep we want this to go and did the bare minimum. Rather than "available sdk packages", it could be a feature support/package matrix—here's what we support and which package(s) you need to install to support the feature.

README.md
glosier: I totally forgot that the repo title isn't injected at the top of the readme. I like maintaining the semantic order of headings, but this sounds good to me otherwise

README.md
```diff
@@ -18,13 +18,18 @@ Before you get started, keep in mind:
...
 To get started, you need to install and initialize the relevant SDK packages in your project. 
 
-To minimize our SDK's impact on your app's size, we offer multiple, separate SDKs. You should only install the packages that you need for your project. 
+We've separated our SDK into packages to minimize our impact on your app's size. You should only install the packages that you need for your project. 
 
-> Tip: Check out our [sample iOS app, Remote Habits](https://github.com/customerio/RemoteHabits-iOS), for a real-world example of how to use our SDK. 
+| Package | Required? | Description |
+| :-- | :---: | :--- |
+| Tracking | Yes | [`identify`](https://customer.io/docs/sdk/ios/identify/) people/devices and [send events](https://customer.io/docs/sdk/ios/track-events/) (to trigger campaigns, track metrics, etc). |
+| MessagingPushAPN | No | [Push](https://customer.io/docs/sdk/ios/push/) and [rich push](https://customer.io/docs/sdk/ios/rich-push/) notifications using Apple's Push Notification service (APNs). |
```
levibostian, line 30:
```suggestion
| MessagingPushAPN | No | [Push](https://customer.io/docs/sdk/ios/push/) and [rich push](https://customer.io/docs/sdk/ios/rich-push/) notifications using Apple's Push Notification service (APNs). |
| MessagingPushFCM | No | [Push](https://customer.io/docs/sdk/ios/push/) and [rich push](https://customer.io/docs/sdk/ios/rich-push/) notifications using Firebase Cloud Messaging (FCM). |
```

levibostian (approved):

PR #91, opened 2021-12-07
feat: automatically run queue after delay

levibostian:
At this time, the background queue only runs if there are >= 10 tasks added to the queue. That is not enough of an opportunity to make sure tasks run. Now, we have added a delay so the queue will be scheduled to run even after only adding 1 single task.


hownowstephen:
https://github.com/customerio/customerio-ios/compare/e2ed358620e3c9f6ddc754bba71e599675042e17...be6ae3d9dca1fc935fd607dc1d79ff2b14c8d341

hownowstephen (approved):

Sources/Tracking/Background Queue/Queue.swift
```diff
@@ -46,55 +46,87 @@ public class CioQueue: Queue {
...
-            logger.verbose("Automatically running background queue")
+            // cancel timer if one running since we will run the queue now
+            queueTimer.cancel()
 
             // not using [weak self] to assert that the queue will complete and callback once started.
             // this might keep this class in memory and not get garbage collected once customer is done using it
             // but it will get released once the queue is done running.
             runRequest.start {
-                self.logger.verbose("Automatic running background queue completed")
+                self.logger.info("queue completed all tasks")
```
hownowstephen, line 115: thinking about it some more, what makes sense to me is:
- try to send all messages now
- if any fail with a 40x, ignore them
- if we hit a 50x, halt and schedule another run

PR #93, opened 2021-12-13
fix: queue tasks require success before it can run

levibostian:
Use case:
* Customer identifies profile A. Task added to queue. 
* Customer registers push device token to profile A. Task added to queue. 
* Queue runs. Fails to run profile A identification. 
* Queue tries to run register push device token task <----- Bad! The API doesn't know about profile A yet! 

This PR addresses the above use case through a concept of *grouping*. In the queue, you can group tasks together so if a task in the queue requires a queue task to run successfully, the task will wait. 

With this PR, the above example would be altered to:
* Customer identifies profile A. Task added to queue. 
* Customer registers push device token to profile A. Task added to queue. 
* Queue runs. Fails to run profile A identification. 
* Queue notices that register device token queue task is grouped with identify profile A which failed. The queue will skip running this register task and will try again later.


Sources/Tracking/Background Queue/QueueRunRequest.swift
```diff
@@ -53,7 +65,8 @@ public class CioQueueRunRequest: QueueRunRequest {
             let success = storage.delete(storageId: nextTaskStorageId)
             logger.debug("deleted task \(nextTaskStorageId) success: \(success)")
 
-            return goToNextTask(query: query, queryTotalNumberTasks: queryTotalNumberTasks)
+            return goToNextTask(query: query, queryTotalNumberTasks: queryTotalNumberTasks,
+                                lastFailedTask: nextTaskToRunInventoryItem)
```
hownowstephen, line 69: seems like this case should get a `nil` for `lastFailedTask` because it is trashed and shouldn't be considered as blocking criteria

Sources/Tracking/Background Queue/QueueRunRequest.swift
```diff
@@ -11,14 +11,22 @@ public class CioQueueRunRequest: QueueRunRequest {
...
     private let logger: Logger
+    private let queryRunner: QueueQueryRunner
 
     private let shortTaskId: (String) -> String = { $0[0 ..< 5] }
 
-    init(runner: QueueRunner, storage: QueueStorage, requestManger: QueueRequestManager, logger: Logger) {
+    init(
+        runner: QueueRunner,
+        storage: QueueStorage,
+        requestManger: QueueRequestManager,
```
hownowstephen, line 21: very festive but 
```suggestion
        requestManager: QueueRequestManager,
```
is probably better

Sources/Tracking/Background Queue/QueueRunRequest.swift
```diff
@@ -35,14 +43,18 @@ public class CioQueueRunRequest: QueueRunRequest {
         runTasks(query: inventory, queryTotalNumberTasks: inventory.count)
     }
 
-    private func runTasks(query: [QueueTaskMetadata], queryTotalNumberTasks: Int) {
-        if query.isEmpty { // we hit the end of the current inventory. Done!
+    private func runTasks(
+        query: [QueueTaskMetadata],
```
hownowstephen, line 47: `query` here feels like a confusing name even moreso now with the filtering

Sources/Tracking/Background Queue/QueueRunRequest.swift
```diff
@@ -92,15 +108,20 @@ public class CioQueueRunRequest: QueueRunRequest {
...
-            return self.goToNextTask(query: query, queryTotalNumberTasks: queryTotalNumberTasks)
+                return self.goToNextTask(query: query, queryTotalNumberTasks: queryTotalNumberTasks,
+                                         lastFailedTask: nextTaskToRunInventoryItem)
+            }
         }
     }
 
-    private func goToNextTask(query: [QueueTaskMetadata], queryTotalNumberTasks: Int) {
+    private func goToNextTask(
+        query: [QueueTaskMetadata],
```
hownowstephen, line 119: this is more like `queue` and `queueTotalNumberTasks` (or just `queueLength`)

Sources/Tracking/Background Queue/QueueQueryRunner.swift
```diff
@@ -0,0 +1,50 @@
...
+
+        let criteria = getCriteria(lastFailedTask: lastFailedTask)
+
+        return queue.first(where: { doesTaskPassCriteria($0, criteria: criteria) })
+    }
+
+    private func getCriteria(lastFailedTask: QueueTaskMetadata) -> QueueQueryCriteria {
+        var criteria = QueueQueryCriteria(excludeGroups: nil)
+
+        if let groupsToExclude = lastFailedTask.groupsParent {
```
hownowstephen, line 26: since this is only considering the last failed task, my reading of it is that if you had (using my suggested terminology - `taskName` below is equiv of `groupsParent` and `blockingTaskNames` is `groupsChild`

```
t1,taskName:X,failed
t2,blockingTaskNames:[X],skipped
t3,taskName:Y,failed
t4,blockingTaskNames:[X]
```

then we would execute `t4` because I don't see anything in here accumulating a list of `taskNames` that are now blocked

Sources/Tracking/Background Queue/Type/QueueTaskMetadata.swift
```diff
@@ -3,13 +3,22 @@ import Foundation
 /// Pointer to full queue task in persistent storage.
 /// This data structure is meant to be as small as possible with the
 /// ability to hold all queue task metadata in memory at runtime.
-public struct QueueTaskMetadata: Codable, Equatable {
+public struct QueueTaskMetadata: Codable, Equatable, AutoLenses {
     let taskPersistedId: String
     let taskType: String
+    /// Groups that this task belongs to where this task is the parent of the group.
+    /// Tasks can be the parent of 0+ groups
```
hownowstephen, line 10: do you have an idea of a case where we'd want something to parent multiple groups? I don't really know that it's a use-case we need to support 🤔

hownowstephen (approved):

Sources/Tracking/CustomerIOImplementation.swift
```diff
@@ -94,7 +94,12 @@ public class CustomerIOImplementation: CustomerIOInstance {
 
         let queueTaskData = IdentifyProfileQueueTaskData(identifier: identifier,
                                                          attributesJsonString: jsonBodyString)
-        let queueStatus = backgroundQueue.addTask(type: QueueTaskType.identifyProfile.rawValue, data: queueTaskData)
+
+        let queueStatus = backgroundQueue.addTask(type: QueueTaskType.identifyProfile.rawValue,
+                                                  data: queueTaskData,
+                                                  groupsParent: [QueueTaskGroup
+                                                      .identifiedProfile(identifier: identifier).string],
+                                                  groupsChild: nil)
```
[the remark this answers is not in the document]
levibostian: Tried something new in [this commit](https://github.com/customerio/customerio-ios/pull/93/commits/4fecb6456c50dedb5127b0454c2dfff37ae92a21).

```
addTask(type: X, data: y, groupStart: .identify(), blockingGroups: [.registerDeviceToken()])
```

Sources/Tracking/Background Queue/QueueRunRequest.swift
levibostian: I didn't use `nil` because if a task can't be found in persistent storage, then it's a failed execution just like failing a HTTP request. 

I do believe that we should mark this task not found as `lastFailedTask`, but we should *not* be deleting tasks from persistent storage if it's not found. 

The way the queue is designed today is that *individual tasks* are only deleted if they run successfully. In the future we may decide to also delete stale tasks in the queue but we should be deleting *groups* of stale tasks.

levibostian: Made change in https://github.com/customerio/customerio-ios/pull/93/commits/f314e936c330e83ef36b6041c3492e76152ab91d

Sources/Tracking/Background Queue/QueueRunRequest.swift
levibostian: fixed

Sources/Tracking/Background Queue/QueueRunRequest.swift
levibostian: Agree. Renamed to `queueInventory`

Sources/Tracking/Background Queue/QueueQueryRunner.swift
levibostian: Good catch. This is now fixed where we keep track of *all* failed group names not just the last one.

Sources/Tracking/Background Queue/Type/QueueTaskMetadata.swift
levibostian: No, I can't think of a scenario where you would need to today. 

In the future if we ever decided to change from `groupStart: String` to `groupStart: [String]` it would require us writing some code to migrate the queue saved in persistent storage on the device. 

I would like to avoid having to write this migration code but we probably need to write it at some time in the future anyways as we will more then likely have a breaking change at some point. 

So, I changed the to only allowing 0 or 1 `groupStart` and if we need to change it in the future, we will migrate the storage to allow it.

Sources/Tracking/CustomerIOImplementation.swift
```diff
@@ -97,9 +97,7 @@ public class CustomerIOImplementation: CustomerIOInstance {
 
         let queueStatus = backgroundQueue.addTask(type: QueueTaskType.identifyProfile.rawValue,
                                                   data: queueTaskData,
-                                                  groupsParent: [QueueTaskGroup
-                                                      .identifiedProfile(identifier: identifier).string],
-                                                  groupsChild: nil)
+                                                  groupStart: .identifiedProfile(identifier: identifier))
```
hownowstephen, line 100: `groupStart` doesn't feel quite right either, since it's more like a critical task in the group, doesn't necessarily need to be the start.. but unless another idea comes to mind I'm fine with it

Sources/Tracking/Background Queue/QueueRunRequest.swift
hownowstephen: I think this results in a stuck point in the inventory, no? if Y and Z depend on task X that can't be found, we should stop trying to find X and just let Y & Z continue, unless there's some way in which we expect X may be found on a subsequent run?

hownowstephen (approved):
looking good, only remaining concern about handling of missing data

Sources/Tracking/Background Queue/QueueRunRequest.swift
levibostian: Let's say that X is the start of a group and Y and Z are members of the group. So, Y and Z depend on X running successfully. 

If X cannot be found in the file system, then we should *not* attempt to run Y or Z since they depend on X. 

Let's say that X ran successfully but Y was not found in the file system. Z will still execute because the queue run request code only skips a task if *the a task that failed was the start of a group*. 

So, the queue will only be *stuck* here if X is not found in the file system and X is the start of a group. In the case, we should come back again to the idea of deleting stale tasks in the queue.

PR #94, opened 2021-12-13
feat: add [String:Any] support to identify & track

hownowstephen:
uses the `StringAnyEncodable` implementation from https://github.com/customerio/customerio-ios/pull/82 to add support for `[String:Any]` to our identify and track calls


Sources/Tracking/CustomerIO.swift
```diff
@@ -88,16 +88,35 @@ public extension CustomerIOInstance {
...
+    ) {
+        identify(identifier: identifier, body: StringAnyEncodable(body), onComplete: onComplete,
+                 jsonEncoder: jsonEncoder)
+    }
+
+    func track(
+        name: String,
+        data: [String: Any],
+        jsonEncoder: JSONEncoder?,
+        onComplete: @escaping (Result<Void, CustomerIOError>) -> Void
```
levibostian, line 105:
```suggestion
        onComplete: @escaping (Result<Void, CustomerIOError>) -> Void,
        jsonEncoder: JSONEncoder? = nil
```

It's best to put parameters with default vales after all non-default params.

levibostian (approved):

PR #96, opened 2021-12-14
feat: subsequent identifies without passing identifier

hownowstephen:
Adds support for `identify(body: [String:Any]) {}` and `identify(body: Encodable) {}` which can be used after an initial identify has been called

### DIFF
--- a/Sources/Tracking/CustomerIO.swift
+++ b/Sources/Tracking/CustomerIO.swift
@@ -3,6 +3,8 @@ import Foundation
 public protocol CustomerIOInstance: AutoMockable {
     var siteId: String? { get }
 
+    var identifier: String? { get }
+
     // sourcery:Name=identify
     func identify<RequestBody: Encodable>(
         identifier: String,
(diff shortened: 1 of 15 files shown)


Sources/Tracking/CustomerIO.swift
```diff
@@ -3,6 +3,8 @@ import Foundation
 public protocol CustomerIOInstance: AutoMockable {
     var siteId: String? { get }
 
+    var identifier: String? { get }
```
levibostian, line 6: I think an in-memory instance of `identifier` isn't very safe as instances of `CustomerIOInstance` can be destroyed and re-created during the lifecycle of an app. 

Instead, I would use [`ProfileStore`](https://github.com/customerio/customerio-ios/blob/4db3dc097e3cf127264ef2d0830499bd10966b4d/Sources/Tracking/Store/ProfileStore.swift)

Sources/Tracking/CustomerIO.swift
```diff
@@ -98,6 +100,32 @@ public extension CustomerIOInstance {
                  jsonEncoder: jsonEncoder)
     }
 
+    func identify(
```
levibostian, line 103: IMO, having 2 functions called `identify` is confusing to the customer wondering when to use what function. I propose having `identify()` stay as it is in the SDK to identify a customer profile and something like:

```swift
customerio.setProfileAttributes(["first_name": "Dana"], onComplete: ...)

// with the background queue, we can do synchronous methods like:
customerio.profileAttributes = ["first_name": "Dana"]
```

Some function naming that implies that 1 function's job is to identify a profile and 1 function's job is to alter attributes of an already identified profile.

hownowstephen: 👍 had to get past my preconceive notions of `identify` on this one, since it's always been the endpoint name, but I agree, and since it's `identify + logic` makes sense to treat it slightly different

Sources/Tracking/CustomerIO.swift
hownowstephen: this was backed by the `profileStore` owned by the `identifyRepository` - I moved it instead to be backed by one on `CustomerIOImplementation` - in either case, my understanding was that Swift didn't cache these values, but called out to the underlying method when defined, but maybe I'm wrong?

levibostian: > my understanding was that Swift didn't cache these values, but called out to the underlying method when defined, but maybe I'm wrong?

I think what you're asking is "Looking at the `ProfileStore` get/set, does Swift call get/set once and cache the returned result?"

```swift
    public var identifier: String? {
        get {
            // ???? Does Swift cache this returned result? 
            keyValueStorage.string(.identifiedProfileId)
        }
        set {
            // ???? Does Swift cache `newValue`?
            keyValueStorage.setString(newValue, forKey: .identifiedProfileId)
        }
    }
```

Since we have explicitly defined a get/set function, Swift will always call those functions. It's like `var identified` is a function, not a class field. 

One way to do a cache is: `lazy var ...`

levibostian: Since the get/set functions are using the thread-safe keyValueStorage class, we can use non-singleton instances of `ProfileStore` all over the SDK and we will always get the latest value across all instances.

levibostian (approved):

Issue #105, opened 2022-01-04
Add timestamp to background queue inventory items 


levibostian:
Internally, we have talked about the potential use case of needing to delete stale/outdated tasks in the background queue. For example: if a task has a 403 and it never gets addressed, we would rather delete it then let it fail indefinitely. 

In order for us to know what tasks to delete, we need to know how old the task is. When the task was *added* to the queue. That's what this task addresses. 

- [ ] Add a `Date` to `QueueTaskMetadata` that gets populated when a task is added to the queue.

PR #106, opened 2022-01-06
fix: add createdAt timestamp to added queue tasks 

levibostian:
Closes: https://github.com/customerio/customerio-ios/issues/105


hownowstephen (approved):

Sources/Tracking/Extensions/DateExtensions.swift
```diff
@@ -0,0 +1,15 @@
...
+    // When decoding code to and from JSON, we lose milliseconds from the Date() objects.
+    // This can be difficult for unit testing because you might have to compare two
+    // identical Date() objects but one of the instances has the milliseconds removed.
+    // Use this to construct a new Date() object without the milliseconds to use in tests.
+    static var nowNoMilliseconds: Date {
+        let formatter = DateFormatter()
+        formatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ssZ"
+        let nowString = formatter.string(from: Date())
+
+        return formatter.date(from: nowString)!
```
ami-aman, line 13: The app/SDK might crash if due of any reason, the return `formatter.date(from: nowString)!` is nil. Better to always safely unwrap the values. Looking at the use case, the chances are rare but always good to write safe code specially while unwrapping.

PR #107, opened 2022-01-07
fix: more safely handle 5xx, 401 status codes

levibostian:
Handle 5xx, 401, 4xx status codes uniquely. 

Tasks to be done to complete this task:
- [x] When receiving a 5xx response:
* Begin an exponential backoff retry on the HTTP task that returned back the 5xx error. The policy will be 100ms, 200ms, 400ms, 800ms, 1600ms, 3200ms. (I made up this policy by reading online that exponential policies usually start at 100ms and have 6 retries. The policy I have here is arbitrary and up for debate.)
* After these 6 attempts, if the queue is still receiving a 5xx response then the queue will terminate early and will sleep for 5 minutes where the queue will not run during this time. 
* After the 5 minutes, the queue is able to be run as normal. No memory of any errors prior. 
- [x] When receiving a 401 response:
* The queue will terminate early and will sleep for 5 minutes as above. 
- [x] Any other 4xx error
* The queue will log the error as it's more then likely a SDK developer error or an error by the customer. 
* The queue will behave as today when the queue experiences an error. Skip that task and any task in it's group and run the next task in the queue. No sleeping of the queue. 

...technically, the queue is not terminated early. Instead, the queue will attempt to run the rest of the queue tasks but will fail instantly because HTTP requests are paused. Same result but and easier to implement.


Sources/Tracking/Service/HttpClient.swift
```diff
@@ -165,4 +174,74 @@ extension CIOHttpClient {
...
+        case 500 ..< 600:
+            if let sleepTimeMilliseconds = retryPolicy.nextSleepTimeMilliseconds {
+                logger
+                    .debug("""
+                    Encountered \(statusCode) HTTP response.
+                    Sleeping \(sleepTimeMilliseconds) and then retrying.
+                    """)
+
+                retryPolicyTimer.schedule(milliseconds: sleepTimeMilliseconds) {
+                    self.request(params, onComplete: onComplete)
```
hownowstephen, line 227: creates a whole different pathway to a request being called, `QueueRunRequest` might be a better context to be handling this sort of logic, versus having a completely separate scheduler here. My imagining was:

- error bubbles back up to the queue. 500 means server unhappy, stop the queue for now
- retries are handled by rescheduling the entire bg queue to the next exponential backoff time, following the retry policy, which then is reset when we get a non-500 / get paused

In this scenario, for 100 unrelated requests in the queue, would we end up spawning 100 timers on their own retry loops? or is this a blocking call?

hownowstephen: https://github.com/customerio/customerio-ios/blob/levi/background-q-errorhandling/Sources/Tracking/Background%20Queue/QueueRunRequest.swift#L80

this was mostly what I got concerned about. From reading just the HTTPClient code, it's not clear that the queue won't continue until `onComplete` is called, which made me concerned that the bg queue might move ahead while this gets deferred and get us into a bit of a mess.

I still don't love having the retry policy & associated timer hidden inside the HttpClient, versus handling this higher up in the call stack, but my concern about multiple spawned timers is not a problem at least.

hownowstephen: fine to leave as-is for now, can revisit the architecture of it at a later date if need be

hownowstephen (approved):

Issue #110, opened 2022-01-12
Push metrics not being reported to the API. No HTTP request being made. 


levibostian:
Version: 1.0.0-alpha.25

Reproduce:
Send yourself a rich push and click on it. Wait until background queue runs. You will see a stacktrace which says

```
Decode key not found. Key: CodingKeys(stringValue: "delivery_id", intValue: nil),
Json path: [], json: {"delivery_id":"RO7nBgAFAX5PajCwhzxlKAfAzMvgNA==","device_id":"c2b2c3398e803626d3bd565e24c9c07208bb0ad5907a1c6ab350fc7cc5545628","event":"opened","timestamp":1642009883}
```

PR #111, opened 2022-01-12
fix: opened push metrics being sent to API again

levibostian:
Fixes: https://github.com/customerio/customerio-ios/issues/110

### DIFF
--- a/Sources/MessagingPush/Service/Request/MetricRequest.swift
+++ b/Sources/MessagingPush/Service/Request/MetricRequest.swift
@@ -8,15 +8,21 @@ public enum Metric: String, Codable {
 
 // https://customer.io/docs/api/#operation/pushMetrics
 internal struct MetricRequest: Codable {
-    let deliveryID: String
+    let deliveryId: String
     let event: Metric
     let deviceToken: String
     let timestamp: Date
 
     enum CodingKeys: String, CodingKey {
-        case deliveryID = "delivery_id"
+        case deliveryId
         case event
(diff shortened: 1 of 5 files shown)


Sources/MessagingPush/Service/Request/MetricRequest.swift
```diff
@@ -8,15 +8,21 @@ public enum Metric: String, Codable {
...
     let event: Metric
     let deviceToken: String
     let timestamp: Date
 
     enum CodingKeys: String, CodingKey {
-        case deliveryID = "delivery_id"
+        case deliveryId
         case event
-        case deviceToken = "device_id"
+        case deviceToken = "deviceId"
```
ami-aman, line 19: Are you sure key is `deviceId` because previously it was `device_id` and also in the test case i see `device_id` as expected json.
`{"delivery_id":"123","device_id":"234","event":"opened","timestamp":1642018466}`

levibostian: Yes I am sure. The test functions test the functionality works as intended. 

I tried to explain why `deviceId` works [in the PR description](https://github.com/customerio/customerio-ios/pull/111#issue-1100764293)

hownowstephen: :grimacing: not a fan of this sort of magic, but seems from some research to be the swift-y way to do it, given that the `convertFromSnakeCase` is a builtin

hownowstephen (approved):

PR #114, opened 2022-01-14
fix: background queue timer scheduling and running

levibostian:
Fixes: https://github.com/customerio/customerio-ios/issues/115

There were a few problems with the queue timer. 
1. There was a bug converting seconds to milliseconds and back to seconds. This has been fixed by removing the complexity of using seconds and milliseconds and now *only* using seconds. No conversion is done between the two now. 
2. The `SimpleTimer` may have scheduled a timer, but it sometimes was not triggering when the timer finished. This has been fixed by using `DispatchQueue.main.async {` in the timer for scheduling and canceling.


ami-aman (approved):
Looks good 👍

hownowstephen (approved):

PR #120, opened 2022-01-18
fix: automatic screenview tracking correct siteid

levibostian:
This fixes a bug found in the SDK where screenview tracking requests would be ignored. 

See logs from the SDK:
```
info	23:39:28.276844+0200	Remote Habits	ℹ️ track event clicked_habit
info	23:39:28.277179+0200	Remote Habits	ℹ️ adding queue task trackEvent
info	23:39:28.863286+0200	Remote Habits	ℹ️ track screen view event HabitDetail
info	23:39:28.863558+0200	Remote Habits	ℹ️ ignoring track screen view event HabitDetail because no profile currently identified
```

The code that tracks events and screenview events is almost identical. So logs like ^^ pointed me to the automatic tracking code.


Sources/Tracking/CustomerIOImplementation.swift
```diff
@@ -177,7 +177,11 @@ public class CustomerIOImplementation: CustomerIOInstance {
         data: RequestBody,
         jsonEncoder: JSONEncoder? = nil
     ) {
-        trackEvent(type: .screen, name: name, data: data, jsonEncoder: jsonEncoder)
+        if sdkConfigStore.config.autoTrackScreenViews {
```
hownowstephen, line 180: this seems wrong - you should definitely be allowed to track screenviews regardless of whether automatic tracking is enabled

Sources/Tracking/CustomerIO.swift
```diff
@@ -326,4 +335,25 @@ public class CustomerIO: CustomerIOInstance {
...
     }
+
+    public func automaticScreenView(
+        name: String,
+        data: [String: Any]
+    ) {
+        automaticScreenView(name: name, data: StringAnyEncodable(data))
+    }
+
+    // Designed to be called from swizzled methods for automatic screen tracking.
```
hownowstephen, line 346: does this mean that in RH we'll end up continuing to track screenviews in other workspaces when the siteId changes? or do we trash the siteId from the datastore when switching?

Sources/Tracking/CustomerIOImplementation.swift
levibostian: Good catch. The automatic tracking is already taken care of when swizzling is enabled.

Sources/Tracking/CustomerIO.swift
levibostian: > does this mean that in RH we'll end up continuing to track screenviews in other workspaces when the siteId changes?

Yes it will. 

The problem is that swizzled methods have no memory of what siteId to use. I can't think of any way to solve the problem?

hownowstephen: we could just flag the most recent siteID in the datastore and decide that auto screen views are only evaluated for the most recently enabled siteId and none others. If someone in future has a use case where they legitimately need CIO creds for two workspaces active for this feature at once, we can re-eval then

levibostian: > If someone in future has a use case where they legitimately need CIO creds for two workspaces active for this feature at once, we can re-eval then

We could keep track of *active* workspaces in the SDK. When the SDK initializes when the app has a cold start, *active* SDKs is empty. Then, every time that an instance of CIO is iniaitlized, we keep track at runtime of *active* siteIds. 

So, we will will *not* report automatic screen view events to workspaces that have not been used in a while.

hownowstephen (approved):

PR #125, opened 2022-01-19
docs: remove alpha references. update install instructions

glosier:
Deleted some unused images while I was at it.


README.md
```diff
@@ -35,23 +33,11 @@ We've separated our SDK into packages to minimize our impact on your app's size.
...
-https://github.com/customerio/customerio-ios.git
-```
-
-3. Select the version that you want to install. While the SDK is its alpha stage, we recommend that you install *an exact* version of the SDK instead of indicating a range. This prevents you from automatically upgrading to a newer alpha version and possibly installing breaking changes on your code base. 
-
-![in xcode select Exact from dropdown when selecting the version of the SDK to install](docs/img/xcode_spm_install_version_blur.png)
-
-4. Lastly, choose the SDK products that you want to install. You can start by selecting `Tracking` for now and adding others later if you need them. 
-
-![](docs/img/xcode_spm_select_sdks.jpeg)
```
levibostian, line 54: We might want to consider keeping these docs. Because we offer 2+ different SDKs, it might be confusing to some customers when they arrive to this screen and wonder, "What ones do I select?". Knowing they just need to select `Tracking` to start with is helpful.

levibostian:
I'll merge this for now as we are in 1.0.0 and it's good enough to ship. 

@glosier, if you don't mind check out [my comment](https://github.com/customerio/customerio-ios/pull/125#discussion_r788058437) to see if you think we should make another change to the docs or not. I would love to hear your ideas.

README.md
glosier: Sorry, I missed this before you headed out! Yeah, we can put this back and update the screenshot.

PR #130, opened 2022-01-19
fix: add siteid to logs help with multi-workspace

levibostian:
When the SDK is used with multiple CIO Workspaces (such as in the Remote Habits app), the SDK logs are difficult to understand. When an event happens, it's hard to see if the event is being recorded for workspace A or workspace B. 

This change adds to the logs the siteId that the event happens for to make debugging issues easier.

### DIFF
--- a/Sources/Tracking/Util/Log.swift
+++ b/Sources/Tracking/Util/Log.swift
@@ -49,13 +49,15 @@ public class ConsoleLogger: Logger {
     private let logSubsystem = "io.customer.sdk"
     private let logCategory = "CIO"
 
+    private let siteId: SiteId
     private let sdkConfigStore: SdkConfigStore
 
     private var minLogLevel: CioLogLevel {
         sdkConfigStore.config.logLevel
     }
 
-    init(sdkConfigStore: SdkConfigStore) {
+    init(siteId: SiteId, sdkConfigStore: SdkConfigStore) {
+        self.siteId = siteId
         self.sdkConfigStore = sdkConfigStore
     }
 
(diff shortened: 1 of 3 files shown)


Sources/Tracking/Util/Log.swift
```diff
@@ -65,12 +67,15 @@ public class ConsoleLogger: Logger {
     private func printMessage(_ message: String, _ level: OSLogType) {
         if !minLogLevel.shouldLog(level) { return }
 
+        let logsPrefix = "(siteid:\(siteId.abbreviatedSiteId))"
```
hownowstephen, line 70: just to avoid too much visual clutter?

hownowstephen (approved):

ami-aman (approved):

Sources/Tracking/Util/Log.swift
levibostian: What is to avoid too much visual clutter? Sorry, I don't understand your question.

ami-aman: I think Stephen wants to ask you that did you use the variable `logsPrefix` to simplify the next line i.e.
`let messageToPrint = "\(logsPrefix) \(message)"` or what was the intention to have a separate variable for this.

levibostian: Ah! 

I think I created this variable during development to fix some issue I was having with getting the code to compile. It isn't needed anymore. I'll remove it as I agree it doesn't add any value here.

PR #143, opened 2022-02-23
feat: adding support for device attributes and custom device attributes

ami-aman:
[Add device attribute collection to SDK](https://github.com/customerio/issues/issues/6510)

Provide support in SDK to send default device attributes and also custom device attributes to customer's workspace

**Use case to enable automatic tracking of device attributes :**
> CustomerIO.config {
>             $0.autoTrackDeviceAttributes = true
>         }

**Use case to add custom device attributes :**
> CustomerIO.shared.deviceAttributes = ["company" : "cio", "checklist" : "complete"]

- [x] [Assign members of your team](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/requesting-a-pull-request-review) to review the pull request. 


Sources/Tracking/DeviceInfo.swift
```diff
@@ -1,15 +1,56 @@
...
+        case .sdkVersion:
+            return SdkVersion.version
+        case .deviceLocale:
+            return Locale.current.identifier
+        }
+    }
+}
+
+/// Provides additional async details about device the SDK is working on
+public class DeviceDetail {
```
levibostian, line 37: I find it confusing to have a class `DeviceInfo` and a class `DeviceDetail` where it seems to me that the only reason that they are not all in 1 class is because you can't make `pushSubscribed()` synchronous and therefore, cannot put it inside of the `DeviceInfo` enum. 

Is there a different way that we can implement capturing device attributes so avoid having 2 separate classes when they each seem to have the same purpose?

Sources/Tracking/DeviceInfo.swift
```diff
@@ -1,15 +1,56 @@
 import Foundation
 #if canImport(UIKit)
 import UIKit
+import UserNotifications
```
levibostian, line 4: Linux doesn't support `UserNotifications` so you have to [do this](https://github.com/customerio/customerio-ios/blob/0ca06183f9c0094019e3e58d0963b5931909c348/Sources/MessagingPush/PushContent.swift#L3)

Sources/MessagingPush/MessagingPushImplementation.swift
```diff
@@ -113,6 +124,28 @@ internal class MessagingPushImplementation: MessagingPushInstance {
                                     data: MetricRequest(deliveryId: deliveryID, event: event, deviceToken: deviceToken,
                                                         timestamp: Date()))
     }
+    func deviceAttributes(deviceToken: String, completionHandler: @escaping([String: String]?) -> Void) {
+        if !sdkConfigStore.config.autoTrackDeviceAttributes {
+            completionHandler(nil)
+            return
+        }
+        #if canImport(UIKit)
```
levibostian, line 132: The `completionHandler` is never called if `#if canImport(UIKit)` is false. 

Read my comment for `DeviceInfo` enum. You may not need this `#if canImport()` at all.

Sources/MessagingPush/MessagingPushImplementation.swift
```diff
@@ -61,13 +66,19 @@ internal class MessagingPushImplementation: MessagingPushInstance {
...
-                                    data: RegisterPushNotificationQueueTaskData(profileIdentifier: identifier,
-                                                                                deviceToken: deviceToken,
-                                                                                lastUsed: Date()),
-                                    groupStart: .registeredPushToken(token: deviceToken),
-                                    blockingGroups: [.identifiedProfile(identifier: identifier)])
+        deviceAttributes(deviceToken: deviceToken) { attributes in
+            var deviceAttributes = attributes ?? customAttributes
+            if let customDeviceAttributes = customAttributes, let defaultDeviceAttributes = attributes {
+                deviceAttributes = defaultDeviceAttributes.merging(customDeviceAttributes) { $1 }
+            }
```
levibostian, line 73: This code took me a while to read it to understand it. 

Could we use extensions or a data structure to make it easier to read? Have code that looks more like this as a result:
```swift
            var deviceAttributes = defaultDeviceAttributes ?? [:] 
            if let customAttributes = customAttributes {
                deviceAttributes = deviceAttributes.merge(customAttributes)
            }
```

Sources/Tracking/Store/SdkConfig.swift
```diff
@@ -44,6 +44,11 @@ public struct SdkConfig {
      */
     public var autoScreenViewBody: (() -> [String: Any])?
 
+    /**
+     Enable this property if you want SDK to automatic tracking of device attributes such as
+     operating system, device locale, device model, app version etc
+     */
+    public var autoTrackDeviceAttributes: Bool = true
```
levibostian, line 51: 1. Our squad decided to have the default value to be `false` like we do with screen view tracking. We want to avoid automatically enabling data collection and allow customers to decide when they are used. 

2. In the squad sync call a couple days ago, we mentioned that instead of having just 1 property for enabling/disabling default device attributes we have 1 property for every type of device attribute. So, something like this:
```swift
CustomerIO.config { 
  $0.deviceAttributeOSVersion = true
  $0.deviceAttributeDeviceModel = true
  ...
}
```

I am unsure about the naming to use? Maybe we do this instead:

```swift
CustomerIO.config { 
  $0.enableDefaultDeviceAttributes = [OSVersion, DeviceModel]
}
```

Also be sure to add to the [APITest](https://github.com/customerio/customerio-ios/blob/0ca06183f9c0094019e3e58d0963b5931909c348/Tests/Tracking/APITest.swift#L87-L95)

Sources/Tracking/hooks/ModuleHook.swift
```diff
@@ -26,3 +27,8 @@ public protocol QueueRunnerHook: AutoMockable {
     /// return `true` if the `task` belongs to that module.
     func runTask(_ task: QueueTask, onComplete: @escaping (Result<Void, CustomerIOError>) -> Void) -> Bool
 }
+
+// Hook to send custom device attributes to workspace
+public protocol DeviceAttributesHook: AutoMockable {
+    func customDeviceAttributesAdded(attributes: [String: String])
```
levibostian, line 33:
```suggestion
    func customDeviceAttributesAdded(attributes: [String: Any])
```

Tests/MessagingPush/MessagingPushImplementationTest.swift
```diff
@@ -42,6 +41,10 @@ class MessagingPushImplementationTest: UnitTest {
         profileStoreMock.identifier = givenIdentifier
         queueMock.addTaskReturnValue = (success: true, queueStatus: QueueStatus.successAddingSingleTask)
 
+        var config = SdkConfig()
+        config.autoTrackDeviceAttributes = true
+        sdkConfigStoreMock.config = config
```
levibostian, line 46: This is copy/pasted code in a few of the test functions. Maybe pulling this out into a function in this class would be handy:

```swift
private func enableDeviceAttributes() {
        var config = SdkConfig()
        config.autoTrackDeviceAttributes = true
        sdkConfigStoreMock.config = config
}
```

Tests/MessagingPush/MessagingPushImplementationTest.swift
```diff
@@ -122,4 +125,41 @@ class MessagingPushImplementationTest: UnitTest {
         XCTAssertEqual(actualQueueTaskData.event, givenEvent)
         XCTAssertEqual(actualQueueTaskData.deviceToken, givenDeviceToken)
     }
+    
+    // MARK: DeviceAttributes
+    func test_deviceAttributes_givenDeviceToken_expectDefaultDeviceAttributesDictionary() {
```
levibostian, line 130:
```suggestion
    func test_deviceAttributes_givenDefaultDeviceAttributesEnabled_expectGetDefaultDeviceMetrics() {
```

Tests/MessagingPush/MessagingPushImplementationTest.swift
```diff
@@ -122,4 +125,41 @@ class MessagingPushImplementationTest: UnitTest {
...
+            XCTAssertEqual(defaultAttributes!["deviceOs"], expectedOS)
+            XCTAssertEqual(defaultAttributes!["deviceLocale"], expectedLocale)
+            XCTAssertEqual(defaultAttributes!["deviceModel"], expectedModel)
+            XCTAssertEqual(defaultAttributes!["pushSubscribed"], "false")
+            XCTAssertEqual(defaultAttributes!["cioSdkVersion"], expectedSDKVersion)
+        }
+        #endif
+    }
+    
+    func test_deviceAttributes_givenDeviceToken_expectNil() {
```
levibostian, line 152:
```suggestion
    func test_deviceAttributes_givenDefaultDeviceAttributesDisabled_expectNil() {
```

Tests/MessagingPush/MessagingPushImplementationTest.swift
```diff
@@ -122,4 +125,41 @@ class MessagingPushImplementationTest: UnitTest {
...
+        var config = SdkConfig()
+        config.autoTrackDeviceAttributes = false
+        sdkConfigStoreMock.config = config
+
+        // This function returns `nil` in case
+        // `autoTrackDeviceAttributes` is set to `false`
+        messagingPush.deviceAttributes(deviceToken: givenDeviceToken) { defaultAttributes in
+            
+            XCTAssertNil(defaultAttributes)
+        }
```
levibostian, line 163: Since this code is async, you will need [your test function to wait for the async call to finish](https://github.com/customerio/customerio-ios/blob/8716b3391f2de135cd7f6a42950b81174e1be39b/Tests/Tracking/Service/HttpClientTest.swift#L53-L63)

Sources/Tracking/DeviceInfo.swift
```diff
@@ -1,15 +1,56 @@
...
+        case .customerAppName:
+            return Bundle.main.infoDictionary?["CFBundleName"] as? String ?? ""
+        case .customerAppVersion:
+            return Bundle.main.infoDictionary?["CFBundleShortVersionString"] as? String ?? ""
+        case .customerBundleId :
+            return Bundle.main.bundleIdentifier ?? ""
+        case .sdkVersion:
+            return SdkVersion.version
+        case .deviceLocale:
+            return Locale.current.identifier
```
levibostian, line 31: All of these do not require `UIKit` to collect these. 

Instead of:
```swift
#if canImport(UIKit)
import UIKit

public enum DeviceInfo {
  ... all enum cases in here
}
#endif 
```

we do
```swift
public enum DeviceInfo {
  var value: String? 
    case .deviceInfo:
       #if canimport(uikit)
         return UIDevice.deviceModelCode
        #else 
        return nil 
         #end 
  }
}
```

Another reason why this is great is because you can then remove the `#if canImport(UiKit)` statements in `MessagingPushImplementation.swift`:

```swift
#if canImport(UIKit)
        let deviceOS = DeviceInfo.osInfo.value
        let deviceModel = DeviceInfo.deviceInfo.value
        let appVersion = DeviceInfo.customerAppVersion.value
        let sdkVersion = DeviceInfo.sdkVersion.value
```

You can do:
```swift
if let deviceOs = DeviceInfo.osInfo.value {
  deviceAttributes["deviceOs"] = deviceOs
}
```

levibostian (changes requested):

Sources/MessagingPush/MessagingPushImplementation.swift
```diff
@@ -50,8 +50,13 @@ internal class MessagingPushImplementation: MessagingPushInstance {
      is no active customer, this will fail to register the device
      */
     public func registerDeviceToken(_ deviceToken: String) {
+        addDeviceAttributes(deviceToken: deviceToken)
+    }
+    /**
+     Adds device default and custom attributes and registers device token.
+     */
+    func addDeviceAttributes(deviceToken: String, customAttributes: [String: String]? = nil) {
```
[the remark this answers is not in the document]
ami-aman: 1. This is a common function to handle custom attributes and get default attributes. Parameter `customAttributes` is optional. Removing the default value will increase the effort by always adding [:] or `nil`. Look at the function call hierarchy to understand the reasoning better. This function is called from `registerDeviceToken` also which doesn't have customAttributes.
2. You are right, I made this a private function initially and mistakenly removed it. Thanks for pointing out. Changes made in recent commits

Sources/MessagingPush/MessagingPushImplementation.swift
```diff
@@ -113,6 +124,28 @@ internal class MessagingPushImplementation: MessagingPushInstance {
                                     data: MetricRequest(deliveryId: deliveryID, event: event, deviceToken: deviceToken,
                                                         timestamp: Date()))
     }
+    func deviceAttributes(deviceToken: String, completionHandler: @escaping([String: String]?) -> Void) {
```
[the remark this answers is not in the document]
ami-aman: ✔️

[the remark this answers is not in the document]
ami-aman: Test case has been written for this function hence not making it private. But removing unwanted parameter.

Sources/MessagingPush/hooks/ModuleHook.swift
```diff
@@ -3,6 +3,9 @@ import Foundation
 
 // sourcery: InjectRegister = "ModuleHookProvider"
 class MessagingPushModuleHookProvider: ModuleHookProvider {
+    var deviceAttributesHook: DeviceAttributesHook? {
```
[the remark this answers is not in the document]
ami-aman: ✔️

Sources/Tracking/DeviceInfo.swift
```diff
@@ -1,15 +1,56 @@
...
+/// Provides additional async details about device the SDK is working on
+public class DeviceDetail {
+    public init() {}
+    // Tells whether user has enabled or disabled push notifications for the app
+    public func pushSubscribed(completion: @escaping(Bool) -> Void) {
+        if let executableBundle = Bundle.main.infoDictionary?["CFBundleExecutable"] as? String,
+           executableBundle == "xctest" {
+            completion(false)
+            return
+        }
```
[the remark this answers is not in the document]
ami-aman: Test case for UNUserNotificationCenter fails when running on simulator. This seems to be an open issue. https://developer.apple.com/forums/thread/54187

Sources/Tracking/DeviceInfo.swift
ami-aman: How did I miss this one 🤔  Thanks !

Sources/Tracking/DeviceInfo.swift
ami-aman: `DeviceInfo` is an enum and `DeviceDetail` is a class. But i think i should club both of them together as even I find it absurd to have two entities performing almost same function. Doing it right away. Changes can be found in latest commit.

ami-aman: ✔️

Sources/MessagingPush/MessagingPushImplementation.swift
ami-aman: Makes sense. Done. 👍

Sources/Tracking/Store/SdkConfig.swift
ami-aman: Yes, we have to implement it in such a way that each property is configurable. But as far as I remember, in sync call I concluded by saying that right now the implementation has a group enable/disable and can be changed. I am planning to do this in a new PR and not block this one for the purpose ! What do you think ?

ami-aman: Tagging you @levibostian so that you can find this suggestion easily and share your thoughts.

+

Default should be `false` Thanks for pointing out

Tests/MessagingPush/MessagingPushImplementationTest.swift
ami-aman: Make sense ! 👍

Tests/MessagingPush/MessagingPushImplementationTest.swift
ami-aman: ✔️

Tests/MessagingPush/MessagingPushImplementationTest.swift
ami-aman: ✔️

Sources/MessagingPush/Background Queue/Task Data/RegisterPushNotificationQueueTaskData.swift
```diff
@@ -1,7 +1,8 @@
 import Foundation
 
 struct RegisterPushNotificationQueueTaskData: Codable {
-    let profileIdentifier: String
     let deviceToken: String
+    let profileIdentifier: String
     let lastUsed: Date
+    let attributes: [String: String]?
```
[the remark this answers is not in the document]
ami-aman: Got confirmation from Colby and made the necessary changes. Custom device attributes now accept `[String: Any]` and also the default attributes can have values of `Any` type.

Sources/Tracking/CustomerIOImplementation.swift
```diff
@@ -84,6 +83,16 @@ public class CustomerIOImplementation: CustomerIOInstance {
             identify(identifier: existingProfileIdentifier, body: StringAnyEncodable(newValue))
         }
     }
+    public var deviceAttributes: [String: String] {
```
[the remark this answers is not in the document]
ami-aman: ✅

Sources/MessagingPush/MessagingPushImplementation.swift
ami-aman: 👍 ✅

Sources/Tracking/DeviceInfo.swift
ami-aman: This has been updated to a new logic !

Sources/Tracking/Store/SdkConfig.swift
ami-aman: https://github.com/customerio/issues/issues/6510#issuecomment-1063800680

Tests/MessagingPush/MessagingPushImplementationTest.swift
ami-aman: ✅

Sources/MessagingPush/MessagingPushImplementation.swift
```diff
@@ -266,3 +311,11 @@ extension MessagingPushImplementation: ProfileIdentifyHook {
         deleteDeviceToken()
     }
 }
+
+extension MessagingPushImplementation: DeviceAttributesHook {
+    // Adds custom device attributes to background queue and sends to workspace
+    func customDeviceAttributesAdded(attributes: [String: Any]) {
+        guard let deviceToken = globalDataStore.pushDeviceToken else { return }
+        addDeviceAttributes(deviceToken: deviceToken, customAttributes: attributes)
```
levibostian, line 319: Because custom device attributes are only added when there is a push token, what would happen in this scenario:

* Custom device attributes set via `CustomerIO.instance.deviceAttributes = ["foo": "bar"]`
* No device token registered yet. 
* Device token registered. 

The custom device attributes will never get added. The request to set will get ignored.

levibostian (commented):
I don't see a lot of tests added for this PR. Is this on purpose?

Sources/MessagingPush/Background Queue/Task Data/RegisterPushNotificationQueueTaskData.swift
```diff
@@ -2,6 +2,5 @@ import Foundation
 
 struct RegisterPushNotificationQueueTaskData: Codable {
     let profileIdentifier: String
-    let deviceToken: String
-    let lastUsed: Date
+    let attributesJsonString: String?
```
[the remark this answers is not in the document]
ami-aman: It was formerly the way you are explaining it should be but doing that was returning hash error. Reason being, the request was being json-ed twice. With this approach, the request is all set to be sent as the payload in the request.

With your approach which is currently being used in [identifying profiles](https://github.com/customerio/customerio-ios/blob/66f5ec2b7138f5eade7a0c9d732646d2bb7ef1c9/Sources/Tracking/CustomerIOImplementation.swift#L113), the data is made ready and sent as payload as is.

Sources/MessagingPush/MessagingPushImplementation.swift
```diff
@@ -61,13 +66,28 @@ internal class MessagingPushImplementation: MessagingPushInstance {
             logger.info("no profile identified, so not registering device token to a profile")
             return
         }
+        getDefaultDeviceAttributes {attributes in
+            var deviceAttributes = attributes ?? [:]
+            if let customDeviceAttributes = customAttributes {
+                deviceAttributes = deviceAttributes.mergeWith(customDeviceAttributes)
+            }
+            let body = StringAnyEncodable(deviceAttributes)
+            let data: AnyEncodable = AnyEncodable(body)
```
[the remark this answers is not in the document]
ami-aman: Can you please elaborate this point. How do you think that this can cause issues ?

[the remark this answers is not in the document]
ami-aman: If you believe that deviceAttributes can be `nil` then that's handled. It will never be nil but [:]

Sources/MessagingPush/MessagingPushImplementation.swift
```diff
@@ -113,6 +133,31 @@ internal class MessagingPushImplementation: MessagingPushInstance {
                                     data: MetricRequest(deliveryId: deliveryID, event: event, deviceToken: deviceToken,
                                                         timestamp: Date()))
     }
+    func getDefaultDeviceAttributes(completionHandler: @escaping([String: Any]?) -> Void) {
```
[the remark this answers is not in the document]
ami-aman: Do you think mocking will work in this case ? Since default attributes are device specific, os specific, sdk specific, by adding mock attributes we would be kind of hard coding the values and tests may fail at times.

Sources/MessagingPush/MessagingPushImplementation.swift
```diff
@@ -113,6 +133,31 @@ internal class MessagingPushImplementation: MessagingPushInstance {
...
+                                    "device_model": deviceModel,
+                                    "app_version": appVersion,
+                                    "cio_sdk_version": sdkVersion,
+                                    "device_locale": deviceLocale,
+                                    "push_subscribed": String(isSubscribed)]
+            completionHandler(deviceAttributes)
+        }
+        #else
+            completionHandler(nil)
+        #endif
```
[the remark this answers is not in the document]
ami-aman: `import UIKit` is always done at file scope. If the condition` #if canImport(UIKit)` is added to the property `osInfo` value then it would work as expected going by your suggestion but the Linux, MacOs test cases will fail in CI/CD. 

I already tried implementing your suggestion but it results in failed test cases. Any other suggestion is always welcome !

Sources/MessagingPush/Service/Request/Device.swift
```diff
@@ -1,14 +1,16 @@
 import Foundation
 
 // Standard device structure
-internal struct Device: Codable {
+internal struct Device<T: Encodable>: Encodable {
```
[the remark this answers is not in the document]
ami-aman: Addressed !

Sources/MessagingPush/Service/Request/RegisterDeviceRequest.swift
```diff
@@ -1,6 +1,6 @@
 import Foundation
 
 // https://customer.io/docs/api/#operation/add_device
-internal struct RegisterDeviceRequest: Codable {
-    let device: Device
+internal struct RegisterDeviceRequest<T: Encodable>: Encodable {
```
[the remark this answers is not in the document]
ami-aman: Addressed !

Sources/MessagingPush/MessagingPushImplementation.swift
levibostian: Yes, you are correct. 

The same issue is present for [profile attributes](https://github.com/customerio/customerio-ios/blob/66f5ec2b7138f5eade7a0c9d732646d2bb7ef1c9/Sources/Tracking/CustomerIOImplementation.swift#L81-L84). So, this shouldn't be a blocker for this PR. I [created a ticket](https://github.com/customerio/issues/issues/6982) to fix attributes for both devices + profile.

Sources/MessagingPush/Background Queue/Task Data/RegisterPushNotificationQueueTaskData.swift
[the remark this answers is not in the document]
levibostian: From trying to do my suggestion, I don't see it possible at this time. We need to keep the current implementation.

PR #152, opened 2022-03-25
fix: locale uses preferred language

levibostian:
Currently, the device locale that the SDK obtains is the current locale of the host app. This has limitations on it because if the user's device is set to Spanish but the host app does not support Spanish, the device locale that the SDK receives is not Spanish. 

We want to improve our SDK's code in regards to obtaining the device locale. We would prefer is we get the locale of the device even if the host app does not support it. That's what this PR is doing. 

### DIFF
--- a/Sources/Tracking/Util/DeviceInfo.swift
+++ b/Sources/Tracking/Util/DeviceInfo.swift
@@ -66,12 +66,22 @@ public class CIODeviceInfo: DeviceInfo {
         SdkVersion.version
     }
 
+    // Requirements:
+    // 1. Use - instead of _
+    // 2. We prefer to get the OS language that the user has set. First try to return that. If that does not succeed
+    // we default to the language set for the host app. The language returned for that will only return languagues
+    // that the app supports. If OS set to "es" but app does not support Spanish, then "es" will not be returned.
     public var deviceLocale: String {
-        guard let preferredIdentifier = Locale.preferredLanguages.first else {
-            return Locale.current.identifier
+        if let osSetLanguage = Locale.preferredLanguages.first {
+            let locale = Locale(identifier: osSetLanguage)
+
+            if let languageCode = locale.languageCode,
+               let regionCode = locale.regionCode {
+                return "\(languageCode)-\(regionCode)"
+            }
         }
 
-        return preferredIdentifier.replacingOccurrences(of: "_", with: "-")
+        return Locale.current.identifier.replacingOccurrences(of: "_", with: "-")
     }
 
     public func isPushSubscribed(completion: @escaping (Bool) -> Void) {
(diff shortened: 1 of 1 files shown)


Sources/Tracking/Util/DeviceInfo.swift
```diff
@@ -66,8 +66,22 @@ public class CIODeviceInfo: DeviceInfo {
         SdkVersion.version
     }
 
+    // Requirements:
+    // 1. Use - instead of _
+    // 2. We prefer to get the OS language that the user has set. First try to return that. If that does not succeed
+    // we default to the language set for the host app. The language returned for that will only return languagues
+    // that the app supports. If OS set to "es" but app does not support Spanish, then "es" will not be returned.
```
jatinn, line 73: `If OS set to "es" but app does not support Spanish, then "es" will not be returned.`

comment seems inconsistent with the code to me as looks like we will always prefer to return the os lang.

Sources/Tracking/Util/DeviceInfo.swift
```diff
@@ -66,8 +66,22 @@ public class CIODeviceInfo: DeviceInfo {
...
     public var deviceLocale: String {
-        Locale.current.identifier.replacingOccurrences(of: "_", with: "-")
+        if let osSetLanguage = Locale.preferredLanguages.first {
+            let locale = Locale(identifier: osSetLanguage)
+
+            if let languageCode = locale.languageCode,
```
jatinn, line 78: could this be simplified to
```
return locale.identifier.replacingOccurrences(of: "_", with: "-")
or
return osSetLanguage.replacingOccurrences(of: "_", with: "-")
```

jatinn, line 78: could this be simplified to
```
return locale.identifier.replacingOccurrences(of: "_", with: "-")
or
return osSetLanguage.replacingOccurrences(of: "_", with: "-")
```

jatinn (approved):

Sources/Tracking/Util/DeviceInfo.swift
levibostian: duplicate. Closing.

levibostian: Looking at the code, I don't currently see an easy way to do that. There are so many possibilities for `null` to happen that's why this code has 3 `if let` statements.

Sources/Tracking/Util/DeviceInfo.swift
levibostian: `if let osSetLanguage = Locale.preferredLanguages.first` has the possibility of returning `null` and the `if` statement being `false`. That's when we will *not* be returning the OS lang.

Sources/Tracking/Util/DeviceInfo.swift
levibostian: Looking at the code, I don't currently see an easy way to do that. There are so many possibilities for null to happen that's why this code has 3 `if let` statements.

PR #167, opened 2022-06-09
refactor: create 1 shared DI graph

levibostian:
Closes: https://github.com/customerio/issues/issues/7439

Things changed
1. Instead of each module having their own DI graph (`DITracking`, `DIMessagingPush`, etc) we just have 1 DI graph class and instance across the entire SDK: `DIGraph`. The code for the graph is still automatically generated as it was before. This greatly simplifies the complexity of the code base. Remembering what dependency is inside of what graph is exhausting and confusing. Also, automated tests are now much more simple to write and maintain. Before, we had to maintain multiple graphs at one time in a test class. 

The implementation chosen (extensions) to do this came from how we created a DI graph on Android SDK. It has worked well over there so the logic is duplicated here. 

2. While working on this task, I noticed that our swift lint was not running against our `Tests/` directory. So, I enabled lint on this and fixed some lint issues to make lint pass. 


.swiftlint.yml
```diff
@@ -1,14 +1,13 @@
 # Rules: https://realm.github.io/SwiftLint/rule-directory.html
+# Config file learn more: https://github.com/realm/SwiftLint#configuration=
+
+excluded: # paths or files to ignore during linting. Takes precedence over `included`.  
+  - .build
+  - Package.swift
```
levibostian, line 6: Instead of `include`, use `exclude` so that there is a smaller chance of lint not running against code in our project that we add over time.

Makefile
```diff
@@ -3,13 +3,13 @@ SHELL = /bin/sh
...
-	sourcery --sources Sources/MessagingPushAPN --templates Sources/Templates --output Sources/MessagingPushAPN/autogenerated --args diAccessLevel=internal,moduleName=MessagingPushAPN,imports=CioMessagingPush:CioTracking:Common
-	sourcery --sources Sources/MessagingPushFCM --templates Sources/Templates --output Sources/MessagingPushFCM/autogenerated --args diAccessLevel=internal,moduleName=MessagingPushFCM,imports=CioMessagingPush:CioTracking:Common
+	sourcery --sources Sources/Common --templates Sources/Templates --output Sources/Common/autogenerated 
+	sourcery --sources Sources/Tracking --templates Sources/Templates --output Sources/Tracking/autogenerated --args imports=Common
+	sourcery --sources Sources/MessagingPush --templates Sources/Templates --output Sources/MessagingPush/autogenerated --args imports=CioTracking-Common
+	sourcery --sources Sources/MessagingPushAPN --templates Sources/Templates --output Sources/MessagingPushAPN/autogenerated --args imports=CioMessagingPush-CioTracking-Common
+	sourcery --sources Sources/MessagingPushFCM --templates Sources/Templates --output Sources/MessagingPushFCM/autogenerated --args imports=CioMessagingPush-CioTracking-Common
```
levibostian, line 12: Bonus: We got to greatly simplify the source code generation code for the DI graph 🎊

Sources/Common/DIGraph.swift
```diff
@@ -0,0 +1,52 @@
+import Foundation
+
+public class DIGraph {
```
levibostian, line 3: The code in this class is almost identical to what used to exist in `DICommon`, `DIMessagingPush`, etc before. The logic is the same in this class compared to the classes before.

Sources/MessagingPush/Background Queue/QueueRunner.swift
```diff
@@ -4,9 +4,9 @@ import Foundation
 // Queue tasks for the MessagingPush module.
 // sourcery: InjectRegister = "QueueRunnerHook"
 public class MessagingPushQueueRunner: ApiSyncQueueRunner, QueueRunnerHook {
-    init(siteId: SiteId, diGraph: DICommon) {
-        super.init(siteId: siteId, jsonAdapter: diGraph.jsonAdapter, logger: diGraph.logger,
-                   httpClient: diGraph.httpClient)
+    override init(siteId: SiteId, jsonAdapter: JsonAdapter, logger: Logger, httpClient: HttpClient) {
```
levibostian, line 7: No more having to do hacks like this where you must inject a DI graph in the constructor. Now all classes can simply inject all dependencies via the constructor!

Tests/.swiftlint.yml
```diff
@@ -0,0 +1,19 @@
...
+  # XCTAssertNotNil(foo)
+  # XCTAssertEqual(foo?.bar, "")
+  # - force_try
+
+  # `as!` in a test can result in tests being halted and not finish executing. Instead, we would rather have a test fail and continue executing the suite. 
+  # Instead of `as!` use: 
+  # let foo = ... as? ABC
+  # XCTAssertNotNil(foo)
+  # XCTAssertEqual(foo?.bar, "")
+  # - force_cast
```
levibostian, line 19: Because this code is all commented out, it can technically be deleted. It exists here on purpose for documentation purposes. I disabled these 2 rules before but then later realizing there are consequences to disabling these rules (the test suite halts instead of just failing tests and continuing to run). So, I documented my findings here to remind developers in the future it's a good idea to keep these rules enabled.

Tests/Common/DIGraphTests.swift
```diff
@@ -0,0 +1,11 @@
...
+import Foundation
+import SharedTests
+import XCTest
+
+class DIGraphTests: IntegrationTest {
+    func testDependencyGraphComplete() {
+        let graph = DIGraph.getInstance(siteId: testSiteId)
+        graph.testDependenciesAbleToResolve() // test will fail if an exception occurs while running this function
+    }
+}
```
levibostian, line 11: This code is copy/pasted into each of the modules of this project (except the `import` statements are unique). It's required that each module have their own copy of this test function instead of just running the test function once in one of the modules (such as `Common`). 

My only idea at this time to prevent this copy/paste is to autogenerate this test function so each time the dependency injection graph is generated, this test gets generated.

Tests/Shared/IntegrationTest.swift
```diff
@@ -17,6 +17,6 @@ open class IntegrationTest: UnitTest {
 
         // To prevent any real HTTP requests from being sent, override http request runner for all tests.
         httpRequestRunnerStub = HttpRequestRunnerStub()
-        diGraph.override(.httpRequestRunner, value: httpRequestRunnerStub, forType: HttpRequestRunner.self)
+        diGraph.override(value: httpRequestRunnerStub, forType: HttpRequestRunner.self)
```
levibostian, line 20: `.override()` was able to be simplified, too 🎊. Only 2 parameters instead of 1!

PR #174, opened 2022-06-27
feat: sdk wrappers modify user-agent

levibostian:
Allows our react native SDK to modify the user-agent in HTTP requests. 

### DIFF
--- a/Sources/Common/Service/HttpClient.swift
+++ b/Sources/Common/Service/HttpClient.swift
@@ -39,7 +39,7 @@ public class CIOHttpClient: HttpClient {
     ) {
         self.httpRequestRunner = httpRequestRunner
         self.session = Self.getSession(siteId: siteId, apiKey: sdkCredentialsStore.credentials.apiKey,
-                                       deviceInfo: deviceInfo)
+                                       deviceInfo: deviceInfo, sdkWrapperConfig: configStore.config.sdkWrapperConfig)
         self.baseUrls = configStore.config.httpBaseUrls
         self.jsonAdapter = jsonAdapter
         self.globalDataStore = globalDataStore
@@ -135,7 +135,8 @@ public class CIOHttpClient: HttpClient {
 }
 
 extension CIOHttpClient {
-    static func getSession(siteId: String, apiKey: String, deviceInfo: DeviceInfo) -> URLSession {
+    static func getSession(siteId: String, apiKey: String, deviceInfo: DeviceInfo,
+                           sdkWrapperConfig: SdkWrapperConfig?) -> URLSession {
         let urlSessionConfig = URLSessionConfiguration.ephemeral
         let basicAuthHeaderString = "Basic \(getBasicAuthHeaderString(siteId: siteId, apiKey: apiKey))"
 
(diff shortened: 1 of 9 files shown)


Sources/Tracking/CustomerIO.swift
```diff
@@ -222,6 +232,8 @@ public class CustomerIO: CustomerIOInstance {
     private func postInitialize(siteId: String) {
         let diGraph = DIGraph.getInstance(siteId: siteId)
 
+        diGraph.logger.info("Customer.io SDK \(SdkVersion.version) initialized successfully.")
```
levibostian, line 235: Here, we will edit the log statement to say something such as:
```
Customer.io SDK \(SdkVersion.version) initialized successfully. SDK being wrapped by \(wrapper.name), \(wrapper.version) \(wrapper.bundleId)
```

Sources/Common/Type/SDKWrapperConfig.swift
```diff
@@ -0,0 +1,9 @@
+import Foundation
+
+/// Metadata about a SDK wrapper/bridge made to use this Customer.io SDK.
+public struct SdkWrapperConfig {
```
levibostian, line 4: I like the idea of adding another property to this struct:

```swift
public struct SdkWrapperConfig {
  let bundleId: String
}
```

Then, for developers who wrap our SDK, they provide a bundle ID such as "io.customer.reactnative", "com.github.cool_customerio_rn_sdk". We can use this value mostly in log statements. Can be useful during customer debugging in case they are using a 3rd party/unofficial SDK made by someone else.

Sources/Common/Type/SDKWrapperConfig.swift
```diff
@@ -0,0 +1,9 @@
...
+    /// What is the name of your SDK wrapper?
+    let name: String
```
ami-aman, line 6: Instead of passing it as a `String`, we should use `Enum` like

```
enum WrapperSource : String {
   case Reactnative = "ReactNative"
   case Flutter // Future = "Flutter"
   case iOS = "iOS"
}
```

Sources/Common/Service/HttpClient.swift
```diff
@@ -166,9 +168,12 @@ extension CIOHttpClient {
      * Otherwise will return
      * `Customer.io iOS Client/1.0.0-alpha.16`
      */
-    static func getUserAgent(deviceInfo: DeviceInfo) -> String {
-        var userAgent = "Customer.io iOS Client/"
-        userAgent += deviceInfo.sdkVersion
+    static func getUserAgent(deviceInfo: DeviceInfo, sdkWrapperConfig: SdkWrapperConfig?) -> String {
+        var userAgent = "Customer.io iOS Client/\(deviceInfo.sdkVersion)"
```
ami-aman, line 172: If we use enum as explained [here](https://github.com/customerio/customerio-ios/pull/174/files#r907806258), then we can directly write the initial value for `userAgent` as :

`var userAgent = "Customer.io \(sdkWrapperConfig.name.rawValue) Client/"`

Since, `Customer.io `and `Client` is same in all cases, only the source of SDK changes followed by the version of the either wrapper or the native SDK.

Sources/Common/Type/SDKWrapperConfig.swift
levibostian: Since we do not encourage customers to create their own wrappers around our SDK, this feature is not needed.

ami-aman (approved):
LGTM 👍

PR #209, opened 2022-11-04
refactor!: singleton API only way to use SDK now

levibostian:
# QA 

This PR touches almost the entire codebase and during my own QA testing, I have found a few bugs. So, a full QA walk through with the RH app is recommended. 

Things to test in particular:
* Install build of RH app *before* [`pr.78` build](https://github.com/customerio/RemoteHabits-iOS/pull/78). Run through the app a little bit to make sure some tasks are added to the background queue and profile and device token have been saved to app's data storage. Then, install `pr.78` build of RH. Open app and make sure that the previously identified profile, device token, and background queue tasks all run as expected. This tests that customers upgrading to this version of SDK will not lose data going from SDK v2 to v3. 
* Open app, do some tasks, force quit app, open it up again. The SDK initialization code has been changed so we want to make sure that it's working correctly when the RH opens and it initializes the SDK. 
* Push notifications and in-app messages are received. 
* Profile and device attributes get added successfully to Fly. RH app adds these automatically. We need to confirm that Fly receives them. 

---

Closes: https://github.com/customerio/issues/issues/6933

This PR removes the ability to use the iOS SDK in a non-singleton way. You used to be able to create 1+ instances of the SDK: `CustomerIO()` but now you have to use `CustomerIO.shared`. 

Notes: 
* This PR will trigger a breaking change to the SDK and will trigger a major version bump to v3.0
* This is a bigger PR since it's a larger refactor. I made comments on parts of the code I find interesting in the code review.


Sources/Common/Store/GlobalDataStore.swift
```diff
@@ -2,11 +2,6 @@ import Foundation
 
 /// SDK data that is common between all site ids.
 public protocol GlobalDataStore: AutoMockable {
-    // site id used for the singleton instance of the SDK.
-    var sharedInstanceSiteId: String? { get set }
-    func appendSiteId(_ siteId: String)
-    // all site ids that have ever been registered with the SDK
-    var siteIds: [String] { get }
```
levibostian, line 9: This code was originally added to the SDK so we could do something such as running background queue tasks for *all* of the site-ids *ever* registered to the SDK. 

However, we have never used this feature so I removed it. We can bring it back in the future if we decide we need this feature.

Sources/Common/Store/SdkConfig.swift
```diff
@@ -66,13 +75,3 @@ public struct SdkConfig {
...
-
-public protocol SdkConfigStore: AutoMockable {
-    var config: SdkConfig { get set }
-}
-
-// sourcery: InjectRegister = "SdkConfigStore"
-// sourcery: InjectSingleton
-public class InMemorySdkConfigStore: SdkConfigStore {
-    @Atomic public var config = SdkConfig()
-}
```
levibostian, line 78: No more need for a SDK store. We now store the SDKConfig instance in the dependency injection graph.

Sources/MessagingInApp/MessagingInAppImplementation.swift
```diff
@@ -0,0 +1,98 @@
+import CioTracking
+import Common
+import Foundation
+import Gist
+
+internal class MessagingInAppImplementation: MessagingInAppInstance {
```
levibostian, line 6: All of this logic was copy/pasted from `MessagingInApp` class. To follow the same pattern that `CustomerIO` and `MessagingPush` classes follow with an `implementation` object.

Sources/MessagingPush/RichPush/RichPushRequestHandler.swift
```diff
@@ -21,7 +22,11 @@ internal class RichPushRequestHandler {
         let existingRequest = requests[requestId]
         if existingRequest != nil { return }
 
-        let diGraph = DIGraph.getInstance(siteId: siteId)
+        let sdkInitializedUtil = SdkInitializedUtilImpl()
+
+        guard let postSdkInitializedData = sdkInitializedUtil.postInitializedData else { return }
+
+        let diGraph = postSdkInitializedData.diGraph
```
levibostian, line 29: This is the new (safe) way to get the DiGraph instance. You see other files in this PR also doing it this way.

README.md
```diff
@@ -44,11 +44,7 @@ We recommend that you set the *Dependency Rule* to *Up to Next Major Version*. W
 
 Before you can use the Customer.io SDK, you need to initialize it. Any calls that you make to the SDK before you initialize it are ignored. 
 
-There are two ways to initialize the SDK. The method you use depends on how you want to use the SDK:
-
-1. Singleton, shared instance (the quick and easy way):
-
-When you use the shared instance, you don't need to manage your own instances of Customer.io SDK classes. To get started, initialize the SDK in the `AppDelegate` `application(_ application: didFinishLaunchingWithOptions)` function: 
+To get started, initialize the SDK in the `AppDelegate` `application(_ application: didFinishLaunchingWithOptions)` function:
```
levibostian, line 47: This change to the README means we need to make an update to the public facing docs for the SDK.

README.md
```diff
@@ -63,6 +59,11 @@ class AppDelegate: NSObject, UIApplicationDelegate {
         // You can optionally provide a Region to set the Region for your Workspace:
         CustomerIO.initialize(siteId: "YOUR SITE ID", apiKey: "YOUR API KEY", region: Region.EU)
 
+        // You can optionally configure the behavior of the SDK:
```
levibostian, line 62: I think it would be a value add to customers to have a link to the docs showing them *all* of the configuration options they have available to them.

Sources/Common/Util/KeyValueStorage.swift
```diff
@@ -20,67 +22,84 @@ public protocol KeyValueStorage: AutoMockable {
...
             appUniqueIdentifier = ".\(appBundleId)"
         }
 
-        return UserDefaults(suiteName: "io.customer.sdk\(appUniqueIdentifier).\(siteId)")
+        var siteIdPart = ".shared" // used for storing global data used for all site-ids.
+        if let siteId = siteId { // if a siteid is given to this instance, we dont store global data with this instance.
+            siteIdPart = ".\(siteId)"
+        }
+
+        return "io.customer.sdk\(appUniqueIdentifier)\(siteIdPart)"
```
levibostian, line 61: Reminder to self: QA test this to make sure that we are *not* changing the name after customer upgrades. To avoid customers losing data.

Tests/Shared/UnitTest.swift
```diff
@@ -122,6 +140,10 @@ open class UnitTest: XCTestCase {
         // The SDK does not use `UserDefaults.standard`, but in case a test needs to,
         // let's delete the data for each test.
         UserDefaults.standard.deleteAll()
+
+        keyValueStorage.deleteAll()
```
jatinn, line 144: called twice now is that expected?
whats suppose to be cleared by calling this after `switchToGlobalDataStore`?

jatinn (approved):
could use review from another engineer more familiar with swift.
overall looks fine main thing that feels odd to me is we pass both `site id & di graph` into many `initialize` calls which feels redundant as some dont use site id and since it seems like can get site id from the di graph.

Sources/Common/Store/GlobalDataStore.swift
```diff
@@ -16,23 +11,7 @@ public protocol GlobalDataStore: AutoMockable {
...
-
-    public var sharedInstanceSiteId: String? {
-        get {
-            keyValueStorage.string(.sharedInstanceSiteId)
-        }
-        set {
-            keyValueStorage.setString(newValue, forKey: .sharedInstanceSiteId)
-        }
-    }
+    private let keyValueStorage: KeyValueStorage
```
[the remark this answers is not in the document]
levibostian: `private` is private to the class. `internal` is scoped to the module. 

I was using `internal` so automated test code can access it to delete data between tests. But,this felt like a hack to me. I improved how tests delete data so this can now be `private`.

Tests/Shared/UnitTest.swift
levibostian: Yes, this is confusing when glancing at it. 

I made a comment explaining what this code is doing. 

`deleteAll()` is needed twice but it deletes a separate set of data.

README.md
levibostian: [Confirmed](https://customer.io/docs/sdk/ios/getting-started/#initialize-the-sdk)

Tests/Common/Util/KeyValueStorageTest.swift
```diff
@@ -4,11 +4,21 @@ import SharedTests
...
+    private func getSiteStorageInstance(siteId: String) -> UserDefaultsKeyValueStorage {
+        UserDefaultsKeyValueStorage(siteId: siteId, deviceMetricsGrabber: deviceMetricsGrabberMock)
+    }
+
+    private func getGlobalStorageInstance() -> UserDefaultsKeyValueStorage {
+        let newInstance = UserDefaultsKeyValueStorage(
+            siteId: testSiteId,
+            deviceMetricsGrabber: deviceMetricsGrabberMock
+        )
+        newInstance.switchToGlobalDataStore()
```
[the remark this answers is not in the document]
levibostian: There are 2 required parameters of `UserDefaultsKeyValueStorage` constructor. So, I do not believe `UserDefaultsKeyValueStorage(deviceMetricsGrabber: deviceMetricsGrabberMock)` is possible.

Sources/MessagingInApp/MessagingInAppImplementation.swift
```diff
@@ -0,0 +1,98 @@
...
+
+        if let deliveryId = getDeliveryId(from: message) {
+            // the state of the SDK does not change if adding this queue task isn't successful so ignore result
+            _ = queue.addTrackInAppDeliveryTask(deliveryId: deliveryId, event: .clicked)
+        }
+    }
+
+    private func getDeliveryId(from message: Message) -> String? {
+        guard let deliveryId = message.gistProperties.campaignId else {
+            logger
```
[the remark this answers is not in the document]
levibostian: Yeah, I would agree to not use an error log. Error logs are logs that we hope customers see and then report to us. 

I do see some sort of log as a good idea. So, unless I hear differently, I will change this to a different log level. 

....knowing that delivery id is not set on test sends, I should probably [make `deliveryId` optional for in-app event listener PR I made](https://github.com/customerio/customerio-ios/pull/211/files#diff-e981dfe7071ab6f834502bd595f26c5b851174240f5b6c3d15b2fb80e42e40c6R16)?

Sources/MessagingInApp/MessagingInApp.swift
```diff
@@ -7,133 +7,50 @@ public protocol MessagingInAppInstance: AutoMockable {
...
-            hooks.add(key: .messagingInApp, provider: moduleHookProvider)
-        }
+    // Initialize SDK module
+    public static func initialize(organizationId: String) {
+        Self.shared.initialize(organizationId: organizationId)
     }
 
+    // This function remains here for backwards compatability since it was introduced in previous versions of SDK. Other
+    // modules use only a static initialize() function.
     public func initialize(organizationId: String) {
```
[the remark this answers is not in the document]
levibostian: Migration steps for customers would be:
```swift
// Before:
MessagingInApp().initialize(...)
// after
MessagingInApp.initialize(...)
```

Just remove the `()`. So, let's remove the function.

Sources/MessagingInApp/MessagingInApp.swift
```diff
@@ -7,133 +7,50 @@ public protocol MessagingInAppInstance: AutoMockable {
...
+    override private init() {
+        super.init()
     }
 
     // for testing
-    internal init(diGraph: DIGraph, siteId: String) {
-        self.diGraphOverride = diGraph
-        self.siteId = siteId
+    internal static func resetSharedInstance() {
+        Self.shared = MessagingInApp()
```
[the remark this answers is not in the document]
levibostian: `Self` is required because this is a `static` function. `Self` is equivalent to using `NameOfClass`. It's a swift convenience.

Sources/Common/Util/KeyValueStorage.swift
```diff
@@ -20,67 +22,84 @@ public protocol KeyValueStorage: AutoMockable {
...
-     We want to separate all of the data for each CIO workspace and app.
+     We want to sandbox all of the data for each CIO workspace and app.
      Therefore, we use the app's bundle ID to be unique to the app and the siteId to separate all of the sites.
 
      We also need to have 1 set of UserPreferences that all workspaces of an app share.
-     For these moments, use `sharedSiteId` as the `siteId` parameter.
+     For these moments, use `shared` as the `siteId` value.
      */
-    private func getUserDefaults() -> UserDefaults? {
+    internal func getFileName() -> String {
```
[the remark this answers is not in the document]
levibostian: This class stores key/value pairs. The SDK can also store data in files in the file system. In the future we may use a database. Many ways to save data. Each feature of the SDK uses a different data store depending on what's best for the use case. 

The BQ at this time uses the file system and not this class.

Sources/Common/Util/KeyValueStorage.swift
```diff
@@ -20,67 +22,84 @@ public protocol KeyValueStorage: AutoMockable {
...
 
-    init(siteId: SiteId) {
+    init(siteId: SiteId, deviceMetricsGrabber: DeviceMetricsGrabber) {
         self.siteId = siteId
+        self.deviceMetricsGrabber = deviceMetricsGrabber
+    }
+
+    // Used for global data that's relevant to *all* of the site-ids (not sandboxed).
+    // Instead of the more common way the SDK stores data by sandboxing all of that data by site-id.
+    public func switchToGlobalDataStore() {
```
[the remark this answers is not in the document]
levibostian: Added comment `See `GlobalDataStore` for data that is relevant for *all* site-ids in the SDK.` to this function to explain why this feature of `KeyValueStorage` is still necessary. 

I wanted to remove it entirely with this refactor but some data is still global. Biggest example: device push token.

Sources/Common/Store/SdkConfig.swift
```diff
@@ -5,12 +5,20 @@ import Foundation
...
+        }
+    }
+
     /**
      Base URL to use for the Customer.io track API. You will more then likely not modify this value.
 
      If you override this value, `Region` set when initializing the SDK will be ignored.
      */
-    public var trackingApiUrl: String = ""
+    public var trackingApiUrl: String
```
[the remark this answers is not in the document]
levibostian: Not sure what you mean by, "...if people don't set region".

Having the default value be `""` is error-prone, IMO. Our SDK should set a value for this when the SDK is initialized. So, by me removing the default, it reminds us to set a value for it at SDK initialization.

Sources/MessagingInApp/MessagingInApp.swift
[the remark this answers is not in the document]
levibostian: After working with the code, I think it's best to leave the non-singleton `initialize(organizationId)` function. The function exists for mocking and tests. We did add a convenience static version of `initialize(organizationId)` to match the API of the other modules that customers can use.

levibostian:
@matt-frizzell came up with an idea of us waiting to deploy this change until [this feature](https://github.com/customerio/issues/issues/8359) is merged as well.

Issue #203, opened 2022-10-12
Could you please update FirebaseMessaging and FirebaseCore dependency to the latest version?


pirrest:
In my project I'm using FirebaseMessaging v9.6.0 but you are 8.10.0. So I have cocoapods config.
```
[!] CocoaPods could not find compatible versions for pod "FirebaseMessaging":
  In Podfile:
    customer_io (from `.symlinks/plugins/customer_io/ios`) was resolved to 0.0.1, which depends on
      CustomerIOMessagingPushFCM (= 1.2.0-beta.3) was resolved to 1.2.0-beta.3, which depends on
        FirebaseMessaging (~> 8.10.0)

    customer_io (from `.symlinks/plugins/customer_io/ios`) was resolved to 0.0.1, which depends on
      FirebaseMessaging (~> 9.6.0)
```
Could you please update it in your .podspec file?

ahood12:
@ami-aman OOC will you all be bumping your RN SDK to pull in the updated iOS SDK version as well? I was going through the process of integrating CIO with our app and was blocked by the out of date `FirebaseMessaging` dep. Thanks!

ami-aman:
Hi @ahood12 
Apologies that the Firebase version is being a blocker for you. We have already added this to our list.. This might take some time as we would need to update iOS & Android SDKs first, test them and then react native package but I want to let you know that our squad is on it !  😄

ami-aman:
Hi @ahood12 
We have already updated React native package with iOS latest stable `1.2.0`. Though, this release does not feature latest Firebase version but I want to let you know that this is on our priority list! 😄

levibostian:
@pirrest Thank you for reporting this to us and being patient with us with getting this task completed. 

I have a branch that is available for you to install now to test that we fixed it. 

In your `Podfile`, install our SDK from git: 

```ruby
pod 'CustomerIO/MessagingPushFCM', :git => 'https://github.com/customerio/customerio-ios.git', :branch => 'levi/update-fcm'
// Note: Follow this format for all CustomerIO dependencies in your Podfile
```

We will also perform our own testing before we get a release out to you. In the meantime, see if this works in your app 😄.

pirrest:
@levibostian Thank you for the work around for us. Waiting for the stable release.

levibostian:
[...]

PR #210, opened 2022-11-04
refactor!: remove FCM dependency from cocoapods

levibostian:
Breaking change introduced: 
* Customers impacted: React Native customers using FCM. That's because if they follow our docs as they appear today, we do not tell customers to install Firebase dependencies on their own with cocoapods. So, by RN customers upgrading the iOS SDK to include this change, they will need to install the firebase SDK themselves. 

~~Needs to be tested:~~ [done](https://github.com/customerio/customerio-ios/pull/210#issuecomment-1305794706)
* Create new iOS app that uses cocoapods
* Install [this PR](https://github.com/customerio/customerio-ios/issues/203#issuecomment-1303787347)
* Integrate FCM
* Run app and see if receive a rich push

This PR fixes [this customer reported issue](https://github.com/customerio/customerio-ios/issues/203). Our original action item was to update the Firebase dependency in our iOS SDK. However, after evaluation, I determined that our code doesn't not actually reference anything from Firebase at this time. Our SPM config files do not contain any Firebase dependency items at all. So, I removed the dependency to fix this customer issue. 

Maybe the Firebase dependency was originally added for react native? If so, I assume that the react native can declare FCM as a dependency in itself instead of inheriting the FCM dependency from the CIO iOS SDK.

### DIFF
--- a/CustomerIOMessagingPushFCM.podspec
+++ b/CustomerIOMessagingPushFCM.podspec
@@ -20,6 +20,5 @@ Pod::Spec.new do |spec|
   spec.source_files  = "Sources/MessagingPushFCM/**/*"
   spec.module_name = "CioMessagingPushFCM" # the `import X` name when using SDK in Swift files 
   
-  spec.dependency "CustomerIOMessagingPush", "= #{spec.version.to_s}"
-  spec.dependency "FirebaseMessaging", '~> 8.10.0'
+  spec.dependency "CustomerIOMessagingPush", "= #{spec.version.to_s}"  
 end


levibostian:
Blocked from merging until testing is done.

levibostian:
Unblocked as I completed testing. I created a new iOS app > Installed this PR > Got a FCM token and sent the token to a profile in CIO.

jatinn:
you mention the dependency may have been for react native so we should also test with a build of the ios react native app using fcm just to confirm

jatinn (approved):
change is straightforward but lets test with the react native/expo ios fcm test app to ensure no impact or to document the extra config steps that may be needed due to this change

levibostian:
Currently blocked while PR gets tested in Ami app to see if RN SDK is impacted by this change. 

Learn more: https://customerio.slack.com/archives/C01QYDKD0SU/p1667939161080969

levibostian:
Keeping branch available so [workaround](https://github.com/customerio/customerio-ios/issues/203#issuecomment-1303787347) continues to work

PR #211, opened 2022-11-09
feat: add in-app event listener

levibostian:
Closes: https://github.com/customerio/issues/issues/8359

Docs change PR: https://github.com/customerio/docs/pull/791
RH PR: https://github.com/customerio/RemoteHabits-iOS/pull/79


Tests/MessagingInApp/APITest.swift
```diff
@@ -19,6 +19,22 @@ class MessagingInAppAPITest: UnitTest {
         try skipRunningTest()
 
         MessagingInApp.shared.initialize(organizationId: "")
-        mock.initialize(organizationId: "")
+        MessagingInApp.shared.initialize(organizationId: "", eventListener: self)
```
ami-aman, line 22: I think it is good to write test case without eventListener as a parameter or by passing a nil parameter to test the other possibility

Sources/MessagingInApp/Type/InAppMessage.swift
```diff
@@ -0,0 +1,18 @@
...
+public struct InAppMessage: Equatable {
+    public let instanceId: String
+    public let messageId: String
+    public let deliveryId: String // (Currently taken from Gist's campaignId property)
+
+    internal init(gistMessage: GistMessage) {
+        self.instanceId = gistMessage.instanceId
+        self.messageId = gistMessage.messageId
+        // The Gist SDK source code always populates the campaign-id. Having a nil campaignId is only common using the
+        // `public init()` for Message.
```
ami-aman, line 15: Though very low probability of this happening but in case it happens then force unwrapping will make the app crash. I am trying to think of a use case but all I can think right now is if customer create own message instance but not sure if that would ever happen !

matt-frizzell:
Can you update the "closes" link to the iOS issue customerio/issues/#8359

Sources/MessagingInApp/Type/InAppMessage.swift
levibostian: I was not aware of this, but it does look like sending test in-app messages can make `deliveryId` null. https://github.com/customerio/customerio-ios/pull/209/files#r1020570037

levibostian:
@matt-frizzell had the idea of deploying this change with https://github.com/customerio/customerio-ios/pull/209

Sources/MessagingInApp/Type/InAppMessage.swift
levibostian: I made `deliveryId` optional.

levibostian:
Do not merge until this change is made. 

Remove `instanceId` from the listener events. 

```
message-shown/dismissed/error:
{
   messageId: string,
   deliveryId: string
}
message-action:
{
   messageId: string,
   deliveryId: string,
   action: string,
   name: string, // what are examples of name/action?
}
```

[Learn more](https://customerio.slack.com/archives/C036FTE6UJ0/p1669747951244759?thread_ts=1669742853.819709&cid=C036FTE6UJ0)

Sources/MessagingInApp/MessagingInApp.swift
```diff
@@ -25,27 +26,58 @@ public class MessagingInApp: ModuleTopLevelObject<MessagingInAppInstance>, Messa
...
 
+    public static func initialize(organizationId: String, eventListener: InAppEventListener) {
+        Self.shared.initialize(organizationId: organizationId, eventListener: eventListener)
+    }
+
     public func initialize(organizationId: String) {
-        initialize() // enables features such as setting up hooks
+        guard let implementation = implementation else {
+            sdkNotInitializedAlert("CustomerIO class has not yet been initialized. Request to initialize the in-app module has been ignored.")
+            return
```
levibostian, line 52: This is a new idea I am proposing in this PR. I prefer to introduce new features in separate PRs so I can pull this out and make a separate PR if that's helpful. 

When the SDK is not initialized yet, we should at a minimum tell the customer in a log message that they might have configured the CIO SDK wrong. 

If we like this idea, we could implement this idea in the other classes such as push and `CustomerIO` classes. We check if `implementation` class is null and if so, log a message to customer that the SDK has not yet been initialized so the request is ignored.

levibostian:
No longer blocked. [This has been addressed](https://github.com/customerio/customerio-ios/pull/211#issuecomment-1332281565).

Sources/Common/Util/Log.swift
```diff
@@ -89,12 +89,31 @@ public class ConsoleLogger: Logger {
...
+        #endif
+    }
+}
+
+// Log messages to customers when the SDK is not initialized.
+// Great for alerts when the SDK may not be setup correctly.
+// Since the SDK is not initialized, dependencies graph is not created.
+// Therefore, this is just a function that's available at the top-level available to all
+// of the SDK without the use of dependencies.
+public func sdkNotInitializedAlert(_ message: String) {
```
Shahroz16, line 117: I like the idea but if all its doing is logging a message why are we calling it, `sdkNotInitializedAlert` because maybe we might have more alerts that could utilize this method?

Sources/MessagingInApp/MessagingInApp.swift
```diff
@@ -25,27 +26,58 @@ public class MessagingInApp: ModuleTopLevelObject<MessagingInAppInstance>, Messa
...
+        }
+
+        initialize()
+
+        implementation.initialize(organizationId: organizationId)
+    }
+
+    public func initialize(organizationId: String, eventListener: InAppEventListener) {
```
Shahroz16, line 60:
```suggestion
   public func initialize(organizationId: String, eventListener: InAppEventListener? = nil) {
        guard let implementation = implementation else {
            sdkNotInitializedAlert("CustomerIO class has not yet been initialized. Request to initialize the in-app module has been ignored.")
            return
        }

        initialize()
        
        if eventListener == nil {
            implementation.initialize(organizationId: organizationId)
        } else {
            implementation.initialize(organizationId: organizationId, eventListener: eventListener!)
        }
    }
```

Do we need 2 methods for this? because seems like a lot of code is similar and it doesn't have `static` so doesn't seem like it is customer facing either.

Shahroz16 (changes requested):

Sources/MessagingInApp/MessagingInApp.swift
levibostian: I [added some comments to the code](https://github.com/customerio/customerio-ios/pull/211/commits/da4fb58dcc6bc5526c7738f4bb61bf8048aa2bbe) to explain this. 

It's mostly because of `MessagingInAppInstance` protocol that we have the multiple initialize functions.

Sources/Common/Util/Log.swift
levibostian: I thought about making the naming more generic. However, I found some cons with that approach. 

In the function body, you see `"⚠️ \(message)", level: .error`. So, we are prefixing all log messages with `⚠️ ` and we are using the log level of error. 

If we were to rename this function to something more generic:
```swift
public func logMessage(_ message: String, level: CioLogLevel) {
	ConsoleLogger.logMessageToConsole(message, level: level)
}
```

* Everywhere in the code that current calls `sdkNotInitializedAlert()` will now have copy/pasted code in it.

```swift
// before
sdkNotInitializedAlert("foo")

// after
logMessage("⚠️ foo", level: .error)
```

* There is a small number of edge cases where we need to log a message and the dependency injection graph is not yet constructed. Therefore, I don't believe at this time we will encounter a problem where the number of functions similar to `sdkNotInitializedAlert` grows. Currently, the only edge case I can consider needing logging without a DI graph is when the SDK is not initialized.

Sources/MessagingInApp/MessagingInApp.swift
Shahroz16: Understood, thanks for clarifying that, my biggest annoyance is seeing the same exact lines of codes in both methods. 

What do you think about maybe, encapsulating 

```
guard let implementation = implementation else {
            sdkNotInitializedAlert("CustomerIO class has not yet been initialized. Request to initialize the in-app module has been ignored.")
            return
        }

        initialize()
```

in a lambda function, which returns true/false and upon its callback, trigger 

```
implementation.initialize(......)
```

On a side note, I added [a comment on another PR of yours](https://github.com/customerio/customerio-ios/pull/252#discussion_r1086535824), which holds true for this as well, regarding organization id with event listeners.

Shahroz16 (approved):
Just added a couple of comments, which we can cater to later, as I feel they are an improvement but shouldn't be blockers to this PR. 

But would appreciate feedback on them.

Sources/Common/Util/Log.swift
levibostian: I am leaning towards the current implementation of `sdkNotInitializedAlert` because
* A function name that's more generic such as `logMessage` doesn't tell the developer when to use that function to log a message and when to use the `Logger` that's inside of the DI graph. A function name such as `sdkNotInitializedAlert` suggests that this only be used when the SDK is not initialized. Otherwise, use the logger in the DI graph. 
* I like how DRY `sdkNotInitializedAlert`'s implementation is where the icon and log level are encapsulated into 1 place in the code. If we were to replace it with: `logMessage(message: "foo", level: .error, icon: .warning)`, it would no longer be DRY.

PR #220, opened 2022-11-21
fix!: make delivered push metric more reliable

levibostian:
Breaking change
* In Notification Service Extension, customers need to change `CustomerIO.initialize()` to `CustomerIO.initializeRichPush()`

Closes: https://github.com/customerio/issues/issues/6633
Closes: https://github.com/customerio/issues/issues/6934

PR contains following changes:
* Add SDK initialization function made just for a Notification Service Extension (NSE). This is required because [iOS doesn't allow a NSE to use our existing `initialize()` function](https://github.com/customerio/customerio-ios/blob/3547076a3694c211f1b11b2a61db7d6debe05a6b/Sources/Tracking/CustomerIO.swift#L160). 
* Run background queue (BQ) tasks before the NSE exits. This makes sure that `delivered` event that was added to queue will attempt to run. 
* Refactor the BQ to re-query the tasks after each task execution. This improves the BQ in this scenario: the BQ is already running and a new task gets added to the BQ. Before, the BQ would not execute that newly added task now it does. This gives us more confidence that the `delivered` event task that gets added will be executed. 
* Refactor the messaging push module code to [have the push logic exist in the tracking module](https://github.com/customerio/issues/issues/6934). 


Tests/Tracking/CustomerIOImplementationTest.swift
```diff
@@ -244,4 +256,136 @@ class CustomerIOImplementationTest: UnitTest {
         XCTAssertTrue(hooksMock.screenViewHooksCalled)
         XCTAssertEqual(hooksMock.screenViewHooksGetCallsCount, 1)
     }
+
+    // MARK: registerDeviceToken
```
levibostian, line 260: The tests in this file were moved from MessagingPushImplementation

Tests/Tracking/APITest.swift
```diff
@@ -65,6 +65,19 @@ class TrackingAPITest: UnitTest {
...
+        mock.registerDeviceToken("")
+
+        // delete push token
+        CustomerIO.shared.deleteDeviceToken()
+        mock.deleteDeviceToken()
+
+        // track push metric
+        let metric = Metric.delivered
+        CustomerIO.shared.trackMetric(deliveryID: "", event: metric, deviceToken: "")
+        mock.trackMetric(deliveryID: "", event: metric, deviceToken: "")
```
levibostian, line 79: New public functions for customers to use if they wish.

Tests/MessagingPush/MessagingPushImplementationTest.swift
```diff
@@ -9,158 +9,20 @@ class MessagingPushImplementationTest: UnitTest {
...
-            deviceAttributesProvider: deviceAttributesMock,
-            dateUtil: dateUtilStub,
-            deviceInfo: deviceInfoMock
+            sdkConfig: sdkConfig,
+            backgroundQueue: queueMock,
+            sdkInitializedUtil: sdkInitializedUtilMock
         )
     }
-
-    // MARK: registerDeviceToken
```
levibostian, line 36: All of these tests have been moved to CustomerIOImplementation tests since that's where this code logic moved to.

Tests/Common/Background Queue/QueueStorageTest.swift
```diff
@@ -92,17 +76,52 @@ class QueueStorageTest: UnitTest {
...
 
-    func test_create_givenFileStorageDoesNotUpdateInventory_expectFalse() {
-        var returnValues = [true, false]
-        fileStorageMock.saveClosure = { _, _, _ in
-            returnValues.removeFirst()
-        }
+// MARK: integration tests
+
+#if !os(Linux) // LINUX_DISABLE_FILEMANAGER
+class QueueStorageIntegrationTest: UnitTest {
```
levibostian, line 88: Moved test functions from unit tests above into integration tests. 

I was having some errors running these test functions since I didn't mock enough. After reading the test functions, I felt the tests were higher quality by being integration tests using a real file system.

Tests/Common/Background Queue/QueueRunRequestTest.swift
```diff
@@ -119,3 +119,125 @@ class QueueRunRequestTest: UnitTest {
         XCTAssertEqual(storageMock.deleteCallsCount, 2)
     }
 }
+
+class QueueRunRequestIntegrationTest: IntegrationTest {
```
levibostian, line 123: Added integration tests to be more confident with the BQ refactor that I did. 

We only mock the queue runner so API calls are not actually performed.

Sources/Common/Background Queue/QueueRunRequest.swift
```diff
@@ -38,27 +38,35 @@ public class CioQueueRunRequest: QueueRunRequest {
...
     private func runTasks(
-        queueInventory: [QueueTaskMetadata],
-        queryTotalNumberTasks: Int,
+        lastRanTask: QueueTaskMetadata?,
         lastFailedTask: QueueTaskMetadata? = nil
     ) {
-        guard let nextTaskToRunInventoryItem = queryRunner.getNextTask(queueInventory, lastFailedTask: lastFailedTask)
+        // get the inventory before running each task. If a task was added to the queue while the last task was being
+        // executed, we can assert that new task will execute during this run.
+        let queueInventory = storage.getInventory()
```
levibostian, line 53: This is the major piece of the BQ refactor. Instead of iterating a snapshot of the queue tasks inventory, we re-query all of the queue inventory tasks before running each task. That way if new tasks get added to the queue while running the queue, they get executed this BQ run instead of waiting until a future run.

Sources/Tracking/CustomerIO.swift
```diff
@@ -140,62 +146,70 @@ public class CustomerIO: CustomerIOInstance {
...
     }
 
     /**
      Initialize the shared `instance` of `CustomerIO`.
-     Call this function when your app launches, before using `CustomerIO.instance`.
+     Call this function in your Notification Service Extension for the rich push feature.
      */
-    @available(iOSApplicationExtension, unavailable)
-    public static func initialize(
+    public static func initializeRichPush(
```
levibostian, line 175: Added an `initialize` that is dedicated for a Notification Service Extension. I tried to create 2 versions of `initialize()` instead of creating a separately named `initializeRichPush` but had no luck. The compiler wouldn't compile the SDK code. 

I tried to create 1 initialize that is only available to NSE and another only available to host iOS apps. However, because the method signature is identical for each target: 

```swift
// For host app
initialize(siteid: String, apiKey: String, region?: Region, config?)

// For NSE
initialize(siteid: String, apiKey: String, region?: Region, config?)
```

the compiler wouldn't allow this type of code. 

Another solution that we could have used is to remove some optional parameters to make the compiler notice there are 2 separate signatures:

```swift
// For host app
initialize(siteid: String, apiKey: String, region: Region, config: (SDKConfig) -> Void)

// For NSE
initialize(siteid: String, apiKey: String, region: Region, config: (RichPushSDKConfig) -> Void)
```

But I wasn't sure that we would like this solution, either. We would still be introducing a breaking change to customers.

Sources/Common/Background Queue/QueueQueryRunner.swift
```diff
@@ -15,18 +16,23 @@ internal class CioQueueQueryRunner: QueueQueryRunner {
...
+        let inventory = getInventoryTasksLeftToRun(inventory: inventory, lastRanTask: lastRanTask)
         if let lastFailedTask = lastFailedTask {
             updateCriteria(lastFailedTask: lastFailedTask)
         }
 
         // log *after* updating the criteria
         logger.debug("queue querying next task. criteria: \(queryCriteria)")
 
-        return queue.first(where: { doesTaskPassCriteria($0) })
+        return inventory.first(where: { doesTaskPassCriteria($0) })
```
ami-aman, line 35: Out of curiosity, what is difference between queue and inventory ?

Sources/Common/Background Queue/QueueRunRequest.swift
ami-aman: So would it be justified to say that if a BG starts executing the tasks, it will keep running until it reaches end of queue, no matter if it had few number of tasks at the time of start ?

Sources/Common/Background Queue/QueueRunRequest.swift
```diff
@@ -38,27 +38,35 @@ public class CioQueueRunRequest: QueueRunRequest {
...
+
+        guard let nextTaskToRunInventoryItem = queryRunner.getNextTask(
+            queueInventory,
+            lastRanTask: lastRanTask,
+            lastFailedTask: lastFailedTask
+        )
         else {
             // we hit the end of the current inventory. Done!
-            logger.debug("queue out of tasks to run.")
```
ami-aman, line 57: This log is helpful. Any reason that you thought of removing it ?

Sources/Common/Background Queue/QueueRunRequest.swift
```diff
@@ -67,14 +75,10 @@ public class CioQueueRunRequest: QueueRunRequest {
...
-            return goToNextTask(
-                queueInventory: queueInventory,
-                queryTotalNumberTasks: queryTotalNumberTasks,
-                lastFailedTask: nextTaskToRunInventoryItem
-            )
+            return goToNextTask(didTaskFail: true)
         }
 
-        logger.debug("queue tasks left to run: \(queueInventory.count) out of \(queryTotalNumberTasks)")
+        logger.debug("queue tasks left to run: \(queueInventory.count)")
```
ami-aman, line 81: I think this change answers my query [here](https://github.com/customerio/customerio-ios/pull/220/files#r1033444324)

Sources/Common/Background Queue/QueueRunRequest.swift
```diff
@@ -131,26 +128,16 @@ public class CioQueueRunRequest: QueueRunRequest {
...
-    ) {
-        var newInventory = queueInventory
-        newInventory.removeFirst()
-        runTasks(
-            queueInventory: newInventory,
-            queryTotalNumberTasks: queryTotalNumberTasks,
-            lastFailedTask: lastFailedTask
-        )
+    private func doneRunning() {
+        logger.debug("queue out of tasks to run.")
```
ami-aman, line 137: Ah, I see! And this answers my query I added [here](https://github.com/customerio/customerio-ios/pull/220/files#r1033445198)

Sources/Common/Background Queue/QueueRunRequest.swift
ami-aman: Feel free to resolve conversation if I understood it right in the comment [below](https://github.com/customerio/customerio-ios/pull/220/files#r1033448843)!

Sources/Common/Store/SdkConfig.swift
```diff
@@ -75,3 +75,60 @@ public struct SdkConfig {
...
+ 1. Instance of SDK inside of a Notification Service Extension does not have as many features to provide
+    compared to running in a host app. Therefore, we don't need to expose as many SDK config options to customers.
+ 2. The SDK code needs to override some configuration options when running inside of a Notication Service Extension.
+    We don't want customers to modify some of these overriden config options as it may effect some features of rich push.
+
+ Note: To not make the SDK code more complex, convert `RichPushSdkConfig` to an instance of `SdkConfig` when SDK is initialized.
+ The SDK should not have conditional logic handling different SDK config objects. The SDK should only have to handle `SdkConfig`.
+ */
+/// SDK configuration just for rich push feature of the SDK.
+public struct RichPushSdkConfig {
```
ami-aman, line 90: I believe, adding a use case code snippet would help !

Sources/Common/Store/SdkConfig.swift
```diff
@@ -75,3 +75,60 @@ public struct SdkConfig {
...
+            return RichPushSdkConfig(
+                trackingApiUrl: defaultSdkConfig.trackingApiUrl,
+                autoTrackPushEvents: defaultSdkConfig.autoTrackPushEvents,
+                logLevel: defaultSdkConfig.logLevel,
+                autoTrackDeviceAttributes: defaultSdkConfig.autoTrackDeviceAttributes
+            )
+        }
+    }
+
+    public func toSdkConfig() -> SdkConfig {
```
ami-aman, line 115: What is `toSdkConfig` used for ?

Sources/MessagingPush/MessagingPushImplementation.swift
```diff
@@ -8,167 +8,52 @@ import UserNotifications
...
 internal class MessagingPushImplementation: MessagingPushInstance {
     let siteId: SiteId
-    let profileStore: ProfileStore
-    let backgroundQueue: Queue
-    var globalDataStore: GlobalDataStore
     let logger: Logger
-    let sdkConfig: SdkConfig
     let jsonAdapter: JsonAdapter
-    let deviceAttributesProvider: DeviceAttributesProvider
-    let dateUtil: DateUtil
```
ami-aman, line 18: `sdkConfig`, `devceAttrbutesProvider`. `globalDataStore` were some of the properties that were seen in customer logs, who reported crashes in their app. I hope this helps in overcoming those issues

Sources/Tracking/CustomerIO.swift
ami-aman: > We use [our API to find your region](https://customer.io/docs/api/#operation/getRegion) for your workspace so we don't need to ask for it anymore.

Not sure if we would like to do this additional step ourselves but I found 👇  useful if one is available only in host target and the other one only in NSE

```
// For host app
initialize(siteid: String, apiKey: String, region: Region, config: (SDKConfig) -> Void)

// For NSE
initialize(siteid: String, apiKey: String, region: Region, config: (RichPushSDKConfig) -> Void) 
```

levibostian: I believe what you're referring to is "Another solution that we could have used is to remove some optional parameters to make the compiler notice there are 2 separate signatures:" from [my original post](https://github.com/customerio/customerio-ios/pull/220#discussion_r1029791315)?

Sources/Common/Background Queue/QueueQueryRunner.swift
```diff
@@ -15,18 +16,23 @@ internal class CioQueueQueryRunner: QueueQueryRunner {
...
-        guard !queue.isEmpty else {
+    func getNextTask(
+        _ inventory: [QueueTaskMetadata],
+        lastRanTask: QueueTaskMetadata?,
+        lastFailedTask: QueueTaskMetadata?
+    ) -> QueueTaskMetadata? {
+        guard !inventory.isEmpty else {
             return nil
         }
+        let inventory = getInventoryTasksLeftToRun(inventory: inventory, lastRanTask: lastRanTask)
```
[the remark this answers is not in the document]
levibostian: I intentionally modified the original parameter value to avoid human error. 

```swift
fun foo(inventory) {
  let modifiedInventory = ...
  
  // Now in this function, you have 2 "inventories": `inventory` and `modifiedInventory`. 
  // If you use the wrong one by accident, that introduces a bug into this function. 
  //
  // The function body at this time doesn't need 2 versions of inventory. It needs to only use a modified version. 
  // Therefore, if we overwrite the parameter with the modified version, we avoid the possible human error of using the wrong variable from happening. 
```

Sources/Common/Background Queue/QueueQueryRunner.swift
levibostian: The queue consists of: [an inventory of tasks](https://github.com/customerio/customerio-ios/blob/606edb3b62dc38e94667e6177bfef470ffbf6e6f/Sources/Common/Background%20Queue/QueueStorage.swift#L27) and [a list of tasks](https://github.com/customerio/customerio-ios/blob/606edb3b62dc38e94667e6177bfef470ffbf6e6f/Sources/Common/Background%20Queue/QueueStorage.swift#L33). 2 different data types for memory efficiency. When I talk about the queue, I often refer to "the inventory of tasks" in the queue. 

I made this variable name change because I felt it made more sense for the change that I made. I find this change very minor and an open to ideas if someone has a different idea.

Sources/Common/Store/SdkConfig.swift
levibostian: I added some comments to the class. See if those added comments answer your question.

Sources/Common/Background Queue/QueueStorage.swift
```diff
@@ -123,10 +123,24 @@ public class FileManagerQueueStorage: QueueStorage {
...
-            return (success: false, queueStatus: beforeCreateQueueStatus)
+            return CreateQueueStorageTaskResult(success: false, queueStatus: beforeCreateQueueStatus, createdTask: nil)
         }
 
-        return (success: true, queueStatus: afterCreateQueueStatus)
+        // It's more accurate for us to get the inventory item from the inventory instead of just returning
+        // newQueueItem. This is because queue storage when saving to storage might modify the metadata object
+        // such as removing milliseconds from Date. By getting the inventory item directly from device storage,
+        // we return the most accurate data on the inventory item.
+        let createdTask = getInventory().last!
```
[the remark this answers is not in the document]
levibostian: Fixed.

Sources/Common/Background Queue/QueueRunRequest.swift
levibostian: > no matter if it had few number of tasks at the time of start 

The queue will keep running until it reaches the end of the queue, even if more tasks *are added* while the queue is running. Tasks do not get removed from the queue so there isn't a scenario where there would be fewer tasks then at the time of start.

Sources/MessagingPush/MessagingPushImplementation.swift
levibostian: Agreed!

Tests/Shared/extension/QueueMockExtensions.swift
```diff
@@ -0,0 +1,11 @@
+@testable import CioTracking
+import Common
+import Foundation
+
+public extension QueueMock {
+    var deviceTokensDeleted: [String] {
+        addTaskReceivedInvocations.map {
```
jatinn, line 7: could use a comment, from what i gather you are filtering the list to non-empty values maybe?

Tests/Tracking/APITest.swift
jatinn: noticed and made a comment in another file regarding if there is a need for this pr to introduce these?

Sources/Tracking/CustomerIO.swift
ami-aman: Yes, correct ! I found that one more convenient to use from a dev's perspective.

Sources/Common/Store/SdkConfig.swift
ami-aman: Thanks for adding the comments. I am trying to understand what this mean ?
`the SDK code does not need to be modified to work with `RichPushSdkConfig`.`

Sources/Tracking/CustomerIO.swift
```diff
@@ -353,4 +367,30 @@ public class CustomerIO: CustomerIOInstance {
...
+     Delete the currently registered device token
+     */
+    public func deleteDeviceToken() {
+        implementation?.deleteDeviceToken()
+    }
+
+    /**
+     Track a push metric
+     */
+    public func trackMetric(
```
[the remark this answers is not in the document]
ami-aman: These are public functions already. Just that the file which contained them changed. Levi might have more to add to this.

Sources/Tracking/autogenerated/AutoDependencyInjection.generated.swift
```diff
@@ -74,6 +77,18 @@ extension DIGraph {
...
 
+    // DeviceAttributesProvider
+    var deviceAttributesProvider: DeviceAttributesProvider {
+        if let overridenDep = overrides[String(describing: DeviceAttributesProvider.self)] {
+            return overridenDep as! DeviceAttributesProvider
+        }
+        return newDeviceAttributesProvider
+    }
+
+    private var newDeviceAttributesProvider: DeviceAttributesProvider {
```
ami-aman, line 88: How about renaming to something meaningful or add a comment ?

Sources/Tracking/CustomerIO.swift
levibostian: Option 2 has been implemented.

Sources/MessagingPush/MessagingPushImplementation.swift
```diff
@@ -8,167 +8,52 @@ import UserNotifications
...
+        self.sdkInitializedUtil = SdkInitializedUtilImpl()
     }
 
-    /**
-     Register a new device token with Customer.io, associated with the current active customer. If there
-     is no active customer, this will fail to register the device
-     */
-    public func registerDeviceToken(_ deviceToken: String) {
-        addDeviceAttributes(deviceToken: deviceToken)
+    func deleteDeviceToken() {
```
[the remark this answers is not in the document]
levibostian: Assuming this is [a duplicate of this conversation](https://github.com/customerio/customerio-ios/pull/220#discussion_r1034126682)

Sources/Tracking/CustomerIO.swift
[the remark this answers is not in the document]
levibostian: These newly added functions are explained better [in this ticket](https://github.com/customerio/issues/issues/6934). We were already planning on doing this work of moving messaging push logic into Tracking module and I decided to include it in this PR because we would need to have both of these functions exist inside of a notification service extension: 

```swift
CustomerIO.initialize(...)
MessagingPush.initialize()
```

I was afraid that customers would forget to call `MessagingPush.initialize()` in a notification service extension and cause push metrics from rich push to not work. 

Since I knew we already made the decision to move the push metrics logic to the tracking module already, I decided the best time to do that was now in this PR. Close that ticket and fix this problem I am talking about here.

Tests/Common/Background Queue/QueueStorageTest.swift
```diff
@@ -92,17 +76,52 @@ class QueueStorageTest: UnitTest {
             blockingGroups: [.identifiedProfile(identifier: String.random)]
         )
 
-        XCTAssertEqual(fileStorageMock.saveCallsCount, 1) // only create task call
+        XCTAssertEqual(fileStorageMock.saveCallsCount, 2)
         XCTAssertFalse(actual.success)
         XCTAssertEqual(actual.queueStatus, QueueStatus(queueId: testSiteId, numTasksInQueue: 0))
```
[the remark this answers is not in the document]
levibostian: Reading this test over again, it's difficult to understand the test function's logic. It's not actually saving 2 tasks, it's saving 1. But in order to save 1 task, 2 calls to `fileStorage.save()` is done. Once for the task and once for the inventory. 

I updated the documentation for the test functions to make this better to understand.

Sources/Tracking/autogenerated/AutoDependencyInjection.generated.swift
levibostian: This file is automatically generated. `newX` property is meant to mean "new instance of X". 

If you look below, all dependencies in this file are ~~written~~ generated in the same way

Sources/Common/Store/SdkConfig.swift
levibostian: I updated the comment. Hope that helps.

ami-aman (approved):

PR #228, opened 2022-12-05
fix: image not shown in push notification using fcm

ami-aman:
Customers using FCM were not able to see the image on push notification despite using the initializer workaround being suggested. This PR has a fix for that (tested and confirmed by a customer). 

closes [#help (Chirag) images not showing (RN SDK + FCM + ObjC) #waiting-for-customer-reply](https://github.com/customerio/issues/issues/8640)

Complete each step to get your pull request merged in. [Learn more about the workflow this project uses](https://github.com/customerio/customerio-ios/blob/develop/docs/dev-notes/GIT-WORKFLOW.md). 
- [ ] [Assign members of your team](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/requesting-a-pull-request-review) to review the pull request. 
- [ ] Wait for pull request status checks to complete. If there are problems, fix them until you see that [all status checks are passing](https://external-content.duckduckgo.com/iu/?u=https%3A%2F%2Fsymfony.com%2Fdoc%2F4.3%2F_images%2Fdocs-pull-request-symfonycloud.png&f=1&nofb=1). 
- [ ] Wait until the pull request has been reviewed *and approved* by a teammate
- [ ] After pull request is approved, and you determine it's ready **add the label "Ready to merge"** to the pull request. A bot will *squash and merge* the pull request for you after the label is added.

PR #230, opened 2022-12-05
fix: push images and processing simple push

levibostian:
Closes: https://github.com/customerio/issues/issues/8640

This PR is the [same as Aman's](https://github.com/customerio/customerio-ios/pull/228) (thanks so much for finding this bug fix, Aman!) PR, but this one: 
* Fixes another bug that I found while testing *a simple push* in an iOS app.  
* This PR is a hotfix PR to release to customers faster.

### DIFF
--- a/Sources/MessagingPush/RichPush/MessagingPush+RichPush.swift
+++ b/Sources/MessagingPush/RichPush/MessagingPush+RichPush.swift
@@ -48,43 +48,67 @@ extension MessagingPushImplementation {
         _ request: UNNotificationRequest,
         withContentHandler contentHandler: @escaping (UNNotificationContent) -> Void
     ) -> Bool {
-        logger.info("did recieve notification request. Checking if message was a rich push sent from Customer.io...")
+        logger.info("did receive notification request. Checking if message was a push sent from Customer.io...")
         logger.debug("notification request: \(request.content.userInfo)")
 
-        if sdkConfigStore.config.autoTrackPushEvents,
-           let deliveryID: String = request.content.userInfo["CIO-Delivery-ID"] as? String,
-           let deviceToken: String = request.content.userInfo["CIO-Delivery-Token"] as? String {
+        guard let deliveryID: String = request.content.userInfo["CIO-Delivery-ID"] as? String,
+              let deviceToken: String = request.content.userInfo["CIO-Delivery-Token"] as? String
+        else {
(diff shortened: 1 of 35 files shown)


Sources/MessagingPush/RichPush/MessagingPush+RichPush.swift
```diff
@@ -48,43 +48,67 @@ extension MessagingPushImplementation {
...
         withContentHandler contentHandler: @escaping (UNNotificationContent) -> Void
     ) -> Bool {
-        logger.info("did recieve notification request. Checking if message was a rich push sent from Customer.io...")
+        logger.info("did receive notification request. Checking if message was a push sent from Customer.io...")
         logger.debug("notification request: \(request.content.userInfo)")
 
-        if sdkConfigStore.config.autoTrackPushEvents,
-           let deliveryID: String = request.content.userInfo["CIO-Delivery-ID"] as? String,
-           let deviceToken: String = request.content.userInfo["CIO-Delivery-Token"] as? String {
+        guard let deliveryID: String = request.content.userInfo["CIO-Delivery-ID"] as? String,
```
levibostian, line 54: The code changes done in this file addresses a bug that I found. If you use the new rich composer and send a simple push (just title and body), this code originally would process the simple push, say that the push was not sent from CIO, then would just return without calling the `contentHandler` resulting in the simple push not showing up until ~30 seconds later when the OS would kill our NSE (by calling `serviceExtensionTimeWillExpire`) and then show the push on the device. 

This change I made uses the deliverId and token to determine if the push was sent by CIO - simple or rich. Then, the processing will end early for simple pushes or continue for rich.

Sources/MessagingPushFCM/MessagingPushFCM.swift
```diff
@@ -92,15 +92,15 @@ public class MessagingPushFCM: MessagingPushFCMInstance {
         _ request: UNNotificationRequest,
         withContentHandler contentHandler: @escaping (UNNotificationContent) -> Void
     ) -> Bool {
-        messagingPush.didReceive(request, withContentHandler: contentHandler)
+        (messagingPush as MessagingPushInstance).didReceive(request, withContentHandler: contentHandler)
```
levibostian, line 95: In iOS v2, I made some refactors that no longer require this casting. But for v1 with the way the code is today, it still requires this casting to be done.

ami-aman (approved):

Issue #242, opened 2023-01-04
Crash in 1.2.8, MessagingInApp


lilidotshi:
**SDK version:** 
1.2.8

**Environment:** Development or Production
Production

**Are logs available?** 

```
CioMessagingInApp/MessagingInApp.swift:21: Fatal error: Unexpectedly found nil while implicitly unwrapping an Optional value
CioMessagingInApp/MessagingInApp.swift:21: Fatal error: Unexpectedly found nil while implicitly unwrapping an Optional value
```

**Describe the bug**

Upgraded from 1.2.7 to 1.2.8, and am seeing a crash when initializing the logger.debug. In particular `DIGraph.getInstance(siteId: siteId)` is returning nil and throwing the error

**To Reproduce**

Upgrade to 1.2.8
Turn on debug logging
Start app and hit the debug logging

**Expected behavior**

Initialization should work

**Screenshots**

No screenshot but stack trace is
```
#4	0x0000000106047e34 in MessagingInApp.diGraph.getter at SourcePackages/checkouts/customerio-ios/Sources/MessagingInApp/MessagingInApp.swift:21
#5	0x0000000106047f1c in MessagingInApp.logger.getter at SourcePackages/checkouts/customerio-ios/Sources/MessagingInApp/MessagingInApp.swift:33
#6	0x000000010604836c in MessagingInApp.initialize(organizationId:) at SourcePackages/checkouts/customerio-ios/Sources/MessagingInApp/MessagingInApp.swift:61
```

**Additional context**

levibostian:
Thank you for reporting this issue, @lilidotshi. I appreciate it. 

Looking through the stacktrace that you sent (thank you!), this does indeed look like a crash could happen. 

I do have 2 ideas for you to try:
* Can you make sure that every place in your code where you are calling `MessagingInApp.initialize(organizationId: )`, you are calling `CustomerIO.initialize()` before? `MessagingInApp.initialize` is assumed that it's being called after `CustomerIO.initialize`. 
* v2 of the iOS SDK has been released. v2 includes some architecture changes to the code which should eliminate this bug from happening all together. [See our migration doc](https://customer.io/docs/sdk/ios/migrate-upgrade/) to help you migrate to this backwards compatible release of our SDK.

levibostian:
Thanks for confirming that for me. 

I believe I have enough information to reproduce this issue. I will take it from here and get back to you when I have an update. 

[...]

PR #243, opened 2023-01-05
fix: 1.2.8 initialize in-app with debug logs no longer crashes SDK

levibostian:
Fixes: https://github.com/customerio/customerio-ios/issues/242

v1.2.8 has a bug where the diGraph instance isn't populated (when it should) in `CustomerIO` class. 

This PR refactors the `CustomerIO` class to populate the diGraph when the SDK is being initialized.

### DIFF
--- a/Sources/Tracking/CustomerIO.swift
+++ b/Sources/Tracking/CustomerIO.swift
@@ -107,22 +107,15 @@ public class CustomerIO: CustomerIOInstance {
     internal var implementation: CustomerIOImplementation?
     // The 1 place that DiGraph is strongly stored in memory for the SDK.
     // Exposed for `SdkInitializedUtil`. Not recommended to use this property directly.
-    internal var diGraph: DIGraph?
+    // This property might be populated when `implementation` is still nil. Do not use this value to check if SDK is initialized or not.
+    internal var diGraph: DIGraph!
 
     internal var globalData: GlobalDataStore = CioGlobalDataStore()
     // strong reference to repository to prevent garbage collection as it runs tasks in async.
     private var cleanupRepository: CleanupRepository?
-//
-//    private var threadUtil: ThreadUtil? {
-//        guard let siteId = siteId else { return nil }
-//
-//        return DIGraph.getInstance(siteId: siteId).threadUtil
-//    }
-//
-    private var logger: Logger? {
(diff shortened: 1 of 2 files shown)


Sources/Tracking/CustomerIO.swift
```diff
@@ -132,7 +125,8 @@ public class CustomerIO: CustomerIOInstance {
      */
     internal init() {
         if let siteId = globalData.sharedInstanceSiteId {
-            let diGraph = DIGraph.getInstance(siteId: siteId)
+            self.diGraph = DIGraph.getInstance(siteId: siteId)
+            
             let credentialsStore = diGraph.sdkCredentialsStore
             let logger = diGraph.logger
```
jatinn, line 131: shouldn't these variables be using `self.diGraph` instead of `diGraph` now?

Sources/Tracking/CustomerIO.swift
```diff
@@ -231,7 +228,7 @@ public class CustomerIO: CustomerIOInstance {
     /**
      Sets credentials on shared or non-shared instance.
      */
-    internal func setCredentials(siteId: String, apiKey: String, region: Region, diGraph: DIGraph) {
+    internal func setCredentials(siteId: String, apiKey: String, region: Region) {
         var credentialsStore = diGraph.sdkCredentialsStore
```
jatinn, line 232: looks like `diGraph` was removed from the params so does this not need to be `DIGraph.getInstance`?

Sources/Tracking/CustomerIO.swift
```diff
@@ -251,7 +248,6 @@ public class CustomerIO: CustomerIOInstance {
     }
 
     private func postInitialize(siteId: String) {
-        let diGraph = DIGraph.getInstance(siteId: siteId)
         let threadUtil = diGraph.threadUtil
```
jatinn, line 251: similar `diGraph` var removed so whats this refer to now

jatinn (approved):

Sources/Tracking/CustomerIO.swift
levibostian: Swift doesn't require you to include `self.` for properties. It automatically determines the scope for you. So, this line of code is referring to `self.`.

PR #245, opened 2023-01-06
fix: prevent stackoverflow while executing background queue with lots of tasks in it

levibostian:
Hotfix release for v2.

Changing algorithm of BQ to no longer use recursion. Recursion was causing a stackoverflow when the number of tasks in the BQ was large (~1000)

Fixes: https://github.com/customerio/customerio-reactnative/issues/54
Fixes: https://github.com/customerio/issues/issues/8917


Tests/Shared/UnitTest.swift
```diff
@@ -66,7 +66,7 @@ open class UnitTest: XCTestCase {
         setUp(enableLogs: false)
     }
 
-    public func setUp(enableLogs: Bool = false, modifySdkConfig: ((inout SdkConfig) -> Void)? = nil) {
+    open func setUp(enableLogs: Bool = false, modifySdkConfig: ((inout SdkConfig) -> Void)? = nil) {
```
jatinn, line 69: more for my curiosity whats the diff between `open & public` and why change it

jatinn (approved):

Tests/Shared/UnitTest.swift
levibostian: In Swift, functions/classes are `final` by default meaning they can't be overriden. 

`UnitTest` functions are overridden by other abstract classes so I changed it to `open` so `setUp` can be overriden by `IntegrationTest` class.

PR #247, opened 2023-01-10
fix: added reusable code for wrapper SDKs

Shahroz16:


Sources/Common/Extensions/StringExtensions.swift
```diff
@@ -71,4 +71,34 @@ public extension String {
...
+        case "debug":
+            return .debug
+        default:
+            return .error
+        }
+    }
+
+    func getEvent() -> Metric? {
+        Metric(rawValue: lowercased())
+    }
```
levibostian, line 103: Instead of these functions being in this file, I think they should exist in the file where the data type exists. 

`func getRegion() -> Region` goes inside of file `Region` is. 
`func getCioLogLevel() -> CioLogLevel` goes inside file `CioLogLevel` is
etc

Since the file that contains the data type has all of the logic for that data type, I think it makes the most sense for extension files to exist in that file.

levibostian: Also, can you add unit tests for each of these functions?

Sources/Common/Store/SdkConfig.swift
```diff
@@ -20,6 +20,19 @@ public struct SdkConfig {
         public static func create(region: Region) -> SdkConfig {
             SdkConfig(trackingApiUrl: region.productionTrackingUrl)
         }
+
+        public static func create(region: Region, params: [String: Any]) -> SdkConfig {
+            var config = SdkConfig(trackingApiUrl: region.productionTrackingUrl)
+            config.autoTrackDeviceAttributes = params[CustomerIOBuilderConfigKeys.Config.autoTrackDeviceAttributes] as? Bool ?? config.autoTrackDeviceAttributes
```
levibostian, line 26:
```suggestion
            if let autoTrackDeviceAttributes = params[CustomerIOBuilderConfigKeys.Config.autoTrackDeviceAttributes] as? Bool {
              config.autoTrackDeviceAttributes = autoTrackDeviceAttributes
            }
```

This suggestion is more of a swift design pattern. 

This idea can be applied to all of the lines here in this file. 

Also, I think it reads better then: `?? defaultValue` logic.

levibostian: Also, automated tests would be great to add for future maintainability of this function. 

Tests that would catch for us: If we ever add a new parameter to the `SdkConfig` object, a test automatically fails reminding us we need to update this function.

Sources/Tracking/CustomerIO.swift
```diff
@@ -173,6 +112,38 @@ public class CustomerIO: CustomerIOInstance {
...
+     Call this function when your app launches, before using `CustomerIO.instance`.
+     */
+    @available(iOSApplicationExtension, unavailable)
+    public static func initialize(
+        siteId: String,
+        apiKey: String,
+        region: Region,
+        params: [String: Any],
+        configure configureHandler: ((inout SdkConfig) -> Void)?
+    ) {
```
levibostian, line 126: Is there a different solution that we can introduce then adding another set of `initialize` functions? I want to avoid us having 3 or 4 different `initialize` functions if we can help it. Already needing 2 requires maintenance. 

# Idea 

What about adding a function to `SdkConfig` object: `SdkConfig.modify([String: String])` that's job is to modify the object? It would be similar to `SdkConfig.create()` that you added, but it modifies the object instead of creating a new one? 

That way, we don't need to add a new `CustomerIO.initialize()` function, but SDK wrappers can use the existing `initialize` functions like this:

```
CustomerIO.initialize(...) { config in 
  config.modify(["foo": "bar"])
}
```

Sources/Tracking/CustomerIOInstance.swift
```diff
@@ -0,0 +1,63 @@
+import Common
+import Foundation
+
+public protocol CustomerIOInstance: AutoMockable {
```
levibostian, line 4: I actually prefer this protocol existing in the `CustomerIO` class file. It's a similar design pattern we do in other classes in the code base and in Android where `interface` and implementation `class` exist in the same file. 

Is there a reason why we should move it to another file? Did you encounter a swift lint error maybe?

Sources/Common/Util/CustomerIOBuilderConfigKeys.swift
```diff
@@ -0,0 +1,18 @@
+import Foundation
+
+struct CustomerIOBuilderConfigKeys {
```
levibostian, line 3: I am not sure we need these constants. 

From this PR, it looks like the constant values are all used inside of the `SdkConfig.create()` function. If the constants are all encapsulated in 1 place for 1 purpose, I don't see a need for a set of constants like this to maintain. 

I prefer us to remove these constants and have hard-coded strings inside of `SdkConfig` since that's the data type that has the knowledge on how to construct an instance of `SdkConfig`.

levibostian (changes requested):

Sources/Common/Store/SdkConfig.swift
matt-frizzell: Question for both of our maybe, but when accessing nested attributes of an object like CustomerIOBuilderConfigKeys.Config.autoTrackDeviceAttributes is some of the syntax that I don't yet understand here protecting against cases where CustomerIOBuilderConfigKeys or CustomerIOBuilderConfigKeys.Config might not exist for some reason?

Sources/Tracking/CustomerIOInstance.swift
matt-frizzell: Hmm, definitely interested in this conversation. Without context having one file do one thing is often a good model. The other thing is if more than one thing implements the interface it makes sense. Like actual implementation and test implementation would be enough in my book. If we have a pattern we should probably stick to it, or if we are redefining it...do that.

Sources/Common/Extensions/StringExtensions.swift
Shahroz16: hmm, I don't think `string extensions` should be part of the `Model` of the class. I would change these extensions to their respective models then.

Sources/Common/Store/SdkConfig.swift
Shahroz16: > Also, I think it reads better then: ?? defaultValue logic.

Agreed, still new to swift design pattern, but this looks better with it 👍

Shahroz16: >CustomerIOBuilderConfigKeys or CustomerIOBuilderConfigKeys.Config might not exist for some reason.

I don't completely understand the questions, but these constants are keys, we expect for the `map`. 
They are hardcoded constants and if the map dont have these constant keys, then we won't read its config value.

Sources/Common/Util/CustomerIOBuilderConfigKeys.swift
Shahroz16: I am okay with both.

Sources/Tracking/CustomerIO.swift
Shahroz16: I am not fully onboard with the idea, of the `modify` method. 

I do understand that we don't want to increase the number of `initialize` methods, so what I am going to do is combine them with a default empty value, since they both do almost the same thing.

This way, we don't create new methods, previous method doesn't break either.

Sources/Tracking/CustomerIOInstance.swift
Shahroz16: It was the linting error, the class is getting bigger and bypassing the number of lines approved by the linter. So this protocol seemed like the best candidate to be taken out of the class. 

I am not against either but wanted to play by the linting rules too. :)

Sources/Common/Store/SdkConfig.swift
```diff
@@ -14,11 +14,47 @@ import Foundation
...
+        static let backgroundQueueSecondsDelay = "backgroundQueueSecondsDelay"
+    }
+
     // Used to create new instance of SdkConfig when the SDK is initialized.
     // Then, each property of the SdkConfig object can be modified by the user.
     public enum Factory {
-        public static func create(region: Region) -> SdkConfig {
-            SdkConfig(trackingApiUrl: region.productionTrackingUrl)
+        public static func create(region: Region, params: [String: Any] = [:]) -> SdkConfig {
+            var config = SdkConfig(trackingApiUrl: region.productionTrackingUrl)
```
ami-aman, line 36: Wondering if `_sdkWrapperConfig` has to be handled form wrappers or it's value also to be set within the native modules. I think the latter one can be done by feeding the values in the parameter itself. Has that been considered or mistakenly missed out ?

Sources/Common/Util/Log.swift
```diff
@@ -98,3 +98,35 @@ public class ConsoleLogger: Logger {
     public func error(_ message: String) {}
     #endif
 }
+
+extension CioLogLevel {
+    static func getLogLevel(for value: Int) -> CioLogLevel {
```
ami-aman, line 103: I know that in react native, we send `int` values of the log levels. Does Flutter send string values as in the next function? I think we can club these two together and maintain consistency across the wrappers i.e. either use strings or ints, because the two functions do the same thing. Thoughts?

Sources/Tracking/CustomerIO.swift
```diff
@@ -173,6 +112,38 @@ public class CustomerIO: CustomerIOInstance {
...
+     Initialize the shared `instance` of `CustomerIO` with optional `configParams`.
+     Call this function when your app launches, before using `CustomerIO.instance`.
+     */
+    @available(iOSApplicationExtension, unavailable)
+    public static func initialize(
+        siteId: String,
+        apiKey: String,
+        region: Region,
+        params: [String: Any] = [:],
+        configure configureHandler: ((inout SdkConfig) -> Void)?
```
ami-aman, line 125: What is the difference between `params` and `configure`? As I understand from the comment above, `param` is an optional configuration but for what ? When configure expects the same kind of configuration ? 🤔

Sources/Common/Store/SdkConfig.swift
Shahroz16: I kept it in wrappers because I feel like, that logic depends very much on wrappers, and it's not shared, e.g, React also caters to the case of expo while flutter doesn't care about it. 

For now, I just pushed the common actors so no logic is done by Native, but open to adding it if helpful.

Sources/Common/Util/Log.swift
Shahroz16: Good point, we can align it, just wanted to be wary of the changes we might have to do to the wrappers, could it be a breaking change? or is it something that can be handled internally?

Sources/Tracking/CustomerIO.swift
Shahroz16: `params` are pre-provided SDKConfig initializing parameters, they just take away the responsibility of parsing from clients to native SDK. 

While `configure` expects clients to give their configuration. 

And in order to reduce the number of initializer methods, we combined them.

Sources/Common/Store/SdkConfig.swift
ami-aman: IMO, we should add `_sdkWrapperConfig` to the common module. Distributing configuring the SDK based on just one property doesn't sound good to me. 

You are right, that in react native wrapper we are handling expo and react native both, but we pass only one value to `SdkWrapperConfig` based on condition

```
 var sdkSource = SdkWrapperConfig.Source.reactNative
        if source.lowercased() == "expo" {
            sdkSource = SdkWrapperConfig.Source.expo
        }
```
`SdkWrapperConfig(source: sdkSource, version: pversion )`

Sources/Common/Util/Log.swift
ami-aman: This can be handled by making a small change in `CioLogLevel` in the package and this won't require any change from a user's end. Let's make a ticket to make and track this change.

Sources/Tracking/CustomerIO.swift
ami-aman: Gotcha. I think it would be helpful to update the comment specifying it's usage only from wrappers, since it is a public function, our users will have access to it too.

Sources/Common/Util/Log.swift
levibostian: Would it be OK for the SDK wrappers to pass in a string instead of an Int data type? 

I am thinking that an Int could cause errors by accident in the future. If, for example, the native SDKs add/remove/edit one of the log levels. By the native SDK making a change to the log levels like this, we would need to modify this function and could introduce a bug into the SDK wrappers by accident. It looks easy to change one of the Int `case` statements and not consider the impact you could be making. 

If we used strings, I wonder if we could better avoid introducing a bug? 

```suggestion
    static func getLogLevel(for value: String) -> CioLogLevel {
      switch value {
      case "error": return .error
      case "info": return .info      
```

Sources/Tracking/CustomerIOInstance.swift
levibostian: Because the code base currently follows a pattern of: Have 1 file that contains both the protocol and it's class that inherits the protocol, I suggest we put this protocol back into `CustomerIO.swift` file. 

I understand that this could create a lint error with the `CustomerIO` class being long. The `CustomerIO` class has lots of functions inside of it and, luckily, this is the only class at the moment that triggers this lint error. 

I am not sure on the best solution for this lint error, but I would prefer that we keep the current code base pattern consistent. I am OK with ignoring the lint error in the `CustomerIO` class at this time because I am not sure on another good solution at the moment and this is the only file at this time that breaks this lint error. Making me think we don't have a major problem.

Sources/Common/Extensions/StringExtensions.swift
levibostian: I like the changes you made with where the functions are located. 

Could you add some automated tests for these functions added?

levibostian:
Could you add automated tests to this PR to cover the code that is added?

Sources/Tracking/CustomerIO.swift
levibostian: Did you make a change yet? I was expecting to no longer see a new `initialize` function in the `CustomerIO` class in this pull request as I currently do. 

Did you have an idea in mind for how we could avoid adding a new `initialize` function?

Sources/Common/Util/CustomerIOBuilderConfigKeys.swift
levibostian: I like how you moved the constants into the `SdkConfig` file where they are being used. I didn't consider that idea and I like it. 

I still don't see a need for the constants inside of `SdkConfig` yet, but I am also OK with the solution currently in the PR.

Sources/Common/Util/Log.swift
levibostian: I am seeing there is already a function that uses a `String` parameter. 

If the SDK wrappers currently use `Int` as the public API of the SDK, I would like to suggest the idea of converting `Int` into `String` inside of the SDK wrapper itself. Keep backwards compatibility while not adding logic to the native SDK that we need to maintain for a specific SDK wrapper?

Sources/Common/Store/SdkConfig.swift
Shahroz16: Added

Sources/Tracking/CustomerIO.swift
Shahroz16: Added

Shahroz16:
> Could you add automated tests to this PR to cover the code that is added?

@levibostian I have added already the tests, can you please explain what automated tests are you referring to?

Sources/Tracking/CustomerIOInstance.swift
Shahroz16: As long as those linting errors might not interrupt the CI workflow? 

Is there any way we can disable a particular lint error on a file? like the way, we have for Android?

levibostian (commented):
My apologies, I didn't see tests written at first. I take back my comment asking for automated tests as I see them now. 

---

One of my biggest concerns about the changes made to `SdkConfig` with being able to create SdkConfig object from `[String: Any]` is that someone may forget to update `SdkConfig.Factory.create` to allow `params` to set that modified SdkConfig property. 

Example: We add a new `SdkConfig` property `enableInApp: Bool`. We could quite easily forget to update `SdkConfig.Factory.create` to not include `params["enableInApp"]` to enable in-app with SDK wrappers. 

I wasn't sure on a good solution to prevent this potential human error so I decided to play with the code to come up with a solution. 

I [made a commit](https://github.com/customerio/customerio-ios/pull/247/commits/5f7e23c717d563e3875b1744141ac35e0a41b19e) to this branch with a suggestion. I added comments to that commit to highlight the logic behind it. To summarize, I modified the `SdkConfig` construction so if you ever add/remove/edit a property inside of `SdkConfig` object, the Swift compiler throws an error in `SdkConfig.Factory.create` reminding you that you need to add a `params` override and default value there.

levibostian:
I am sorry Shahroz if me pushing a commit to the branch caused confusion or frustration compared to an alternative action such as typing a suggestion in a comment in the PR. 

Sometimes I am most productive by digging into the code instead of giving up a concept. Commits can always be modified and reverted anyway 😉. 

I see that I broke tests by my push. I didn't consider that happening and I made a commit to attempt to fix them. 

Trying to make this PR a collaborative experience. We're learning on how best to do that.

levibostian:
Tests should be passing now. 

From the tests I wrote I found a couple bugs 👍!

Sources/Common/Store/SdkConfig.swift
```diff
@@ -14,11 +14,55 @@ import Foundation
...
+        static let logLevel = "logLevel"
+        static let autoTrackPushEvents = "autoTrackPushEvents"
+        static let backgroundQueueMinNumberOfTasks = "backgroundQueueMinNumberOfTasks"
+        static let backgroundQueueSecondsDelay = "backgroundQueueSecondsDelay"
+    }
+
+    public enum Package {
+        static let source = "source"
+        static let sourceVersion = "version"
+    }
```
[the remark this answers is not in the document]
Shahroz16: Open to it, I don't mind whether it's one or multiple, RN just has this division on their JS config maps, so wanted to be respectful of that. But it's cleaner and doesn't break their functionality, so we can go ahead with it.

Sources/Common/Util/Log.swift
Shahroz16: Removing the int option, as only RN uses `int` and it's not customer facing so they can change it internally.

Tests/Common/Extensions/CIOExtensionsTest.swift
```diff
@@ -0,0 +1,54 @@
...
+        let givenLogLevelNone = CioLogLevel.none
+        let expectedLogLevelNone = CioLogLevel.getLogLevel(for: "none")
+
+        XCTAssertEqual(givenLogLevelNone, expectedLogLevelNone)
+
+        let givenLogLevelError = CioLogLevel.error
+        let expectedLogLevelError = CioLogLevel.getLogLevel(for: "error")
+
+        XCTAssertEqual(givenLogLevelError, expectedLogLevelError)
+    }
```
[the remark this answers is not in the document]
Shahroz16: That's good to know, didn't know about `CaseIterable`, will add it.

Sources/Tracking/CustomerIO.swift
```diff
@@ -153,7 +92,24 @@ public class CustomerIO: CustomerIOInstance {
...
+
+     Currently this constructor is of internal use only.
+     */
+    @available(iOSApplicationExtension, unavailable)
+    public static func initialize(
+        siteId: String,
+        apiKey: String,
+        region: Region,
+        configParams: [String: Any] = [:],
+        configure configureHandler: ((inout SdkConfig) -> Void)?
```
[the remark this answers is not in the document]
Shahroz16: I don't have any better solution than that, I just wanted to take away the responsibility from wrapper clients to call the modify, as it might create confusion about when to modify it, i.e, after making the changes to `configureHandler` or which changes needs to go first. 

The order might be confusing, that's why I wanted to take away this confusion. 

But have both paramterts: `configParams`, `configure` might be confusing too, because Aman asked the same thing about whats the difference, so maybe,  `SdkConfig.modify()` might be a better call.

Shahroz16:
@levibostian thanks for the suggestions and the commit, the only confusion it causes was the implementation was a bit different than one of the suggested comments, for example, the `Keys` enum structure, the commit removed the enum completely. 

The only pending comment seems to be [regarding the initialization](https://github.com/customerio/customerio-ios/pull/247#discussion_r1081567422), so I'll update that, but it would also affect the changes you did it commit. So once you have time to review new changes, we can revisit it.

Tests/Tracking/APITest.swift
```diff
@@ -1,4 +1,5 @@
 import CioTracking // do not use `@testable` so we can test functions are made public and not `internal`.
+@testable import Common
```
levibostian, line 2:
```suggestion
```

The `APITest.swift` files are unique in that they are the only test classes in the code base that should not
* use `@testable` on any of the imports. 
* import a module that we don't expect customers to use themselves. 

These rules are put into place to prevent false positive tests. 

`APITest`s are designed to test the public API that customers would use. 

I tried adding some comments to these files with these rules but we can probably improve them for future devs besides yourself. Thanks for bringing this to our attention on something we can improve.

Tests/Tracking/APITest.swift
```diff
@@ -107,4 +108,120 @@ class TrackingAPITest: UnitTest {
...
+            CioSdkConfig.Keys.trackingApiUrl.rawValue: trackingApiUrl,
+            CioSdkConfig.Keys.autoTrackPushEvents.rawValue: false,
+            CioSdkConfig.Keys.backgroundQueueMinNumberOfTasks.rawValue: backgroundQueueMinNumberOfTasks,
+            CioSdkConfig.Keys.backgroundQueueSecondsDelay.rawValue: backgroundQueueSecondsDelay,
+            CioSdkConfig.Keys.backgroundQueueExpiredSeconds.rawValue: backgroundQueueExpiredSeconds,
+            CioSdkConfig.Keys.logLevel.rawValue: logLevel,
+            CioSdkConfig.Keys.autoTrackScreenViews.rawValue: autoTrackScreenViews,
+            CioSdkConfig.Keys.autoTrackDeviceAttributes.rawValue: autoTrackDeviceAttributes,
+            CioSdkConfig.Keys.source.rawValue: sdkWrapperSource,
+            CioSdkConfig.Keys.sourceVersion.rawValue: sdkWrapperVersion
```
levibostian, line 136: We will need to go back to using hard-coded strings in this test function. 

If this was a unit test, using constants would be nice. But because this is an API test, we should be using hard-coded strings. 

Why? 

If someone was to, for example, modify `CioSdkConfig.Keys.trackingApiUrl` to `CioSdkConfig.Keys.trackingUrl`. If this test function used constants, tests would pass. But if it was hard-coded, this test would fail. 

Since this test function's purpose is to make sure that we don't, by accident, introduce breaking changes to the public API of our SDK, we would want the test to fail reminding us that we might have a Flutter developer out there using `"trackingApiUrl"` in their app.

Tests/Tracking/APITest.swift
```diff
@@ -107,4 +108,120 @@ class TrackingAPITest: UnitTest {
...
+        XCTAssertEqual(actual.backgroundQueueMinNumberOfTasks, backgroundQueueMinNumberOfTasks)
+        XCTAssertEqual(actual.backgroundQueueSecondsDelay, backgroundQueueSecondsDelay)
+        XCTAssertEqual(actual.backgroundQueueExpiredSeconds, backgroundQueueExpiredSeconds)
+        XCTAssertEqual(actual.logLevel.rawValue, logLevel)
+        XCTAssertEqual(actual.autoTrackScreenViews, autoTrackScreenViews)
+        XCTAssertEqual(actual.autoTrackDeviceAttributes, autoTrackDeviceAttributes)
+        XCTAssertNotNil(actual._sdkWrapperConfig)
+    }
+
+    func test_SdkConfigFromMap_givenWrongKeys_expectDefaults() {
```
levibostian, line 190: I like this test. Great idea!

Tests/Common/Util/LogTest.swift
```diff
@@ -0,0 +1,22 @@
...
+            let givenString = logLevelCase.rawValue
+            let expected = logLevelCase
+            let actual = CioLogLevel.getLogLevel(for: givenString)
+            XCTAssertEqual(expected, actual)
+        }
+    }
+
+    func test_loglevel_givenIncorrectString_expectLogLevel() {
+        let givenLogLevelError: CioLogLevel? = nil
+        let expectedLogLevelError = CioLogLevel.getLogLevel(for: "debuggable")
```
levibostian, line 18:
```suggestion
        let expected: CioLogLevel? = nil
        let actual = CioLogLevel.getLogLevel(for: "debuggable")
```

The order of the 2 variables should be swapped. Our test gives "debuggable" as the test value and we then expect to get back `nil`. 

But, great test besides that.

Sources/Common/Util/Log.swift
```diff
@@ -98,3 +98,20 @@ public class ConsoleLogger: Logger {
...
+    static func getLogLevel(for value: String) -> CioLogLevel? {
+        switch value.lowercased() {
+        case "none":
+            return CioLogLevel.none
+        case "error":
+            return .error
+        case "info":
+            return .info
+        case "debug":
+            return .debug
```
levibostian, line 112:
```suggestion
        case CioLogLevel.none.rawValue:
            return CioLogLevel.none
        case CioLogLevel.error.rawValue:
            return .error
        case CioLogLevel.info.rawValue:
            return .info
        case CioLogLevel.debug.rawValue:
            return .debug
```

Since `CioLogLevel` now uses `: String`, we can use `rawValue`. This seems more maintainable, but maybe it would be a bad idea? 

The only flaw I see if we might by accident introduce a breaking change into the public API for our SDK wrappers. 

I suggest we add a test function to Common/APITest which makes sure that we keep the hard-coded string values "none", "error", "info", "debug" getting a log level. So we don't, by accident, break logging for SDK wrappers when they configure the SDK.

Sources/Tracking/CustomerIO.swift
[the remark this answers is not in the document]
levibostian: When I suggest the SDK wrappers contain code similar to:
```
// Code that could exist in the SDK wrappers: 
CustomerIO.initialize(siteId, apiKey, region} { config in 
  config.modify(["foo": "bar"]) 
}
```

I think it would be best if this code wasn't called directly from a customer in their app. I suggest that we provide a nice API to customers to use then the SDK wrapper internally calls this code. 

So, here might be an example piece of code for the RN SDK:

```js
// Here is code that a customer would have in their app. 
// This is a convenient javascript API that javascript developers would be familiar with. 
let config = new CIOConfig()
config.backgroundQueueExpiredSeconds = 0.5

CustomerIO.initialize(siteId: "", apiKey: "", region: "US", config)

// Here is code then that is internal in the RN SDK:
initialize(siteId, apikey, region, configObject) {
  // Note: this is not valid javascript code, but you get the example I am trying to show
  CustomerIO.initialize(siteId, apiKey, region} { config in 
     config.modify(configObject) 
  }
}
```

Would this solution work?

Sources/Common/Store/SdkConfig.swift
```diff
@@ -17,11 +17,58 @@ public struct SdkConfig {
...
+
+            // Construct object with all required parameters. Each config option should be provided from `params` or a default value.
+            // Define default values here in constructor instead of in struct properties. This is by design so in the future if we add
+            // a new SDK config option to the struct, we get a compiler error here in the constructor reminding us that we need to
+            // add a way for `params` to override the SDK config option.
+            var config = SdkConfig(
+                trackingApiUrl: (params[Keys.trackingApiUrl.rawValue] as? String) ?? region.productionTrackingUrl,
+                autoTrackPushEvents: (params[Keys.autoTrackPushEvents.rawValue] as? Bool) ?? true,
+                backgroundQueueMinNumberOfTasks: (params[Keys.backgroundQueueMinNumberOfTasks.rawValue] as? Int) ?? 10, backgroundQueueSecondsDelay: params[Keys.backgroundQueueSecondsDelay.rawValue] as? Seconds ?? 30,
+                backgroundQueueExpiredSeconds: params[Keys.backgroundQueueExpiredSeconds.rawValue] as? Seconds ?? Seconds.secondsFromDays(3),
```
levibostian, line 39: Something I want to point out, Shahroz, is that `Seconds` data type in a typealias we have setup to be a Double. 

I am not sure if the SDK Wrappers will be able to convert nicely to a Double? Maybe we need to instead have the SDK wrappers provide to us milliseconds in an Int data type that we convert into a Double in native SDK? 

I am thinking the conversion between SDK wrappers and native might not work well with Double data types?

Sources/Tracking/CustomerIO.swift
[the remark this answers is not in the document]
Shahroz16: This portion is used by our iOS native binding, Customer isn't going to see them, they send `config` once. 

But i have changed the structure.

Tests/Tracking/APITest.swift
```diff
@@ -124,57 +123,20 @@ class TrackingAPITest: UnitTest {
...
-            CioSdkConfig.Keys.backgroundQueueMinNumberOfTasks.rawValue: backgroundQueueMinNumberOfTasks,
-            CioSdkConfig.Keys.backgroundQueueSecondsDelay.rawValue: backgroundQueueSecondsDelay,
-            CioSdkConfig.Keys.backgroundQueueExpiredSeconds.rawValue: backgroundQueueExpiredSeconds,
-            CioSdkConfig.Keys.logLevel.rawValue: logLevel,
-            CioSdkConfig.Keys.autoTrackScreenViews.rawValue: autoTrackScreenViews,
-            CioSdkConfig.Keys.autoTrackDeviceAttributes.rawValue: autoTrackDeviceAttributes,
-            CioSdkConfig.Keys.source.rawValue: sdkWrapperSource,
-            CioSdkConfig.Keys.sourceVersion.rawValue: sdkWrapperVersion
+            "trackingApiUrl": trackingApiUrl,
+            "autoTrackPushEvents": false,
```
levibostian, line 127:
```suggestion
            "autoTrackPushEvents": autoTrackPushEvents,
```

levibostian (approved):

Issue #248, opened 2023-01-10
Memory/CPU Hang in 2.0.2


janremes:
**SDK version:** 2.0.2

**Environment:**  Production

**Are logs available?** 
In logs I see just theese

```
[CIO] (siteid:6cbe9) ℹ️ Customer.io SDK 2.0.2 initialized and ready to use for site id: xxx
[CIO] (siteid:6cbe9) ℹ️ registering device token xxx
[CIO] (siteid:6cbe9) ℹ️ adding queue task registerPushToken
[CIO] (siteid:6cbe9) ℹ️ queue met criteria to run automatically
```

**Describe the bug**
SDK is taking all available memory and CPU resources in version 2.0.2. App is then unusable. This is not happening in 2.0.1

**To Reproduce**
For me is enough just to launch app on device. In stack trace it seems to be failing when trying to update push token with customer.io

Here is stack trace I see on client.
<img width="791" alt="Screenshot 2023-01-10 at 14 35 42" src="https://user-images.githubusercontent.com/1276036/211565979-600ff9a7-7be0-4760-816d-d0cf4bf2f0a7.png">

levibostian:
Thanks for reporting this to us, @janremes. 

Sorry that you're experiencing a problem. We have this issue noted and am trying to reproduce it on our end. We'll get back to you with updates or questions when we have them. 

I am curious, how did you capture the stack trace and SDK logs that you attached? Are you running the app locally for testing purposes? If so, [can you enable debug logs for us to get more log messages and send those to us](https://github.com/customerio/customerio-ios/blob/develop/docs/SUPPORT.md#how-to-capture-logs-from-sdk-to-send-to-customerio-team)? 

In the mean time, I hope that version 2.0.1 will work for you!

janremes:
Yes I'm running app locally on a device. Logs via console are the same I attached before: 

[logs.txt](https://github.com/customerio/customerio-ios/files/10391988/logs.txt)

A few seconds after start, customer.io sdk will take all available memory and CPU. I can try to isolate it into small xcodeproject, it seems to be related to updating push token.

levibostian:
Thanks for including the logs. 

[If you enable the `debug` log level](https://github.com/customerio/customerio-ios/blob/develop/docs/SUPPORT.md#how-to-capture-logs-from-sdk-to-send-to-customerio-team) instead of the default, I believe you should see more logs in the console then you are seeing right now. 

[...]

PR #250, opened 2023-01-11
fix: async running BQ operations in loop

levibostian:
Closes: https://github.com/customerio/customerio-ios/issues/248

2.0.2 could encounter an infinite loop and eventually stackoverflow while executing the background queue. This was because the BQ while loop would has an infinite loop running while not waiting for the *async HTTP request* to finish. Running async code in a while loop requires some extra logic that was previously not present. 

Automated unit test added that verifies the BQ while loop runs async tasks in sequential order without causing an infinite loop. 

Still todo: 
- [X] QA smoke tests. RH branch `levi/2.0.2-stackoverflow` created to perform testing for this branch. Use builds from that branch.

### DIFF
--- a/Sources/Common/Background Queue/QueueRunRequest.swift
+++ b/Sources/Common/Background Queue/QueueRunRequest.swift
@@ -44,13 +44,14 @@ public class CioQueueRunRequest: QueueRunRequest {
     // Disable swiftlint because function at this time isn't too complex to need to make it smaller.
     // Many of the lines of this function are logging related.
     // swiftlint:disable:next function_body_length
-    private func runTasks() {
+    internal func runTasks() {
         var lastRanTask: QueueTaskMetadata?
         var lastFailedTask: QueueTaskMetadata?
         var continueRunnning = true
+        let whileLoopWait = DispatchGroup()
 
         // call when you're done with task
-        func goToNextTask(didTaskFail: Bool, taskJustExecuted: QueueTaskMetadata) {
+        func updateWhileLoopLogicVariables(didTaskFail: Bool, taskJustExecuted: QueueTaskMetadata) {
             lastFailedTask = didTaskFail ? taskJustExecuted : nil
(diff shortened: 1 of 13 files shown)


Sources/Common/Background Queue/QueueRunRequest.swift
```diff
@@ -38,106 +41,124 @@ public class CioQueueRunRequest: QueueRunRequest {
...
-        else {
-            // we hit the end of the current inventory. Done!
-            return doneRunning()
-        }
+    internal func runTasks() {
+        // Variables that power the logic of the queue run loop
+        var lastRanTask: QueueTaskMetadata?
+        var lastFailedTask: QueueTaskMetadata?
+        var continueRunnning = true
+        let whileLoopWait = DispatchGroup() // make async operations perform synchronously in while loop
```
Shahroz16, line 57: is there any specific reason we opted for `DispatchGroup` rather `Semaphores` I believe they do serve somewhat a similar purpose?

Shahroz16 (approved):
Added one comment, the rest looks good.

Sources/Common/Background Queue/QueueRunRequest.swift
levibostian: Using a semaphore didn't come to mind. In fact, I have never used a semaphore in iOS before. 

`DispatchGroup` is popular in iOS/mac development and will more then likely see the class used. I am not sure about the history of `DispatchGroup`, but I would guess that it's an abstraction that is provided by Apple to provide a more convenient API during development. 

The same could be said about using `DispatchQueue` for creating and executing code on a thread. Or, using `Core Data` for writing to a database. All are abstraction APIs provided by Apple that are popular to use in the development community. However, there are usually alternatives to these abstractions. 

Long story short, `DispatchGroup` is popular, iOS devs are familiar with it, and it's simple to use.

PR #253, opened 2023-01-21
fix: gist set user token incase identifier exists

Shahroz16:

### DIFF
--- a/Sources/MessagingInApp/MessagingInAppImplementation.swift
+++ b/Sources/MessagingInApp/MessagingInAppImplementation.swift
@@ -23,7 +23,7 @@ internal class MessagingInAppImplementation: MessagingInAppInstance {
     func initialize(organizationId: String) {
         inAppProvider.initialize(organizationId: organizationId, delegate: self)
 
-        if let identifier {
+        if let identifier = identifier {
             inAppProvider.setProfileIdentifier(identifier)
         }
     }
--- a/Sources/MessagingInApp/MessagingInAppImplementation.swift
+++ b/Sources/MessagingInApp/MessagingInAppImplementation.swift
@@ -10,15 +10,22 @@ internal class MessagingInAppImplementation: MessagingInAppInstance {
     private var inAppProvider: InAppProvider
     private var eventListener: InAppEventListener?
 
+    private var identifier: String?
+
     init(diGraph: DIGraph) {
         self.logger = diGraph.logger
         self.queue = diGraph.queue
         self.jsonAdapter = diGraph.jsonAdapter
         self.inAppProvider = diGraph.inAppProvider
+        self.identifier = diGraph.profileStore.identifier
     }
 
     func initialize(organizationId: String) {
         inAppProvider.initialize(organizationId: organizationId, delegate: self)
+
+        if let identifier {
+            inAppProvider.setProfileIdentifier(identifier)
+        }
     }
 
     func initialize(organizationId: String, eventListener: InAppEventListener) {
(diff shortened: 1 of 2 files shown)


Sources/MessagingInApp/MessagingInAppImplementation.swift
```diff
@@ -10,15 +10,22 @@ internal class MessagingInAppImplementation: MessagingInAppInstance {
...
     private var eventListener: InAppEventListener?
 
+    private var identifier: String?
+
     init(diGraph: DIGraph) {
         self.logger = diGraph.logger
         self.queue = diGraph.queue
         self.jsonAdapter = diGraph.jsonAdapter
         self.inAppProvider = diGraph.inAppProvider
+        self.identifier = diGraph.profileStore.identifier
```
levibostian, line 20:
```suggestion
        self.profileStore = diGraph.profileStore
```

Instead of storing the identifier as a property in the `MessagingInAppImplementation` class, I suggest that we store the dependency itself. 

Biggest reason being that it makes the automated tests easier to maintain. Logic inside of class constructors (such as accessing property `.identifier` from profile store dependency) makes test classes/functions more work to setup since you'll need to mock `profileStore.identifier` on every test function, even if that function doesn't need the identifier. I find it less work to only need to mock `profileStore.identifier` in the functions that access `profileStore.identifier`.

Tests/MessagingInApp/MessagingInAppImplementationTest.swift
```diff
@@ -31,6 +34,18 @@ class MessagingInAppImplementationTest: UnitTest {
...
+    func test_initialize_givenIdentifier_expectGistSetProfileIdentifier() {
+        profileStoreMock.identifier = "identifier"
+
+        let givenId = String.random
+
+        let instance = MessagingInAppImplementation(diGraph: diGraph)
+        instance.initialize(organizationId: givenId)
+        XCTAssertTrue(inAppProviderMock.initializeCalled)
+        XCTAssertTrue(inAppProviderMock.setProfileIdentifierCalled)
+        XCTAssertEqual(inAppProviderMock.setProfileIdentifierReceivedArguments, "identifier")
```
levibostian, line 46:
```suggestion
        let givenProfileIdentifiedInSdk = String.random
        profileStoreMock.identifier = givenProfileIdentifiedInSdk

        let givenId = String.random

        let instance = MessagingInAppImplementation(diGraph: diGraph)
        instance.initialize(organizationId: givenId)
        XCTAssertTrue(inAppProviderMock.initializeCalled)
        XCTAssertTrue(inAppProviderMock.setProfileIdentifierCalled)
        XCTAssertEqual(inAppProviderMock.setProfileIdentifierReceivedArguments, givenProfileIdentifiedInSdk)
```

I have encountered false positive test results before using hard-coded values in tests. I try to make values random to prevent these scenarios from happening. I find it makes test quality a small bit better.

Sources/MessagingInApp/MessagingInAppImplementation.swift
```diff
@@ -10,15 +10,22 @@ internal class MessagingInAppImplementation: MessagingInAppInstance {
...
     }
 
     func initialize(organizationId: String) {
         inAppProvider.initialize(organizationId: organizationId, delegate: self)
+
+        if let identifier = identifier {
```
levibostian, line 26: If you could add a comment to this similar to the comment added in the Android PR, that would be greatly appreciated.

levibostian (approved):

Sources/MessagingInApp/MessagingInAppImplementation.swift
Shahroz16: That's a good point, I would say, not just in the case of "testing" but this also promotes loose coupling between classes. 👍

PR #258, opened 2023-02-02
ci: test deployments on each pr

levibostian:
Closes: https://github.com/customerio/issues/issues/9289

# This PR does 2 great things for us:

1. Helps us feel more confident that a deployment will execute successfully once we merge a PR. 
2. Make the deployment process less intimidating to team members. Now anyone on the team can make a new branch, modify the deployment configuration, open up a draft PR, and see what would happen. All before merging to production. 
3. We get to preview what the new semantic version of the software will be before we merge. 

I added lots of comments on the CI configuration files in an effort to make the process easy to understand by everyone on the team. 

# How it works

On every commit added to a PR, the CI server....

* *Simulates* merging the PR. This means that a temporary branch is created on the github repository where it's git commit history resembles the commit history if a PR was to be squashed + merged. 
* On this new branch, the deployment process is executed in dry-run mode meaning that the deployment process is executed, but nothing is pushed to production servers. 
* The CI server makes really convenient comments on the PR during the whole process. Here are what these messages look like:

CleanShot 2023-02-16 at 12 14 45@2x
> indicating that the test deployment has begun

CleanShot 2023-02-16 at 12 15 07@2x
> indicating the test deployment has begun

CleanShot 2023-02-16 at 12 15 21@2x
> after dry-run, a summary of what would happen

CleanShot 2023-02-16 at 12 16 52@2x
> after dry-run, a summary of what would happen


levibostian:
Made this a PR so I can test the CI server (it only runs on PRs). 

This PR not yet ready for review and merging.

levibostian:
`develop` branch is being removed as part of the strategy to no longer do alpha or beta deployments. This PR is now pointing to `beta` as that's what the old `develop` was.

.github/workflows/simulate-merge-pr.yml
```diff
@@ -0,0 +1,95 @@
+name: Simulate merging a PR 
+
+# Trigger when PR created, when new commits added to it (synchronize), or when the PR title has been modified since the PR title changes behavior of a simulated merge 
+on:
+  pull_request:
+    types: [opened, edited, synchronize]
```
nunofgs, line 6: This is only relevant when [creating a PR against](https://github.com/customerio/customerio-ios/blob/beta/.github/workflows/deploy-sdk.yml#L5) `main` and `v1` right?

Could we do this?

```suggestion
  pull_request:
    types: [opened, edited, synchronize]
    branches:
      - main
      - v1
```

.github/workflows/test-deployment.yml
```diff
@@ -0,0 +1,129 @@
+name: Test deployment
```
nunofgs, line 1: What do you think about renaming this to `test-release`?

No deployment is actually simulated, only versioning and releasing (ie: no assets are uploaded, etc.)

.github/workflows/simulate-merge-pr.yml
```diff
@@ -0,0 +1,95 @@
+name: Simulate merging a PR
```
nunofgs, line 1: 🤔 feels like this file and `test-deployment.yml` could be merged. I don't imagine we'd run the `test-deployment` workflow separately.

If so, I would suggest renaming the file to `simulate-release.yml`.

What do you think?

nunofgs (changes requested):
Pretty cool. I can see how it could be useful. Left some comments for you.

nunofgs:
> This question comes from a place that lacks knowledge, so bear with me. Are the yaml files here specifically for this repository or do they represent generic steps that we plan to implement everywhere and so could be part of a CI repo that this might draw from?

@matt-frizzell at the moment it would only apply to this repository. It can definitely be abstracted to a separate actions repo for reusability, but I would recommend trying out the workflow first and making sure that it's useful for the team.

matt-frizzell:
> > This question comes from a place that lacks knowledge, so bear with me. Are the yaml files here specifically for this repository or do they represent generic steps that we plan to implement everywhere and so could be part of a CI repo that this might draw from?
> 
> @matt-frizzell at the moment it would only apply to this repository. It can definitely be abstracted to a separate actions repo for reusability, but I would recommend trying out the workflow first and making sure that it's useful for the team.

Got it. I had hoped that all work for moving our CI forward would be done generically and drawn from project by project, as we were ready. I think I probably just need a 5-10 catch-up meeting on the current plan so I can understand where we are going from here.

Great steps forward, and very exciting to see!

.github/workflows/test.yml
```diff
@@ -1,12 +1,40 @@
...
+          filters: |
+            src:
+              - 'Sources/**'
+              - 'Tests/**'
+
   older-xcode-test:
+    needs: files-changed
+    if: needs.files-changed.outputs.src == 'true' 
     strategy:
       matrix:
```
Shahroz16, line 39: Thank you for this, I didn't know and had to look up what it does 😄 
But from the example, I have seen, it looks like, we don't have to create two jobs, we can use 1 job and not duplicate the steps. You can also provide OS information in `matric`

```
matrix:
        os: [macos-10.14, macos-10.15]
        xcode-version: [11.3, 11.4]
```

.github/workflows/simulate-merge-pr.yml
```diff
@@ -0,0 +1,95 @@
...
+  PR_NUMBER: ${{ github.event.pull_request.number }}
+
+jobs:
+  simulate-merge-pr:
+    runs-on: ubuntu-latest
+    name: Simulate merging PR
+    permissions: 
+      contents: write # for pushing commits and tags 
+    steps:
+    - uses: actions/checkout@v3
```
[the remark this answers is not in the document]
nunofgs: 🤔 interesting. I haven't found that to be the case. They even list this on [the readme](https://github.com/actions/checkout#checkout-v3): 

> Only a single commit is fetched by default, for the ref/SHA that triggered the workflow. 

In any case, I'd prefer not to use older versions unless absolutely necessary. We should measure the perf impact and investigate the underlying issue.

.github/workflows/simulate-merge-pr.yml
```diff
@@ -0,0 +1,95 @@
...
+
+    # Squash all commits of the current branch into 1 commit. Each command below is executed as 1 step in this job 
+    # to act as comments explaining each step of the process. 
+    # Help with how to squash and merge a PR: https://stackoverflow.com/a/5201642
+    - name: Fetch all branches so we can reference origin/main branch
+      run: git fetch --depth=1
```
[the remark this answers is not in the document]
nunofgs: This can be moved to the actions/checkout above (with `fetchDepth: 0`)

.github/workflows/simulate-merge-pr.yml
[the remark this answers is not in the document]
Shahroz16: [This](https://github.com/actions/checkout/compare/v2.4.2...v3.0.2#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L17) is where I got this from. 

But maybe it was just a bad PR change.

levibostian:
Closing, at least for now. Solution may need a different solution implementation. Will address when task becomes priority for team.

Issue #262, opened 2023-02-07
Universal links are not properly supported on push notifications


tfcporciuncula:
**SDK version:** 2.0.4

**Environment:** Development or Production

**Are logs available?** 
N/A

**Describe the bug**
When tapping on a push notification with a universal link as the deep link URL, the app opens but then the browser opens right afterwards.

**To Reproduce**
- Send a universal link as a deep link in a push notification
- Tap on the notification
- Observe the app opens and then the browser opens right afterwards

**Expected behavior**
When tapping on a notification with a universal link, the app opens and the browser doesn't. Universal links work great outside of notifications, so my expectation is that they should work in notifications as well so we can use the same URLs everywhere instead of having to use a custom scheme for the deep links we send in notifications.

The current documentation isn't very clear about this, but based on what I see here it seems like our only option might be to use a custom scheme?

![Screen Shot 2023-02-02 at 10 01 28 AM](https://user-images.githubusercontent.com/750114/217283701-eed3cfc2-0963-4b0c-b885-ab61fffae2ec.png)

HTTPS URLs work great on Android, so it would be nice if they work well on iOS as well. And based on what your docs say [here](https://customer.io/docs/sdk/ios/push/#deep-links), it really seems like universal links were supposed to be supported.

**Screenshots**
N/A

**Additional context**
According to [Apple's docs](https://developer.apple.com/library/archive/documentation/General/Conceptual/AppSearch/UniversalLinks.html) (referenced by you [here](https://github.com/customerio/customerio-ios/issues/205#issuecomment-1283794187) as well):

>It’s important to understand that if your app uses [openURL:](https://developer.apple.com/documentation/uikit/uiapplication/1622961-openurl) to open a universal link to your website, the link does not open in your app. In this scenario, iOS recognizes that the call originates from your app and therefore should not be handled as a universal link by your app.

However, it seems that's what the SDK is doing [here](https://github.com/customerio/customerio-ios/blob/43b3e9706b152fe2dead882336a3f416ba400b73/Sources/MessagingPush/MessagingPush%2BAppDelegate.swift#L68):

```swift
if let deepLinkurl = pushContent.deepLink {
    UIApplication.shared.open(url: deepLinkurl)
}
```

Is there anything more sophisticated the SDK could do to make sure universal links will be handled properly (e.g. something like [this](https://stackoverflow.com/questions/46834222/how-can-i-use-universal-links-when-i-call-openurl-inside-my-own-app))? Or anything we can do on our end?

[...]

PR #265, opened 2023-02-09
fix: universal links when touch a push notification open host app

levibostian:
Closes: https://github.com/customerio/customerio-ios/issues/262

Before merging
- [x] Code review
- [X] ~Docs change helping customers setup universal links deep linking. Including the need to implement the function `func application(_ application: UIApplication, continue userActivity: NSUserActivity, restorationHandler: @escaping ([UIUserActivityRestoring]?) -> Void) -> Bool`.~ - [docs pr made](https://github.com/customerio/docs/pull/938)

This PR fixes this bug by:
* If customer's app implements the function `func application(_ application: UIApplication, continue userActivity: NSUserActivity, restorationHandler: @escaping ([UIUserActivityRestoring]?) -> Void) -> Bool` in their app to handle deep links (and they return the correct value from the function), we will attempt to call that function first. 
* If they do not implement that function *or* they return `false` from the function, we fallback to opening the URL system-wide as the SDK does already today. 

This solution should keep backwards compatibility with the SDK customers today - as long as customers are using `func application(_ application: UIApplication, continue userActivity: NSUserActivity, restorationHandler: @escaping ([UIUserActivityRestoring]?) -> Void) -> Bool` correctly (returning an accurate value), this SDK should not be introducing a breaking change for anyone. 

QA tests:
- [x] Touch a push without a deep link in it. Expect host app opens, doesn't show "Switch workspace" screen. 
- [x] Touch a push with deep link using app scheme url `remote-habits://switch_workspace?site_id=app&api_key=scheme`. Expect switch workspace screen shows with "app" and "scheme" filled into the text fields on the screen. 

This tests backwards compatibility of our SDK. 

- [x] Touch a push with a deep link using universal links `https://remotehabits.page.link/switch_workspace?site_id=universal&api_key=link`. Expect switch workspace screen shows with "universal" and "link" filled into the text fields on the screen. 

This change _might_ not work for SwiftUI apps. We do not officially support SwiftUI apps at this time so we can fix this at a later time if needed.


Sources/MessagingPush/Util/DeepLinkUtil.swift
```diff
@@ -0,0 +1,21 @@
...
+import Foundation
+
+protocol DeepLinkUtil: AutoMockable {
+    func isLinkValidNSUserActivityLink(_ url: URL) -> Bool
+}
+
+// sourcery: InjectRegister = "DeepLinkUtil"
+class DeepLinkUtilImpl: DeepLinkUtil {
+    // When using `NSUserActivity.webpageURL`, only certain URL schemes are allowed. An exception will be thrown otherwise which is why we have this function.
+    func isLinkValidNSUserActivityLink(_ url: URL) -> Bool {
```
Shahroz16, line 11: maybe I don't understand the problem statement correctly, here for which i apologize. 

But from what I understand, 

```
// to ger a URL from string, the method that gets used is 
URL(string: stringUrl)
```

and this already validates if it's a valid URL, that is it has the `allowedSchemes`

```
if let url = URL(string: stringUrl) {
    print(url)
} else {
    print("Invalid URL")
}
```

and the method, `isLinkValidNSUserActivityLink` is expecting an URL as well, so wouldn't this be already vetted? otherwise, the URL would always be `nil` ?

Sources/MessagingPush/MessagingPush+AppDelegate.swift
```diff
@@ -65,7 +65,27 @@ extension MessagingPushImplementation {
...
-                UIApplication.shared.open(url: deepLinkurl)
+                // First, try to open the link inside of the host app. This is to keep compatability with Universal Links.
+                // Learn more of edge case: https://github.com/customerio/customerio-ios/issues/262
+                // Fallback to opening the URL system-wide if fail to open link in host app.
+                // Customers with Universal Links in their app will need to add this function to their `AppDelegate` which will get called with deep link:
+                // func application(_ application: UIApplication, continue userActivity: NSUserActivity, restorationHandler: @escaping ([UIUserActivityRestoring]?) -> Void) -> Bool
+
+                var didHostAppHandleLink = false
+                if deepLinkUtil.isLinkValidNSUserActivityLink(deepLinkurl) {
+                    logger.debug("Found a deep link inside of the push notification. Attempting to open deep link in host app, first.")
```
Shahroz16, line 76: Like mentioned in later [comment](https://github.com/customerio/customerio-ios/pull/265/files#r1102059231) 

If its not valid URL, it will be `nil`

Sources/MessagingPush/MessagingPush+AppDelegate.swift
```diff
@@ -65,7 +65,27 @@ extension MessagingPushImplementation {
         switch response.actionIdentifier {
         case UNNotificationDefaultActionIdentifier: // push notification was touched.
             if let deepLinkurl = pushContent.deepLink {
-                UIApplication.shared.open(url: deepLinkurl)
+                // First, try to open the link inside of the host app. This is to keep compatability with Universal Links.
+                // Learn more of edge case: https://github.com/customerio/customerio-ios/issues/262
+                // Fallback to opening the URL system-wide if fail to open link in host app.
+                // Customers with Universal Links in their app will need to add this function to their `AppDelegate` which will get called with deep link:
+                // func application(_ application: UIApplication, continue userActivity: NSUserActivity, restorationHandler: @escaping ([UIUserActivityRestoring]?) -> Void) -> Bool
+
```
Shahroz16, line 73: I saw there is already a method available `UIApplication.shared.canOpenURL`. Can we utilize it maybe? 

```suggestion
if UIApplication.shared.canOpenURL(deepLinkurl) {
            UIApplication.shared.open(deepLinkurl, options: [:], completionHandler: nil)
            return true
        } else {
            // Fall back to opening the URL system-wide
            UIApplication.shared.open(deepLinkurl, options: [:], completionHandler: nil)
            return false
        }
```

levibostian: I previously tried `canOpenURL` but it doesn't solve the problem we are having. At least, `canOpenURL` doesn't work for universal links. It might work for app schemes, but we are not experiencing any issues with app schemes at the moment.

Sources/MessagingPush/Util/DeepLinkUtil.swift
levibostian: I wonder if a class or function rename would help better understand the problem. 

This function's job is not to test if a URL is valid. The function's job is to verify that a URL is using only `http` or `https` for the scheme part of the URL string. 

This is needed because `MessagingPush+AppDelegate` file's code changes are using `NSUserActivity.webpageURL` and an exception is thrown if [a URL is using the wrong scheme](https://developer.apple.com/documentation/foundation/nsuseractivity/1418086-webpageurl). 

Do the tests for this class help explain the use of this function? 

Would a better comment help? Renaming of this function?

Shahroz16: Thanks, for the explanation, the [fixed test](https://github.com/customerio/customerio-ios/pull/265/commits/f68e813f7846d96b1b2e10a65192a1bc07f4b9a1) removed the confusion. 

A class renaming might help here because from my understanding there are two types of links, `deeplink` and `universal link`. And I was unsure where `NSUserActivity` lies in all of this. 

But i can't think of any other name either, so I am okay if you go with this or if you are able to think of another one.

Shahroz16 (approved):

Sources/MessagingPush/Util/DeepLinkUtil.swift
levibostian: Deep links is an umbrella term. A deep link could be a universal link or an app scheme link. 

So, it's not deep link *or* universal link as universal links are deep links. I did add more details to deep links docs [in this pr](https://github.com/customerio/docs/pull/938) might be helpful for you to read those changes.

tfcporciuncula:
>This change might not work for SwiftUI apps. We do not officially support SwiftUI apps at this time so we can fix this at a later time if needed.

Can you elaborate on why this change might not work for SwiftUI apps? Also, I'm surprised to learn you don't officially support them. Is this mentioned anywhere in the docs? I can't find anything about this [there](https://customer.io/docs/sdk/ios/getting-started/).

Sources/MessagingPush/MessagingPush+AppDelegate.swift
```diff
@@ -65,7 +65,27 @@ extension MessagingPushImplementation {
...
+                var didHostAppHandleLink = false
+                if deepLinkUtil.isLinkValidNSUserActivityLink(deepLinkurl) {
+                    logger.debug("Found a deep link inside of the push notification. Attempting to open deep link in host app, first.")
+
+                    let openLinkInHostAppActivity = NSUserActivity(activityType: NSUserActivityTypeBrowsingWeb)
+                    openLinkInHostAppActivity.webpageURL = deepLinkurl
+
+                    let didHostAppHandleLink = UIApplication.shared.delegate?.application?(UIApplication.shared, continue: openLinkInHostAppActivity, restorationHandler: { _ in }) ?? false
```
tfcporciuncula, line 81: Do we really want to redefine `didHostAppHandleLink` here? This means we're never changing the original `didHostAppHandleLink` from line 74, so it'll always be `false` when we check below on line 84.

Universal links are still not working for us, and based on our tests this is the culprit.

levibostian:
To reply [to your message](https://github.com/customerio/customerio-ios/pull/265#issuecomment-1427705322), @tfcporciuncula. 

I can see how my original comment regarding this pull request and SwiftUI could trigger some questions. My original comment was poorly stated and was not accurate in regards to our SDK and it's features. 

We do not expect issues to be encountered when using a SwiftUI app with our SDK. If issues do arise, we are open to resolving them. If you or someone you know use SwiftUI with our SDK and have problems, please submit issues to us and we will help get those resolved!

PR #275, opened 2023-02-24
build: remove SPM lockfile from git repo

levibostian:
Build system lock files are meant to be used by applications to make builds predictable and consistent. Our SDK code base is designed to be compatible with with a range of our dependencies (example: Gist SDK >= 3.0, < 4.0) and therefore, a lock file such as `Package.resolved` (1) does not add any value to our code and (2) is annoying when it gets added to PR code reviews. 

To do: 
- [x] Confirm that this branch is able to be installed into RH with this file removed. I don't expect any issues, but SPM might require this file when installing dependencies? I did review the git repos of some popular iOS libraries and many did not have this file in their git repos.


.gitignore
```diff
@@ -1,6 +1,7 @@
 .DS_Store
 test-report.*
 *.xcresult/
+Package.resolved
```
Shahroz16, line 4: Thank you for this Levi, I have been reading up a lot on the SPM lock file and its benefits and its advantages outweigh the annoyance. 

- Even though we allow a range of internal dependencies and imagineing there was a faulty version in it. It can help user to reproduce and diagnose any issues that arise due to changes in the dependencies. Because the lock file specifies the exact versions of your dependencies that were used when building your SDK. 
- Secondly, the lock file can help to ensure that all collaborators are using the same versions of dependencies when building and testing the SDK.

Shahroz16: I am okay with trying this.

Shahroz16 (approved):

PR #276, opened 2023-03-01
fix: delete tasks that returns http 400 

Shahroz16:
closes : https://github.com/customerio/issues/issues/9443


Sources/Common/Background Queue/QueueRunRequest.swift
```diff
@@ -135,7 +135,14 @@ public class CioQueueRunRequest: QueueRunRequest {
                         self.logger.info("queue is quitting early because all HTTP requests are paused.")
 
                         doneRunning()
-                        break
```
Shahroz16, line 138: It was very confusing, took me some time to grasp it as well. 

So there is a loop and inside it, there is a switch statement, `break` statement caused confusion so I opted for exhaustive approach.

Sources/Common/Background Queue/QueueRunRequest.swift
```diff
@@ -135,7 +135,14 @@ public class CioQueueRunRequest: QueueRunRequest {
...
+                    } else if case .badRequest = error {
+                        self.logger.debug("queue task \(nextTaskStorageId) failed with 400")
+
+                        self.logger.debug("queue deleting task \(nextTaskStorageId) because it will always fail")
+                        _ = self.storage.delete(storageId: nextTaskToRunInventoryItem.taskPersistedId)
+
+                        // since failed task isn't being saved, no need to update `lastFailedTask`
+                        updateWhileLoopLogicVariables(didTaskFail: false, taskJustExecuted: nextTaskToRunInventoryItem)
```
levibostian, line 145: 🤔 what would happen in this scenario? 

BQ has these tasks in it:
* identify profile
* track event for that profile

identify profile returns a 400. The SDK deletes that task from the BQ and says that it was successful. So, the SDK now attempts to track an event for that profile (that didnt get created on the API) and sends an HTTP request. 

Maybe queue groups need to be accounted for in this work? If a 400 is encountered by a task at the start of a queue group, then all tasks in that group also get deleted?

levibostian (changes requested):
Rejecting to make sure that code review suggestion that I find important (labeled with `⚠️` emoji) gets addressed. 

Overall, I like the implementation and the tests added!

Sources/Common/Background Queue/QueueRunRequest.swift
Shahroz16: Very good point, thanks for bringing this up.

Sources/Common/Service/HttpRequestError.swift
```diff
@@ -14,6 +14,8 @@ public enum HttpRequestError: Error {
     case noOrBadNetwork(_ urlError: URLError)
     /// Request was not able to get a response from server. Maybe no network connection?
     case noRequestMade(_ error: Error?)
+    /// 400 status code. The request body contains errors or is malformed.
+    case badRequest
```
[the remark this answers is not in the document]
mrehan27: I disagree with this. See my comment: https://github.com/customerio/customerio-android/pull/177#discussion_r1123134468

Sources/Common/Background Queue/QueueRunRequest.swift
```diff
@@ -135,7 +135,14 @@ public class CioQueueRunRequest: QueueRunRequest {
                         self.logger.info("queue is quitting early because all HTTP requests are paused.")
 
                         doneRunning()
-                        break
+                    } else if case .badRequest = error {
+                        self.logger.debug("queue task \(nextTaskStorageId) failed with 400")
+
+                        self.logger.debug("queue deleting task \(nextTaskStorageId) because it will always fail")
```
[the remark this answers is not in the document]
Shahroz16: I am going to replace the debug, with an error so we don't overload on commenting and make readability harder.

Sources/Common/Service/HttpRequestError.swift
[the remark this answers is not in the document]
Shahroz16: I am okay with any, will prefer `badRequest` as well, as it looks more natural to me. So if Levi, is okay, we can but if they feel its something that can really improve readability then we can go that way.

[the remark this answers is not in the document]
mrehan27: I just wanted to share my opinion here. As I think it is little too much detail in name and not really needed. Because in my mind, when I see `badRequest` in any HTTP class, it automatically relates to `400` because of being standard practice. 

I would prefer not to include the code in name here, but I'm also okay with continuing what you both folks think looks better. And whatever we decide for this, we should keep the same for Android as well.

Sources/Common/Background Queue/QueueRunRequest.swift
```diff
@@ -135,11 +135,17 @@ public class CioQueueRunRequest: QueueRunRequest {
...
 
                         doneRunning()
-                    } else if case .badRequest = error {
-                        self.logger.debug("queue task \(nextTaskStorageId) failed with 400")
+                    } else if case .badRequest400 = error {
+                        self.logger.debug("queue deleting task \(nextTaskStorageId) because it failed with 400 and will always fail")
 
-                        self.logger.debug("queue deleting task \(nextTaskStorageId) because it will always fail")
-                        _ = self.storage.delete(storageId: nextTaskToRunInventoryItem.taskPersistedId)
+                        self.logger.error("Received HTTP 400 response while trying to <queue task type>. 400 responses never succeed and therefore, the SDK is deleting this SDK request and not retry. Error message from API: <error message string>, request data sent: <queue task data JSON>")
```
levibostian, line 141: Reminder to replace the placeholders in this string with the actual values.

Sources/Common/Background Queue/QueueRunRequest.swift
```diff
@@ -135,11 +135,17 @@ public class CioQueueRunRequest: QueueRunRequest {
...
+
+                        // check if its a groupStart task, if so delete the group otherwise just delete the task
+                        if let groupStart = nextTaskToRunInventoryItem.groupStart {
+                            _ = self.storage.deleteGroup(groupStartTask: groupStart)
+                        } else {
+                            _ = self.storage.delete(storageId: nextTaskToRunInventoryItem.taskPersistedId)
+                        }
```
levibostian, line 148: I think this logic would benefit from belonging somewhere else. `QueueRunRequest` is large and adding more logic in it can make even it harder to test. I think moving this logic into `QueueStorage` itself would help that. 

Some ideas that we could do. 

### Modify `delete()`

We could modify `QueueStorage.delete` to take a new optional parameter. 

```
func delete(storageId: String, deleteGroupIfStartOfGroup: Bool = false) -> Bool {
  // delete the task. 

  if deleteGroupIfStartOfGroup && task.groupStart != nil {
    // delete all tasks in group
  }
} 
```

### Add a function to `QueueStorage` to do this logic 

Add a function `QueueStorage.deleteGroupIfStartOfGroup(storageId: )` that does the logic here in this file. 

---

This makes the test for `QueueRunRequest` easier since we just need to confirm that when a 400 happens, we call a specific function on `QueueStorage`, but we don't need to worry about the logic of that function in `QueueStorage`. 

I don't have a preference for the solution. If you have better ideas for naming I am open.

Tests/Common/Background Queue/QueueRunRequestTest.swift
```diff
@@ -286,6 +287,30 @@ class QueueRunRequestIntegrationTest: IntegrationTest {
         XCTAssertTrue(queueStorage.getInventory().isEmpty)
     }
 
+    func test_givenMultipleErrorRequest_expectTasksToBeDeletedFor400BadRequestOnly() {
```
levibostian, line 290: Good test to test what happens when a 400 occurs for a task at the start of a group. 

We also need to have a test that tests when a 400 happens and the task is *not* the start of a group.

Tests/Common/Background Queue/QueueStorageTest.swift
```diff
@@ -239,6 +239,76 @@ class QueueStorageIntegrationTest: UnitTest {
...
+            type: String.random,
+            data: Data(),
+            groupStart: nil,
+            blockingGroups: [givenStartOfAnotherGroup]
+        )
+
+        let itemsDeleted = storage.deleteGroup(groupStartTask: givenStartOfTheGroup.string)
+
+        XCTAssertEqual(itemsDeleted.count, 2)
+        XCTAssertEqual(storage.getInventory().count, 2)
```
levibostian, line 309: I would add to this check that the 2 items that are left in the inventory are the 2 tasks that I expect to exist.

levibostian (commented):
great changes!

Sources/Common/Service/HttpRequestError.swift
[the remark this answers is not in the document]
levibostian: I would agree that naming the enum with 400 and with `bad request` feels verbose and not adding value. 

My intention with the naming is to make sure that "bad request" is *a server response* and not a generic "bad HTTP request" that the SDK made. When someone says, "That was a bad request" maybe they meant that the SDK has a bug in it and the request "was bad" and a request wasn't even made. Maybe they meant the SDK made a successful request but the server responded back with a 400 response. 

I am OK with a different name that doesn't feel verbose. My concern with the name "bad request" is that it could be interrupted in a wrong way by someone reading the code. As long as the name is showing that a response came back from the server and the response was a 400 response, I am happy with the name 👍.

levibostian:
I think it would be a good idea to not merge this PR until [the Android PR](https://github.com/customerio/customerio-android/pull/177/) is also ready. That's because I brought up some concerns in the Android PR that are relevant here and vice-versa. 

It might be a good idea to wait in case 1 of the 2 PRs brings up an edge case that the other PR didn't consider.

Sources/Common/Background Queue/QueueRunRequest.swift
Shahroz16: Conclusion after sync: 

// delete task
// conditional check if group, then delete group

levibostian (approved):

PR #277, opened 2023-03-01
fix: cocoapods app extension targets able to compile

levibostian:
Fixes: https://github.com/customerio/issues/issues/9451

Complete each step to get your pull request merged in. [Learn more about the workflow this project uses](https://github.com/customerio/customerio-ios/blob/develop/docs/dev-notes/GIT-WORKFLOW.md). 
- [ ] [Assign members of your team](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/requesting-a-pull-request-review) to review the pull request. 
- [ ] Wait for pull request status checks to complete. If there are problems, fix them until you see that [all status checks are passing](https://external-content.duckduckgo.com/iu/?u=https%3A%2F%2Fsymfony.com%2Fdoc%2F4.3%2F_images%2Fdocs-pull-request-symfonycloud.png&f=1&nofb=1). 
- [ ] Wait until the pull request has been reviewed *and approved* by a teammate
- [ ] After code reviews are approved and you determine this PR is ready to merge, select *Squash and Merge* button on this screen, leave the title and description to the default values, then merge the PR.

PR #278, opened 2023-03-02
test: add cococapods sample app to compile as a test

levibostian:
Closes: https://github.com/customerio/issues/issues/9450

Method used to test [PR](https://github.com/customerio/customerio-ios/pull/277). 

To reproduce the bug addressed by [this PR](https://github.com/customerio/customerio-ios/pull/277), you can create an iOS app that installs our SDK via CocoaPods (Remote Habits uses SPM) and then you try and compile the iOS app. 

This PR adds a iOS app that installs our SDK via CocoaPods. I then setup the CI server to compile this app on each PR. Which means that if the compilation bug is present in the SDK, compiling this sample app should fail.


.github/workflows/build-sample-apps.yml
```diff
@@ -0,0 +1,70 @@
+name: Build sample apps
```
levibostian, line 1: Cool fact, this CI workflow is the base structure required to add multiple sample apps into our SDK code base! 

Great to get part of this work done 😄

PR #284, opened 2023-04-12
feat: in app click tracking

Shahroz16:
closes: https://github.com/customerio/issues/issues/8658


Tests/Common/Background Queue/QueueTest.swift
```diff
@@ -87,6 +87,28 @@ class QueueTest: UnitTest {
         XCTAssertEqual(queueTimerMock.cancelCallsCount, 1)
     }
 
+    // MARK: addTrackInAppDeliveryTask
+
+    func test_addTrackInAppDeliveryTask_expectStartQueueAfterSuccessfullyAddingTask() {
```
levibostian, line 92: The idea of testing *expect queue to start after successfully adding task* is a good test for the generic function `addTask`. Since that test function already exists, I don't see this test function as adding any value. 

Looking at the unit and integration tests of the Queue today, I currently don't see any tests for the `addTrackInAppDeliveryTask` function. The only test functions that I can currently think of adding is testing that when you call `addTrackInAppDeliveryTask`, a task gets added to the queue. [The Android PR added this test](https://github.com/customerio/customerio-android/pull/187/files#diff-5708bc858e1c12a1fed33106ccd7d271b359c72dfa3ba4c4a4ff2abeb90877daR96).

levibostian (commented):
The comments I have about this PR are identical to [the Android PR](https://github.com/customerio/customerio-android/pull/187). I'll approve once those conversations are resolved there and decisions are replicated on this PR.

Tests/Common/Background Queue/QueueTest.swift
Shahroz16: The reason I added this test is that there is no test that currently confirms, calling `addTrackInAppDeliveryTask` would eventually call `addTask`. This test confirms that if you call `addTrackInAppDeliveryTask ` would start the queue as well even though internally calls `addTask` for which we have tests but we still need to be certain that there is a link thats not broken. 

Second test is added too.

Tests/MessagingInApp/MessagingInAppImplementationTest.swift
```diff
@@ -188,4 +205,29 @@ class MessagingInAppImplementationTest: UnitTest {
         XCTAssertEqual(eventListenerMock.messageActionTakenReceivedArguments?.actionValue, givenAction)
         XCTAssertEqual(eventListenerMock.messageActionTakenReceivedArguments?.actionName, givenName)
     }
+
+    func test_inAppTracking_givenCustomAction_expectBQTrackInAppClicked() {
```
levibostian, line 209: It might be a good idea to add another test function that confirms that when the gist close action occurs, we do not add an item to the BQ.

Tests/Common/Background Queue/QueueTest.swift
levibostian: Thanks for the test added. Looks good! 

I understand the value of making sure that the full implementation of the `Queue` is in place. However, I don't think that a unit test is the best tool for that. I would prefer for us to create some integration tests to confirm behavior like this. The test we are talking about here in `QueueTest` seems redundant as far as a unit test goes and would require maintenance that I would like to avoid. I would prefer to only need to maintain unit tests for the `addTask` function to test this logic.

levibostian (approved):

Tests/MessagingInApp/MessagingInAppImplementationTest.swift
Shahroz16: Added that as part of the current test for close, which ensures `BQ` doesn't get updated.

PR #285, opened 2023-04-13
fix: sdk wrappers not having device token registered because of application lifecycle

Shahroz16:
closes: https://github.com/customerio/issues/issues/9641


levibostian (approved):
Assuming that [this comment I left](https://github.com/customerio/issues/issues/9641#issuecomment-1507222266) is correct, then I approve of this PR. 

That comment is important to make sure that FCM behavior is what I am expecting.

levibostian:
Adding labels to not merge this PR until all CI status checks on this PR runs successfully. This is especially important for the CocoaPods FCM app compiling to run successful for us to test the `@available(iOSApplicationExtension, unavailable)` annotation use.

levibostian:
Update: This PR is still blocked and should not be merged. Shahroz is changing the implementation as the previous solution did not work for FCM.

Shahroz16:
@levibostian would appreciate another review here.

Sources/Common/DIGraph.swift
```diff
@@ -22,10 +22,25 @@ public class DIGraph {
...
+
+        // Attempt to get and cast the overridden instance from the dictionary.
+        if let overriddenInstance = overrides[typeName] as? T {
+            return overriddenInstance
+        }
+
+        // If the instance is not found or cannot be cast, return nil.
+        return nil
+    }
+
```
levibostian, line 43: I suggest we remove the optional casting code from this function and instead do a force cast, `as!`. 

This function is only called during automated tests and the force cast will only fail if the developer made a mistake setting up the dependency override. If tests crash, it's easy for the developer to catch the override issue and fix it. Otherwise, the override could go unnoticed and cause tests to be much harder to debug.

Sources/MessagingPush/MessagingPush.swift
```diff
@@ -50,7 +46,20 @@ public class MessagingPush: ModuleTopLevelObject<MessagingPushInstance>, Messagi
...
+            // Call the registerDeviceToken method on the implementation.
+            // This method is responsible for registering the device token and updating the globalDataStore as well.
+            if let implementation = implementation {
+                implementation.registerDeviceToken(deviceToken)
+            } else {
+                // Update the globalDataStore with the new device token.
+                // The implementation may be nil due to lifecycle issues in wrappers SDKs.
+                globalDataStore.pushDeviceToken = deviceToken
+            }
+        }
```
levibostian, line 62:
```suggestion
        // The SDK might not yet have been initialized. Save the device token so when the SDK does get initialized, the token can get registered to a profile. 
        globalDataStore.pushDeviceToken = deviceToken
        implementation?.registerDeviceToken(deviceToken)
```

Are there any consequences to simplifying the logic of this function to these 2 lines of code?

Sources/Tracking/ModuleTopLevelObject.swift
```diff
@@ -14,17 +14,21 @@ open class ModuleTopLevelObject<ImplementationClass> {
...
     }
 
+    public var globalDataStore: GlobalDataStore
+
     private let sdkInitializedUtil: SdkInitializedUtil
 
     // for writing tests
     // provide a nil implementation if you want `sdkInitializedUtil` logic to run and real instance of implementation to run in tests
-    public init(implementation: ImplementationClass?, sdkInitializedUtil: SdkInitializedUtil) {
+    public init(implementation: ImplementationClass?, globalDataStore: GlobalDataStore, sdkInitializedUtil: SdkInitializedUtil) {
```
levibostian, line 23: Does `globalDataStore` need to be part of this `init` function and class? 

Since I don't see this class using this dependency, I would prefer we don't keep an instance of it in the class.

levibostian:
Changed title of PR because I believe this is a bug fix and not a feature. 

Of course, if anyone believes this is not a bug fix, change the title to suggest something different.

Sources/Common/DIGraph.swift
Shahroz16: hmm, can do that, even though I would prefer logging an error than throwing it.  
We actually use the `getOverriddenInstance` in the sample app at times to override the logger etc. But open to change.

Sources/Tracking/ModuleTopLevelObject.swift
Shahroz16: `MessagingPush` is a singleton class. So to make it injectable, and testable I think this was the ideal way. But I might not be fully integrated with the iOS structure. So if there is any other way I can make it injectable in `MessagingPush` class?

Sources/Common/DIGraph.swift
levibostian: 🤔 I think it would be a good idea to add a comment to the `overrides` property saying that it's suggested to not use overrides in the SDK source code but only in tests (and sample apps if there is a need). 

If we as a team agree that overriding a dependency should only be used for tests (aka: non production code in the SDK), then doing things such as force casting in a function like this is OK. 

---

Back to your original comment, I agree that a log would be more useful then simply a crash without a log explaining the reason for the crash. 

I would be OK to log a message and then call `fatalError` if the cast is not successful. As long as we cause a crash, I am OK with the solution.

Sources/MessagingPush/MessagingPush.swift
levibostian: The iOS SDK [already stores the device token to global data storage](https://github.com/customerio/customerio-ios/blob/f462325f9774dba01bbfcd1a96d2f41b461eb02e/Sources/Tracking/CustomerIOImplementation.swift#L226) every time that `registerDeviceToken` is called on the SDK. 

Therefore, after this PR gets merged with the changes made to `CustomerIO` class:
```swift
if let deviceToken = globalDAtaStore.deviceToken {
  registerDeviceToken(deviceToken)
}
```

we will be performing a redundant call even if `MessagingPush.registerDeviceToken` doesn't get called. Each time the SDK gets initialized after registering a device token the first time in the app's history, the device token will get registered again. So I am not sure the logic in this function as the PR currently has will make a difference. 

I think it's ideal to avoid redundant network calls. I would prefer it myself. I am not sure that we can accomplish that at this moment without adding a lot of extra logic to the SDK code?

Sources/Tracking/ModuleTopLevelObject.swift
levibostian: I believe if you modify [the constructor of `MessagingPush`](https://github.com/customerio/customerio-ios/blob/3dbc9bd9e68c8fab57d8c13578f3be1d89387c17/Sources/MessagingPush/MessagingPush.swift#L13-L14) to something like this:

```swift
	private let globalDataStore: GlobalDataStore

    // testing constructor
    override internal init(implementation: MessagingPushInstance?, globalDataStore: GlobalDataStore, sdkInitializedUtil: SdkInitializedUtil) {
        super.init(implementation: implementation, sdkInitializedUtil: sdkInitializedUtil)
		self.globalDataStore = globalDataStore
    }
```

Then you wouldn't need to include `globalDataStore` in this `ModuleTopLevelObject` class.

Shahroz16: I don't think that will work, a subclass constructor cannot have more parameters than its superclass constructor.
It will return `Initializer does not override a designated initializer from its superclass`

Sources/MessagingPush/MessagingPush.swift
Shahroz16: I am not sure I follow, the following line of code in `MessagingPush.registerDeviceToken` 
`if deviceToken != globalDataStore.pushDeviceToken`

It saves the extra network call. 

The later logic just saved another insertion to "GlobablDataStore" which will already happen in [the method you shared](https://github.com/customerio/customerio-ios/blob/f462325f9774dba01bbfcd1a96d2f41b461eb02e/Sources/Tracking/CustomerIOImplementation.swift#L226)

Sources/Common/DIGraph.swift
Shahroz16: Sure, I'll change it.

Sources/MessagingPush/MessagingPush.swift
levibostian: Ok. Now that I think about it again, I can see how this would work. 

Thanks for making the comments in the function explaining why we are doing this logic.

Sources/Tracking/ModuleTopLevelObject.swift
levibostian: I am trying something out in Xcode right now. I'll update this comment or push a commit if I find something that will work.

levibostian: [Found a solution](https://github.com/customerio/customerio-ios/pull/285/commits/01f66a437a1596e10e45cce78e3a4d517dee0204#diff-d322168af6d12b45ff1a57021d6a686396fe280ff72e00f2c1872bc9e4abff3fR11-R21). 

As always, we can revert this commit if we don't like the solution. 

Otherwise, let's resolve this conversation.

levibostian (approved):

Shahroz16:
@levibostian could use a re-review here after the merge. I don't like the fact we are using `sdkConfig` for the KeyValueStorage and internally just using `siteID`. But i didn't make the change here, because you specifically removed it so wanted to have your opinion first. Would do it in a separate PR.

levibostian (approved):

PR #290, opened 2023-04-18
refactor: make getting instances of classes before SDK initialization easier

levibostian:
In some work that Shahroz is doing, he is going to need an instance of GlobalDataStore before the SDK is initialized. 

This refactor is to address this. In particular, see the changes made to `CustomerIO` class where there is now an instance of `GlobalDataStore` that the SDK can use before SDK initialization but also still allow mocking this instance in `CustomerIO` in automated tests.


Shahroz16:
To make it a faster collaborative thing to unblock my ticket, I am changing the base to my ticket and will be making changes in there. Otherwise, we will have to wait for review comments to be answered/catered in this PR and then for it gets merged and then updating my PRs.

PR #292, opened 2023-04-20
refactor: create cocoapods script to make sdk override easier in sdk wrappers

levibostian:
This PR adds a re-usable script for `Podfile` that our SDK wrappers can use. My hope is the comments in the script explain this PR. If there are questions or concerns about this PR, I will update the description in the script to try and explain it better. 

For an example on how to use this script, see [this PR](https://github.com/customerio/amiapp-reactnative/pull/54).

### DIFF
--- a/scripts/cocoapods_override_sdk.rb
+++ b/scripts/cocoapods_override_sdk.rb
@@ -1,49 +1,59 @@
-def override_cio_sdk 
-  ###
-  # How to configure what version of the iOS SDK is installed on your machine. 
-  #
-  # This is the order of priority for installing:
-  # 
-  # 1. Install local version of CIO SDK, if the source code is found on your local computer. 
-  #    To enable this feature, pass environment variable with the path set to where on your computer the iOS SDK source code is. 
-  install_ios_sdk_local_path = ENV['INSTALL_IOS_SDK_LOCAL'] || nil 
-  # 2. Install from a CIO SDK git branch. 
-  #    To install from a branch, pass environment varible with branch name
-  install_ios_sdk_branch_name = ENV['INSTALL_IOS_SDK_BRANCH'] || nil 
-  # 3. Install version of CIO SDK specified in `/ios/customer_io.podspec` file. 
-  #
-  ###
+# This is a Ruby file designed to be used by our SDK wrapper sample apps if they want to install 
+# a non-production build of the iOS native SDK. 
+# 
+# Here is an example Podfile showing you how to use this script: 
+=begin
 
-  # All of the CIO pods that are listed as dependencies in `/ios/customer_io.podspec` need to be listed here in this array:
-  all_ios_pods_sdk_wrapper_needs = [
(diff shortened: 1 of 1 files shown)


scripts/cocoapods_override_sdk.rb
```diff
@@ -0,0 +1,59 @@
...
+
+def get_all_cio_pods
+  return [
+    'CustomerIOCommon',
+    'CustomerIOTracking',
+    'CustomerIOMessagingInApp',
+    'CustomerIOMessagingPush',
+    'CustomerIOMessagingPushAPN',
+    'CustomerIOMessagingPushFCM'
+  ]
```
Shahroz16, line 58:
```suggestion
cio_pods = [
  'CustomerIOCommon',
  'CustomerIOTracking',
  'CustomerIOMessagingInApp',
  'CustomerIOMessagingPush',
  'CustomerIOMessagingPushAPN',
  'CustomerIOMessagingPushFCM'
]
```

scripts/cocoapods_override_sdk.rb
```diff
@@ -0,0 +1,59 @@
...
+=end 
+
+def install_non_production_ios_sdk_git_branch(branch_name)
+  puts ""
+  puts "⚠️ Installing CIO iOS SDK from git branch #{branch_name}"
+  puts ""
+
+  get_all_cio_pods.each { |podname| 
+    pod podname, :git => "https://github.com/customerio/customerio-ios.git", :branch => branch_name
+  }
```
Shahroz16, line 35:
```suggestion
   cio_pods.each { |podname| 
    pod podname, :git => "https://github.com/customerio/customerio-ios.git", :branch => branch_name
  }  
```

scripts/cocoapods_override_sdk.rb
```diff
@@ -0,0 +1,59 @@
...
+def install_non_production_ios_sdk_local_path(local_path)
+  local_path = File.expand_path(local_path, Dir.pwd)
+
+  puts ""
+  puts "⚠️ Installing local version of the iOS SDK. Path of iOS SDK: #{local_path}"
+  puts ""
+
+  get_all_cio_pods.each { |podname| 
+    pod podname, :path => local_path
+  }
```
Shahroz16, line 47:
```suggestion
  cio_pods.each { |podname| 
    pod podname, :path => local_path
  }
```

Shahroz16 (approved):
Thanks for this, looks good, just believe this can be simplified with an array rather than having a method, especially when we don't have any code logic either.

scripts/cocoapods_override_sdk.rb
levibostian: Unfortunately, I tried that and I was receiving `cio_pods` is undefined when I was calling either of the 2 `install` functions in the file. That's why I turned it into a function. 

I am assuming it's how `load` in the `Podfile` behaves where maybe only functions are available? I am not sure. I decided not to spend much time on it since this is for internal purposes only and not customer facing.

PR #293, opened 2023-04-21
refactor: create dashboard for FCM sample app

levibostian:
Created dashboard screen using SwiftUI for the FCM/SwiftUI sample app. 

Part of this ticket: https://github.com/customerio/issues/issues/9818


Apps/CocoaPods-FCM/src/Extensions/StringExtensions.swift
```diff
@@ -0,0 +1,11 @@
+import Foundation
+
+extension String {
+    static var abcLetters: String {
+        "abcdefghijklmnopqrstuvwxyz"
+    }
+
+    static func random(length: Int = 10) -> String {
```
Shahroz16, line 8: Should we expose the method in SDK? if all the sample libraries are going to use it? we can probably have some kind of warning on it for not to be used in production.

Apps/CocoaPods-FCM/src/DashboardView.swift
```diff
@@ -0,0 +1,110 @@
...
+            }
+            ColorButton(title: "Send Custom Event") {
+                showingCustomEventSheet.toggle()
+            }.sheet(isPresented: $showingCustomEventSheet, content: {
+                VStack(spacing: 15) {
+                    TextField("Event name", text: $customEventName)
+                    TextField("Property name", text: $customEventPropertyName)
+                    TextField("Property value", text: $customEventPropertyValue)
+                    Button("Send event") {
+                        CustomerIO.shared.track(name: customEventName, data: [
```
Shahroz16, line 42: would there be any validation for these events? i remember adding them in RH, i think it would be good idea.

Shahroz16 (approved):
Really enjoyed reviewing the SwiftUI PRs, I played with it a couple of months back, and its a good refresher.

PR #295, opened 2023-04-21
refactor: add FCM sample app SDK setting screen

levibostian:
Part of this ticket: https://github.com/customerio/issues/issues/9818

# Dependencies 

Do not merge this PR until the following PRs are merged: 

- https://github.com/customerio/customerio-ios/pull/298
- https://github.com/customerio/customerio-ios/pull/299
- https://github.com/customerio/customerio-ios/pull/300

This PR contains all of the commits from the above PRs. So, it could be recommended to wait on reviewing this PR until the above have been merged in. Otherwise, you will be reviewing this PR plus all of the above PRs at once. 

# This PR made these big changes

* Created a settings screen in the SwiftUI FCM sample app to dynamically change the CIO SDK at runtime. 
* Created a module (`/Apps/Common`) that contains backend logic that all of our iOS sample apps can take advantage of. From reading Aman's APN UIKit sample app PRs, I already think that her PRs could have lots of code removed and instead use this Common module!

PR #300, opened 2023-04-25
refactor: access siteId and apiKey from config object

levibostian:
This PR addresses [a need in the sample apps](https://github.com/customerio/customerio-ios/pull/295#discussion_r1176721639) where the sample apps can access the current siteid and apikey registered with the CIO SDK. 

Closes: https://github.com/customerio/issues/issues/9240


Tests/Tracking/APITest.swift
```diff
@@ -135,7 +135,7 @@ class TrackingAPITest: UnitTest {
             "version": sdkWrapperVersion
         ]
 
-        var actual = CioSdkConfig.Factory.create(region: .US)
+        var actual = CioSdkConfig.Factory.create(siteId: "", apiKey: "", region: .US)
```
Shahroz16, line 138: can we use an internal convenience initializer for testing only, instead of providing empty values every time?

Tests/Tracking/CustomerIOTest.swift
```diff
@@ -21,7 +21,7 @@ class CustomerIOTest: UnitTest {
     }
 
     func test_initialize_expectAddModuleHooks_expectRunCleanup() {
-        customerIO.postInitialize(siteId: testSiteId, diGraph: diGraph)
+        customerIO.postInitialize(diGraph: diGraph)
```
Shahroz16, line 24: I couldn't find a test case, that checks that the `siteID` and `apiKey` are correctly inserted and retrieved from `Config`

Shahroz16 (approved):

Tests/Common/Util/KeyValueStorageTest.swift
```diff
@@ -5,180 +5,171 @@ import XCTest
...
+        let storage = UserDefaultsKeyValueStorage(sdkConfig: sdkConfig, deviceMetricsGrabber: deviceMetricsGrabberMock)
+        storage.switchToGlobalDataStore()
 
         XCTAssertEqual(storage.getFileName(), "io.customer.sdk.com.foo.bar.shared")
     }
 
     func test_getFileName_givenNotGlobalStore_expectGetFileNameForSiteIdStoredData() {
-        let givenSiteId: SiteId = "485895958"
+        let givenSiteId = "485895958"
+        changeTo(siteId: givenSiteId)
```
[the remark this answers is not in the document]
levibostian: I am glad you brought this up. As I was typing up my response to you explaining why it was done this way, I didn't like my response and I ended up refactoring the test class. 

The test class should now behave like it did before and follows the same test class patterns as the rest of the codebase.

Tests/Tracking/CustomerIOTest.swift
levibostian: Added https://github.com/customerio/customerio-ios/pull/300/commits/b3cad3af618b04742fb9fb9be0b725487fc5a563

PR #306, opened 2023-04-28
ci: allow manually deploying to cocoapods from existing tag

levibostian:
Because cocoapods deployments are flaky at the moment, we still need the ability to manually triggger cocoapods deployments on the CI server. 

We should be able to merge this PR and then manually trigger a 2.5 deployment.

### DIFF
--- a/.github/workflows/deploy-sdk.yml
+++ b/.github/workflows/deploy-sdk.yml
@@ -1,6 +1,13 @@
 name: Deploy SDK
 
 on:
+  workflow_dispatch: # manually run this workflow. This allows you to manually deploy things like cococapods, not manually create a git tag. The tag needs to already be created to run this. 
+    inputs:
+      existing-git-tag:
+        description: 'Type name of existing git tag (example: 1.0.3) to checkout and manually deploy'
+        required: true
+        type: string 
+
   push:
     branches: [beta, main, v1] # all branches where deployments currently occur. Make sure this list matches list of branches in  `.releaserc` file.
 
@@ -13,6 +20,7 @@ jobs:
   deploy-git-tag:
     name: Deploy git tag
     runs-on: ubuntu-latest
+    if: ${{ github.event_name == 'push' }} # only deploy tag when triggered automatically by merging a PR. 
     outputs:
       new_release_git_head: ${{ steps.semantic-release.outputs.new_release_git_head }}
       new_release_published: ${{ steps.semantic-release.outputs.new_release_published }}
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy-sdk.yml
```diff
@@ -1,6 +1,13 @@
...
 on:
+  workflow_dispatch: # manually run this workflow. This allows you to manually deploy things like cococapods, not manually create a git tag. The tag needs to already be created to run this. 
+    inputs:
+      existing-git-tag:
+        description: 'Type name of existing git tag (example: 1.0.3) to checkout and manually deploy'
+        required: true
+        type: string 
+
   push:
     branches: [beta, main, v1] # all branches where deployments currently occur. Make sure this list matches list of branches in  `.releaserc` file.
```
Shahroz16, line 12: do we still need other branches other than `main` ?

.github/workflows/deploy-sdk.yml
```diff
@@ -119,16 +127,26 @@ jobs:
           SLACK_WEBHOOK_TYPE: INCOMING_WEBHOOK
   
   deploy-cocoapods:
-    needs: [deploy-git-tag] # run after git tag is made 
-    if: needs.deploy-git-tag.outputs.new_release_published == 'true' # only run if a git tag was made.
-    env:    
-      COCOAPODS_TRUNK_TOKEN: ${{ secrets.COCOAPODS_TRUNK_TOKEN }}
     name: Deploy SDK to Cocoapods 
+    needs: [deploy-git-tag] # run after git tag is made
```
Shahroz16, line 131: I am still a little confused, 

We are still dependent on `deploy-git-tag` and it would only run 

```
if: ${{ github.event_name == 'push' }} # only deploy tag when triggered automatically by merging a PR.
```

so how will this work?

.github/workflows/deploy-sdk.yml
levibostian: We will need `v1` listed. We could just remove `beta` from this list.

.github/workflows/deploy-sdk.yml
levibostian: I have not tested this myself, but I believe that the workflow should still work. [The docs for `needs`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds) suggests that because we added an `if` to deploy cocoapods job, this should not be a problem.

Shahroz16 (approved):

PR #308, opened 2023-05-04
build: fix SDK wrappers not compiling because in-app SDK installing in NSE

levibostian:
@Shahroz16 only merge if you find that Ami RN iOS builds are fixed and no longer show the Gist compile-time errors.


Shahroz16 (approved):
Seems to be working.

PR #309, opened 2023-05-09
refactor: fcm sample app rich push

levibostian:


.github/workflows/build-sample-apps.yml
```diff
@@ -26,13 +26,25 @@ jobs:
...
         working-directory: ${{ env.GITHUB_WORKSPACE }}
         ruby-version: '3.0'
         bundler-cache: true # cache tools to make builds faster in future 
+    
+    - name: Setup build environment to prepare for building 
+      working-directory: ${{ env.GITHUB_WORKSPACE }} # root directory of source code 
+      run: |
+        make setup_sample_app app=${{ matrix.sample-app }}
+        sd CUSTOMERIO_WORKSPACE_SITE_ID ${{ secrets.CUSTOMERIO_WORKSPACE_SITE_ID }} "Apps/${{ matrix.sample-app }}/BuildEnvironment.swift"
+        sd CUSTOMERIO_WORKSPACE_API_KEY ${{ secrets.CUSTOMERIO_WORKSPACE_API_KEY }} "Apps/${{ matrix.sample-app }}/BuildEnvironment.swift"
```
Shahroz16, line 47: With the challenges, we have faced with `sed` or other text-replacing tools in the past. The only thing that really worked for us is adding the complete file as an environment and inserting it. Worked seamless for RH and Ami apps. 

Do you see any advantage with this approach rather then just inserting Env file completely?

Shahroz16 (approved):

.github/workflows/build-sample-apps.yml
levibostian: I plan to never use sed again, haha! We have indeed faced lots of issues. 

I did find a [popular sed replacement called `sd`](https://github.com/chmln/sd) which is what we are using here. It seems to be easier to use and doesn't have the macos/linux issues we have always had. 

I came up with this solution of text replacement of a file because I encountered problems with creating a whole file as a secret. If you ever change the syntax of that Swift file, you have to update the GH secret and then all commits in your code base up to that point will not compile since the syntax changed. This happened once where Jatin updated the GH secret with a new syntax update and all previous releases would no longer compile. 

I think this solution is less fragile to break and is easier to be backwards compatible. 

Android has gradle build config files that make this sort of thing easier to do. iOS has something kind od similar (xcconfig files) but those have lots of boilerplate involved and complexity when using with cocoapods. Seems overkill. This is the most simple solution that I could come up with.

PR #316, opened 2023-05-15
build: setup git hooks add formatted files

levibostian:
Setup a better git hooks experience for the iOS code base. We have been using lefthook on other repos with good success so I am introducing it to iOS here. 

This is a long-term fix to prevent [breaking Lint on main branch](https://github.com/customerio/customerio-ios/pull/314/files). I know some members of the team have had issues getting git hooks in this code base working on their computer. This improved tool should fix those issues.


Shahroz16 (approved):
Thanks for the PR. We can try it out, for me personally git hooks worked, not sure why it might not for others thats why i am a bit skeptical that this problem might be because of something else.

levibostian:
Note to the team: Now that this PR has been merged, run these commands:

```
git checkout main
git pull 
rm -rf .git/hooks
brew install lefthook 
lefthook install 
```

That will remove the old git hooks and install the new ones.

PR #317, opened 2023-05-17
ci: bot comment on PRs with sample app build numbers

levibostian:
iOS SDK tasks for this ticket: https://github.com/customerio/issues/issues/9028

This PR's job is to do 2 things: 
1. Compile all of the sample apps that are inside of the source code of the SDK. 
2. Automate the communication between engineer and QA for what sample app build to install to test the latest changes in a PR. 

This PR adds the following features to our SDK's repository: 

### On every commit pushed to a pull request....

* All sample apps are built and uploaded to Firebase App Distribution for downloading. 
* The `Version.swift` source code file inside of the SDK's source code is updated with build information so when QA finds a bug in a sample app build, we can see at runtime what SDK commit that bug exists for. 
* Bot [creates and updates a comment inside of pull request](https://github.com/customerio/customerio-ios/pull/317#issuecomment-1551455858) indicating the latest sample app version to install to test the PR. 
  * Bot messages are updated when a new commit is pushed, when a new build is compiled, and when a new build fails to compile. By updating the bot comment with all of these states, we help prevent miscommunication between QA and engineers as to what build is the latest.


scripts/update-version.sh
```diff
@@ -9,26 +9,17 @@
 set -e 
 
 NEW_VERSION="$1"
```
Shahroz16, line 11: can we possibly add these checks to point out what issue can occur? 

```
if [ -z "$1" ]
then
  echo "Error: No version number supplied"
  echo "Usage: $0 [new version number]"
  exit 1
fi

if ! command -v sd &> /dev/null
then
    echo "sd command not found. Install it using 'brew install sd'"
    exit 1
fi
```

Shahroz16 (approved):

scripts/update-version.sh
levibostian: If the script was going to be used by developers on their development machines, I could see value in adding code for error checking. 

Because this is a script only used by the CI, IMO, adding error checking code is over engineering and adds complexity to the bash script. Instead, documentation in the script provides what the developer needs to do for the CI to run the script successfully.

PR #320, opened 2023-05-18
feat: in-app dismiss message

ami-aman:
closes https://github.com/customerio/issues/issues/9930

PR for in-app dismiss message. Import `CioMessagingInApp` and  call the method `MessagingInApp.shared.dismissMessage()` to dismiss active in-app message shown on the screen.

________________________________


Sources/MessagingInApp/MessagingInApp.swift
```diff
@@ -12,6 +12,8 @@ public protocol MessagingInAppInstance: AutoMockable {
     @available(*, deprecated, message: "Parameter organizationId no longer being used. Remove the parameter from your function call to migrate to new function.")
     // sourcery:Name=initializeOrganizationId
     func initialize(organizationId: String)
+
+    func dismissMessage()
```
levibostian, line 16: According to the [design proposal for this feature](https://github.com/customerio/issues/issues/9761#issuecomment-1502286550), this function should be a static function added to `MessagingInApp` and not an instance function as it appears here in this PR. 

I was expecting to not see this function here in the protocol. Is there a reason it's needed in the protocol?

Sources/MessagingInApp/MessagingInApp.swift
```diff
@@ -102,4 +104,9 @@ public class MessagingInApp: ModuleTopLevelObject<MessagingInAppInstance>, Messa
     override public func getImplementationInstance(diGraph: DIGraph) -> MessagingInAppInstance {
         MessagingInAppImplementation(diGraph: diGraph)
     }
+
+    // Dismiss in-app message
+    public func dismissMessage() {
+        implementation?.dismissMessage()
+    }
```
levibostian, line 111: According to the [design proposal for this feature](https://github.com/customerio/issues/issues/9761#issuecomment-1502286550), this function should be a static function added to `MessagingInApp` and not an instance function as it appears here in this PR. 

I was expecting to not see this function here as a non-static function. I think we need to modify it to be a static function?

levibostian (changes requested):
Because the PR implementation doesn't align with the [design of the feature](https://github.com/customerio/issues/issues/9761#issuecomment-1502286550), I will request changes for now.

Sources/MessagingInApp/MessagingInApp.swift
Shahroz16: @levibostian good call out. Now that I am thinking about it, I added it as a `static` method because I think, the `initialize` method isn't compulsory anymore. as we don't take organization id and even callbacks are optional. 

If that is indeed the case, then `static` would just do, otherwise, if the earlier hypothesis wasn't correct and we do indeed need a `initialize` call then I can change the proposal to be an `instance` function rather than a `static function` as this method is internally going to call a method provided by the `gist SDK` and we need gist SDK to be initialized. 

What do you both feel, i don't feel very strongly about either option.

levibostian: I don't feel strongly 1 way or the other. Assuming that all 3 of us do not feel strongly towards 1 solution, I suggest that we keep the API consistent with the rest of the SDK. 

Because the rest of the SDK uses the naming pattern: 

```swift
// non-static functions: 
MessagingInApp.shared.foo()
```

Then that is my current vote.

If that sounds good to others, then my original comment can be dismissed.

Shahroz16 (approved):

Sources/MessagingInApp/MessagingInApp.swift
ami-aman: [This](https://github.com/customerio/customerio-ios/pull/320/files/7e93f6f182adabf5c87ead0d0b20e3178e2952fc#r1200614832) conversation resolves current conversation as well.

PR #323, opened 2023-05-19
ci: build sample apps on main branch pushes

levibostian:
At this time, sample apps are only being built for pull requests. However, these can be unstable. We need the ability to download *stable* versions of our sample apps for work that is completed. 

This PR makes builds of the sample apps when PRs are merged to the `main` branch.

### DIFF
--- a/.github/workflows/build-sample-apps.yml
+++ b/.github/workflows/build-sample-apps.yml
@@ -2,6 +2,8 @@ name: Build sample apps
 
 on: 
   pull_request: # build sample apps for every commit pushed to an open pull request (including drafts)
+  push: 
+    branches: [main]
   release: # build sample apps for every git tag created. These are known as "stable" builds that are suitable for people outside the mobile team. 
     types: [published]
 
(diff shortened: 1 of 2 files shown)


.github/workflows/build-sample-apps.yml
```diff
@@ -66,11 +68,6 @@ jobs:
     steps:
     - uses: actions/checkout@v3
 
-    - name: Install tools used in workflow
-      run: |
-        brew install xcbeautify    
-        brew install sd
-
```
Shahroz16, line 73: why don't we need these anymore? I remember i previously saw your comment on one of the PRs where you mention "You expected X to be available on Linux but that was not the case and had to install it manually"

Shahroz16 (approved):

.github/workflows/build-sample-apps.yml
levibostian: This is removing a duplicate entry. If you look further down in the workflow file you see that this code is copy/pasted. I made a mistake and added this code twice.

PR #328, opened 2023-05-23
fix: internal module Common clashing with other Common modules in customers app

levibostian:
Fixes: https://github.com/customerio/issues/issues/10268

Rename internal module `Common` to something more less generic. The Swift language itself does not allow an app to install 2 modules of the same name. One of our customers has an internal module named `Common`.


Package.swift
```diff
@@ -54,7 +54,7 @@ let package = Package(
 
         // Messaging Push 
         .target(name: "CioMessagingPush",
-                dependencies: ["Common", "CioTracking"],
+                dependencies: ["CioTracking"],
```
Shahroz16, line 57: why did we remove it from here? was it never needed ?

Shahroz16 (approved):
I believe we should maybe get it smoke tested to be sure.

Package.swift
levibostian: It was never needed. .

Because `CioTracking` depends on `Common` already, then any module that depends on `CioTracking` automatically include `Common` as a dependency. 

I removed it thinking that it would make the file easier to read since our team communicates like, *FCM depends on Tracking module*.

PR #331, opened 2023-06-06
ci: fix Info.plist file editing to only modify project files, not dependency files

levibostian:
This improves compatibility of our iOS CI configuration for building sample apps to work better with the SDK wrappers.

### DIFF
--- a/Apps/fastlane/Fastfile
+++ b/Apps/fastlane/Fastfile
@@ -45,9 +45,14 @@ lane :build do
     build_number: new_build_number,
     app_version: new_app_version
   )
+  
+  version_update_script_path = '../../../scripts/update-version.sh'
+  # Since this Fastfile is used by SDK wrapper projects that might not contain this script file, we skip running it if the file does not exist. 
+  if File.exist?(version_update_script_path)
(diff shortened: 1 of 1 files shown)


Apps/fastlane/Fastfile
```diff
@@ -47,7 +47,7 @@ lane :build do
   )
 
   UI.important("Updating the SDK's source code version to non-production version. This allows the sample apps to show the SDK version at runtime for app user to better understand the version of the SDK they are running.")
-  sh("../../../scripts/update-version.sh \"#{new_app_version} (#{new_build_number})\"")
+  sh("../../../scripts/update-version.sh \"#{new_app_version} (#{new_build_number})\" || true") # we add "|| true" to the end to prevent the script from throwing any errors. SDK wrappers use this Fastfile and they might not have this script in the project.
```
Shahroz16, line 50: I believe this is done not to halt the process. if so why not just add a check instead of this? instead of `|| true` which gives it a blanket error pass 

```
version_update_script_path = '../../../scripts/update-version.sh'
if File.exist?(version_update_script_path)
  sh("#{version_update_script_path} \"#{new_app_version} (#{new_build_number})\"")
```

levibostian: I like that idea. 

I implemented the change in latest commit.

PR #332, opened 2023-06-06
ci: sample apps to stable test group when merge on main

levibostian:
When building sample apps, builds are either considered stable or unstable. We like to separate the two so unstable builds go to our squad members and stable goes to everyone else. That way we can avoid non-squad members from installing unstable builds of our apps we make. 

Previously, I setup the CI to make stable app builds when a new git tag was made. However, I have since noticed that this method has consequences to it and takes more work. 

This PR instead changes our CI sample apps building to consider all builds on the `main`  branch to be stable. 

# Notes: 
- Before merging, make a PR on Android SDK repo to test against this branch. Make sure sample apps build on Android successfully from this branch. 
- I have tested this[ branch against RH](https://github.com/customerio/RemoteHabits-iOS/pull/88/) with success.

### DIFF
--- a/.github/workflows/build-sample-apps.yml
+++ b/.github/workflows/build-sample-apps.yml
@@ -4,8 +4,6 @@ on:
   pull_request: # build sample apps for every commit pushed to an open pull request (including drafts)
   push: 
     branches: [main]
-  release: # build sample apps for every git tag created. These are known as "stable" builds that are suitable for people outside the mobile team. 
-    types: [published]
 
 concurrency: # cancel previous workflow run if one exists. 
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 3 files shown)


Shahroz16 (commented):
This looks okay, but this class is getting very bloated and the a lot of information is down to the `if/else` checks. I understand its supposed to be used by all of the repos hence the number of cases but can we maybe try to break it apart? 

- Utility methods e.g if ENV["CI"]  check is seen couple of times
- Since we are changing information based on "Github action", maybe create separate classes for those actions? and instead of checking we just use those class methods to get notes etc? 

Otherwise, this file even though very well documented will be a lot harder to maintain.

levibostian:
I took [the conversation](https://github.com/customerio/customerio-ios/pull/332/#pullrequestreview-1469403292) into Slack.

PR #333, opened 2023-06-15
feat: include FCM SDK as a dependency to make FCM push setup easier

levibostian:
This PR was inspired by recent conversations in Slack and also [in a RN SDK PR](https://github.com/customerio/customerio-reactnative/pull/149#discussion_r1230945251). 

Including the FCM SDK as a dependency in our iOS SDK adds benefits to the SDK. Mostly, easier setup. 

The consequences of adding the FCM SDK as a dependency is version issues. If the FCM SDK is specified multiple times in a `Podfile` and the version requirements do not match, then cocoapods throws an error. We experienced this in the past which is why we removed the FCM SDK. This PR does not include a version requirement at all meaning that we support the latest version. 

# SDK wrapper compatibility

Native iOS customers and SDK wrapper customers all are asked to manually add the FCM SDK to their `Podfile`. We made this recommendation to allow customers to specify their own version for the FCM SDK. This PR still allows customers to specify a version if they need to. 

That means that if a native or SDK wrapper customer already has the FCM SDK in their Podfile, this PR will not break anything for them. 

# Testing 

In the CocoaPods sample app, I commented and uncommented out [this line](https://github.com/customerio/customerio-ios/blob/1e439e0e31a7dc514e3128067b933a4aa56bdec4/Apps/CocoaPods-FCM/Podfile#L18), ran `pod update` to update the lock file, then diffed the lock file results. 

The only difference that was made in the lock file was:

```
PODS:
  - FirebaseMessaging (10.11.0): <--- this line
    - ...
```

When the line is commented, the lock file shows 10.11.0 which is the latest version. 
When the line is uncommented, the lock file shows 10.6 which is the version the Podfile explicitly specified.

### DIFF
--- a/CustomerIOMessagingPushFCM.podspec
+++ b/CustomerIOMessagingPushFCM.podspec
@@ -25,5 +25,5 @@ Pod::Spec.new do |spec|
   # Add FCM SDK as a dependency, as our SDK is designed to be compatible with it. 
   # No version is specified which means that by default, the latest version is installed for customers. 
   # Customers can override this by adding the pod to their Podfile themselves to specify a version they want to use. 
-  spec.dependency "FirebaseMessaging"
+  spec.dependency "FirebaseMessaging", ">= 8.7.0", "< 11"
 end
--- a/CustomerIOMessagingPushFCM.podspec
+++ b/CustomerIOMessagingPushFCM.podspec
@@ -21,4 +21,9 @@ Pod::Spec.new do |spec|
   spec.module_name = "CioMessagingPushFCM" # the `import X` name when using SDK in Swift files 
   
   spec.dependency "CustomerIOMessagingPush", "= #{spec.version.to_s}"  
+
+  # Add FCM SDK as a dependency, as our SDK is designed to be compatible with it. 
+  # No version is specified which means that by default, the latest version is installed for customers. 
+  # Customers can override this by adding the pod to their Podfile themselves to specify a version they want to use. 
+  spec.dependency "FirebaseMessaging"
 end
--- a/Apps/CocoaPods-FCM/Podfile
+++ b/Apps/CocoaPods-FCM/Podfile
@@ -7,11 +7,18 @@ sdk_local_path = File.join(File.dirname(__FILE__), '../../')
 target 'test cocoapods' do
   pod 'CustomerIOCommon', :path => sdk_local_path
   pod 'CustomerIOTracking', :path => sdk_local_path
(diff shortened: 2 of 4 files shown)


Shahroz16 (commented):
I added the comments in the [slack thread](https://customerio.slack.com/archives/C01QYDKD0SU/p1686852783619059) as they were not about the implementation but rather "if" we should do this.

CustomerIOMessagingPushFCM.podspec
```diff
@@ -21,4 +21,9 @@ Pod::Spec.new do |spec|
   spec.module_name = "CioMessagingPushFCM" # the `import X` name when using SDK in Swift files 
   
   spec.dependency "CustomerIOMessagingPush", "= #{spec.version.to_s}"  
+
+  # Add FCM SDK as a dependency, as our SDK is designed to be compatible with it. 
+  # No version is specified which means that by default, the latest version is installed for customers. 
+  # Customers can override this by adding the pod to their Podfile themselves to specify a version they want to use. 
+  spec.dependency "FirebaseMessaging"
```
ami-aman, line 28: Does this mean that the installer would give priority to the version mentioned in the app's podfile over latest version of FirebaseMessaging. 

Also, bit curious to learn, how would this make our(devs/QA) lives easier? One point i can think of is that the this will help us test the compatibility of latest FirebaseMessaging dependency with our SDK, is there any other point to support this change ?

Shahroz16 (approved):

khurram18:
@levibostian does this change makes any impact on customers who are not using the FCD SDK?
We are using customer.io SDK through SPM and after updating to version 2.7.0 I see a lot of google and other SDKs added to our project `Package Dependencies` which we are not using at all, these dependencies are below
- abseil
- Firebase
- GoogleAppMeasurement
- GoogleDataTransport
- GoogleUtilities
- gPRC
- GTMSessionFetcher
- InteropForGoogle
- levels
- nanopb
- Promises
- SwiftProtobuf

levibostian:
Thanks for asking, @khurram18. 

I agree, it is not ideal that Xcode will download some google dependencies, even when you're not using the FCM SDK in your app. 

[This is a SPM behavior](https://github.com/customerio/customerio-ios/issues/371#issuecomment-1679138421), not a Customer.io behavior. Even though Xcode downloads these google dependencies onto your machine, it does not mean that any of those SDKs are installed into your app binary at compile-time.

PR #334, opened 2023-06-15
refactor: setup Appium tags for Swift UI sample app

levibostian:
Part of: https://github.com/customerio/issues/issues/9818

Sets identifiers to each of the existing Views in the app for Appium tests to reference. 

I am not sure how to run Appium tests against the sample app to verify that I used the correct method of adding IDs to views for Appium to use. So I am not sure [if the method I used](https://github.com/customerio/customerio-ios/pull/334/files#diff-d67db103c287986efd5bb55c8b0d335544d1e5ac4aa1c57cd6364a33c44fbe4aR7) is correct. However, since the logic is encapsulated, we should be able to fix issues with 1 line of code. 

# Note: 

I see that linter has failed on PR because of a `TODO` I added. I plan to add the back button in a future PR. This PR is not being merged into `main` so I plan to keep this lint error and fix it in future PR before `main` merge.

### DIFF
--- a/Apps/CocoaPods-FCM/src/View/Appium.swift
+++ b/Apps/CocoaPods-FCM/src/View/Appium.swift
@@ -1,0 +1,9 @@
+import Foundation
+import SwiftUI
+
+extension View {
+    // Centralized way to set the identifier on a View so Appium can find the view.
+    func setAppiumId(_ viewId: String) -> some View {
+        accessibility(identifier: viewId)
+    }
+}
--- a/Apps/Common/Source/Model/CioSettings.swift
+++ b/Apps/Common/Source/Model/CioSettings.swift
@@ -10,6 +10,7 @@ public struct CioSettings: Codable {
     public var bqSecondsDelay: TimeInterval
     public var bqMinNumberTasks: Int
     public var trackScreens: Bool
+    public var debugSdkMode: Bool
     public var trackDeviceAttributes: Bool
 
     public func configureCioSdk(config: inout CioSdkConfig) {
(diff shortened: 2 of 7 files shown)


Apps/CocoaPods-FCM/src/View/Appium.swift
```diff
@@ -0,0 +1,9 @@
+import Foundation
+import SwiftUI
+
+extension View {
+    // Centralized way to set the identifier on a View so Appium can find the view.
+    func setAppiumId(_ viewId: String) -> some View {
+        accessibility(identifier: viewId)
+    }
+}
```
Shahroz16, line 9: curious to know, why don't we directly use `accessibility(identifier: viewId)`

There is no other logic inside this method, why not just use this method directly from wherever we are calling `setAppiumId`.

Shahroz16 (approved):

Apps/CocoaPods-FCM/src/View/Appium.swift
levibostian: 2 reasons. 

1. Documentation of the code. If someone was reading the code and saw `accessibility(identifier: "Login button")`, they might wonder why we are calling that function. Seeing `setAppiumId` gives a reason for why the code is called. 
2. I am not yet sure if `accessibility(identifier: viewId)` is the correct way to add appium view tagging. If I am wrong, it's 1 line of code to fix the code to work for appium.

Apps/Common/Source/Model/CioSettings.swift
```diff
@@ -18,6 +19,7 @@ public struct CioSettings: Codable {
         config.backgroundQueueMinNumberOfTasks = bqMinNumberTasks
         config.autoTrackScreenViews = trackScreens
         config.autoTrackDeviceAttributes = trackDeviceAttributes
+        config.logLevel = debugSdkMode ? .debug : .error
```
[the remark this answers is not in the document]
levibostian: Glad you asked about this. I don't like how I added that code either. 

I am not sure why I didn't think about [this as a solution](https://github.com/customerio/customerio-ios/pull/334/commits/5f93f4223cd04d4f16b4206fa5134432f6ec7909) at the time of writing but I didn't, haha!

PR #336, opened 2023-06-21
refactor: Swift UI FCM app consistency

levibostian:
Part of: https://github.com/customerio/issues/issues/9818

Misc changes to make the UI of the sample app more consistent to other sample apps.

### DIFF
--- a/Apps/CocoaPods-FCM/src/Extensions/BindingExtensions.swift
+++ b/Apps/CocoaPods-FCM/src/Extensions/BindingExtensions.swift
@@ -16,7 +16,7 @@ extension Binding where Value == Double {
 
 extension Binding where Value == Int {
     func toStringBinding() -> Binding<String> {
-        cast(to: { String($0) }, from: { $0 == "" ? 0 : Int($0)! })
+        cast(to: { String($0) }, from: { $0 == "" ? 0 : Int($0) ?? 0 })
     }
 }
 
--- a/Apps/CocoaPods-FCM/src/Extensions/BindingExtensions.swift
+++ b/Apps/CocoaPods-FCM/src/Extensions/BindingExtensions.swift
@@ -16,7 +16,7 @@ extension Binding where Value == Double {
 
 extension Binding where Value == Int {
     func toStringBinding() -> Binding<String> {
-        cast(to: { String($0) }, from: { Int($0)! })
+        cast(to: { String($0) }, from: { $0 == "" ? 0 : Int($0)! })
     }
 }
 
--- a/Apps/CocoaPods-FCM/src/View/CustomAttributeView.swift
+++ b/Apps/CocoaPods-FCM/src/View/CustomAttributeView.swift
@@ -40,38 +40,38 @@ struct CustomAttributeView: View {
     @State private var name: String = ""
     @State private var value: String = ""
 
+    var close: () -> Void
     var done: (_ name: String, _ value: String) -> Void
 
     @State private var showConfirmationAlert: Bool = false
 
     var body: some View {
-        VStack {
-            Text(headerText).bold().font(.system(size: 20))
-
-            VStack(spacing: 8) {
-                LabeledTextField(title: "Attribute Name", appiumTextFieldId: "Attribute Name Input", value: $name)
-                LabeledTextField(title: "Attribute Value", appiumTextFieldId: "Attribute Value Input", value: $value)
-            }.padding([.vertical], 40)
-
-            ColorButton(buttonText) {
-                showConfirmationAlert = true
-            }.setAppiumId(appiumButtonText)
-        }.padding([.horizontal], 20)
-            .alert(isPresented: $showConfirmationAlert) {
-                Alert(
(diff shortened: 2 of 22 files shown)


Apps/CocoaPods-FCM/src/View/DashboardView.swift
```diff
@@ -40,107 +37,120 @@ struct DashboardView: View {
...
-                        VStack(spacing: 15) {
-                            TextField("Event name", text: $customEventName)
-                            TextField("Property name", text: $customEventPropertyName)
-                            TextField("Property value", text: $customEventPropertyValue)
-                            Button("Send event") {
-                                CustomerIO.shared.track(name: customEventName, data: [
-                                    customEventPropertyName: customEventPropertyValue
+                }
+
+                Group { // hack to get around Swift UI's max of 10 Views rule.
```
matt-frizzell, line 76: Why not create a group for all the buttons? Yes, there is a limit of 10 views in the various Stacks, but isn't the Group a structure to collect similar views (elements)? I might be wrong here, but I think it isn't exactly a hack if used with a purpose.

Apps/CocoaPods-FCM/src/View/CustomAttributeView.swift
```diff
@@ -0,0 +1,77 @@
+import CioTracking
+import SwiftUI
+
+struct CustomAttributeView: View {
```
matt-frizzell, line 4: - What did you gain by having a conditional view?
- Could you have made it even MORE conditional and not hardcoded so much? It seems like if the View knew which type of Attribute it was tracking that it could have a single String for everything that just replaces a param in each with the Attribute type when it came to it. 
Something like

```
    let attributeType: AttributeType
    
    private var headerText: String
    private var buttonText: String

    init(attributeType: String) {
        self.headerText = "Set Custom \(attributeType) Attribute"
        self.buttonText = "Send \(attributeType) Attributes"
    }
```

Apps/CocoaPods-FCM/src/View/DashboardView.swift
```diff
@@ -3,26 +3,23 @@ import SwiftUI
...
-    @State private var showingProfileAttributesSheet = false
-    @State private var showingProfileAttributesAlert = false
-    @State private var profileAttributeName: String = ""
-    @State private var profileAttributeValue: String = ""
-
     @EnvironmentObject var userManager: UserManager
 
     @State private var showSettings: Bool = false
-    @State private var showAskForPushPermissionButton = false
+
```
matt-frizzell, line 17: I'm trying to think about all these booleans. They seem really heavy for what they are accomplishing. Could you not set a "current view" or something and do a logic check instead of tracking individual booleans for everything? The answer might be no, and if it is I probably need to learn a little more in here to show a possible second path.

Apps/CocoaPods-FCM/src/View/SettingsView.swift
```diff
@@ -1,60 +1,93 @@
...
+                    Button(action: {
+                        UIPasteboard.general.string = viewModel.pushToken
+                    }) {
+                        Image(systemName: "list.clipboard.fill").font(.system(size: 24))
+                    }
                 }
 
-                done()
-            }.setAppiumId("Save Settings Button")
+                Group {
```
matt-frizzell, line 31: If I was using the Groups to connect similar views, I think I would have created a few groups here. This might just be something to learn with swift UI.

matt-frizzell (approved):
I think there are ways this code could be cleaner and probably closer to a Swift UI standard but at the moment I wouldn't hold this code back for that.

matt-frizzell:
I saw enough in testing this build that I know the issue won't close after this push, but I am fine with this commit happening and more clean-up following up behind it. https://www.notion.so/custio/073e1db8d6954fa0b53008c77b153add?v=675d4a8a76c041958b390b1e1b4ffdf3

Shahroz16 (approved):

Apps/CocoaPods-FCM/src/Extensions/BindingExtensions.swift
```diff
@@ -16,7 +16,7 @@ extension Binding where Value == Double {
 
 extension Binding where Value == Int {
     func toStringBinding() -> Binding<String> {
-        cast(to: { String($0) }, from: { Int($0)! })
+        cast(to: { String($0) }, from: { $0 == "" ? 0 : Int($0)! })
```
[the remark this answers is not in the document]
Shahroz16: similarly, there are a lot of forced unwrapping in other files, i didn't comment on all of them but maybe something we can avoid to avoid crash on edge cases?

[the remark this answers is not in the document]
levibostian: For sample apps, I am OK with using forced unwrapping where the non-forced unwrapping solution adds complexity that is not valuable to the app. I say this because sometimes a non-forced unwrap solution feels like overkill for an internal project like a sample app.

PR #339, opened 2023-06-23
refactor: sample app updates

ami-aman:
closes https://github.com/customerio/issues/issues/9679


Apps/APN-UIKit/APN UIKit/Router/LoginRouter.swift
```diff
@@ -4,7 +4,7 @@ import UIKit
 protocol LoginRouting {
     func routeToDashboard()
     func routeToSettings()
```
Shahroz16, line 6: wonder if this is necessary? 
we now have `routeToSettings(_ withInfo: [String: String]?)` with optional info? 

I believe both of them take it to same screen.

Apps/APN-UIKit/APN UIKit/Util/DeepLinksHandlerUtil.swift
```diff
@@ -19,18 +19,32 @@ extension AppDeepLinksHandlerUtil {
...
+            }
+            if let _ = storage.userEmailId {
+                NotificationCenter.default
+                    .post(
+                        name: Notification.Name("showSettingsScreenOnDashboard"),
+                        object: nil,
+                        userInfo: userInfo
+                    )
+                return true
+            }
```
Shahroz16, line 44: I dont't think we should hardcode the positioning optional parameters. 

maybe, something like this,

```
if let urlComponents = URLComponents(string: url.absoluteString), let host = urlComponents.host, host == "settings" {
    var userInfo: [String: String] = [:]
    
    urlComponents.queryItems?.forEach { queryItem in
        if queryItem.name == "site_id" || queryItem.name == "api_key" {
            userInfo[queryItem.name] = queryItem.value
        }
    }
    
    if storage.userEmailId != nil {
        NotificationCenter.default.post(name: Notification.Name("showSettingsScreenOnDashboard"), object: nil, userInfo: userInfo)
        return true
    }
}
```

Apps/APN-UIKit/APN UIKit/View/DashboardViewController.swift
```diff
@@ -1,88 +1,106 @@
...
     }
 
     @IBAction func sendRandomEvent(_ sender: UIButton) {
-        let randomEventName = String.generateRandomString(ofLength: 10)
-        CustomerIO.shared.track(name: randomEventName)
-        showAlert(withMessage: "Random event '\(randomEventName)' tracked successfully")
+        let randomInt = Int.random(in: 0..<3)
+        let randomEventInfo = randomData[randomInt]
+        guard let name = randomEventInfo["name"] as? String else {
+            return
```
Shahroz16, line 96: do we show any error if this is not present?

Shahroz16 (approved):
Added some questions and ideas but none blockers.

Apps/APN-UIKit/APN UIKit/Router/LoginRouter.swift
ami-aman: Ah! Yes... good catch! Thanks Shahroz. I've fixed it.

Apps/APN-UIKit/APN UIKit/Util/DeepLinksHandlerUtil.swift
ami-aman: Thanks for the suggestion, Shahroz. The deeplink that we receive is `apn-uikit://settings&site_id=xxx&api_key=yyy` of which `apn-uikit` is scheme and `settings&site_id=xxx&api_key=yyy` is the host. On trying to fetch queryItems it returns nil since the deep link format is not query friendly. If the decided deeplink had been `apn-uikit://?settings&site_id=xxx&api_key=yyy` i.e in format of a query string then the above suggestion would have worked.

Apps/APN-UIKit/APN UIKit/View/DashboardViewController.swift
ami-aman: Name will always be present for random data. This check is just to unwrap the name data inside `randomData`.

Apps/APN-UIKit/APN UIKit/Util/DeepLinksHandlerUtil.swift
levibostian: I agree with Shahroz with the suggestion. 

The deep link URL is not `apn-uikit://settings&site_id=xxx&api_key=yyy` (not a valid URL) but the URL that our sample apps receive is: `apn-uikit://settings?site_id=xxx&api_key=yyy`

Therefore, I believe his code sample would work.

PR #342, opened 2023-06-27
fix: bad memory access crash

Shahroz16:
closes: https://github.com/customerio/customerio-reactnative/issues/140

discussion: https://customerio.slack.com/archives/C01QYDKD0SU/p1687434815544309?thread_ts=1685640694.944389&cid=C01QYDKD0SU

since this is an SDK release, it should be done after the wrapper version fix goes out.


mrehan27 (approved):

Sources/Tracking/CustomerIO.swift
```diff
@@ -230,7 +230,6 @@ public class CustomerIO: CustomerIOInstance {
         // run cleanup in background to prevent locking the UI thread
         threadUtil.runBackground { [weak self] in
             self?.cleanupRepository?.cleanup()
-            self?.cleanupRepository = nil
```
ami-aman, line 233: How about updating the comment above since there is no call to the method `cleanup` now.

ami-aman (approved):
Approved with a suggestion but not a blocker for this PR.

levibostian (approved):

PR #345, opened 2023-07-05
feat: deep link handling for swiftui users

levibostian:
Closes: https://github.com/customerio/issues/issues/10179

PR in draft form to begin early PR reviews of implementation.


Sources/MessagingPush/MessagingPush+AppDelegate.swift
```diff
@@ -67,7 +67,7 @@ extension MessagingPushImplementation {
                 // A hack to get an instance of deepLinkUtil without making it a property of the MessagingPushImplementation class. deepLinkUtil is not available to app extensions but MessagingPushImplementation is.
                 // We get around this by getting a instance in this function, only.
                 if let deepLinkUtil = sdkInitializedUtil.postInitializedData?.diGraph.deepLinkUtil {
-                    deepLinkUtil.handleDeepLink(deepLinkUrl)
+                    deepLinkUtil.handleDeepLink(deepLinkUrl, deepLinkDelegate: customerIO?.deepLinkDelegate)
                 }
```
levibostian, line 71: This code block only works for push. 

If the Gist migration happens before this PR gets merged, we can re-use deeplinkutil and this logic for in-app. Otherwise, we could copy/paste this same delegate logic into in-app code as a temporary solution to get this PR merged.

levibostian:
Closing as we will be changing the implementation.

PR #350, opened 2023-07-10
refactor: fcm sample show non-blocking view for successful events

levibostian:


Apps/CocoaPods-FCM/src/View/Widget/Toast.swift
```diff
@@ -0,0 +1,35 @@
+import Foundation
+import SwiftUI
+
+struct ToastView: View {
+    @Binding var message: String?
+//    @Binding var isShowing: Bool
```
mrehan27, line 6: We can probably remove it now?

mrehan27 (approved):

Shahroz16 (approved):

ami-aman (approved):

Apps/CocoaPods-FCM/src/View/Widget/Toast.swift
levibostian:
```suggestion
```

PR #351, opened 2023-07-12
fix: wrong thread when invoking webview 

Shahroz16:
fixes: https://github.com/customerio/customerio-reactnative/issues/161


Sources/MessagingInApp/MessagingInAppImplementation.swift
```diff
@@ -68,7 +70,10 @@ extension MessagingInAppImplementation: ScreenTrackingHook {
     public func screenViewed(name: String) {
         logger.debug("setting route for in-app to \(name)")
 
-        inAppProvider.setRoute(name)
+        // Gist expects webview to be launched in UI thread and chaging route will trigger locally stored in-app messages for that route
```
ami-aman, line 73:
```suggestion
        // Gist expects webview to be launched in main thread and changing route will trigger locally stored in-app messages for that route.
```

PR #353, opened 2023-07-13
fix: prevent api calls when identifier is empty

ami-aman:
Fixes issue for https://secure.helpscout.net/conversation/2267893231/217421/
closes https://github.com/customerio/opsbugs/issues/3603

### DIFF
--- a/Sources/Common/Extensions/StringExtensions.swift
+++ b/Sources/Common/Extensions/StringExtensions.swift
@@ -58,6 +58,10 @@ public extension String {
         return self + characters
     }
 
+    func isBlankOrEmpty() -> Bool {
+        trimmingCharacters(in: .whitespaces).isEmpty
+    }
+
     /// Substrings
     /// let s = "hello"
     /// s[0..<3] // "hel"
--- a/Sources/Tracking/CustomerIOImplementation.swift
+++ b/Sources/Tracking/CustomerIOImplementation.swift
@@ -225,8 +225,12 @@ internal class CustomerIOImplementation: CustomerIOInstance {
         // we can reference the token and register it to a new profile.
         globalDataStore.pushDeviceToken = deviceToken
 
-        guard let identifier = profileStore.identifier, !identifier.isEmpty else {
+        guard let identifier = profileStore.identifier else {
             logger.info("no profile identified, so not registering device token to a profile")
+            return
+        }
+        guard !identifier.isEmpty else {
+            logger.error("profiles cannot be created without an identifier. Empty string was provided as the identifier, therefore, request to identify profile in the SDK is ignored. Try identifying a profile again with a non-empty identifier value.")
             return
         }
         // OS name might not be available if running on non-apple product. We currently only support iOS for the SDK
--- a/Sources/Tracking/CustomerIOImplementation.swift
+++ b/Sources/Tracking/CustomerIOImplementation.swift
@@ -229,12 +229,12 @@ internal class CustomerIOImplementation: CustomerIOInstance {
         // we can reference the token and register it to a new profile.
         globalDataStore.pushDeviceToken = deviceToken
 
-        guard let identifier = profileStore.identifier?.isBlankOrEmpty() else {
-            logger.error("profile cannot be identified: Identifier is empty, so not registering device token to a profile")
-            return
-        }
         guard let identifier = profileStore.identifier else {
             logger.info("no profile identified, so not registering device token to a profile")
+            return
+        }
+        if identifier.isBlankOrEmpty() {
+            logger.error("profile cannot be identified: Identifier is empty, so not registering device token to a profile")
             return
         }
         // OS name might not be available if running on non-apple product. We currently only support iOS for the SDK
--- a/Sources/Tracking/CustomerIOImplementation.swift
+++ b/Sources/Tracking/CustomerIOImplementation.swift
@@ -80,7 +80,7 @@ internal class CustomerIOImplementation: CustomerIOInstance {
         body: RequestBody
     ) {
         if identifier.isEmpty {
-            logger.error("profiles cannot be created without an identifier. Empty string was provided as the identifier, therefore, request to identify profile in the SDK is ignored. Try identifying a profile again with a non-empty identifier value.")
+            logger.error("profile cannot be identified: Identifier is blank. Please retry with a valid, non-empty identifier.")
             return
         }
         logger.info("identify profile \(identifier)")
--- a/Sources/Tracking/CustomerIOImplementation.swift
+++ b/Sources/Tracking/CustomerIOImplementation.swift
@@ -225,7 +225,7 @@ internal class CustomerIOImplementation: CustomerIOInstance {
         // we can reference the token and register it to a new profile.
         globalDataStore.pushDeviceToken = deviceToken
 
-        guard let identifier = profileStore.identifier else {
+        guard let identifier = profileStore.identifier, !identifier.isEmpty else {
             logger.info("no profile identified, so not registering device token to a profile")
             return
         }
--- a/Sources/Tracking/CustomerIOImplementation.swift
+++ b/Sources/Tracking/CustomerIOImplementation.swift
@@ -223,12 +223,7 @@ internal class CustomerIOImplementation: CustomerIOInstance {
      Adds device default and custom attributes and registers device token.
      */
     private func addDeviceAttributes(deviceToken: String, customAttributes: [String: Any] = [:]) {
-        logger.info("registering device token \(deviceToken)")
-        logger.debug("storing device token to device storage \(deviceToken)")
-        // no matter what, save the device token for use later. if a customer is identified later,
-        // we can reference the token and register it to a new profile.
-        globalDataStore.pushDeviceToken = deviceToken
-
+        // check if the identifier is not null, blank or empty
         guard let identifier = profileStore.identifier else {
             logger.info("no profile identified, so not registering device token to a profile")
             return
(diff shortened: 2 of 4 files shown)


Sources/Tracking/CustomerIOImplementation.swift
```diff
@@ -225,7 +225,7 @@ internal class CustomerIOImplementation: CustomerIOInstance {
         // we can reference the token and register it to a new profile.
         globalDataStore.pushDeviceToken = deviceToken
 
-        guard let identifier = profileStore.identifier else {
+        guard let identifier = profileStore.identifier, !identifier.isEmpty else {
```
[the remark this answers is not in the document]
ami-aman: Already done in [this commit](https://github.com/customerio/customerio-ios/pull/353/commits/eb79fdfb277eeed2ecef89aecb3ca234dc64f0cc).

Sources/Tracking/CustomerIOImplementation.swift
```diff
@@ -225,7 +225,7 @@ internal class CustomerIOImplementation: CustomerIOInstance {
...
             logger.info("no profile identified, so not registering device token to a profile")
             return
         }
```
levibostian, line 231:
```suggestion
        guard let identifier = profileStore.identifier else {
            logger.info("no profile identified, so not registering device token to a profile")
            return
        }
        guard !identifier.isEmpty else {
            logger.error("profiles cannot be created without an identifier. Empty string was provided as the identifier, therefore, request to identify profile in the SDK is ignored. Try identifying a profile again with a non-empty identifier value.")
            return
        }
```

The message logged to the customer when they provide an empty value would be, "no profile identified, so not registering device token to a profile" which is not accurate to them. The customer did identify a profile, but the identifier is an invalid value. 

Therefore, I suggest we create a 2nd guard statement with a more helpful error message logged. 

I also used `logger.error` in hopes the customer notices the log message because it's an error log level.

Sources/Tracking/CustomerIOImplementation.swift
```diff
@@ -229,6 +229,10 @@ internal class CustomerIOImplementation: CustomerIOInstance {
...
+        guard !identifier.isEmpty else {
+            logger.error("not registering device token to a profile since profiles cannot be created without an identifier. Empty string was provided as the identifier, therefore, request to identify profile in the SDK is ignored. Try identifying a profile again with a non-empty identifier value.")
+            return
+        }
```
levibostian, line 235: Sorry I didn't notice this in my first PR review, but I think that we want the empty identifier check to go into the identify profile function. Not in the register device token function. 

I don't think we need to check if the `profileStore.identifier` is empty because I don't think we should set `profileStore.identifier` with a non-empty value in the first place (in the SDK identify profile function).

levibostian (changes requested):

Sources/Tracking/CustomerIOImplementation.swift
levibostian: I think `identifier.isEmpty` check should go [in the first line of the identify function](https://github.com/customerio/customerio-ios/blob/6e53c1b1688a913666869372168acdcc5663abfb/Sources/Tracking/CustomerIOImplementation.swift#L82).

ami-aman: That makes sense. I have added a check to identify function and removed from registerDeviceToken method.

Sources/Tracking/CustomerIOImplementation.swift
```diff
@@ -79,6 +79,10 @@ internal class CustomerIOImplementation: CustomerIOInstance {
         identifier: String,
         body: RequestBody
     ) {
+        if identifier.isEmpty {
+            logger.error("profiles cannot be created without an identifier. Empty string was provided as the identifier, therefore, request to identify profile in the SDK is ignored. Try identifying a profile again with a non-empty identifier value.")
+            return
+        }
```
Shahroz16, line 85:
```suggestion
        if identifier.isEmpty {
            logger.error("Profile cannot be identified: Identifier is blank. Please retry with a valid, non-empty identifier.")
            return
        }
```

Sources/Tracking/CustomerIOImplementation.swift
```diff
@@ -79,6 +79,10 @@ internal class CustomerIOImplementation: CustomerIOInstance {
         identifier: String,
         body: RequestBody
     ) {
+        if identifier.isEmpty {
```
Shahroz16, line 82: also should we check for both both blank and empty?

ami-aman: What is difference between blank and empty ? 🤔

Shahroz16: let identifier = "   " // a blank string

the `identifier.isEmpty` above will return false, but i dont think we should even allow this

ami-aman: ✅

Sources/Tracking/CustomerIOImplementation.swift
```diff
@@ -79,6 +79,10 @@ internal class CustomerIOImplementation: CustomerIOInstance {
         identifier: String,
         body: RequestBody
     ) {
+        if identifier.isBlankOrEmpty() {
```
levibostian, line 82:
```suggestion
        guard !identifier.isBlankOrEmpty() else {q
```

I prefer `guard` statements because the compiler forces you to include a `return` statement in the body. A great reminder to exit the function early.

Shahroz16 (approved):

Sources/Tracking/CustomerIOImplementation.swift
```diff
@@ -229,6 +233,10 @@ internal class CustomerIOImplementation: CustomerIOInstance {
             logger.info("no profile identified, so not registering device token to a profile")
             return
         }
+        if identifier.isBlankOrEmpty() {
+            logger.error("profile cannot be identified: Identifier is empty, so not registering device token to a profile")
+            return
+        }
```
levibostian, line 239:
```suggestion
```

The identifier should never be blank or empty because the only place that `profileStore.identifier` gets populated is in the `identify()` function. So this `if` statement will never be `true`. 

Also, I don't think that it's the responsibility of `registerDeviceToken` to assert that a profile identifier is not empty. If we add this `if` check here, we might as well add that check everywhere that a profile identifier is attached to the HTTP request `CustomerIO.track()`, `CustomerIO.screen()`, etc. 

Therefore, I suggest we remove this.

Tests/Tracking/CustomerIOImplementationTest.swift
```diff
@@ -129,6 +129,32 @@ class CustomerIOImplementationTest: UnitTest {
         XCTAssertEqual(profileIdentifyHookMock.profileIdentifiedCallsCount, 1)
     }
 
+    func test_identify_givenEmptyIdentifier_givenNoProfilePreviouslyIdentified_expectDoNotRunHooks() {
```
levibostian, line 132:
```suggestion
    func test_identify_givenEmptyIdentifier_givenNoProfilePreviouslyIdentified_expectRequestIgnored() {
```

We want to check for more then just hooks executing (example: `XCTAssertNil(profileStoreMock.identifier)` is a great test) so I suggest a function name change.

Tests/Tracking/CustomerIOImplementationTest.swift
```diff
@@ -129,6 +129,32 @@ class CustomerIOImplementationTest: UnitTest {
...
+        backgroundQueueMock.addTaskReturnValue = (
+            success: true,
+            queueStatus: QueueStatus.successAddingSingleTask
+        )
+
+        customerIO.identify(identifier: givenIdentifier)
+
+        XCTAssertFalse(hooksMock.mockCalled)
+        XCTAssertTrue(backgroundQueueMock.deviceTokensDeleted.isEmpty)
+    }
```
levibostian, line 156:
```suggestion
```

Test not needed because empty identifier check not needed in `registerDeviceToken()`.

Sources/Common/Extensions/StringExtensions.swift
```diff
@@ -58,6 +58,10 @@ public extension String {
         return self + characters
     }
 
+    func isBlankOrEmpty() -> Bool {
+        trimmingCharacters(in: .whitespaces).isEmpty
+    }
```
levibostian, line 63: I would like some unit tests written to test this function. Even though the function is only 1 line of code, we do need to feel confident that it returns `false` for blank or empty strings.

Sources/Tracking/CustomerIOImplementation.swift
Shahroz16: We already have users, who could have identified with a `blank` identifier. But I do think we should always check for this, as this is a breaking app behavior. Even though our only source right now is through `identify` but that `repository` to store identifiers is still exposed to other classes and by mistake or anything that can still be updated. 

So i think, this checks adds a safe-guard.

ami-aman: Users can also register device token using public method `registerDeviceToken`. They may not provide a nil value as this method needs a non-nil value but sending `"   "` is not an exception. The chances of a user sending such values are grim but not to be ignored. I think having this check is a good approach here.

Tests/Tracking/CustomerIOImplementationTest.swift
ami-aman: Added a note [here](https://github.com/customerio/customerio-ios/pull/353#discussion_r1263864194) to support this.

Sources/Tracking/CustomerIOImplementation.swift
levibostian: > We already have users, who could have identified with a blank identifier. 

Yes, that is true that scenario could exist. To help get this fix out, we could make a new PR to support the scenario of fixing SDK installs today where the identifier is blank. 

I suggest a future PR because (1) I do believe that use case is beyond the scope of this PR and (2) I would rather use a different solution to handle that use case then what is currently implemented in this PR.

ami-aman: I think we should let go this PR and remove this code snippet when we make another PR. Same check has been done on Android as well and it is now live. This is an OpsBug issue and need to be closed as soon as possible when the fix is done.

levibostian (approved):
Approved so you can merge when you're done reviewing the remaining unresolved suggestions 👍

Sources/Tracking/CustomerIOImplementation.swift
ami-aman: There is not logical difference between the current implementation and the suggestion, hence keeping the code as is. Thanks for the suggestion.

Sources/Common/Extensions/StringExtensions.swift
ami-aman: Added test cases.

PR #358, opened 2023-07-20
fix: support json array in attributes

mrehan27:
fixes: https://github.com/customerio/issues/issues/10475

### Changes

- Updated `StringAnyEncodable` to encode arrays recursively

### Tested On

- [x] React Native (APN Sample App)
- [x] Flutter (Sample App)
- [x] iOS (CocoaPods-FCM)

### DIFF
--- a/Sources/Common/Util/StringAnyEncodable.swift
+++ b/Sources/Common/Util/StringAnyEncodable.swift
@@ -4,19 +4,24 @@ public struct StringAnyEncodable: Encodable {
     private let data: [String: AnyEncodable]
 
     public init(_ data: [String: Any]) {
-        var builtValue = [String: AnyEncodable]()
-        for (key, value) in data {
+        func encode(value: Any) -> AnyEncodable? {
             switch value {
             case let enc as Encodable:
-                builtValue[key] = AnyEncodable(enc)
+                return AnyEncodable(enc)
+            
             case let dict as [String: Any]:
-                builtValue[key] = AnyEncodable(StringAnyEncodable(dict))
+                return AnyEncodable(StringAnyEncodable(dict))
+            
+            case let list as [Any]:
+                return AnyEncodable(list.compactMap { encode(value: $0) })
+
             default:
                 // XXX: logger error
-                continue
+                return nil;
             }
         }
-        self.data = builtValue
+
+        self.data = data.compactMapValues { encode(value: $0) }
     }
 
     public func encode(to encoder: Encoder) throws {
--- a/Sources/Common/Util/StringAnyEncodable.swift
+++ b/Sources/Common/Util/StringAnyEncodable.swift
@@ -1,9 +1,10 @@
 import Foundation
 
 public struct StringAnyEncodable: Encodable {
+    private let logger: Logger
     private let data: [String: AnyEncodable]
 
-    public init(_ data: [String: Any]) {
+    public init(logger: Logger, _ data: [String: Any]) {
         // Nested function to convert the ‘Any’ values to ‘AnyEncodable’ recursively
         func encode(value: Any) -> AnyEncodable? {
             switch value {
@@ -11,18 +12,19 @@ public struct StringAnyEncodable: Encodable {
                 return AnyEncodable(enc)
 
             case let dict as [String: Any]:
-                return AnyEncodable(StringAnyEncodable(dict))
+                return AnyEncodable(StringAnyEncodable(logger: logger, dict))
 
             case let list as [Any]:
                 // If the value is an array, recursively encode each element
                 return AnyEncodable(list.compactMap { encode(value: $0) })
 
             default:
-                // XXX: logger error
+                logger.error("Tried to convert \(data) into [String: AnyEncodable] but the data type is not Encodable.")
                 return nil
             }
         }
 
+        self.logger = logger
         self.data = data.compactMapValues { encode(value: $0) }
     }
 
--- a/Sources/Common/Util/StringAnyEncodable.swift
+++ b/Sources/Common/Util/StringAnyEncodable.swift
@@ -4,6 +4,7 @@ public struct StringAnyEncodable: Encodable {
     private let data: [String: AnyEncodable]
 
     public init(_ data: [String: Any]) {
+        // Nested function to convert the ‘Any’ values to ‘AnyEncodable’ recursively
         func encode(value: Any) -> AnyEncodable? {
             switch value {
             case let enc as Encodable:
@@ -13,6 +14,7 @@ public struct StringAnyEncodable: Encodable {
                 return AnyEncodable(StringAnyEncodable(dict))
             
             case let list as [Any]:
+                // If the value is an array, recursively encode each element
                 return AnyEncodable(list.compactMap { encode(value: $0) })
 
             default:
--- a/Sources/Common/Util/StringAnyEncodable.swift
+++ b/Sources/Common/Util/StringAnyEncodable.swift
@@ -9,17 +9,17 @@ public struct StringAnyEncodable: Encodable {
             switch value {
             case let enc as Encodable:
                 return AnyEncodable(enc)
-            
+
             case let dict as [String: Any]:
                 return AnyEncodable(StringAnyEncodable(dict))
-            
+
             case let list as [Any]:
                 // If the value is an array, recursively encode each element
                 return AnyEncodable(list.compactMap { encode(value: $0) })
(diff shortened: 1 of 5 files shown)


ami-aman:
@mrehan27 Please test this with wrapper packages as well. I don't see this as a note in the description.

mrehan27:
Thanks @ami-aman. Already tested with React Native and sample data in linked ticket, will test with Flutter too. Updated descriptions for clarity.

Sources/Common/Util/StringAnyEncodable.swift
```diff
@@ -4,19 +4,26 @@ public struct StringAnyEncodable: Encodable {
...
             case let dict as [String: Any]:
-                builtValue[key] = AnyEncodable(StringAnyEncodable(dict))
+                return AnyEncodable(StringAnyEncodable(dict))
+
+            case let list as [Any]:
+                // If the value is an array, recursively encode each element
+                return AnyEncodable(list.compactMap { encode(value: $0) })
+
             default:
                 // XXX: logger error
```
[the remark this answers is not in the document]
mrehan27: I didn't change anything in the current implementation in this block. However, if we want, we can log some error here using `logger` object.

Tests/Common/Util/StringAnyEncodable.swift
```diff
@@ -85,14 +87,16 @@ class StringAnyEncodableTest: UnitTest {
...
 
+        // React native wrap some values in AnyHashable
         let data = [
             "testValue": "foo",
             "numeric": 1,
             "boolean": true,
             "array": [1, 2, 4],
-            "dict": ["test": "value"] as [String: Any]
+            "dict": ["test": "value"] as [String: Any],
+            "hashables": [AnyHashable("color"): [AnyHashable("Red"), AnyHashable("Green"), AnyHashable("Blue")]] as [AnyHashable: [AnyHashable]]
```
Shahroz16, line 99: curious why do we need to provide type `AnyHashable` here? 

maybe add another test if you want to check for `AnyHashable`, but for this we should check we are expect and getting the same map?

Sources/Common/Util/StringAnyEncodable.swift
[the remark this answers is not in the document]
Shahroz16: Yup, we should definitely log here otherwise customers won't know whats failing.

Shahroz16 (approved):
LGTM, added couple of comments

Sources/Common/Util/StringAnyEncodable.swift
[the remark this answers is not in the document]
levibostian: I do believe logging using `error()` is a good idea here. I suggest that the error message tries to convert `value` into a String and include that in the error message: `Tried to convert <data.toString()> into [String: AnyEncodable] but the data type is not Encodable.`

Tests/Common/Util/StringAnyEncodable.swift
```diff
@@ -9,13 +9,15 @@ struct DummyData: Codable, Equatable {
     let testValue: String
     let dict: [String: String]
     let array: [Int]
+    let hashables: [String: [String]]
```
levibostian, line 12: Could you explain the name `hashables` and why this data type is important to test? Are you trying to test nested lists? If so, maybe use the name `hashableNestedList`? 

I bring this up because from understanding the bug we are trying to fix, I was expecting to see us test nesting but the test function doesn't include "nest" in the naming?

Tests/Common/Util/StringAnyEncodable.swift
```diff
@@ -85,14 +87,16 @@ class StringAnyEncodableTest: UnitTest {
     }
 
     func test_stringanyencodable_encodes_complex_data() {
-        let expect = DummyData(boolean: true, numeric: 1, testValue: "foo", dict: ["test": "value"], array: [1, 2, 4])
+        let expect = DummyData(boolean: true, numeric: 1, testValue: "foo", dict: ["test": "value"], array: [1, 2, 4], hashables: ["color": ["Red", "Green", "Blue"]])
 
+        // React native wrap some values in AnyHashable
```
levibostian, line 92: Why does RN wrap values in AnyHashable? When does RN do this and when does it not?

levibostian (approved):

Tests/Common/Util/StringAnyEncodable.swift
mrehan27: Sorry for not making it that clear earlier, But I tried the following code on both react native and Flutter.

```
CustomerIO.setProfileAttributes(attributes: {
  'dietary_restrictions-1': 'VE',
  'dietary_restrictions-2': JSON.stringify(['VE', 'P']),
  'dietary_restrictions-3': ['VE', 'P'],
});
```

And both of these wrappers had similar output as follows:

```
["attributes": AnyHashable([AnyHashable("dietary_restrictions-3"): AnyHashable([AnyHashable("VE"), AnyHashable("P")]), AnyHashable("dietary_restrictions-2"): AnyHashable("[VE, P]"), AnyHashable("dietary_restrictions-1"): AnyHashable("VE")])]
```

So looks like both of these wrappers pass data wrapped in `AnyHashable`. This might be for some flexibility, but I'm not really sure about the reason. Couldn't even find any relevant link too but this was just result of running and debugging different values from both of these wrappers. 🤐 

Adding separate test function and DummyData class for wrapper might make it more clear? Let me know if you folks have more ideas on it.

Tests/Common/Util/StringAnyEncodable.swift
mrehan27: I [responded above](https://github.com/customerio/customerio-ios/pull/358#discussion_r1273337734) to another comment which might help, let me know if you have more questions.

Tests/Common/Util/StringAnyEncodable.swift
mrehan27: I think both RN and Flutter do this for all values, but not sure. More details on this can be [read above](https://github.com/customerio/customerio-ios/pull/358#discussion_r1273337734)

Tests/Common/Util/StringAnyEncodable.swift
mrehan27: With new log message, the output for case 3 is

```
Tried to convert ["dietary_restrictions-3": AnyHashable([AnyHashable("VE"), AnyHashable("P")])] into [String: AnyEncodable] but the data type is not Encodable.
```

Sources/Common/Util/StringAnyEncodable.swift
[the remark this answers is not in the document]
mrehan27: Added logger which required passing `Logger` to `StringAnyEncodable`, let me know if any further improvements are needed for it

Tests/Common/Util/StringAnyEncodable.swift
mrehan27: Updated test cases which might help addressing this

Tests/Common/Util/StringAnyEncodable.swift
```diff
@@ -96,10 +96,38 @@ class StringAnyEncodableTest: UnitTest {
...
+
+        guard let result: DummyData = jsonAdapter.fromJson(actual) else {
+            XCTFail("data did not decoded to a DummyData object")
+            return
+        }
+
+        XCTAssertEqual(expect, result)
+    }
+
+    func test_stringanyencodable_encodes_complex_datafromwrappers() {
```
levibostian, line 117: I am not sure that this test function adds any value to our test suite?

The idea that a value gets converted to `AnyHashable` should not break our SDK's code so I can't think of a reason to test against it. 

A value has to be Hashable in order to be stored in an array. `AnyHashable` is identical to `Hashable`, but generic. 

Therefore, since you added `"dictWithArray": ["color": ["Red", "Green", "Blue"]] as [String: [Any]]` to the test function above, I think that tests `AnyHashable` well enough and we can delete this test function.

Tests/Common/Util/StringAnyEncodable.swift
levibostian: The latest change that you made to this test function we are discussing looks like we have resolved this. 

I made a [comment](https://github.com/customerio/customerio-ios/pull/358#discussion_r1273529294) about how I believe using `AnyHashable(X)` anywhere in our test class doesn't add value. That can be discussed there. I'll resolve this since it looks like Shahroz and I's comment has been addressed.

PR #361, opened 2023-07-25
refactor: fcm swiftui sample app improvements

levibostian:
More changes to the FCM SwiftUI sample app to be consistent with other sample apps.

### DIFF
--- a/Apps/CocoaPods-FCM/src/View/DashboardView.swift
+++ b/Apps/CocoaPods-FCM/src/View/DashboardView.swift
@@ -160,6 +160,11 @@ struct DashboardView: View {
         .overlay(
             ToastView(message: $nonBlockingMessage)
         )
+        .onAppear {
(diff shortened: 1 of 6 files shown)


Apps/CocoaPods-FCM/src/View/DashboardView.swift
```diff
@@ -124,17 +134,37 @@ struct DashboardView: View {
                 EnvironmentText()
             }
             .padding()
-            .alert(isPresented: .notNil(blockingMessage)) {
-                Alert(
-                    title: Text(blockingMessage!),
+        }
+        // Can only use 1 alert() in a View so we combine the different types of Alerts into 1 function.
```
Shahroz16, line 138: Is it you can only use 1 `alert()` or is it only 1 `alert()` would show at a time? because if latter is the case, you can have two alerts and make their display optional on `blockingAlert!.callToActionButton` and only the one with true would show at a time.

Shahroz16 (approved):

mrehan27 (approved):

PR #362, opened 2023-07-26
refactor: minor improvement to uikit sample app

ami-aman:
- [x] Fixed timestamp issue on iOS UIKit APN Sample App. 

 
### DIFF
--- a/Apps/APN-UIKit/APN UIKit/View/Extension/DateExtension.swift
+++ b/Apps/APN-UIKit/APN UIKit/View/Extension/DateExtension.swift
@@ -1,0 +1,11 @@
+import Foundation
+
+extension Date {
+    func addDaysToCurrentDate(days: Int) -> Date? {
+        guard let newDate = Calendar.current.date(byAdding: .day, value: days, to: self) else {
+            // return current date
+            return Date()
+        }
+        return newDate
+    }
+}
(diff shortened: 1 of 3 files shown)


Apps/APN-UIKit/APN UIKit/View/Extension/DateExtension.swift
```diff
@@ -0,0 +1,11 @@
+import Foundation
+
+extension Date {
+    func addDaysToCurrentDate(days: Int) -> Date? {
+        guard let newDate = Calendar.current.date(byAdding: .day, value: days, to: self) else {
+            // return current date
+            return Date()
+        }
+        return newDate
```
levibostian, line 9:
```suggestion
        return Calendar.current.date(byAdding: .day, value: days, to: self)!
```

Because this is a sample app and not our SDK, I think it's OK to use `!` in this logic. I prefer it over `guard...else return Date()` because I don't agree with returning `Date()` when the caller of the function asked to add days, but the function could return a day with nothing added.

levibostian (approved):

Apps/APN-UIKit/APN UIKit/View/Extension/DateExtension.swift
ami-aman: Thanks for the suggestion, Levi. I understand that force unwrapping can be convenient, but I prefer to use guard let or if let for safety reasons. I believe that it is important to always check for nil values, even in sample apps. Hence I am not committing the suggestion. Thanks again :)

PR #365, opened 2023-07-27
refactor: migrate SDK data from sandboxed storage to non-sandboxed

levibostian:
Part of ticket: https://github.com/customerio/issues/issues/10797

Today, the SDK data is all sandboxed by the siteid. The goal of ticket [sdk always initialized](https://github.com/customerio/issues/issues/10797) is that the SDK no longer requires a siteid in order to function. Therefore, we can no longer sandbox data by siteid and we need to migrate existing SDK data previously saved to new non-sandboxed storage. 

Note: Some parts of this code are rough around the edges and tests have been written but are not yet passing. I worked on cleaning up the code and debugging the tests to get them passing and I realized that it would be easier to do both once some other tasks in this ticket were to complete. Therefore, some questions asked in this PR might be fixed in future PRs for this ticket.


Sources/Common/Store/FileStorage.swift
```diff
@@ -0,0 +1,137 @@
...
+    let filePath: URL
+}
+
+public protocol FileStorage: AutoMockable {
+    /// return `true` if the save was successful and no errors were caught/logged
+    /// `fileId` - the file name. `nil` if
+    func save(type: FileType, contents: Data, fileId: String?) -> Bool
+    /// return `nil` if an error was caught and logged *or* if the file simply doesn't exist
+    func get(type: FileType, fileId: String?) -> Data?
+    func delete(type: FileType, fileId: String) -> Bool
```
ami-aman, line 15: Wondering if `fileId` parameter for `delete` is non-optional or it has been left out mistakenly. I am asking because `save` and `get` have `fileId` as an optional parameters.

Sources/Common/Store/Migrations/SDKAlwaysInitializedMigration.swift
```diff
@@ -0,0 +1,148 @@
...
+
+ Before this change to the SDK, the SDK sandboxed the SDK data by siteid. The background queue and profile id for profile identified to name a few examples.
+ After this change, the SDK stores all of it's SDK data in a single place. The SDK does not support using the SDK with multiple workspaces and therefore, no need to sandbox.
+
+ Run after the SDK has been initialized so we have a site id.
+ */
+public class SDKAlwaysInitializedMigration: BaseMigration {
+    let siteId: String
+
+    let sandboxedFileStorage: SandboxedSiteIdFileStorage
```
Shahroz16, line 14: should we drop these instances after the migration is completed? because i suspect them to be big in memory?

Shahroz16 (commented):
Thank you again for the great work here, but I am a bit skeptical about this change. I believe it introduces a lot of things that have a possibility to break. I want us to further discuss what can we do to avoid this and maybe look into strategies to avoid these.  

We might eventually decided to proceed with this strategy but i think we can discuss the pros-cons about this migration approach.

levibostian (commented):
Oh! Sorry everyone. I think I forgot to click the button to make all of my comments be shown. I see they are all "pending". So, sorry!

levibostian:
I am going to change the implementation of this PR a bit. After I do so, I'll remove blocked label and ask for a re-review.

Ahmed-CIO (changes requested):
I am requesting a change for this as I am unable to tell whether it is ready for review.
If it is ready for review:
* Please break it down into smaller PRs given there are several separate API changes that each can be in their own PR
* Include passing tests for the changed parts

If it is not ready for review:
* Please change the PR to be a draft so we know it doesn't need immediate review

levibostian:
Closing. Since this task began, our team has changed our priorities, opinion about this problem of sandboxed data by siteid. If we decide to work on this task again, we should re-define the scope and problem and start fresh.

PR #366, opened 2023-08-09
fix: disable automatic screenview tracking for SwiftUI apps

levibostian:
Update: [I have opened a PR](https://github.com/customerio/customerio-ios/pull/367) that is an alternative approach to this solution. I suggest you review that PR first because we may decide to merge that PR and not this one. 

Part of https://github.com/customerio/issues/issues/10950

automatic screenview tracking is not possible for SwiftUI apps. Therefore, to avoid problems for customers, we disable it at runtime for those apps. 

If customers use a combination of UIKit and SwiftUI in their app, automatic screenview tracking is still not recommended to enable because the SwiftUI framework can still send lots of screenview tracking events that are not part of the customer's app.


Sources/Templates/AutoMockable.stencil
```diff
@@ -204,6 +207,9 @@ class ExampleTest: XCTestCase {
 {% endmacro %}
 
 {% macro mockVariable variable %}
+{% if variable|annotated:"IfCanImport" %}
+    #if canImport({{ variable.annotations["IfCanImport"] }})
+{% endif %}
```
levibostian, line 212: This is not new logic, we already do this for mocked functions. This PR makes it so we also do this logic for mocked properties.

Sources/Common/autogenerated/AutoMockable.generated.swift
```diff
@@ -9,6 +9,9 @@ import FoundationNetworking
 #if canImport(UserNotifications)
 import UserNotifications
 #endif
+#if canImport(UIKit)
+import UIKit
+#endif
```
levibostian, line 14: Because we modified the `Sources/Templates/AutoMockable.stencil` file, all `AutoMockable.generated.swift` files in the codebase gets this added to it. 

This is required because of `UIKitWrapper` modification.

Sources/Tracking/CustomerIO.swift
```diff
@@ -166,8 +166,16 @@ public class CustomerIO: CustomerIOInstance {
         }
 
         Self.initialize(config: newSdkConfig)
+        // OK to force unwrap here because (1) digraph is initialized by this line and (2) when SDK always initialized work complete, the DIGraph instance will no longer be optional.
+        let environment = Self.shared.diGraph!.environmentUtil
+        let logger = Self.shared.diGraph!.logger
 
         if newSdkConfig.autoTrackScreenViews {
+            guard !environment.isSwiftUIApp else {
+                logger.info("SDK is detecting that this is a SwiftUI app. Automatic screenview tracking feature is disabled for SwiftUI apps. We suggest using manual screenview tracking instead. Disable this feature in SDK configuration to silence this warning.")
```
levibostian, line 175: I plan on making docs changes to give SwiftUI examples for manual screenview tracking. I hope that helps customers if they see this log message.

Sources/Common/Util/EnvironmentUtil.swift
```diff
@@ -0,0 +1,35 @@
...
+    public var isSwiftUIApp: Bool {
+        #if canImport(SwiftUI)
+
+        guard let uiApplication = uiKit.uiApplication, let delegate = uiApplication.delegate else {
+            return false
+        }
+
+        // The description will contain module name "SwiftUI" if the class belongs to that module.
+        // We must use a string because 'SwiftUI.AppDelegate' is not public so we cannot do something such as: `delegate is SwiftUI.AppDelegate`
+        return delegate.description.contains("SwiftUI.")
```
Shahroz16, line 29: can you please provide the source for this check?

Shahroz16 (commented):
I would generally lean towards the other PR, which you mentioned as well. Thank you for giving options.

levibostian:
closing in favor of https://github.com/customerio/customerio-ios/pull/367

PR #367, opened 2023-08-09
feat: filter automatic screenview tracking events

levibostian:
Part of https://github.com/customerio/issues/issues/10950

The SDK's automatic screenview tracking feature was designed with UIKit in mind. While doing testing with SwiftUI apps, we discovered that the automatic screenview tracking feature is noisey. The SwiftUI framework is a wrapper around UIKit and because of that, our SDK may report screenview events for Views that live inside of the SwiftUI framework. These Views are probably not wanted by customers and therefore, they should not be reported as screenview events. 

This PR attempts to filter screenview events that we believe are irrelevant to customers. We do give customers the flexibility of overriding the SDK's filtering by providing their own filter function via SdkConfig. 

# Tasks before merging

- [x] QA testing in SwiftUI to verify automatic screenview tracking events get filtered out. 
- [x] QA test UIKit app to verify events do not get filtered out. 
- [X] [Add docs changes](https://github.com/customerio/docs/pull/1490) about the changes made to auto screenview tracking and how to override behavior.


Sources/Common/Extensions/UIKitExtensions.swift
```diff
@@ -10,4 +10,14 @@ public extension UIApplication {
     }
 }
 
+public extension UIViewController {
+    var moduleNameOfView: String? {
+        Bundle(for: type(of: self)).bundleIdentifier
+    }
+
+    var isSwiftUIView: Bool {
+        moduleNameOfView == "com.apple.SwiftUI"
```
Shahroz16, line 19: I think this is true considering, all SwiftUI view controllers are subclasses of UIHostingController only then they would have `bundleIdentifier` equal to "com.apple.SwiftUI". Can there be a case where this is not true ?

Sources/Tracking/CustomerIOImplementation+ScreenViews.swift
```diff
@@ -58,6 +59,13 @@ internal extension UIViewController {
             return
         }
         let nameOfViewControllerClass = String(describing: type(of: viewController))
+
+        // SwiftUI is a wrapper around UIKit classes. So, SwiftUI apps may have noisey screenview tracking.
+        // Filter out screenview tracking events for Views that belong to the SwiftUI framework.
+        guard !viewController.isSwiftUIView else {
```
Shahroz16, line 65: Do we have any instrumentation test suite that can run and validate this for us?

Sources/Common/Extensions/UIKitExtensions.swift
levibostian: This is checking what module a class belongs to. That means that any class that is in the SwiftUI framework will have this same module name. 

That's why I like this solution and am confident in it - it's based on module names (this wouldn't change on each iOS release) instead of something more fragile like a class name (apple could rename "UIHostingController", for example).

Sources/Tracking/CustomerIOImplementation+ScreenViews.swift
levibostian: I thought of that and unfortunately, no, we do not have a test suite at this time for this type of test. 

There is quite a bit of work involved to introduce this kind of test suite.[ I made a ticket](https://github.com/customerio/issues/issues/11047) for introducing this type of test suite.

Shahroz16 (approved):

levibostian:
I changed the implementation to include a SdkConfig option. Asking for code review again because of this.

Sources/Tracking/Util/File.swift
```diff
@@ -0,0 +1 @@
+import Foundation
```
Shahroz16, line 1: What is this file for?

Sources/Common/Extensions/UIKitExtensions.swift
```diff
@@ -0,0 +1,20 @@
...
+    }
+}
+
+public extension UIViewController {
+    // find the bundleId of the framework this View belongs in. This can be used to differentiate Views that belong to host app and Views that belong to SDKs/frameworks.
+    var bundleIdOfView: String? {
+        // TODO: create unit test for this function
+        Bundle(for: type(of: self)).bundleIdentifier
+    }
+}
```
Shahroz16, line 19: I am not sure I understand this use case fully, I believe this will (like you have mentioned) not track any 3rd party SDKs either. 

1- We haven't received any requirements for this feature.
2- People might want to track it. for example, someone has an external SDK to handle chat/help screens etc/ All those views will be filtered out as well due to the fact they have different `bundleIdentifier` and might not be using swift ui at all?

Sources/Tracking/CustomerIOImplementation+ScreenViews.swift
```diff
@@ -78,6 +85,24 @@ internal extension UIViewController {
...
+        guard let diGraph = SdkInitializedUtilImpl().postInitializedData?.diGraph else {
+            return false // SDK not initialized yet. Therefore, we ignore event.
+        }
+
+        let filter: (UIViewController) -> Bool = diGraph.sdkConfig.filterAutoScreenViewEvents ?? { viewController in
+            let deviceInfo = diGraph.deviceInfo
+
+            let doesViewControllerBelongToHostApp = viewController.bundleIdOfView == deviceInfo.customerBundleId
+
+            return doesViewControllerBelongToHostApp
```
Shahroz16, line 98: Can't we just do something like this?

```
func isSwiftUIViewController(_ viewController: UIViewController) -> Bool {
    return viewController is UIHostingController<AnyView>
}
```

Shahroz16 (changes requested):

Sources/Tracking/CustomerIOImplementation+ScreenViews.swift
Shahroz16: I think if we are only dynamically checking for UIHostingController, Instead of checking for specific SwiftUI views.

This addresses the concern of dealing with many private SwiftUI view classes. You're only concerned about whether a screen is backed by SwiftUI or not, not the specifics of the view itself. This was the essence of the function we intended to write.

Regarding, 
> there are many more classes that SwiftUI can use. The list of statements could get long

No matter how many classes are there, they should be in `UIHostingController ` so they should be ignored, if its a pure Swift UI-based application then there won't be any UIViewController and hence we shouldn't even care as our method only checks for UIViewController classes?

> This solution adds a cat and mouse risk where each iOS update that comes out might add/remove/modify classes in the framework that we would need to update for.

I think it's been 3-4 years since Swift-UI is out and the `UIHostingController` so that does seem a bit stable route, we can still keep the custom config method that you exposed along with it as a fallback.

Tests/Tracking/CustomerIOImplementation+ScreenViewsTest.swift
```diff
@@ -0,0 +1,54 @@
...
+    func test_shouldTrackAutomaticScreenviewEvent_givenViewFromHostApp_expectTrue() {
+        // View that does not belong to a 3rd party framework.
+        class ViewInsideOfHostApp: UIViewController {}
+
+        let viewController = ViewInsideOfHostApp()
+        deviceInfoStub.customerBundleId = viewController.bundleIdOfView! // set mock that controls the filter function.
+
+        XCTAssertTrue(viewController.shouldTrackAutomaticScreenviewEvent())
+    }
+}
```
Shahroz16, line 54: maybe we should add a negative test as well?

Shahroz16 (approved):

Sources/Tracking/CustomerIOImplementation+ScreenViews.swift
```diff
@@ -73,11 +58,8 @@ extension CustomerIO {
             return // event has been filtered out. Ignore it.
         }
 
-        guard let data = CustomerIOImplementation.autoScreenViewBody?() else {
-            automaticScreenView(name: name, data: defaultScreenViewBody)
-            return
-        }
-        automaticScreenView(name: name, data: data)
+        let addionalScreenViewData = CustomerIOImplementation.autoScreenViewBody?() ?? [:]
```
ami-aman, line 61: Spell check - `additional`

Sources/Tracking/CustomerIOImplementation+ScreenViews.swift
```diff
@@ -24,18 +25,50 @@ extension CustomerIO {
         guard let swizzledMethod = class_getInstanceMethod(forClass, new) else { return }
         method_exchangeImplementations(originalMethod, swizzledMethod)
     }
+
+    func performScreenTracking(onViewController viewController: UIViewController) {
```
levibostian, line 29: I moved this function to make the code more testable in automated tests. Separating the `performAutomaticScreenTracking` logic with the logic that is given a UIViewController instance to track is what helped.

Sources/Tracking/CustomerIOImplementation+ScreenViews.swift
```diff
@@ -46,36 +79,39 @@ extension UIViewController {
...
         }
         guard let viewController = getVisibleViewController(fromRootViewController: rootViewController) else {
             return
         }
-        let nameOfViewControllerClass = String(describing: type(of: viewController))
+
+        CustomerIO.shared.performScreenTracking(onViewController: viewController)
+    }
+
+    func getNameForAutomaticScreenViewTracking() -> String? {
```
levibostian, line 97: Moved this logic into it's own function to make the code easier to read. I added tests to this function, too.

Sources/Tracking/CustomerIOImplementation+ScreenViews.swift
```diff
@@ -112,6 +94,26 @@ extension UIViewController {
...
+        let nameOfViewControllerClass = String(describing: type(of: self))
+
+        var name = nameOfViewControllerClass.replacingOccurrences(
+            of: "ViewController",
+            with: "",
+            options: .caseInsensitive
+        )
+        if name.isEmpty || name == "" {
+            if let title = title {
```
ami-aman, line 106: Wouldn't it be a good idea to use guard let ? That'll make the code look cleaner.

Tests/Shared/extension/QueueStorageExtension.swift
```diff
@@ -0,0 +1,9 @@
+@testable import CioInternalCommon
+@testable import CioTracking
+import Foundation
+
+public extension QueueStorage {
+    func filterTrackEvents(_ type: CioTracking.QueueTaskType) -> [QueueTaskMetadata] {
```
levibostian, line 6: Reusable function in test classes.

Sources/Common/Store/SdkConfig.swift
```diff
@@ -144,6 +148,17 @@ public struct SdkConfig {
      */
     public var autoTrackScreenViews: Bool
 
+    #if canImport(UIKit)
+    /**
+     Filter automatic screenview events to remove events that are irrelevant to your app.
+
+     Return `true` from function if you would like the screenview event to be tracked.
```
ami-aman, line 155: `     Return `true` from function if you would like the screenview event to be tracked.`

Is this specific to SwiftUI? Having `autoTrackScreenViews` and `filterAutoScreenViewEvents` seem to be a bit confusing to me.

levibostian: `autoTrackScreenViews` is to disable the feature entirely. 
`filterAutoScreenViewEvents` is to configure the feature to behavior that you want, assuming that `autoTrackScreenViews` is true. 

Therefore, both SDK config options provide a different value.

Sources/Tracking/CustomerIOImplementation+ScreenViews.swift
```diff
@@ -46,36 +79,39 @@ extension UIViewController {
...
+
+    func getNameForAutomaticScreenViewTracking() -> String? {
+        let nameOfViewControllerClass = String(describing: type(of: self))
+
         var name = nameOfViewControllerClass.replacingOccurrences(
             of: "ViewController",
             with: "",
             options: .caseInsensitive
         )
         if name.isEmpty || name == "" {
```
Shahroz16, line 105: isn't it redundant? shouldn't 
```
if name.isEmpty
```

just work?

Shahroz16 (approved):

PR #373, opened 2023-08-18
ci: build with xcode 15 beta

levibostian:
Part of ticket https://github.com/customerio/issues/issues/11139

To verify that customers can compile their apps with our SDK installed, this PR enables Xcode15 beta builds on the CI. 

I expect that once 15 goes stable, we can make 15 the default and drop compiling on CI with 14. 

I checked the [CI logs](https://github.com/customerio/customerio-ios/actions/runs/5904453500/job/16016675787?pr=373) and it looks like it ran as expected. The latest Xcode 15 beta 6 got installed and fastlane used xcode 15 for compiling the app.

### DIFF
--- a/.github/workflows/build-sample-apps.yml
+++ b/.github/workflows/build-sample-apps.yml
@@ -51,6 +51,9 @@ jobs:
     strategy:
       fail-fast: false # if one sample app fails to build, let the other sample apps continue to build and not cancel them. 
       matrix: # Use a matrix allowing us to build multiple apps in parallel. Just add an entry to the matrix and it will build! 
+        xcode-version:
+        - "14"
+        - "15.0-beta"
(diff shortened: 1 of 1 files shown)


Shahroz16 (commented):
I am happy to approve, but I think it's probably unnecessary to do it every single time with the beta version. Once we have verified it compiles with 15, we are good and the task is completed. 

We can maybe replace 14 with 15 once it moved to stable but before that, it just adds to noise, 4 releases being made, extra time for workflow to be completed etc.

levibostian:
@Shahroz16 Good point. I made a change to PR to use 15 beta by default for all sample app builds instead. 

If compiling works today, that's enough for me to give a 👍 and say let's begin using it for `main` builds.

PR #376, opened 2023-08-30
fix: added url path encoding

Shahroz16:
closes: https://github.com/customerio/customerio-reactnative/issues/191

### DIFF
--- a/Sources/Common/Extensions/StringExtensions.swift
+++ b/Sources/Common/Extensions/StringExtensions.swift
@@ -31,6 +31,11 @@ public extension String {
         String((0 ..< length).map { _ in abcLetters.randomElement()! })
     }
 
+    static func randomStringWithSpecialCharacters(length: Int = 20) -> String {
+        let characters = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789-._~:/?#[]@!$&'()*+,;=%"
+        return String((0 ..< length).map { _ in characters.randomElement()! })
+    }
+
     /**
      Checks if the string matches the regex pattern.
 
--- a/Sources/Common/Extensions/StringExtensions.swift
+++ b/Sources/Common/Extensions/StringExtensions.swift
@@ -31,11 +31,6 @@ public extension String {
         String((0 ..< length).map { _ in abcLetters.randomElement()! })
     }
 
-    static func randomStringWithSpecialCharacters(length: Int = 20) -> String {
-        let characters = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789-._~:/?#[]@!$&'()*+,;=%"
-        return String((0 ..< length).map { _ in characters.randomElement()! })
-    }
-
     /**
      Checks if the string matches the regex pattern.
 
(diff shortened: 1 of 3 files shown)


Tests/Common/Service/HttpEndpointTest.swift
```diff
@@ -44,4 +44,29 @@ class HttpEndpointTest: UnitTest {
 
         XCTAssertEqual(actual, expected)
     }
+
+    func test_getUrlString_givenRandomPathNeedsEncoding_expectEncodedPath() {
+        let identifierWithSpecialChar = String.randomStringWithSpecialCharacters()
```
levibostian, line 49: I would prefer if we hard-coded test data instead of randomly generate it. Maybe the randomly generated string gives us a char with no special chars, for example. 

The other test function is great in that it includes a special character that we have identified causes issues for customers. I like the idea of hard-coding test strings that we know could cause an issue.

Tests/Common/Service/HttpEndpointTest.swift
```diff
@@ -44,4 +44,29 @@ class HttpEndpointTest: UnitTest {
...
+
+    func test_getUrl_givenUnencodedPathWithSpecialCharacter_expectValidUrl() {
+        let identifierWithSpecialChar = "social-login|1234567890abcde"
+        let endpoint = CIOApiEndpoint.identifyCustomer(identifier: identifierWithSpecialChar)
+
+        setHttpBaseUrls(trackingApi: "https://customer.io")
+
+        let actualUrl = endpoint.getUrl(baseUrls: httpBaseUrls)
+
+        XCTAssertNotNil(actualUrl, "Expected valid URL but got nil. Ensure path is encoded correctly.")
```
levibostian, line 70: I prefer that we do a AssertEquals() here with a hard-coded URL that is generated. Would help me feel more confident in this function. Also, gives good examples for devs reading this code's function to understand it's behavior.

Sources/Common/Extensions/StringExtensions.swift
```diff
@@ -31,6 +31,11 @@ public extension String {
         String((0 ..< length).map { _ in abcLetters.randomElement()! })
     }
 
+    static func randomStringWithSpecialCharacters(length: Int = 20) -> String {
```
levibostian, line 34: If we do end up using this function, move it into `Tests/Shared` since it's a function used only for tests.

levibostian (approved):

Tests/Common/Service/HttpEndpointTest.swift
mrehan27: Good point, but random values does extend range for testing more characters. We can perhaps update `randomStringWithSpecialCharacters` to must include special character every time.

Tests/Common/Service/HttpEndpointTest.swift
levibostian: I believe that using `assertequals` will increase the quality of the test for us to have more confidence in it. AssertNil makes sense in that we want to try and prevent a behavior the SDK has today. My suggestion is because I believe we can take this test function a step further and make the test even better then it is.

Shahroz16: I understand that but we are doing exactly that in the previous test, this test is to ensure we don't get a `nil`, which was the case for this behavior that has special characters produced `nil` as a URL. 

Now we have two tests, first, that asserts equal of making sure we get the URL we expect and the second to ensure having characters doesn't cause nil.

PR #378, opened 2023-09-01
feat: remove need to register token manually & fetch device token 

ami-aman:
closes https://github.com/customerio/issues/issues/11179
closes a part of https://github.com/customerio/issues/issues/10298 (iOS)

###  What is included in this PR ?
**Reduced efforts to register device**
With this pull request, customers will no longer need to write native code or call our SDK methods to register their devices. Devices will be automatically registered when a token is generated. 

**Works for FCM & APNS**
This PR works for both Apple Push Notification Service (APNS) and Firebase Cloud Messaging (FCM) tokens.

**Fetch device token**
Additionally, this PR adds a new feature method called `registeredDeviceToken` to the iOS SDK. This method can be used to fetch the current device token for the user.

**Updated sample apps to test the feature add-on**
Both the sample apps in this repository have been updated to test the feature. Go to settings screen, the device token shown on the screen is now fetched from the new method. It was previously saved & fetched from local storage.

### What is pending in this PR?
- Test cases for the new methods.
- Minor Code cleanup


CustomerIOTracking.podspec
```diff
@@ -20,5 +20,10 @@ Pod::Spec.new do |spec|
...
 
-  spec.dependency "CustomerIOCommon", "= #{spec.version.to_s}"  
-end
\ No newline at end of file
+  spec.dependency "CustomerIOCommon", "= #{spec.version.to_s}"
+  
+  # Add FCM SDK as a dependency, as our SDK is designed to be compatible with it.
+  # No version is specified which means that by default, the latest version is installed for customers.
+  # Customers can override this by adding the pod to their Podfile themselves to specify a version they want to use.
+  spec.dependency "FirebaseMessaging", ">= 8.7.0", "< 11"
```
ami-aman, line 28: Added Firebase dependency in CioTracking to swizzle FCM method.

levibostian, line 28: 🚧 (blocker review comment) 

I think that by adding FCM dependency in the tracking module, all customers who install our tracking module but do not install our MessagingPushFCM module will now have the FCM SDK installed in their app. 

I don't think this is an acceptable solution to add the FCM SDK as a dependency outside of the MessagingPushFCM module.

Sources/Tracking/CustomerIO.swift
```diff
@@ -238,6 +246,59 @@ public class CustomerIO: CustomerIOInstance {
...
+            let swizzledSelector = #selector(self.messaging(_:didReceiveRegistrationToken:))
+            let originalSelector = #selector(MessagingDelegate.messaging(_:didReceiveRegistrationToken:))
+            guard let swizzledMethod = class_getInstanceMethod(CustomerIO.self, swizzledSelector) else {
+                return
+            }
+            if let originalMethod = class_getInstanceMethod(messagingDelegateClass, originalSelector) {
+                method_exchangeImplementations(originalMethod, swizzledMethod)
+            }
+        }
+    }
```
levibostian, line 300: The `CustomerIO.swift` file is already a large file. I would prefer this code be moved to a different file. 

I suggest seeing how [the auto screenview tracking was done](https://github.com/customerio/customerio-ios/blob/main/Sources/Tracking/CustomerIOImplementation%2BScreenViews.swift) by using an `extension`, we are able to move the logic into another file.

Sources/Tracking/CustomerIO.swift
```diff
@@ -20,7 +24,7 @@ public protocol CustomerIOInstance: AutoMockable {
         // sourcery:TypeCast="AnyEncodable(body)"
         body: RequestBody
     )
-
+    func registeredDeviceToken() -> String?
```
levibostian, line 27: 🚧 (blocker comment)

```suggestion
    var registeredDeviceToken: String? { get }
```

I prefer that we use a property accessor like we do for other property getters such as `CustomerIO.shared.config`. The API is more Swift-y.

Sources/Tracking/CustomerIO.swift
```diff
@@ -299,6 +360,10 @@ public class CustomerIO: CustomerIOInstance {
         implementation?.identify(identifier: identifier, body: body)
     }
 
+    public func registeredDeviceToken() -> String? {
```
levibostian, line 363: 🚧  

Reminder to add code to the `Tracking/APITest` class for this new public API change.

Sources/Tracking/CustomerIO.swift
```diff
@@ -172,6 +176,10 @@ public class CustomerIO: CustomerIOInstance {
             // Only call this code when not possibly being called from a NSE.
             shared.setupAutoScreenviewTracking()
         }
+
+        // Swizzle apns/fcm token methods to register device.
+        Self.shared.swizzleDidReceiveRemoteNotification()
+        Self.shared.swizzleDidReceiveFCMRegistrationToken()
```
levibostian, line 182: 🚧 

Like we do for auto screenview tracking, I think we need to add to the `SdkConfig` the ability to disable swizzling of register device token.

Sources/Tracking/CustomerIO.swift
```diff
@@ -238,6 +246,59 @@ public class CustomerIO: CustomerIOInstance {
...
+            let appDelegate = UIApplication.shared.delegate
+            let appDelegateClass: AnyClass? = object_getClass(appDelegate)
+            let originalSelector = #selector(UIApplicationDelegate.application(_:didRegisterForRemoteNotificationsWithDeviceToken:))
+            let swizzledSelector = #selector(self.application(_:didRegisterForRemoteNotificationsWithDeviceToken:))
+            guard let swizzledMethod = class_getInstanceMethod(CustomerIO.self, swizzledSelector) else {
+                return
+            }
+            if let originalMethod = class_getInstanceMethod(appDelegateClass, originalSelector) {
+                method_exchangeImplementations(originalMethod, swizzledMethod)
+            }
```
levibostian, line 271: The `CustomerIO` class [has an existing function to swizzle methods](https://github.com/customerio/customerio-ios/blob/f304189cc5504ac763c3d69307da783213fe903d/Sources/Tracking/CustomerIOImplementation%2BScreenViews.swift#L11-L27) that can be re-used here.

levibostian (changes requested):
Great start so far. 

Few 🚧 comments which indicate comments that I consider blockers to merging. 

I understand that some of these comments may require change to the current implementation of the PR. Some may also require conversations with the team to possibly only enable this feature for APN and not FCM, for example. There are many ways that this PR could go based on conversations in here.

CustomerIOTracking.podspec
ami-aman: There are different ways we could achieve swizzling the fcm token -

- **Hooks**: One of the way to get fcm token swizzled is using Hooks. With hooks, the swizzling would happen within CioFCM module. I did a quick & rough PoC to check if it would be feasible to implement it in our current SDK and I found that using hooks might make the SDK prone to error & makes the implementation quite complicated.
- **Swizzling within Tracking module**: Another solution is to add Firebase dependency in Tracking module and perform swizzling within tracking without the need of hooks. This simplifies the implementation but comes at a cost to add Firebase dependency even when the customer is not using CioFCM module. We can avoid swizzling FCM token when APN library is used by adding a configurable property during initialisation of our SDK but that still does not prevent installation of FCM dependency.

An idea to handle this to NOT add swizzling for FCM at all & have this feature only for APNS. But I find the feature incomplete without having to save & fetch FCM token but APNS.

Do you have any suggestion to handle this better? I would appreciate ideas.

Sources/Tracking/CustomerIO.swift
```diff
@@ -238,6 +246,59 @@ public class CustomerIO: CustomerIOInstance {
...
+        let token = String(apnDeviceToken: deviceToken)
+        Self.shared.registerDeviceToken(token)
+    }
+
+    // Implement method swizzling for APN device token. Swizzling will ensure that the
+    // device token gets registered without needing the developer to add call respective
+    // Customer.io SDK methods to register the device.
+    @available(iOSApplicationExtension, unavailable)
+    func swizzleDidReceiveRemoteNotification() {
+        DispatchQueue.main.async {
```
Shahroz16, line 261: Is there a specific reason we use `main` over here?

Sources/Tracking/CustomerIO.swift
```diff
@@ -238,6 +246,59 @@ public class CustomerIO: CustomerIOInstance {
...
+        }
+    }
+
+    // Implement method swizzling for FCM token. Swizzling the FCM token method will ensure
+    // that the token is registered with CIO even when the developer doesn't call the respective
+    // CIO method after token generation.
+    @available(iOSApplicationExtension, unavailable)
+    func swizzleDidReceiveFCMRegistrationToken() {
+        DispatchQueue.main.async {
+            let messagingDelegate = Messaging.messaging().delegate
```
Shahroz16, line 289: should we probably keep a weak reference for messaging delegates to avoid any possible memory leakages?

```
weak var messagingDelegate
```

Sources/Tracking/CustomerIO.swift
ami-aman: As mentioned in description, I would be cleaning this up and putting this in a separate file.

Sources/Tracking/CustomerIO.swift
ami-aman: Thanks for the suggestion. But I do not understand how this is a blocker ? This is a suggestion and shouldn't be called out as a blocker IMO.

Sources/Tracking/CustomerIO.swift
ami-aman: Let's decide on [this](https://github.com/customerio/customerio-ios/pull/378#discussion_r1314538993) and then go write test cases. As mentioned in the description, test cases is a part of pending list, hence a draft PR.

Sources/Tracking/CustomerIO.swift
ami-aman: https://github.com/customerio/customerio-ios/pull/378#discussion_r1314728612

Sources/Tracking/CustomerIO.swift
ami-aman: My bad! I implemented this code in react native package first to check feasibility. I should bring the code out of main thread.

ami-aman:
Closing this PR and will create a new one when the feature is added.

PR #379, opened 2023-09-06
fix: reduce memory and cpu usage while running background queue

levibostian:
part of: https://github.com/customerio/issues/issues/10950

While profiling the SDK, I noticed that the `QueueRunner` was using a lot of resources while the BQ executed. I found that for every background task item that was executed, a lot of new dependencies were being created because a new `QueueRunner` instance (and it's subdepenencies) were being constructed by the hooks manager. 

I tried a couple of solutions to prevent new `QueueRunner` instances being created for each BQ task execution, but each solution I came up with would introduce a new edge case to the BQ execution where tasks might not get picked up by a `QueueRunner` instance with hooks. 

Instead, I added commits later on in this branch that introduced the solution of no longer using hooks for executing the background queue. I don't see a reason to use hooks for BQ execution in the long-term and because it's what is causing the issues fixed by this PR, I removed them.


Shahroz16 (approved):
Good catch!

levibostian:
Asking for a re-review because of an [addition](https://github.com/customerio/customerio-ios/pull/379/commits/052c005e423a0e4c6648caff698270c70af9529a) to the implementation to avoid an edge case that could occur.

Sources/Common/Background Queue/QueueRunner.swift
```diff
@@ -36,30 +29,19 @@ public class CioQueueRunner: ApiSyncQueueRunner, QueueRunner {
...
-            switch queueTaskType {
-            case .trackDeliveryMetric: trackDeliveryMetric(task, onComplete: onComplete)
-            }
-
-            return
-        }
-
-        var hookHandled = false
+        guard let queueTaskType = QueueTaskType(rawValue: task.type) else {
+            logger.error("task \(task.type) not handled by any module")
```
levibostian, line 33:
```suggestion
            // not being able to compose a QueueTaskType is unexpected. All types are expected to be handled by this runner. Log an error so we get notified of this event. 
            logger.error("task \(task.type) not handled by the queue runner.") 
```

levibostian:
PR implementation changed so I asked for a re-review. 

I updated the PR description talking about it.

Sources/Common/Background Queue/QueueRunner.swift
```diff
@@ -13,20 +13,13 @@ public protocol QueueRunner: AutoMockable {
```
levibostian, line 13: This file contains the most changes for the PR. I moved code from Tracking module's QueueRunner into this one. Now, all of the Tracking API HTTP calls are in 1 module (which, I think is a good thing thinking about modularity in the code base).

Sources/Common/Background Queue/Task Data/DeletePushNotificationQueueTaskData.swift
```diff
@@ -0,0 +1,16 @@
+import Foundation
+
+public struct DeletePushNotificationQueueTaskData: Codable {
```
levibostian, line 3: Lots of files in this PR are structs that git says are new files. 

They are not new, they are moved and I modified them from `internal` to `public` so code in tracking module could access them.

Sources/Common/Background Queue/Type/QueueTaskType.swift
```diff
@@ -1,5 +1,10 @@
 import Foundation
 
-enum QueueTaskType: String {
+public enum QueueTaskType: String {
     case trackDeliveryMetric
+    case identifyProfile
+    case trackEvent
+    case registerPushToken
+    case deletePushToken
+    case trackPushMetric
```
levibostian, line 9: Part of moving code from Tracking QueueRunner into 1 QueueRunner.

Shahroz16 (approved):
This is what I understand. Please correct me If I am wrong and ask for a re-review.

Before we had two Queue Runners, one in common and one in tracking. We utilized hooks to communicate between modules (common and tracking). 

Now with this change, we are getting rid of the Queue runner in the tracking module and shifting all the logic from there to the common queue logic also moved the data models from tracking to common so they can be utilized in the updated Queue runner in common. 

So there is no new code that is being added but rather code being shifted from one module to another and we are getting rid of a particular hook that was used for this purpose.

levibostian:
@Shahroz16, [100% correct](https://github.com/customerio/customerio-ios/pull/379#pullrequestreview-1615603943). I couldn't have put it better myself. 

Communication between modules is valuable and does help us keep our SDKs lightweight and modular. There is value in it. However, I no longer believe that a QueueRunner is the best use of the hooks feature. Originally, separate QueueRunners were to try and keep the SDK extra slim by putting specific Track API calls into modules that use them. But, doing that has added more complexity then what it's worth. 

Also, from internal discussions we have had about splitting our Common module into more modules, we have mentioned having a module that contains all HTTP code for communicating with the Track API. So, I see a day where all CIO Track API code is in 1 module. Therefore, another reason to not spread Track API code across modules.

PR #381, opened 2023-09-08
test: prevent false positives in url encoding tests

levibostian:
I noticed a test function in the code where the function under test `getUrlString` and the test function `test_getUrlString_givenPathWithSpecialCharacters_expectEncodedPath` both contained some identical code `addingPercentEncoding`. 

When a test function and the code under test contain identical code, I tend to have less confidence in the test function because if, for example, there is a bug in the SDK's use of `addingPercentEncoding`, the test function would not catch this bug because the test function itself would also contain the bug. 

To prevent false positives such as this, I modified the test function body to use a hard-coded value as the expected result. This should help us feel more confident in the test function and it's purpose. Bonus: anyone viewing the test function and the hard-coded expected value can see an example return value from the `getUrlString` function which is good for developer experience. 

# todo
 
- [ ] PR blocked until [slack conversation](https://customerio.slack.com/archives/C01QYDKD0SU/p1694200319312109) resolved. PR may require some modifications after conversation.

### DIFF
--- a/Tests/Common/Service/HttpEndpointTest.swift
+++ b/Tests/Common/Service/HttpEndpointTest.swift
@@ -7,13 +7,15 @@ class HttpEndpointTest: UnitTest {
     private let defaultEndpoint = CIOApiEndpoint.findAccountRegion
     private var httpBaseUrls: HttpBaseUrls!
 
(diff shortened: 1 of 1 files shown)


levibostian:
Closing due to age

PR #384, opened 2023-09-15
ci: enable codecov uploading again

levibostian:
Get test coverage working again on this repo after it broke a while back.


.github/workflows/test.yml
```diff
@@ -27,8 +27,8 @@ jobs:
       with:
         xcode-version: ${{ matrix.xcode }}
     # If running tests fails, sometimes it's because of scheme name is wrong. This gives us all available schemes. 
-    - name: Get XCode schemes (xcode ${{ matrix.xcode }})
-      run: xcrun xcodebuild -list
+    # - name: Get XCode schemes (xcode ${{ matrix.xcode }})
+    #   run: xcrun xcodebuild -list
```
levibostian, line 31:
```suggestion
    - name: Get XCode schemes (xcode ${{ matrix.xcode }})
      run: xcrun xcodebuild -list
```

.github/workflows/test.yml
```diff
@@ -41,18 +41,19 @@ jobs:
...
         lane: 'scan'
 
-    # Read generate script file to learn more about this CI step. 
-    - name: Generate code coverage report from .xcresult/
-      run: ./scripts/generate-code-coverage-report.sh 
+    # https://about.codecov.io/blog/pre-converting-xcresult-files-for-codecov-using-xcresultparser/
+    - name: Generate code coverage report from .xcresult/ that can be converted to codecov format
+      run: |
+        brew tap a7ex/homebrew-formulae && brew install xcresultparser
+        xcresultparser --output-format cobertura "Customer.io-Package.xcresult" > ".build/coverage.xml"
```
levibostian, line 48: This is the most important part of this PR. The previous tool that we were using to convert `xcresult` file into a format that codecov understands was no longer working. So, I went out and tried to find a popular tool that we can depend on into the future. 

Because codecov made a blog post that included this tool, it made me feel confident in this tool.

.github/workflows/test.yml
```diff
@@ -41,18 +41,19 @@ jobs:
...
-    #     fail_ci_if_error: true # fail if upload fails so we can catch it and fix it right away.
-    #     verbose: true 
-    #     directory: .build/generated
+    - name: Upload code coverage report 
+      uses: codecov/codecov-action@v3
+      with: 
+        token: ${{ secrets.CODECOV_UPLOAD_TOKEN }} # not required for public repos, but sometimes uploads fail without it so include it anyway
+        files: .build/coverage.xml
+        fail_ci_if_error: true # fail if upload fails so we can catch it and fix it right away.
+        verbose: true
```
levibostian, line 56: Updated action to latest v3 and modified the file path to an up-to-date one.

Package.swift
```diff
@@ -7,20 +7,25 @@
...
+var products: [PackageDescription.Product] = [
+    .library(name: "Tracking", targets: ["CioTracking"]),
+    .library(name: "MessagingPushAPN", targets: ["CioMessagingPushAPN"]),
+    .library(name: "MessagingPushFCM", targets: ["CioMessagingPushFCM"]),
+    .library(name: "MessagingInApp", targets: ["CioMessagingInApp"])
+]
+
+if (ProcessInfo.processInfo.environment["CI"] != nil) {
+    products.append(.library(name: "InternalCommon", targets: ["CioInternalCommon"]))
+}
```
levibostian, line 21: I plan on writing a comment in this file to explain what this logic is about. 

A gist is: xcode requires including `InternalCommon` as a library in order to get test coverage data on the module's code. 

But we want to hide the `InternalCommon` from customers when they try to install our SDK into their app using SPM. Therefore, I am making this file dynamic so the `InternalCommon` module only shows up when the SDK is being installed on the CI server.

codecov.yml
```diff
@@ -13,4 +13,11 @@ coverage:
...
-        informational: true 
\ No newline at end of file
+        informational: true 
+
+# The file that contains code coverage information may contain coverage for files that we do not want to keep track of. 
+# Example: We do not want to know the coverage for the Tests/ directory because that does not contain source code of our SDK. 
+# https://docs.codecov.com/docs/ignoring-paths
+ignore: 
+  - "Tests" # ignore code coverage of Test/ files 
+  - "**/*.generated.swift" # Ignore code coverage of generated files.
```
levibostian, line 23: Added some ignore blocks to make our test coverage data more accurate. These files are not relevant to test coverage so we should ignore them.

Shahroz16 (approved):

Package.swift
levibostian: Docs wrote.

PR #387, opened 2023-09-25
ci: xcode15 stable instead of beta

levibostian:
I noticed on `main` that sample apps are not building anymore. Looks like it's because xcode 15 is no longer in beta and so xcode version `15-beta` is not installing. This PR attempts to fix CI building of sample apps to fix `main`.

### DIFF
--- a/.github/workflows/build-sample-apps.yml
+++ b/.github/workflows/build-sample-apps.yml
@@ -70,7 +70,7 @@ jobs:
     - name: Set up Xcode
       uses: maxim-lobanov/setup-xcode@v1
       with:
-        xcode-version: "15.0-beta"
+        xcode-version: "15"
 
     - name: Install CLI tools used in CI script 
       run: |


Shahroz16 (approved):

.github/workflows/deploy-sdk.yml
```diff
@@ -32,12 +30,6 @@ jobs:
       - name: Install sd CLI to use later in the workflow 
         uses: kenji-miyake/setup-sd@v1
 
-      # once this issue is resolved, we can remove this workaround. https://github.com/cycjimmy/semantic-release-action/issues/159#issuecomment-1490892625
```
jatinn, line 35: issue still seems to be open, are we sure this has been fixed?

jatinn (approved):

.github/workflows/deploy-sdk.yml
levibostian: I have not been experiencing this issue in a while which is why I removed it. 

Since adding this workaround, the github action, the tool, and all plugins have gone through multiple major version updates. Making me think that there is a chance this issue no longer exists.

ami-aman (approved):

PR #383, opened 2023-09-13
build: setup binny to install and run CLI binary tools with hard-coded versions

levibostian:
Part of [fall cleaning project](https://www.notion.so/custio/Devs-and-CI-use-same-version-of-iOS-CLI-tools-9b7a8cfb42f14fdf9e2de93114fa2e89?pvs=4). 

This PR fixes the problem that we have today where all developers and CI servers use a different version of development tools from one another. In order to fix the problems that we currently experience, we need to assert that everyone is using the same version of, for example, `swiftformat` to avoid unnecessary lint errors and complicated merge conflicts. 

# todo
- [x] Experiment with deliverability of the script to team members to avoid them needing to install Deno to execute script.

PR #388, opened 2023-09-28
ci: fix the min version for semantic release

Shahroz16:
version should be greater than the 22.0.1 (https://github.com/semantic-release/semantic-release/releases/tag/v22.0.1)

### DIFF
--- a/.github/workflows/deploy-sdk.yml
+++ b/.github/workflows/deploy-sdk.yml
@@ -45,11 +45,9 @@ jobs:
           # version numbers below can be in many forms: M, M.m, M.m.p
           # version should be greater than the 22.0.1 (https://github.com/semantic-release/semantic-release/releases/tag/v22.0.1)
           # because previous version had a bug in commit analyzer
-          semantic_version: 22.0.5
+          semantic_version: latest
           extra_plugins: |
             conventional-changelog-conventionalcommits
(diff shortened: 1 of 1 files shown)


mrehan27 (approved):

.github/workflows/deploy-sdk.yml
```diff
@@ -42,6 +42,10 @@ jobs:
         uses: cycjimmy/semantic-release-action@v4
         id: semantic-release
         with:
+          # version numbers below can be in many forms: M, M.m, M.m.p
+          # version should be greater than the 22.0.1 (https://github.com/semantic-release/semantic-release/releases/tag/v22.0.1)
+          # because previous version had a bug in commit analyzer
+          semantic_version: 22.0.5
```
levibostian, line 48:
```suggestion
          semantic_version: latest
```

npm supports keyword `latest` to always get the latest version of the tool. 

I prefer we use `latest` instead of a hard-coded version so we can use future bug fixes introduced by the tool, too.

levibostian:
I see lint check failed. I would assume that it's because swiftformat came out with a new version that `main` didn't get executed against? 

If this PR gets merged into `main` https://github.com/customerio/customerio-ios/pull/383, that could fix this lint check instead of having to manually lint to get status checks to pass.

PR #391, opened 2023-10-05
feat: auto fetch device token and auto register device (apns)

ami-aman:
Refers: https://github.com/customerio/issues/issues/11179
Refers: https://github.com/customerio/issues/issues/10298

This feature PR simplifies device token registration for customers. Instead of writing multiple code snippets, customers can simply add a configuration to their AppDelegate file, and the SDK will automatically fetch and register the device token. The following is the use case for an app using APN.

```
MessagingPushAPN.initialize { config in
}
```
 
 ### Note

The file changed count is 13 and includes:
- Feature implementation
- Sample app update with the new feature
- autogenerated files

### Pending
Test cases


Sources/Tracking/CustomerIO.swift
```diff
@@ -1,5 +1,9 @@
 import CioInternalCommon
+import FirebaseMessaging
```
levibostian, line 2:
```suggestion
```

I think this was forgotten to be removed.

Sources/Tracking/CustomerIO.swift
```diff
@@ -1,5 +1,9 @@
...
 import Foundation
+#if canImport(UIKit)
+import UIKit
+#endif
```
levibostian, line 6:
```suggestion
```

I think this was forgotten to be removed.

Sources/Tracking/CustomerIO.swift
```diff
@@ -20,7 +24,7 @@ public protocol CustomerIOInstance: AutoMockable {
         // sourcery:TypeCast="AnyEncodable(body)"
         body: RequestBody
     )
-
+    func registeredDeviceToken() -> String?
```
levibostian, line 27:
```suggestion
    var registeredDeviceToken: String? { get }
```

As mentioned in a previous PR, it would be ideal if this was a property instead of a function call.

Sources/MessagingPushAPN/MessagingPushAPN.swift
```diff
@@ -76,6 +76,24 @@ public class MessagingPushAPN: MessagingPushAPNInstance {
         messagingPush.trackMetric(deliveryID: deliveryID, event: event, deviceToken: deviceToken)
     }
 
+    /**
+     Configure `MessagingPushAPN`.
+     Call this function in your app if you want to configure the module.
+     */
+    @available(iOSApplicationExtension, unavailable)
+    public static func configure(configOptions configureHandler: ((inout MessagingPushConfigOptions) -> Void)?
```
levibostian, line 84: I am leaning towards having this function be inside of `MessagingPush` class instead of in `MessagingPushAPN`. I am assuming the FCM implementation of this feature will need this function therefore, I dont see a reason to also have a separate `MessagingPushFCM.configure` function.

Sources/MessagingPushAPN/MessagingPushAPN+PushConfigs.swift
```diff
@@ -0,0 +1,35 @@
+import Foundation
+import UIKit
+
+extension MessagingPushAPN {
+    func setupAutoFetchDeviceToken() {
+        // Register for push notifications
+        UIApplication.shared.registerForRemoteNotifications()
+        // Swizzle method `didRegisterForRemoteNotificationsWithDeviceToken`
+        swizzleDidRegisterForRemoteNotifications()
```
levibostian, line 9:
```suggestion
        swizzleDidRegisterForRemoteNotifications()
        UIApplication.shared.registerForRemoteNotifications()
```

Shouldn't these 2 function calls be switched in order? If not, I think a comment explaining why this is the order would be helpful. 

I was expecting that the SDK needed to swizzle *first* in order to receive the callback when `registerForRemoteNotifications` gets called?

Sources/MessagingPushAPN/MessagingPushAPN+PushConfigs.swift
```diff
@@ -0,0 +1,35 @@
...
+
+    private func swizzleDidRegisterForRemoteNotifications() {
+        let appDelegate = UIApplication.shared.delegate
+        let appDelegateClass: AnyClass? = object_getClass(appDelegate)
+        let originalSelector = #selector(UIApplicationDelegate.application(_:didRegisterForRemoteNotificationsWithDeviceToken:))
+        let swizzledSelector = #selector(application(_:didRegisterForRemoteNotificationsWithDeviceToken:))
+        swizzle(forOriginalClass: appDelegateClass, forSwizzledClass: MessagingPushAPN.self, original: originalSelector, new: swizzledSelector)
+    }
+
+    private func swizzle(forOriginalClass: AnyClass?, forSwizzledClass: AnyClass?, original: Selector, new: Selector) {
```
levibostian, line 20: [We have a swizzle function already in the SDK](https://github.com/customerio/customerio-ios/blob/b0f309f1da66a6717a48feba2a0edc7555c8b629/Sources/Tracking/CustomerIOImplementation%2BScreenViews.swift#L23-L26). Could you refactor the code so the code base only has 1 swizzle function that gets re-used?

Sources/MessagingPushAPN/MessagingPushAPN.swift
levibostian: Since this is a new public API being added, add an entry to `MessagingPushAPN/APITests` file.

Sources/MessagingPushAPN/MessagingPushAPN+PushConfigs.swift
```diff
@@ -0,0 +1,35 @@
+import Foundation
+import UIKit
+
+extension MessagingPushAPN {
+    func setupAutoFetchDeviceToken() {
```
levibostian, line 5: I think we can write automated tests for this: 

```swift
// Note: non-working pseudocode 

let app = UIApplication()
app.delegate = ....
// before swizzling, we do not expect CIO SDK to be called
delegate.application(didRegisterForRemoteNotificationsWithDeviceToken:)
expect(cioSdk.registerDeviceToken).not.called()

MessagingPushAPN.configure {}
delegate.application(didRegisterForRemoteNotificationsWithDeviceToken:givenToken)
expect(cioSdk.registerDeviceToken).called(givenToken)
```

Could you try adding automated tests to verify this logic?

Apps/APN-UIKit/APN UIKit/AppDelegate.swift
```diff
@@ -80,11 +83,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
     }
 
-    func application(_ application: UIApplication, didRegisterForRemoteNotificationsWithDeviceToken deviceToken: Data) {
-        storage.deviceToken = String(apnDeviceToken: deviceToken)
-        MessagingPush.shared.application(application, didRegisterForRemoteNotificationsWithDeviceToken: deviceToken)
-    }
-
     func application(_ application: UIApplication, didFailToRegisterForRemoteNotificationsWithError error: Error) {
         MessagingPush.shared.application(application, didFailToRegisterForRemoteNotificationsWithError: error)
     }
```
levibostian, line 88: Can we remove this function, too? Or also swizzle it?

Sources/Common/Store/MessagingPushConfigOptions.swift
```diff
@@ -0,0 +1,21 @@
+public struct MessagingPushConfigOptions {
```
levibostian, line 1: I don't think this belongs in `Common` module. It seems like a file best suited for `MessagingPush` module.

Sources/Tracking/CustomerIO.swift
ami-aman: ✅

Sources/Tracking/CustomerIO.swift
ami-aman: ✅

Apps/APN-UIKit/APN UIKit/AppDelegate.swift
```diff
@@ -51,6 +51,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
         // Add event listeners for in-app. This is not to initialise in-app but event listeners for in-app.
         MessagingInApp.initialize(eventListener: self)
+        MessagingPushAPN.configure { config in
```
Shahroz16, line 54: can we maybe combine `configure` and `initialize`? maybe, initializing returns `config` that can be used?

levibostian: To +1 to Shahroz's comment, the SDK does have a naming pattern that we could duplicate in the Push module, too. 

Since we already use `initialize` in other places: 

`CustomerIO.initialize()`
`MessagingInApp.initialize()`

It would be good if the push module was also `initialize`.

ami-aman: I named the method `configure` because initialization is not a mandatory step in this case. The main purpose of this API is to configure the `MessagingPushAPN` module. However, since two votes are in favor of `initialize`, I will change the name to `initialize`. Also, note that we do not currently have an initialization method in `MessagingPushAPN`.

Sources/MessagingPushAPN/MessagingPushAPN.swift
ami-aman: Adding this method in MessagingPush might create issues with FCM implementation as Firebase libraries are not accessible in MessagingPush. Also, there is no way to communicate from MessagingPush -> MessagingPushFCM to help passing the control to FCM module.

Hence I am keeping the configure method (name changed to initialize) as is and keeping the common code in MessagingPush.

Sources/MessagingPushAPN/MessagingPushAPN+PushConfigs.swift
ami-aman: Thanks for the feedback. I do not see any difference in functionality, however I have no issue to swizzle first and then register for remote notifications. Order changed.

Apps/APN-UIKit/APN UIKit/AppDelegate.swift
ami-aman: ✅

Sources/MessagingPushAPN/MessagingPushAPN.swift
levibostian: My original comment saying, "have this function inside of `MessagingPush` class" doesn't describe the idea I had well. Sorry about that. 

I see a day where we remove the need for our SDK to have modules initialized. So I am OK with having `MessagingPushAPN.initialize` and `MessagingPushFCM.initialize`.

Sources/MessagingPushAPN/MessagingPushAPN+PushConfigs.swift
ami-aman: Yeah, I noted this too. I am wondering how can we achieve this with minimal efforts. One way is to put this code in Common module as `CioMethodSwizzle`, that can then be accessible through all other modules. But with this I think, MessagingPushAPN/MessagingPushFCM might still not be able to access CioMethodSwizzle class because there is no diGraph in APN/FCM module. 🤔 
Just thinking out loud, because now there 3 occurrences of swizzle.

Apps/APN-UIKit/APN UIKit/AppDelegate.swift
ami-aman: Swizzle.d

Sources/Common/Store/MessagingPushConfigOptions.swift
ami-aman: ✅

Sources/MessagingPushAPN/MessagingPushAPN+PushConfigs.swift
levibostian: Good points. 

Because we are swizzling inside of a top-level class (`MessagingPushFCM`), directly referencing `DIGraph` instance from `SdkInitializedUtil` would be OK to do. 

Therefore, you could create a simple `Swizzle` class in the Common module, add that new class to the DiGraph, then inside of `MessagingPushFCM` classes, get a reference of `DiGraph` to then get an instance of this new Swizzle class. 

The `MessagingPush` module does have a [class that is a `ModuleTopLevelObject`](https://github.com/customerio/customerio-ios/blob/main/Sources/MessagingPush/MessagingPush.swift#L9) which has an instance of `DiGraph` ready to use. However, using that class in the `MessagingPushFCM` module might be complicated?

Sources/MessagingPush/MessagingPushConfigOptions.swift
```diff
@@ -0,0 +1,21 @@
...
+    public enum Keys: String { // Constants used to map each of the options in MessagingPushConfigOptions
+        case autoFetchDeviceToken
+    }
+
+    /**
+     Enable automatic fetching of device token by the SDK without the need to write custom code by the customer.
+     On fetching the token, SDK will auto-register the device. This value is `true` by default.
+     */
+    public var autoFetchDeviceToken: Bool
+}
```
levibostian, line 21:
```suggestion
public struct MessagingPushConfigOptions {
    /**
     Enable automatic fetching of device token by the SDK without the need to write custom code by the customer.
     On fetching the token, SDK will auto-register the device. This value is `true` by default.
     */
    public var autoFetchDeviceToken: Bool = true 
}
```

I think this struct can be simplified to provide the default value by assigning it to the property directly.

levibostian (approved):
Logic of the PR looks good so I am approving. 

Before merging, I think it would be good to review and resolve the PR comments I left previously that are still open.

Sources/MessagingPushAPN/MessagingPushAPN+PushConfigs.swift
ami-aman: As discussed, this can be taken care at a later point. Not a blocker for this PR.

Sources/MessagingPush/MessagingPushConfigOptions.swift
ami-aman: This change will require code change in FCM and APN implementation as both modules use the same logic. There could be possibly other keys as well to be added and I believe we should let this as is as this is a tested logic in use for other configurations as well.

PR #392, opened 2023-10-06
fix: prevent registering device with empty identifier existing BQ tasks

levibostian:
Quick bug fix for customer to prevent `/api/v1/customers//devices/` (empty profile identifier) requests from being made.


Sources/Common/Background Queue/QueueRunner.swift
```diff
@@ -92,6 +92,13 @@ private extension CioQueueRunner {
             return onComplete(failureIfDontDecodeTaskData)
         }
 
+        // If a HTTP request is made to register device token with an empty identifier, either (1) a 302 will occur or (2) a "devices" profile will be created in a CIO workspace. Neither are ideal.
+        // The SDK has logic to prevent *new* register device tasks being added to the BQ if identifier is empty but existing tasks in the BQ will still occur.
+        // This logic is to clear those tasks from the BQ.
```
levibostian, line 97: Hopefully, this comment explains why this PR was made and why this change is needed.

Sources/Common/Background Queue/QueueRunner.swift
```diff
@@ -92,6 +92,13 @@ private extension CioQueueRunner {
...
+        if taskData.profileIdentifier.isBlankOrEmpty() {
+            return onComplete(.success(()))
```
Shahroz16, line 99: is there a reason, we return `success` for this task, which will eventually also delete instead of just `deleting`? 

The register device token is blocked by the Identify task, so it should run in the first place, but in case if identify task has been run and was faulty e.g. `empty identifier` then why not just delete this task?

Tests/Common/Background Queue/QueueRunnerTest.swift
```diff
@@ -0,0 +1,36 @@
...
+
+    override func setUp() {
+        super.setUp()
+
+        runner = CioQueueRunner(jsonAdapter: jsonAdapter, logger: log, httpClient: httpClientMock, hooksManager: diGraph.hooksManager, sdkConfig: sdkConfig)
+    }
+
+    // MARK: registerPushToken
+
+    func test_registerPushToken_givenIdentifierEmpty_expectHttpRequestNotMade_expectBQDeletesTask() {
```
Shahroz16, line 19: can we have a test for the `failure` of the `result` as well?

Sources/Common/Background Queue/QueueRunner.swift
levibostian: > is there a reason, we return success for this task, which will eventually also delete instead of just deleting?

You are correct. The reason for returning success is to tell the BQ to delete the task. 

> but in case if identify task has been run and was faulty e.g. empty identifier 

I am not sure if it would be faulty? The problem that we have been experiencing is HTTP requests with empty identifier ID for this register device endpoint. I don't recall the SDK having issues with identifying a profile with the API with an empty identifier. Therefore, I don't think we need to modify identify profile BQ tasks? 

---

Are you suggesting that we modify the solution to a SDK migration where we iterate the BQ and delete tasks from it? 

Assuming that is what your comment is suggesting...From my experience building BQs, I have learned that BQ code gets complex quickly when you begin modifying tasks that are already in the BQ. An approach I have found to be less complex is to do the modify logic when that BQ task is executing. Therfore, that's where this solution comes from. When the BQ task runs, tell the BQ to delete it.

Shahroz16: I was suggesting, something like this? 

```
if taskData.profileIdentifier.isBlankOrEmpty()
queuStore.delete(taskID)
return
```

Shahroz16 (approved):
Approving as these aren't blockers but rather different approaches.

Tests/Common/Background Queue/QueueRunnerTest.swift
levibostian: Ah, good call-out. 

I added a function to Swift's built-in Result class. Looking at the diff, I do see that I didn't name the file correctly and I did not write tests for the added function. 

I just pushed a commit taking care of those things. Thanks for bringing that up.

PR #393, opened 2023-10-18
fix: in-app positioning and persistence

Shahroz16:
More details on it can be found on [Slack](https://customerio.slack.com/archives/C036FTE6UJ0/p1697540277891429).


BernardGatt:
LGTM

PR #395, opened 2023-10-20
feat: fetch fcm device token automatically 

ami-aman:
Merges into https://github.com/customerio/customerio-ios/pull/391
Refers: https://github.com/customerio/issues/issues/11179
Refers: https://github.com/customerio/issues/issues/10298

This feature PR simplifies FCM device token registration for customers. Instead of writing multiple code snippets, customers can simply add a initialization to their AppDelegate file, and the SDK will automatically fetch and register the device token. The following is the use case for an app using FCM.

```
MessagingPushFCM.initialize { config in
}
```
 
 ### Note
PR has:
- Feature implementation
- Sample app update with the new feature
- autogenerated files


Apps/CocoaPods-FCM/src/AppDelegate.swift
```diff
@@ -33,7 +33,9 @@ class AppDelegate: NSObject, UIApplicationDelegate {
         // Now that the Firebase and Customer.io SDK's are initialized, follow the rest of the required steps for the FCM push setup.
         UNUserNotificationCenter.current().delegate = self
 
-        Messaging.messaging().delegate = self
+        MessagingPushFCM.initialize { config in
+            config.autoFetchDeviceToken = true
+        }
```
levibostian, line 38:
```suggestion
        // Initialize Customer.io push messaging allows 
        // the SDK to automatically send FCM push tokens to 
        // Customer.io! 
        MessagingPushFCM.initialize()
```

Adding a comment that could be helpful to customers. 

I also like the idea of using default values in the sample apps but I am open to what others prefer.

Sources/MessagingPushFCM/MessagingPushFCM+PushConfigs.swift
```diff
@@ -0,0 +1,46 @@
...
+    @objc
+    func application(_ application: UIApplication, didRegisterForRemoteNotificationsWithDeviceToken deviceToken: Data) {
+        Messaging.messaging().apnsToken = deviceToken
+        Messaging.messaging().token(completion: { token, _ in
+            guard let token = token else {
+                return
+            }
+            Self.shared.registerDeviceToken(fcmToken: token)
+        })
+    }
```
levibostian, line 45:
```suggestion
        // Registers listener with FCM SDK to always have the latest FCM token. Used to automatically register it with the SDK. 
        Messaging.messaging().token(completion: { token, _ in
            guard let token = token else {
                return
            }
            Self.shared.registerDeviceToken(fcmToken: token)
        })
    }
```

Added comment.

levibostian: I wonder, since this code is registering a callback function with the FCM SDK, it feels more natural to me for this block of code to exist in `MessagingPushFCM.initialize` function instead of in a swizzled function that could be called multiple times.

levibostian (approved):
Awesome!

Apps/CocoaPods-FCM/src/AppDelegate.swift
ami-aman: Updated in latest commit

Sources/MessagingPushFCM/MessagingPushFCM+PushConfigs.swift
ami-aman: Thanks for the review, Levi ! If I add this FCM callback method in `MessagingPushFCM.initialize` then it may npt work as expected as the token may be generated after initialization method.

ami-aman: Added comment.

levibostian:
To fix the CI building sample apps, try to merge in `main` branch.

Sources/MessagingPushFCM/MessagingPushFCM+PushConfigs.swift
levibostian: > may not work as expected as the token may be generated after initialization method.

This is a non-blocker for this PR, but I still want to learn more about why putting `Messaging.messaging().token(completion: )` in `MessagingPushFCM.initialize` would not work. Can you provide more details on why this would not work? Maybe steps that could reproduce the behavior of it not working? 

If `Messaging.messaging().token(completion: )` will only work in certain places in the code, I think a comment should be added to this code block explaining that. So we know in the future where this block of code should go.

Shahroz16 (approved):

Sources/MessagingPushFCM/MessagingPushFCM+PushConfigs.swift
ami-aman: This is because that the token might get generated after a delay and not immediately. One such reason is that FCM token needs APN token and it could get created later as in my case it is. 

Steps to reproduce:
- Make the change as above
- Open FCM-CocoaPods app
- Run FCM-CocoaPods app from this branch

You'll notice that when placed in initialization, the token is not generated. I can get on a call to explain this if needed.

I hope this helps.

PR #399, opened 2023-10-27
fix: memory exception during SDK initialization async tasks

levibostian:
This is a bug fix to resolve [issue reported by React Native customer](https://github.com/customerio/customerio-reactnative/issues/168). 

- I created integration test to reproduce the issue. The test suite crashes when running the newly added test. 
- I wrote the bug fix in the `CustomerIO` class. Test passes after bug fix implemented.

### DIFF
--- a/Sources/Tracking/CustomerIO.swift
+++ b/Sources/Tracking/CustomerIO.swift
@@ -118,7 +118,7 @@ public class CustomerIO: CustomerIOInstance {
     var diGraph: DIGraph?
 
     // strong reference to repository to prevent garbage collection as it runs tasks in async.
-    private var cleanupRepository: CleanupRepository?
+    @Atomic private var cleanupRepository: CleanupRepository?
 
     // private constructor to force use of singleton API
     private init() {}
@@ -215,8 +215,6 @@ public class CustomerIO: CustomerIOInstance {
         let logger = diGraph.logger
         let siteId = diGraph.sdkConfig.siteId
 
-        cleanupRepository = diGraph.cleanupRepository
-
         // Register Tracking module hooks now that the module is being initialized.
         hooks.add(key: .tracking, provider: TrackingModuleHookProvider())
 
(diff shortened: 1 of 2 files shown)


Sources/Tracking/CustomerIO.swift
```diff
@@ -227,9 +225,20 @@ public class CustomerIO: CustomerIOInstance {
...
+        // - New instance of CleanupRepository created by thread B.
+        // - Async operation in background thread begins. Tries to reference repository instance that was created by thread A, where it got scheduled.
+        // - Memory exception thrown because old repository instance from thread A is gone.
+        if cleanupRepository == nil { // Using cleanupRepository instance to determine if this has been run before.
+            cleanupRepository = diGraph.cleanupRepository
+
+            // run cleanup in background to prevent locking the UI thread
+            threadUtil.runBackground { [weak self] in
+                // Crash occurs on line below if repository gets re-assigned
+                self?.cleanupRepository?.cleanup()
```
Shahroz16, line 240: should we also make it nil? 

```
strongSelf.cleanupRepository = nil
```

to remove the strong reference afterwards?

levibostian: We only want the cleanup logic to execute one time while the app is in memory. If we were to set `strongSelf.cleanupRepository = nil`, then the added `if cleanupRepository == nil {` would be `true` and could run all of this cleanup again. 

I currently do not see a need to remove this strong reference at this time. If you can think of a use case,  I would love to hear it.

Shahroz16: My only concern at the moment from the description of the ticket, it says 

>memory exception when cleanup repo created by multiple threads 

so it seems like a concurrency issue, while the solution doesn't cater to that?  maybe we should synchronize the block of code?

Tests/Tracking/CustomerIOIntegrationTests.swift
```diff
@@ -156,4 +156,11 @@ class CustomerIOIntegrationTests: IntegrationTest {
         XCTAssertEqual(CustomerIO.shared.config!.siteId, givenSiteId)
         XCTAssertEqual(CustomerIO.shared.config!.apiKey, givenApiKey)
     }
+
+    // Note: It's very important that this is an integration test that uses *real* instances of ThreadUtil to run code in a real background thread and the dependencies in SDK initialization, such as CleanupRepository, are all real instances.
+    func test_initializeSdk_givenCallMultipleTimes_expectNoExceptions() {
```
Shahroz16, line 161: can we have a test for multiple threads as well maybe using dispatch queue?

Sources/Tracking/CustomerIO.swift
levibostian: I can see how that is confusing. The best explanation for this bug is [documented here](https://github.com/customerio/opsbugs/issues/3618#issuecomment-1640910138). 

Now that I think about it, the title is actually inaccurate. The bug is caused because of async operations, not from multiple threads. 

I updated the title to try and be more accurate.

Tests/Tracking/CustomerIOIntegrationTests.swift
levibostian: That is a great suggestion. 

[From this comment](https://github.com/customerio/customerio-ios/pull/399/files#r1385062478), I tried to explain that there was a miscommunication with what this bug is. 

Based on that, this test is a good test that reproduces the crash stacktrace reported to us from customers.

Shahroz16: Awesome, thanks for clarifying that. 🙏

Shahroz16 (approved):
Thanks for updating the description.

levibostian:
PR for React Native to include this bug fix: https://github.com/customerio/customerio-reactnative/pull/217

PR #403, opened 2023-11-09
feat: automatic push click handling

levibostian:
iOS part of ticket: https://github.com/customerio/issues/issues/11289

I suggest [starting here](https://github.com/customerio/customerio-ios/pull/403#discussion_r1417966913) in your PR review.


Sources/MessagingPush/ClickHandling/NotificationCenterDelegate.swift
```diff
@@ -0,0 +1,52 @@
+import UserNotifications
```
levibostian, line 1: I am in the middle of moving code out of this file into `Sources/MessagingPush/ClickHandling/MessagingPush+ClickHandling.swift`. So, view that file instead 😄

Sources/MessagingPush/PushClickHandler.swift
```diff
@@ -0,0 +1,220 @@
...
+
+        // TODO: duplicate metrics could be reported because we might be swizzling multiple delegates and so this function gets called X number of times.
+        // https://github.com/customerio/issues/issues/11150 should we do this?
+
+        // continue swizzle
+        cio_swizzle_didReceive(center, didReceive: response, withCompletionHandler: completionHandler)
+    }
+}
+
+// TODO: cleanup duplicate swizzle functions all over codebase.
```
levibostian, line 173: To avoid scope creep, cleaning up of duplicate swizzled functions will be done in another PR. 

Work has begun on branch `levi/duplicate-swizzling`

Sources/MessagingPush/PushClickHandler.swift
```diff
@@ -0,0 +1,220 @@
...
+    if didAddMethod {
+        class_replaceMethod(forClass, new, method_getImplementation(originalMethod), method_getTypeEncoding(originalMethod))
+    } else {
+        method_exchangeImplementations(originalMethod, swizzledMethod)
+    }
+}
+
+// https://github.com/OneSignal/OneSignal-iOS-SDK/blob/5ac5927fca8361a3af9cee1262ac11fd49e218ee/iOS_SDK/OneSignalSDK/OneSignalCore/Source/Swizzling/OneSignalSelectorHelpers.m#L33
+func swizzle(targetClass: AnyClass, targetSelector: Selector, myClass: AnyClass, mySelector: Selector) {
+    // TODO: make this runtime safe.
```
levibostian, line 192: This TODO is important to do in this PR. Will push a commit to fix this.

Sources/MessagingPush/Store/PushHistory.swift
```diff
@@ -0,0 +1,47 @@
+import CioInternalCommon
+import Foundation
+
+protocol PushHistory: AutoMockable {
+    func hasHandledPushClick(deliveryId: String) -> Bool
+    func handledPushClick(deliveryId: String)
+}
+
+// sourcery: InjectRegister = "PushHistory"
+class PushHistoryImpl: PushHistory {
```
levibostian, line 10: New class meant to prevent duplicate push metrics and duplicate deep link handling when a push is clicked on. 

We have encountered this problem for a while. This is meant to be a long-term solution to the problem by keeping track of last 100 delivery-ids the SDK has handled.

Sources/MessagingPush/PushClickHandler.swift
```diff
@@ -0,0 +1,220 @@
...
+        // so our SDK still gets notified when a push is clicked, even though the new delegate is setup to handle the push click.
+        swizzle(
+            targetClass: type(of: delegate),
+            targetSelector: #selector(UNUserNotificationCenterDelegate.userNotificationCenter(_:didReceive:withCompletionHandler:)),
+            myClass: PushClickHandlerImpl.self,
+            mySelector: #selector(PushClickHandlerImpl.cio_swizzle_didReceive(_:didReceive:withCompletionHandler:))
+        )
+    }
+
+    func pushClicked(response: UNNotificationResponse) -> CustomerIOParsedPushPayload? {
```
levibostian, line 71: This function does not yet compile as I am not yet done with this refactor. 

The idea here is to encapsulate the logic of click handling into 1 class to (1) prevent duplicate push metrics and deep link handling when a push is clicked (2) make this code more testable being in it's own class then `MessagingPush` extension functions, and (3) encapsulate automatic and manual push handling logic into 1 place in the code.

Sources/MessagingPush/PushClickHandler.swift
```diff
@@ -0,0 +1,200 @@
...
+            .parse(
+                notificationContent: response.notification.request.content,
+                jsonAdapter: jsonAdapter
+            )
+        else {
+            // Push not sent by CIO. Exit early. Return nil to indicate that the push not from CIO.
+            return nil
+        }
+
+        // TODO: prevent duplicate push metrics and deep link handling.
```
levibostian, line 87: To keep this PR small, a new PR will be opened to address this TODO.

Sources/MessagingPush/PushClickHandler.swift
```diff
@@ -0,0 +1,189 @@
...
+    func pushClicked(_ response: UNNotificationResponse) -> CustomerIOParsedPushPayload?
+    func setupClickHandling()
+}
+
+// Make class a singleton to avoid being garbage collected. A strong reference needs to be held of instance.
+// We are setting this class to be UNUserNotificationCenter.delegate instance and delegates are usually weak.
+//
+// sourcery: InjectRegister = "PushClickHandler"
+// sourcery: InjectSingleton
+class PushClickHandlerImpl: NSObject, PushClickHandler, UNUserNotificationCenterDelegate {
```
levibostian, line 16: New class added which encapsulates all push click behavior of the SDK. This includes the automated push click handling setup as well as what to do once a push does get clicked.

Sources/MessagingPush/PushClickHandler.swift
```diff
@@ -0,0 +1,213 @@
...
+    // There are files that are created just for displaying a rich push. After a push is interacted with, those files
+    // are no longer needed.
+    // This function's job is to cleanup after a push is no longer being displayed.
+    func cleanupAfterPushInteractedWith(pushContent: CustomerIOParsedPushPayload) {
+        pushContent.cioAttachments.forEach { attachment in
+            let localFilePath = attachment.url
+
+            try? FileManager.default.removeItem(at: localFilePath)
+        }
+    }
```
levibostian, line 145: This logic was moved here from `Sources/MessagingPush/MessagingPush+AppDelegate.swift` file.

Sources/MessagingPush/PushClickHandler.swift
```diff
@@ -0,0 +1,213 @@
...
+}
+
+// Swizzle functions
+extension PushClickHandlerImpl {
+    // Swizzled method that gets called when a new UNUserNotificationCenter.delegate gets set.
+    @objc dynamic func cio_swizzled_setDelegate(delegate: UNUserNotificationCenterDelegate?) {
+        guard let delegate = delegate else {
+            cio_swizzled_setDelegate(delegate: delegate) // continue swizzle
```
levibostian, line 153: Explanation of `continue swizzling` is in another branch [here](https://github.com/customerio/customerio-ios/commit/6050e7b6cb4f3329c7a1b539ed14e0e7970eae3a#diff-aafef34d456fef5f84dba8d1dc5f3751b89d3dcdedfbfa1572d4806b8f70b426R3-R22).

Shahroz16 (approved):
Looks okay to me, swizzling scared me a little so i believe, manual testing and verification would help here alot.

Sources/MessagingPush/MessagingPush+AppDelegate.swift
```diff
@@ -53,60 +53,26 @@ extension MessagingPushImplementation {
...
-                // A hack to get an instance of deepLinkUtil without making it a property of the MessagingPushImplementation class. deepLinkUtil is not available to app extensions but MessagingPushImplementation is.
-                // We get around this by getting a instance in this function, only.
-                if let deepLinkUtil = sdkInitializedUtil.postInitializedData?.diGraph.deepLinkUtil {
-                    deepLinkUtil.handleDeepLink(deepLinkUrl)
-                }
-            }
-        default: break
+        // A hack to get an instance of pushClickHandler without making it a property of the MessagingPushImplementation class. pushClickHandler is not available to app extensions but MessagingPushImplementation is.
+        // We get around this by getting a instance in this function, only.
+        if let pushClickHandler = sdkInitializedUtil.postInitializedData?.diGraph.pushClickHandler {
```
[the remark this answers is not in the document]
levibostian: > This gets called from AppDelegate only right?

Yes. The push click handling swizzling *must* occur during `AppDelegate.didFinishLaunching` function. It's a requirement by the OS to do that. 

> otherwise not sure how reliable would it be for wrappers

You are correct. Currently, the SDK being dependent to app lifecycle is a blocker for this feature to launch on RN and Flutter. The only way to be able to launch this feature on those SDKs is to have customers initialize the iOS SDK in the `AppDelegate` *and* in their dart/js code, which is not ideal. This is not a blocker for Expo because we generate all AppDelegate code so it's easier for us to add the required iOS initialize function as the workaround. 

We have not yet prioritized RN and Flutter for this feature. We may decide to launch it after we make the SDK lifecycle independent. But iOS and Expo, this method works.

Apps/APN-UIKit/APN UIKit/AppDelegate.swift
```diff
@@ -44,12 +43,15 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+
+            config.autoTrackPushEvents = true
+
             if let trackUrl = self.storage.trackUrl, !trackUrl.isEmpty {
                 config.trackingApiUrl = trackUrl
             }
         }
 
-        // Add event listeners for in-app. This is not to initialise in-app but event listeners for in-app.
+        // Initialize messaging features after initializing Customer.io SDK
```
ami-aman, line 54:
```suggestion
        // Initialize in-app messaging features after initializing Customer.io SDK
```

Apps/APN-UIKit/APN UIKit/AppDelegate.swift
```diff
@@ -44,12 +43,15 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
         MessagingInApp.initialize(eventListener: self)
         MessagingPushAPN.initialize { config in
```
ami-aman, line 56: Looks like the above comment is more suitable at this initialisation method.

`Initialize messaging features after initializing Customer.io SDK`

Apps/APN-UIKit/APN UIKit/AppDelegate.swift
```diff
@@ -85,31 +87,11 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
 
 extension AppDelegate: UNUserNotificationCenterDelegate {
-    // Delegate called when user responds to a notification. Set delegate in
-    // application:didFinishLaunchingWithOptions: method.
     func userNotificationCenter(_ center: UNUserNotificationCenter, didReceive response: UNNotificationResponse, withCompletionHandler completionHandler: @escaping () -> Void) {
-        let handled = MessagingPush.shared.userNotificationCenter(
-            center,
-            didReceive: response,
-            withCompletionHandler: completionHandler
+        CustomerIO.shared.track(
```
ami-aman, line 91: Is there a reason to add manual push tracking here?

Apps/CocoaPods-FCM/src/AppDelegate.swift
```diff
@@ -23,65 +23,45 @@ class AppDelegate: NSObject, UIApplicationDelegate {
...
-    ) {
-        // Send Customer.io SDK click event to process. This enables features such as
-        // push metrics and deep links.
-        let handled = MessagingPush.shared.userNotificationCenter(
-            center,
-            didReceive: response,
-            withCompletionHandler: completionHandler
+extension AppDelegate: UNUserNotificationCenterDelegate {
+    func userNotificationCenter(_ center: UNUserNotificationCenter, didReceive response: UNNotificationResponse, withCompletionHandler completionHandler: @escaping () -> Void) {
+        CustomerIO.shared.track(
```
ami-aman, line 61: [Same as comment above](https://github.com/customerio/customerio-ios/pull/403/files#r1403182539)

Sources/MessagingPush/PushClickHandler.swift
```diff
@@ -0,0 +1,58 @@
...
+    private let customerIO: CustomerIOInstance
+
+    init(sdkConfig: SdkConfig, deepLinkUtil: DeepLinkUtil, pushHistory: PushHistory, customerIO: CustomerIOInstance) {
+        self.sdkConfig = sdkConfig
+        self.deepLinkUtil = deepLinkUtil
+        self.pushHistory = pushHistory
+        self.customerIO = customerIO
+    }
+
+    func pushClicked(_ parsedPush: CustomerIOParsedPushPayload) {
```
levibostian, line 24: I have encapsulated all of the logic of when a push is clicked into this class. 

Automated tests can be written for this class. That is next on my list to do.

Sources/MessagingPush/PushHandling/AutomaticPushClickHandling.swift
```diff
@@ -0,0 +1,49 @@
+import CioInternalCommon
+import Foundation
+
+/**
+ Sets up the CIO SDK to automatically handle events related to push notifications such as when a push is clicked or deciding if a push should be shown while app is in foreground.
```
levibostian, line 5: Because this feature is a bit complex, I thought that it would be good to document in the code how this feature is implemented, the problems we have to solve when making it, and the goals that this feature has. 

Ask questions if you read this and do not understand part of the feature. I hope these docs can be helpful far into the future as we maintain this feature.

Sources/MessagingPush/PushHandling/ManualPushHandling.swift
```diff
@@ -4,6 +4,12 @@ import Foundation
 import UserNotifications
```
levibostian, line 4: I renamed this file and moved it into the `PushHandling` folder. The new file name and the comments I added to the top of the file are meant to better describe that these functions are separate from the automatic push click handling feature.

Sources/MessagingPush/PushHandling/ManualPushHandling.swift
```diff
@@ -53,60 +59,36 @@ extension MessagingPushImplementation {
...
         withCompletionHandler completionHandler: @escaping () -> Void
     ) -> Bool {
-        guard let pushContent = userNotificationCenter(center, didReceive: response) else {
-            // push did not come from CIO
-            // Do not call completionHandler() because push did not come from CIO. Another service might have sent it so
-            // allow another SDK
-            // to call the completionHandler()
-            return false
-        }
+        guard let _: CustomerIOParsedPushPayload = userNotificationCenter(center, didReceive: response) else {
```
levibostian, line 62: A lot of the logic in this file has been moved to `PushClickHandler` class.

Sources/MessagingPush/PushHandling/NotificationCenterDelegateProxy.swift
```diff
@@ -0,0 +1,37 @@
+import Foundation
+import UserNotifications
+
+// TODO: add comment expkaining tjhs
```
levibostian, line 4:
```suggestion
/*
Because the CIO SDK forces itself to be the app's only push click handler, we want our SDK to still be compatible with other SDKs that also need to handle pushes being clicked. 

This class is a proxy that forwards requests to all other click handlers that have been registered with the app. Including 3rd party SDKs. 
*/
```

Shahroz16 (approved):
While the PR looks generally good, the previous merged PRs and new approach changes make it difficult to distinguish between the two and ripple. So a bit of caution 😅 and would appreciate more reviewers in case i am missing something.

Apps/APN-UIKit/APN UIKit/AppDelegate.swift
```diff
@@ -11,11 +11,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
         // Override point for customization after application launch.
 
-        // Register for remote notifications using APN. On successful registration,
-        // didRegisterForRemoteNotifications delegate method will be called and it
-        // provides a device token. In case, registration fails then
-        // didFailToRegisterForRemoteNotifications will be called.
         initializeCioAndInAppListeners()
+
+        // Register a 2nd push click listener besides the Customer.io SDK.
+        // Tests that the SDK is able to handle push clicks when there are multiple click listeners.
```
mrehan27, line 17: Are we going to push these changes to `main`? Could this be confusing for customers looking into sample when not using multiple click listeners?

Apps/CocoaPods-FCM/src/AppDelegate.swift
```diff
@@ -23,65 +23,45 @@ class AppDelegate: NSObject, UIApplicationDelegate {
...
         MessagingPushFCM.initialize { config in
             config.autoFetchDeviceToken = true
         }
+
+        // Manually get FCM device token. Swizzling hasn't been working for me.
+        Messaging.messaging().delegate = self
+
+        // Register a 2nd push click listener besides the Customer.io SDK.
+        // Tests that the SDK is able to handle push clicks when there are multiple click listeners.
+        UNUserNotificationCenter.current().delegate = self
```
mrehan27, line 43: Is this final for `main`? If yes, can we improve comments by adding some explanation from customer's perspective.

Sources/MessagingPush/Extensions/NotificationCenterExtensions.swift
```diff
@@ -0,0 +1,12 @@
+import Foundation
+import UserNotifications
+
+extension UNNotificationResponse {
```
mrehan27, line 4: Some comments in this file might be helpful

Sources/MessagingPush/PushHandling/AutomaticPushClickHandling.swift
```diff
@@ -0,0 +1,51 @@
...
+
+ 2. When the CIO SDK's instance of `UNUserNotificationCenterDelegate` is called, the SDK forwards that push click event to all other `UNUserNotificationCenterDelegate` instances registered to the app. This allows other SDKs the ability to handle a push click event for pushes not sent by CIO.
+
+ Related code for this logic:
+ * `NotificationCenterDelegateProxy` - stores all other `UNUserNotificationCenterDelegate` instances that have been registered with the host app.
+ * `iOSPushEventListener` - is what calls the proxy class when a push is clicked that was not sent by CIO.
+
+ The goal of this feature is:
+ 1. The customer does not need to interact with `UNUserNotificationCenter` themselves to get the CIO SDK to process when a push is clicked. The CIO SDK should be able to set this up itself.
+ 2. The CIO SDK should be able to stay compatible with other SDKs that also want to handle push click events. A customer should be able to install 2+ push notification SDKs in an app and all of them are able to work, even though iOS only allows 1 `UNUserNotificationCenterDelegate` instnace to be set in the app.
```
mrehan27, line 32: Fixed typo

```suggestion
 2. The CIO SDK should be able to stay compatible with other SDKs that also want to handle push click events. A customer should be able to install 2+ push notification SDKs in an app and all of them are able to work, even though iOS only allows 1 `UNUserNotificationCenterDelegate` instance to be set in the app.
```

Sources/MessagingPush/PushHandling/AutomaticPushClickHandling.swift
```diff
@@ -0,0 +1,51 @@
...
+ */
+@available(iOSApplicationExtension, unavailable)
+protocol AutomaticPushClickHandling: AutoMockable {
+    func start()
+}
+
+@available(iOSApplicationExtension, unavailable)
+// sourcery: InjectRegister = "AutomaticPushClickHandling"
+class AutomaticPushClickHandlingImpl: AutomaticPushClickHandling {
```
mrehan27, line 41: Appreciate all the notes you added, really helpful 👏🏻 But I'm not sure why we introduced `AutomaticPushClickHandling` as it looks like `PushEventListener` could alone handle this?

Sources/MessagingPush/PushHandling/iOSPushEventListener.swift
```diff
@@ -0,0 +1,138 @@
...
+
+        // Forward request to the original implementation that we swizzled. So that the app finishes setting UNUserNotificationCenter.delegate.
+        //
+        // Instead of providing the given 'delegate', provide CIO SDK's click handler.
+        // This will force our SDK to be the 1 push click handler of the app instead of the given 'delegate'.
+        cio_swizzled_setDelegate(delegate: diGraph.pushEventListener.delegate)
+    }
+}
+
+// TODO: cleanup duplicate swizzle functions all over codebase.
```
mrehan27, line 124: Is this still pending?

mrehan27 (approved):
I've added some small comments and don't have any major concerns. Since we're already testing this in sample apps, I'm okay with the changes, provided they can be disabled by customers and work with all our sample apps (including scenarios for single and multiple push providers), and assuming the changes aren't imposing any new limitations on customer implementation.

Apps/APN-UIKit/APN UIKit/AppDelegate.swift
levibostian: Yes, I was planning on pushing these changes to `main` so we can continue to test this important edge case of having multiple push services installed. 

I can see how this could make our example apps more complex to read. I think writing some comments directed towards customers would help. If you have a different idea, please share.

Apps/CocoaPods-FCM/src/AppDelegate.swift
levibostian: I believe this response [here](https://github.com/customerio/customerio-ios/pull/403#discussion_r1425853552) also answers this question?

Sources/MessagingPush/Extensions/NotificationCenterExtensions.swift
levibostian: Any comments in particular that you're wondering? 

I have some bias since I am familiar with the iOS UserNotifications framework so it's more difficult for me to guess what questions you may have.

Sources/MessagingPush/PushHandling/AutomaticPushClickHandling.swift
levibostian: > But I'm not sure why we introduced AutomaticPushClickHandling as it looks like PushEventListener could alone handle this?

That is a fair question and one that I was also asking myself when I created the file. 

The decision behind this file is to house the comments that I wrote describing this feature. I felt the `PushEventListener` file was very large already and I didn't want to make it any larger in size by putting the documentation in there. 

I liked the idea of having a file named "AutomaticPushClickHandling" and the documentation of that feature going inside of this file. That seems easy to find for future developers working in the code base. Felt easier to find and maintain compared to adding docs somewhere else like a notion doc. 

If we don't like this idea, perhaps creating a `README.md` inside of the `Sources/MessagingPush/PushHandling/` or `Sources/MessagingPush/` directory would be more welcome?

Sources/MessagingPush/PushHandling/iOSPushEventListener.swift
levibostian: thanks for asking. 

This is taken care of in the lint fix PR.

Apps/APN-UIKit/APN UIKit/AppDelegate.swift
mrehan27: I'm good with comments to keep it easier for customers to understand 👍🏻

Apps/CocoaPods-FCM/src/AppDelegate.swift
mrehan27: Yeah, sounds good

Sources/MessagingPush/Extensions/NotificationCenterExtensions.swift
mrehan27: Not really anything particularly. But because I'm on the other side and less familiar iOS UserNotifications framework, its hard for me to pick how these checks actually work or what kind of extensions go to these files. The variable names are descriptive enough to guess, but if there is any additional info that you think might help the reader, that would be great.

Sources/MessagingPush/PushHandling/AutomaticPushClickHandling.swift
mrehan27: All makes sense to me. I would have preferred the `README.md` inside `Sources/MessagingPush/`, but I'm okay with this too. And since you have already put efforts writing and testing it, I think we should stick with it for now.

Apps/APN-UIKit/APN UIKit/AppDelegate.swift
levibostian: This isn't push tracking, but instead simply tracking an event for easier QA testing. 

We expect that when clicking on a push that was not sent by CIO, the host app is able to receive the push notification and handle it. By tracking an event, this makes our QA testing easier so you click on a push and expect to see an event tracked.

Ahmed-CIO (changes requested):
the scope of these changes is massive, let's break it down and describe each change in its own right

Ahmed-Ali:
Had a chat with Levi about this one. Given this was something took place a while ago and was already reviewed, I won't block it farther.
Moving forward though:
- When we stack PRs and we are not using any tools that helps keep each PR in the stack reflecting its own change, then we start merge from the bottom to make sure the rest of the PRs don't end up looking like they changed the entire world.
- Let's also avoid a situation where approved work hangs for too long, as this can cause two issues:
  - Might need to rebase on main that has conflicting changes, then resolving the conflicts with a stale PR can introduce unwanted behavior and there would be either no review, or a need to a second review that could have been avoided
  - It is better to keep the list of PRs tidy. If there is something there then it is either a draft, or something need an action.

Ahmed-Ali (approved):
Please also double check any existing new comments that took place after putting it all together and address them before landing

Ahmed-CIO:
Had a chat with Levi about this one. Given this was something took place a while ago and was already reviewed, I won't block it farther.
Moving forward though:
- When we stack PRs and we are not using any tools that helps keep each PR in the stack reflecting its own change, then we start merge from the bottom to make sure the rest of the PRs don't end up looking like they changed the entire world.
- Let's also avoid a situation where approved work hangs for too long, as this can cause two issues:
  - Might need to rebase on main that has conflicting changes, then resolving the conflicts with a stale PR can introduce unwanted behavior and there would be either no review, or a need to a second review that could have been avoided
  - It is better to keep the list of PRs tidy. If there is something there then it is either a draft, or something need an action.

Ahmed-CIO (approved):
Please also double check any existing new comments that took place after putting it all together and address them before landing

P.s looks like I did a review with my personal account, which definitely did not do much other than a comment :D

PR #406, opened 2023-11-15
refactor: parsed push payload works for simple push as well as rich

levibostian:
Merges into: https://github.com/customerio/customerio-ios/pull/403

Small refactor that takes advantage of new `deliveryId` and `deviceToken` properties added to `CustomerIOParsedPushPayload` object. 

This refactor makes `CustomerIOParsedPushPayload` more helpful by having it be able to parse simple and rich push. Previously, it only parsed rich so push metrics reporting for simple pushes was a more manual process.


Sources/MessagingPush/CustomerIOParsedPushPayload.swift
```diff
@@ -29,19 +29,27 @@ public class CustomerIOParsedPushPayload {
...
         get {
-            cio.push.link?.url
+            cio?.push.link?.url
         }
         set {
-            cioPush = cioPush.linkSet(newValue?.absoluteString)
+            if cioPush == nil {
+                cio = CioRichPushPayload(push: CioRichPushPayload.Push(link: newValue?.absoluteString, image: nil))
+            } else {
+                cioPush = cioPush?.linkSet(newValue?.absoluteString)
```
Shahroz16, line 38:
```
private func updateCioPush(link: String?, image: String?) {
    if let cioPush = cioPush {
        if let link = link {
            self.cioPush = cioPush.linkSet(link)
        }
        if let image = image {
            self.cioPush = cioPush.imageSet(image)
        }
    } else {
        cio = CioRichPushPayload(push: CioRichPushPayload.Push(link: link, image: image))
    }
}

```

does combining logic from both methods simplify this? i find this version easier to understand but its totally subjective.

Sources/MessagingPush/CustomerIOParsedPushPayload.swift
```diff
@@ -29,19 +29,27 @@ public class CustomerIOParsedPushPayload {
...
         get {
-            cio.push.image?.url
+            cio?.push.image?.url
         }
         set {
-            cioPush = cioPush.imageSet(newValue?.absoluteString)
+            if cioPush == nil {
+                cio = CioRichPushPayload(push: CioRichPushPayload.Push(link: nil, image: newValue?.absoluteString))
+            } else {
+                cioPush = cioPush?.imageSet(newValue?.absoluteString)
```
Shahroz16, line 51:
```suggestion
        updateCioPush(link: nil, image: newValue?.absoluteString)
```

Shahroz16, line 51: can we add test for this? as codecov pointed out and i couldnt find this case in the test suite either

Shahroz16 (approved):

ami-aman (approved):

Sources/MessagingPush/CustomerIOParsedPushPayload.swift
mrehan27: I like the suggestion, easier to understand for me as well

Sources/MessagingPush/CustomerIOParsedPushPayload.swift
levibostian: Tests added. Thanks for the callout.

Sources/MessagingPush/CustomerIOParsedPushPayload.swift
levibostian: Thanks for the suggestion. 

I was able to reduce the complexity of the getter/setter to not need the original logic or suggested logic.

PR #407, opened 2023-11-15
feat: show a push while app in foreground

levibostian:
Merges into: https://github.com/customerio/customerio-ios/pull/403

Because the SDK will now be automatically handling when a push from CIO is clicked on, the SDK is also responsible for determining if a push is shown while app is in foreground or not. 

Previously, [customers specified this in their host app themselves](https://github.com/customerio/customerio-ios/blob/a2ccebf1dae6d891c8c699005a083a4147241130/Apps/CocoaPods-FCM/src/AppDelegate.swift#L56-L64). Now that customers will no longer need these code in their AppDelegate file, then customers can configure this behavior via SDK config. 

# todo
- add config option to docs

### DIFF
--- a/Sources/MessagingPush/MessagingPush.swift
+++ b/Sources/MessagingPush/MessagingPush.swift
@@ -10,6 +10,9 @@ public class MessagingPush: ModuleTopLevelObject<MessagingPushInstance>, Messagi
     @Atomic public private(set) static var shared = MessagingPush()
     private var globalDataStore: GlobalDataStore
 
+    // singleton instance of module configuration
+    @Atomic public static var moduleConfig: MessagingPushConfigOptions = .init()
+
     // testing constructor
     init(implementation: MessagingPushInstance?, globalDataStore: GlobalDataStore, sdkInitializedUtil: SdkInitializedUtil) {
         self.globalDataStore = globalDataStore
@@ -29,7 +32,11 @@ public class MessagingPush: ModuleTopLevelObject<MessagingPushInstance>, Messagi
 
     // Gets called when MessagingPushAPN.initialize() or MessagingPushFCM.initialize() called.
     @available(iOSApplicationExtension, unavailable)
-    public static func initialize() {
+    public static func initialize(config: MessagingPushConfigOptions? = nil) {
+        if let newConfig = config {
+            moduleConfig = newConfig
+        }
+
         MessagingPush.shared.initializeModuleIfSdkInitialized()
     }
 
--- a/Sources/MessagingPush/MessagingPushConfigOptions.swift
+++ b/Sources/MessagingPush/MessagingPushConfigOptions.swift
@@ -1,16 +1,9 @@
-public struct MessagingPushConfigOptions {
-    // Used to create new instance of MessagingPushConfigOptions when the MessagingPushAPN module is configured.
-    // Each property of the MessagingPushConfigOptions object can be modified by the user.
-    public enum Factory {
-        public static func create() -> MessagingPushConfigOptions {
-            MessagingPushConfigOptions(
-                autoFetchDeviceToken: true
-            )
-        }
-    }
+import CioInternalCommon
 
-    public enum Keys: String { // Constants used to map each of the options in MessagingPushConfigOptions
-        case autoFetchDeviceToken
+public struct MessagingPushConfigOptions {
+    public init() {
+        self.autoFetchDeviceToken = true
+        self.showPushAppInForeground = true
     }
 
     /**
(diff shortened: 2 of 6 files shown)


Sources/MessagingPush/MessagingPushConfigOptions.swift
```diff
@@ -1,21 +1,33 @@
...
     /**
      Enable automatic fetching of device token by the SDK without the need to write custom code by the customer.
      On fetching the token, SDK will auto-register the device. This value is `true` by default.
      */
     public var autoFetchDeviceToken: Bool
+
+    /**
+     Display push notifications sent by Customer.io while app is in foreground. This value is `true` by default.
+     */
+    public var showPushAppInForeground: Bool
```
levibostian, line 18: Added this SDK configuration to the messaging module config instead of SdkConfig since we are planning on encapsulating config options into modules.

Sources/MessagingPush/MessagingPush.swift
```diff
@@ -10,6 +10,9 @@ public class MessagingPush: ModuleTopLevelObject<MessagingPushInstance>, Messagi
     @Atomic public private(set) static var shared = MessagingPush()
     private var globalDataStore: GlobalDataStore
 
+    // singleton instance of module configuration
+    @Atomic public static var moduleConfig: MessagingPushConfigOptions = .init()
```
levibostian, line 14: This is meant to be an in-memory store for the MessagingPush config options. The DiGraph references this value to provide to dependencies when they need it.

Shahroz16 (approved):

Sources/MessagingPush/PushClickHandler.swift
```diff
@@ -170,6 +180,25 @@ extension PushClickHandlerImpl {
...
+                completionHandler([.badge, .sound])
+            }
+        }
+
+        // continue swizzle
+        cio_swizzle_willPresent(center, willPresent: notification, withCompletionHandler: completionHandler)
+    }
 }
 
 // TODO: cleanup duplicate swizzle functions all over codebase.
```
ami-aman, line 204: Is this still required ?

Sources/MessagingPush/PushClickHandler.swift
```diff
@@ -170,6 +180,25 @@ extension PushClickHandlerImpl {
         // continue swizzle
         cio_swizzle_didReceive(center, didReceive: response, withCompletionHandler: completionHandler)
     }
+
+    // Swizzled method that gets called when a push notification gets clicked or swiped away
```
ami-aman, line 184: This is a swizzled method for `willPresentNotification` that basically means that this method will be called only if the application is in the foreground and is used to present the push with a sound, badge, alert etc. Reading the comment confuses me as how does this method gets called when a push notification gets clicked or swiped away - interaction??? 🤔

Sources/MessagingPush/PushClickHandler.swift
```diff
@@ -170,6 +180,25 @@ extension PushClickHandlerImpl {
...
+        if messagingPushConfig.showPushAppInForeground {
+            if #available(iOS 14.0, *) {
+                completionHandler([.list, .banner, .badge, .sound])
+            } else {
+                completionHandler([.badge, .sound])
+            }
+        }
+
+        // continue swizzle
+        cio_swizzle_willPresent(center, willPresent: notification, withCompletionHandler: completionHandler)
```
ami-aman, line 200: Does this method call itself? I do not see any other occurrence of this method in the changed files. Am I missing out something ?

levibostian: See https://github.com/customerio/customerio-ios/pull/403#discussion_r1394524344

Sources/MessagingPush/PushClickHandler.swift
levibostian:
```suggestion
    // Swizzled method that gets called before the OS displays the push. Used to determine if a push gets displayed while app is in foreground or not. 
```

Sources/MessagingPush/PushClickHandler.swift
levibostian: See https://github.com/customerio/customerio-ios/pull/403#discussion_r1394484635

ami-aman (approved):

Sources/MessagingPush/MessagingPushConfigOptions.swift
```diff
@@ -1,21 +1,33 @@
...
+}
+
+// Add MessagingPush config options to the DIGraph like we do for SdkConfig.
+// Allows dependencies to easily access module configuration via dependency injection
+// in constructor.
+extension DIGraph {
+    var messagingPushConfigOptions: MessagingPushConfigOptions {
```
mrehan27, line 25: In Android, we have an option like `moduleConfigOptions` which is scoped to module. So classes within in-app module can use the same name for its config, while push can access its config with same name at same time. Can we apply the same here?

levibostian: Unfortunately, the naming of the variable must match the name of the protocol, in camelCase. That's the naming convention that the DI graph code generator expects. 

However, this name is only for *including* the dependency in the graph. When dependencies need to use this dependency, the name can be whatever you want. 

Example: 
```swift
class Foo {
  let moduleConfigOptions: MessagingPushConfigOptions

  init(whatever: MessagingPushConfigOptions} {
    self.moduleConfigOptions = whatever
  }
```

So, for the majority of the code base, yes, you can use `moduleConfigOptions` naming.

mrehan27: Makes sense. Thanks for the clarification 👍🏻

PR #408, opened 2023-11-20
chore: extract classes needed to be shared

Shahroz16:
Third draft PR for [[CDP] iOS Refactor: Base module#11645](https://github.com/customerio/issues/issues/11645)

### Problem statement:
Current `DIGraph` is bound by `sdkConfig` which is bound by SDK initialization
Push and in-app should be able to access classes without being dependent on SDK being initialized
Push and in-app should be able to access classes that are not bound by the tracking module initialing the SDK
Removed the dependency on `SdkInitializedUtil` We are holding a strong reference to digraph just for this purpose
```
    // The 1 place that DiGraph is strongly stored in memory for the SDK.
    // Exposed for `SdkInitializedUtil`. Not recommended to use this property directly.
    var diGraph: DIGraph?
```
### Benefits:
Breaking sdkConfig to not be sandboxed and to be to utilize digraph would be a scope creep and a lot bigger change and this separate class removes that.
We have now added a concept of `scoped` graphs, previously everything was bound to sdkConfig scope hence the issue. Now we have a digraph that is not bound to anything, and we have a graph bound to sdkConfig  and a template to add more graphs if needed for each module for the classes in that module
`Events/Journeys` needs to be shared instance among all modules to make sure multiple instances aren't created and events are not missed.
This approach makes sure that even if end up deciding we are going to keep tracking code (separate ticket), digraph or any of the older implementations doesn't break

We aren't modifying the `DiGraph` but rather adding another one, independent of sdkConfig that push and in-app are utilizing.


levibostian (changes requested):
I have some questions regarding this PR. I'll ask in Slack.

mrehan27 (approved):
Looks good to me. I can see this has been [resolved in Slack](https://customerio.slack.com/archives/C05UY5J4VLK/p1700589181633699?thread_ts=1700504463.533699&cid=C05UY5J4VLK), so looks like we are good merging this.

PR #409, opened 2023-11-20
chore: migrate classes to common

Shahroz16:
First draft PR for [[CDP] iOS Refactor: Base module#11645](https://github.com/customerio/issues/issues/11645)

What it does is move `CustomerIOInstance` to `common`. 

[PR](https://github.com/customerio/customerio-ios/pull/411) to follow next is what moves the `CustomerIO` object to `common`.


Tests/Tracking/CustomerIOMockTest.swift
```diff
@@ -1,3 +1,4 @@
+@testable import CioInternalCommon
```
levibostian, line 1:
```suggestion
```

I do understand why this import was added. Because `CustomerIOInstance` was moved into a different module, some parts of the codebase as well as customer's apps will be expected to find `CustomerIOInstance` and `CustomerIOInstanceMock` in Track module but will not be able to anymore. 

It's not ideal for a customer to need to have to import internal modules into their code base so to get around that, we have [these public aliases](https://github.com/customerio/customerio-ios/blob/a2ccebf1dae6d891c8c699005a083a4147241130/Sources/Tracking/Type/Aliases.swift). I anticipate you will need to add `CustomerIOInstance` and `CustomerIOInstanceMock` to that file, too.

Sources/Common/CustomerIOInstance.swift
```diff
@@ -0,0 +1,66 @@
...
+    var deviceAttributes: [String: Any] { get set }
+
+    func registerDeviceToken(_ deviceToken: String)
+
+    func deleteDeviceToken()
+
+    func trackMetric(
+        deliveryID: String,
+        event: Metric,
+        deviceToken: String
```
levibostian, line 64: 🤔 I had an idea? 

The reason for moving this interface into the Common module is for the CDP module to be able to add new public functions this interface (and I think CustomerIO class?). 

These functions highlighted here I think are functions that are directly tied to the Track module, right? 

Would it be worth having the `CustomerIOInstance` in the Common module be almost empty and then have the Track module add all of these functions via Track module?

Tests/Tracking/CustomerIOMockTest.swift
Shahroz16: > some parts of the codebase as well as customer's apps will be expected to find CustomerIOInstance and CustomerIOInstanceMock in the Track module but will not be able to anymore.

I tried to cover it in this [comment](https://github.com/customerio/issues/issues/11645#issuecomment-1810153108) 

Why would a customer be using `CustomerIOInstance`? 
- We don't use it in our sample apps
- It's not part of our documentation
- None of the sample snippets ever shared with us include `CustomerIOInstance`
- It's not a general practice to use interface interface `CustomerIOInstance` instead of `CustomerIO` while adding an implementation. 

The only usage that we currently have is testing as well, so that makes me think, that maybe we are worrying about something customers aren't even possibly using. What are your thoughts on it?

Sources/Common/CustomerIOInstance.swift
Shahroz16: That's a very good point, this interface does have indeed things that are `tracking` module only and they would indeed need to be moved into implementation by `tracking` modules. 

I'll think a bit more about this to make sure, we are not adding redundancy while also adding the `data pipeline` specific extension. But i do see alot of value in keeping it bare minimum and letting track then add more specific implementation.

PR #410, opened 2023-11-21
chore: module-wise configuration support

mrehan27:
helps: https://github.com/customerio/issues/issues/11648

### Changes

- Decoupled `ModuleTopLevelObject` from `SdkConfig`
- Moved `ModuleTopLevelObject` to `Common` in the process of decoupling feature modules from `Tracking`
- Removed dependencies on `SdkInitializedUtil` from modules
- Added `ModuleTopLevelConfig` to be used as a placeholder for config classes
- Added an empty `MessagingInAppConfigOptions` for the in-app module
- Commented code for for creating module instances, to be filled in by later PRs


Sources/Common/Module/ModuleTopLevelConfigOptions.swift
```diff
@@ -0,0 +1,2 @@
+public protocol ModuleTopLevelConfig {
```
levibostian, line 1: I currently don't see a use for this protocol in this PR. Perhaps it goes to a different PR?

Sources/MessagingInApp/MessagingInApp.swift
```diff
@@ -21,8 +21,8 @@ public class MessagingInApp: ModuleTopLevelObject<MessagingInAppInstance>, Messa
     // constructor that is called by test classes
     // This function's job is to populate the `shared` property with
     // overrides such as DI graph.
-    override init(implementation: MessagingInAppInstance?, sdkInitializedUtil: SdkInitializedUtil) {
-        super.init(implementation: implementation, sdkInitializedUtil: sdkInitializedUtil)
+    override init(implementation: MessagingInAppInstance?) {
```
levibostian, line 24: If I remember correctly, `ModuleTopLevelObject` was created to initialize modules in a way that worked with SDK initialization. By making push and in-app no longer tied to SDK initialization, I think that we can make `MessagingPush` and `MessagingInApp` no longer extend `ModuleTopLevelObject` and possibly even remove `ModuleTopLevelObject` entirely?

levibostian (approved):

Sources/Common/Module/ModuleTopLevelConfigOptions.swift
mrehan27: I was actually using this in upcoming PRs, but I no longer find this helpful, so removed this for now.

Sources/MessagingInApp/MessagingInApp.swift
mrehan27: Yes, we might be able to do that. But I'm a bit hesitant to remove this entirely at this point. I want to hold it till the feature modules are actually fully independent, we can then reevaluate this and remove if no longer needed. I have added a TODO so we don't forget this.

PR #411, opened 2023-11-21
chore: migrated customerio to common

Shahroz16:
Second draft PR for [[CDP] iOS Refactor: Base module#11645](https://github.com/customerio/issues/issues/11645)

What it does is it moves the `CustomerIO` object to `common`.


Apps/APN-UIKit/NotificationServiceExtension/NotificationService.swift
```diff
@@ -7,9 +7,9 @@ class NotificationService: UNNotificationServiceExtension {
     var bestAttemptContent: UNMutableNotificationContent?
 
     override func didReceive(_ request: UNNotificationRequest, withContentHandler contentHandler: @escaping (UNNotificationContent) -> Void) {
-        CustomerIO.initialize(siteId: BuildEnvironment.CustomerIO.siteId, apiKey: BuildEnvironment.CustomerIO.apiKey, region: Region.US) { config in
-            config.autoTrackPushEvents = true
-        }
+//        CustomerIO.initialize(siteId: BuildEnvironment.CustomerIO.siteId, apiKey: BuildEnvironment.CustomerIO.apiKey, region: Region.US) { config in
```
Shahroz16, line 10: This line is resulting in ` Ambiguous use of 'autoTrackPushEvents'` that i need to resolve

Sources/Common/Util/CleanupRepository.swift
```diff
@@ -0,0 +1,5 @@
+import Foundation
+
+public protocol CleanupRepository: AutoMockable {
```
Shahroz16, line 3: had to be shifted to `common` so we can hold a strong reference in CustomerIO class which is also is `common`. Implementation lives in tracking.

Sources/Tracking/CustomerIO.swift
```diff
@@ -1,4 +1,4 @@
-import CioInternalCommon
+@_exported import CioInternalCommon
```
Shahroz16, line 1: This is the magic line that helps us not break customers methods, they wouldn't need to specifically add import for common for the methods they were using previously.

Apps/APN-UIKit/NotificationServiceExtension/NotificationService.swift
levibostian: I could be wrong, but I think this error is happening because the Swift compiler is seeing finding multiple `CustomerIO.initialize` functions that it could call. The compiler is not sure what one to choose, so it gives this error.

Sources/Tracking/CustomerIO.swift
levibostian: I have not seen `@_exported` before so I looked it up. 

From my research thus far, it appears that this is a private/internal Swift feature not meant to be used in an iOS app. 

Sometimes this isn't a problem, sometimes it can result in people's apps getting rejected by Apple when they find non-public code in an app. 

Because our codebase does have an alternative solution that should fix this same problem, I strongly suggest that we remove this line and instead add a new line to [this file](https://github.com/customerio/customerio-ios/blob/main/Sources/Tracking/Type/Aliases.swift): 

```swift
public typealias CustomerIO = CioInternalCommon.CustomerIO
```

that way we can avoid a possible scenario where customer's apps get rejected because of our SDK.

Sources/Tracking/CustomerIO.swift
```diff
@@ -21,339 +21,55 @@ public extension CustomerIOInstance {
...
-
-    /**
-     Initialize the shared `instance` of `CustomerIO`.
-     Call this function when your app launches, before using `CustomerIO.instance`.
-     */
-    @available(iOSApplicationExtension, unavailable)
-    public static func initialize(
+public extension CustomerIO {
+    // Initialize for general usage
+    static func initialize(
```
levibostian, line 26: I think this function needs `@available(iOSApplicationExtension, unavailable)` added to the top. 

That may fix [this](https://github.com/customerio/customerio-ios/pull/411/files#r1400703149).

Sources/Tracking/CustomerIOImplementation+ScreenViews.swift
```diff
@@ -59,7 +59,7 @@ extension CustomerIO {
         }
 
         let addionalScreenViewData = CustomerIOImplementation.autoScreenViewBody?() ?? [:]
-        automaticScreenView(name: name, data: addionalScreenViewData)
+        screen(name: name, data: addionalScreenViewData)
```
levibostian, line 62: This change might break this feature? `automaticScreenView` has a different implementation then the `screen` function.

Sources/Tracking/CustomerIO.swift
Shahroz16: Absolutely, not going to advocate for the usage of it. Even though App Store guidelines do not advise the usage against it and but its meant to be used with swift and not external apps. This looks like a neat solution that is meant for this purpose. 

I can definitely try 
```
public typealias CustomerIO = CioInternalCommon.CustomerIO
```

Apps/APN-UIKit/NotificationServiceExtension/NotificationService.swift
Shahroz16: @levibostian yup, before I dig deep can you think of any reason why this might be happening?

Sources/Tracking/CustomerIOImplementation+ScreenViews.swift
Shahroz16: @levibostian from the code it seems like it is calling the `screen` from implementation, can you point to the code chunk which it is pointing but has a different implementation than `screen`?

Shahroz16:
```
 func automaticScreenView(
        name: String,
        data: [String: Any]
    ) {
        guard let logger = diGraph?.logger else {
            return
        }
        automaticScreenView(name: name, data: StringAnyEncodable(logger: logger, data))
    }

    // Designed to be called from swizzled methods for automatic screen tracking.
    // Because swizzled functions are not able to determine what siteId instance of
    // the SDK the app is using, we simply call `screen()` on all siteIds of the SDK
    // and if automatic screen view tracking is not setup for that siteId, the function
    // call to the instance will simply be ignored.
    func automaticScreenView<RequestBody: Encodable>(
        name: String,
        data: RequestBody
    ) {
        implementation?.screen(name: name, data: data)
    }
```

levibostian: ^^^ Yes, this is the code I am referring to. 

My apologies. I read the comment above the `automaticScreenView` function but skipped to read the function body. I was not aware that the function body changed.

Issue #178, opened 2022-07-15
Won't compile when installed via CocoaPods and using Notification Extension.


ddanielczyk:
Hi,

When I add a Notification Service Extension to handle Rich Pushes, I need to include there `CioMessagingPushAPN` pod.

The problem is, it won't compile, since there is a code in SDK that is not available to App Extensions in general.

As an example, `UIApplication.shared` is not available in extensions. Probably these methods that are not needed from within extension, should be marked with `@available(iOSApplicationExtension, unavailable)`.

![Zrzut ekranu 2022-07-15 o 10 10 52](https://user-images.githubusercontent.com/939650/179182260-d8af7123-fbbf-47ce-9145-560337c940f4.png)

When SDK is installed via SPM - it works fine.

Could you please check and fix this, so that it's possible to use the SDK via CocoaPods ? Thank you!

levibostian:
Hey there, @ddanielczyk. Thanks for asking this question. 

I have recently experienced this myself actually. I have a couple of solutions for you. 

1. I have created a branch with a potential bug fix to address this problem. Branch name is `levi/compile-notification-service`. If you could, install that branch from your `Podfile` and see if that fixes the compilation issue you're having. I have not been able to test this branch very well yet so having you try it would be very helpful. 

I don't recommend using the branch in production as we have not QA tested all the commits inside of that branch yet. So, I do have this recommendation as a workaround to get your app moving closer to production:
2. At the end of your `Podfile`, insert this code:
```ruby
cio_target_names = [
  'CustomerIOTracking',
  'CustomerIOMessagingPushAPN',
  'CustomerIOMessagingPush'
]

post_install do |installer|
  installer.pods_project.targets.each do |target|
    if cio_target_names.include? target.name
      puts "Modifying target #{target.name}"

      target.build_configurations.each do |config|
        puts "Setting build config settings for #{target.name}"
        config.build_settings['APPLICATION_EXTENSION_API_ONLY'] ||= 'NO'
      end
    end
  end
end
```

This code's job is to disable this compiler warning that's being done for the application extension just for the Customer.IO SDK targets.

ddanielczyk:
Hi @levibostian , Thank you.

So solution 1 does not seem to work, is this maybe that .podspec file on this branch is not up to date? Possibly it's missing the Common module. I couldn't get it to compile.

Solution 2 works, as a temporary solution that's fine, but xCode now shows warnings that it's not safe, and I wouldn't go to production with it.

[...]

PR #414, opened 2023-11-22
fix: cocoapods APN customers app not compiling

levibostian:
Fixes: https://github.com/customerio/customerio-ios/issues/178#issuecomment-1821464983

The CI server was not able to catch this bug for us because this bug is specific to APN and cocoapods. Our CI server is only testing cocoapods and FCM. 

This PR modifies CI tests to now test for APN compilation as well to prevent bugs like this in future.

### DIFF
--- a/Sources/MessagingPushAPN/MessagingPushAPN+PushConfigs.swift
+++ b/Sources/MessagingPushAPN/MessagingPushAPN+PushConfigs.swift
@@ -2,6 +2,7 @@ import Foundation
 import UIKit
 
 extension MessagingPushAPN {
+    @available(iOSApplicationExtension, unavailable)
     func setupAutoFetchDeviceToken() {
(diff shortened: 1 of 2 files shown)


Sources/MessagingPushAPN/MessagingPushAPN+PushConfigs.swift
```diff
@@ -2,13 +2,15 @@ import Foundation
 import UIKit
 
 extension MessagingPushAPN {
+    @available(iOSApplicationExtension, unavailable)
```
levibostian, line 5: The additions to this file match what the [FCM equivalent file](https://github.com/customerio/customerio-ios/blob/a2ccebf1dae6d891c8c699005a083a4147241130/Sources/MessagingPushFCM/MessagingPushFCM%2BPushConfigs.swift) contains.

PR #415, opened 2023-11-27
chore: separate data pipeline module

Shahroz16:
PR for [[CDP] iOS Native Refactor: Creation of Data-pipeline module#11646](https://github.com/customerio/issues/issues/11646)

Goal:
Have a separate module for Segment implementation Data Pipeline that we can utilize from `Tracking` and extension methods on the `CustomerIO` class, that customers can call.


Package.swift
```diff
@@ -41,7 +42,8 @@ let package = Package(
         // https://web.archive.org/web/20220525200227/https://www.timc.dev/posts/understanding-swift-packages/
         //
         // Update to exact version until wrapper SDKs become part of testing pipeline.
-        .package(name: "Firebase", url: "https://github.com/firebase/firebase-ios-sdk.git", "8.7.0"..<"11.0.0")
+        .package(name: "Firebase", url: "https://github.com/firebase/firebase-ios-sdk.git", "8.7.0"..<"11.0.0"),
+        .package(name: "Segment", url: "https://github.com/segmentio/analytics-swift.git", .branch("main"))
```
Shahroz16, line 46: This will be changed to the forked version, rest should be the same

Sources/DataPipeline/CIODataPipelineConfig.swift
```diff
@@ -0,0 +1,12 @@
+import CioInternalCommon
+import Foundation
+import Segment
+
+extension Configuration {
+    static var defaultConfiguration: Configuration {
```
Shahroz16, line 6: This is dependent on the `DIGraph` work, so we can get either the `write key` or `site id or API key` to have the config initialized as default.

Sources/DataPipeline/CIODataPipeline.swift
```diff
@@ -0,0 +1,167 @@
+import CioInternalCommon
+import Foundation
+import Segment
+
+public enum CIODataPipeline {
+    private static var analytics: Analytics = .init(configuration: Configuration.defaultConfiguration)
```
Shahroz16, line 7: `analytics` object can't be optional, as `Segment` always expect it be non nullable. `CIODataPipeline` will be initialized from `CustomerIOImplementation` on SDK launch so there shouldn't be a case where this would be nullable.

Package.swift
```diff
@@ -16,6 +16,7 @@ import Foundation
 // Therefore, it's important that we only expose modules that we want customers to use. Internal modules should not be included in this array.
 var products: [PackageDescription.Product] = [
     .library(name: "Tracking", targets: ["CioTracking"]),
+    .library(name: "DataPipeline", targets: ["CioDataPipeline"]),
```
levibostian, line 19:
```suggestion
    .library(name: "DataPipeline", targets: ["CioDataPipelines"]),
```

Somewhat nit-picky, but important to me since it's customer facing. 

Since [we call the product "Data Pipelines"](https://customer.io/data-pipelines/) I think it makes sense for the public SDK module name to match that naming.

Package.swift
```diff
@@ -16,6 +16,7 @@ import Foundation
 // Therefore, it's important that we only expose modules that we want customers to use. Internal modules should not be included in this array.
```
levibostian, line 16: Friendly reminder to also update cocoapods files. 

You will probably need to create a new `podspec` file, update `CustomerIO.podspec` file, and also update the CI cocoapods deployments script to deploy the new `podspec` file.

Sources/DataPipeline/CIODataPipeline.swift
```diff
@@ -0,0 +1,173 @@
...
+
+public class CIODataPipeline {
+    // Private static instance of Analytics, created only once.
+    private static var analytics: Analytics = .init(configuration: Configuration.defaultConfiguration)
+
+    // Private initializer to enforce singleton usage, prevents external instantiation.
+    private init() {}
+
+    // Static method to access the singleton Analytics instance.
+    public static func shared() -> Analytics {
```
levibostian, line 13: I was expecting `CIODataPipeline.shared` to return an instance of `CIODataPipeline` but this is returning the Segment SDK instance. 

Do we need to publicly expose the Segment SDK? If so, is `shared` the right name for exposing it?

Sources/DataPipeline/CIODataPipeline.swift
```diff
@@ -0,0 +1,173 @@
...
+
+    // Method to reinitialize Analytics with a new configuration.
+    public static func initialize(configuration: Configuration) {
+        analytics = Analytics(configuration: configuration)
+    }
+}
+
+// MARK: - System Modifiers
+
+public extension CustomerIO {
```
levibostian, line 25: I am assuming that all of these extensions are added for Segment drop-in requirement? 

If so, I think it would be good to move this `extension` block into another file to keep the drop-in code encapsulated somewhere.

One idea is using a common Swift pattern of using the file name to indicate it's purpose. Perhaps creating a new file `CustomerIO+Segment.swift` and putting this block of code inside could be helpful?

Sources/DataPipeline/CIODataPipeline.swift
```diff
@@ -0,0 +1,173 @@
...
+
+    /// Resets this instance of Analytics to a clean slate.  Traits, UserID's, anonymousId, etc are all cleared or reset.  This
+    /// command will also be sent to each plugin present in the system.
+    func reset() {
+        CIODataPipeline.shared().reset()
+    }
+
+    /// Retrieve the version of this library in use.
+    /// - Returns: A string representing the version in "BREAKING.FEATURE.FIX" format.
+    func version() -> String {
```
levibostian, line 86: Do we want to expose the Segment SDK version? 

It feels weird to me for a customer to call `CustomerIO.version` and `CIODataPipeline.version` and get 2 different returned results? Also feels weird to expose the version of a dependency that our SDK has?

Sources/DataPipeline/CIODataPipelineConfig.swift
levibostian: Good call-out. 

Could you add a `TODO` or `FIXME` comment on this line stating that this file is going to change after some other PR is created?

Sources/DataPipeline/CIODataPipelineEvents.swift
```diff
@@ -0,0 +1,116 @@
+import CioInternalCommon
+import Foundation
+
+// MARK: - Typed Event Signatures
+
+public extension CustomerIO {
+    // make a note in the docs on this that we removed the old "options" property
+    // and they need to write a middleware/enrichment now.
+    // the objc version should accomodate them if it's really needed.
```
levibostian, line 9:
```suggestion
    // TODO: make a note in the docs on this that we removed the old "options" property
    // and they need to write a middleware/enrichment now.
    // the objc version should accomodate them if it's really needed.
```

Is this comment still valid? If so, I suggest adding `TODO` to make it easier to find and fix in a future PR.

Sources/DataPipeline/CIODataPipelineEvents.swift
```diff
@@ -0,0 +1,116 @@
+import CioInternalCommon
+import Foundation
+
+// MARK: - Typed Event Signatures
+
+public extension CustomerIO {
```
levibostian, line 6: I think it would be a good idea to create an [APITest](https://github.com/customerio/customerio-ios/blob/main/Tests/Tracking/APITest.swift) for this new module. It has been really helpful to make sure that the public-facing API compiles for customers as intended. 

If you agree, could you make sure this is added as a task to a CDP ticket to do this? It's OK if this is not part of this PR if that's out of scope.

levibostian (approved):

Package.swift
Shahroz16: Good call-out.

Sources/DataPipeline/CIODataPipeline.swift
Shahroz16: if we are to use the instance of `Analytics` in extension method of `CustomerIO` we need to have it as a static member of `CIODataPipeline`. 

Open to naming it whatever makes the most sense, since this is not going to be accessed by the users directly, shared seemed like a good candidate.

Sources/DataPipeline/CIODataPipeline.swift
Shahroz16: I just feature matched the segments APIs, we can drop/keep or make any decisions in the `ready for production` tickets, otherwise the discussion might scope creep the PRs. What do you think?

Sources/DataPipeline/CIODataPipelineEvents.swift
Shahroz16: This is the comment from Segment about this method, i copied it right as is so we can make a more informed decision.

Sources/DataPipeline/CIODataPipelineEvents.swift
Shahroz16: Agreed, I was thinking it could be added to Ready for Production tickets where we can add even more test cases related.

Sources/DataPipeline/CIODataPipeline.swift
levibostian: `shared` is a commonly used Swift name when naming singleton instances. I think that is what is throwing me off. If I saw Swift code `CIODataPipeline.shared`, I was assume that an instance of CIODataPipeline would be returned. 

Instead of having this `shared` function, could we instead re-use the static Segment variable above? 

```swift
public private(set) static var analytics: Analytics = .init(configuration: Configuration.defaultConfiguration)
```

Then, `CIODataPipeline.analytics` would do the same thing as `CIODataPipeline.shared` does currently.

Sources/DataPipeline/CIODataPipeline.swift
levibostian: I am OK with deciding later. 

Idea: We make a decision in project slack channel? This seems like a short and quick decision to me. Could be answered async together.

Sources/DataPipeline/CIODataPipelineEvents.swift
levibostian: I am OK with that. 

If that's the case, could you add a // TODO or // FIXME comment to this file specifying that we are planning on that?

PR #416, opened 2023-11-28
refactor: feature modules initialization

mrehan27:
helps: https://github.com/customerio/issues/issues/11650

### Changes

- Updated `CocoaPods-FCM` app to use new initialization for `MessagingInApp`
- Refactored `ModuleTopLevelObject` to support optional initialization for modules requiring user input
- Added more classes to `DIGraphShared` to decouple modules from `SdkConfig`
- Updated auto generated files
- Updated `MessagingInAppImplementation`
- Updated `MessagingPushImplementation`
- Minor fixes and improvements


Sources/Common/Module/ModuleTopLevelObject.swift
```diff
@@ -7,33 +7,49 @@ import Foundation
...
-// TODO: [CDP] Remove class if no longer needed
 open class ModuleTopLevelObject<ImplementationClass> {
     private(set) var alreadyCreatedImplementation: ImplementationClass?
     public var implementation: ImplementationClass? {
-        alreadyCreatedImplementation ?? createAndSetImplementationInstance()
+        let instance = alreadyCreatedImplementation ?? createAndSetImplementationInstance()
+        if instance == nil {
+            logger.info("Module \(moduleName) is not yet initialized")
+        }
+        return instance
```
mrehan27, line 17: This ensures safe access to an already initialized instance of the module and generates a log message if any method from the module is accessed before initialization.

Sources/MessagingInApp/MessagingInApp.swift
```diff
@@ -2,32 +2,27 @@ import CioInternalCommon
...
 
 public protocol MessagingInAppInstance: AutoMockable {
-    // sourcery:Name=initialize
-    func initialize()
-    // sourcery:Name=initializeEventListener
-    func initialize(eventListener: InAppEventListener)
-
-    @available(*, deprecated, message: "Parameter organizationId no longer being used. Remove the parameter from your function call to migrate to new function.")
-    // sourcery:Name=initializeOrganizationId
-    func initialize(organizationId: String)
```
mrehan27, line 12: Removed unused and deprecated initialization methods

Sources/MessagingInApp/hooks/HookProvider.swift
```diff
@@ -9,16 +9,16 @@ class MessagingInAppModuleHookProvider: ModuleHookProvider {
     }
 
     var profileIdentifyHook: ProfileIdentifyHook? {
-        guard let diGraph = diGraph else { return nil }
+        // guard let diGraph = diGraph else { return nil }
 
         // FIXME: [CDP] Find workaround by attaching hook to Journeys or reusing existing instance to utilize customer provided siteid
-        return MessagingInAppImplementation(diGraph: diGraph, moduleConfig: MessagingInAppConfigOptions.Factory.create())
+        MessagingInApp.shared.implementation as? MessagingInAppImplementation
```
mrehan27, line 15: Temporary fixes, note is still there as reminder

Sources/MessagingPush/MessagingPushConfigOptions.swift
```diff
@@ -62,4 +62,12 @@ public struct MessagingPushConfigOptions {
...
      */
     public var autoTrackDeviceAttributes: Bool
+
+    /// Applies new configuration to this object
+    /// - Important: Ensure all mutable properties are updated accordingly
+    mutating func apply(_ newConfig: MessagingPushConfigOptions) {
+        autoFetchDeviceToken = newConfig.autoFetchDeviceToken
+        autoTrackPushEvents = newConfig.autoTrackPushEvents
+        autoTrackDeviceAttributes = newConfig.autoTrackDeviceAttributes
+    }
```
mrehan27, line 72: For safely updating config without updating reference

Sources/MessagingPush/MessagingPush.swift
```diff
@@ -7,43 +7,70 @@ import Foundation
...
+        configure configureHandler: ((inout MessagingPushConfigOptions) -> Void)? = nil
+    ) -> MessagingPushInstance {
+        var moduleConfig = MessagingPushConfigOptions.Factory.create()
+
+        if let configureHandler = configureHandler {
+            configureHandler(&moduleConfig)
+        }
+
+        shared.initializeModule(moduleConfig: moduleConfig)
+        return shared
```
mrehan27, line 46: Similar to current initialization, push module configuration can be updated using any of the following option

1. Using optional initialization

```swift
MessagingPush.initialize { config in
    config.autoTrackPushEvents = true
}
```

2. Using `configure` method

```swift
MessagingPush.shared.configure { config in
    config.autoTrackPushEvents = true
}
```

Sources/MessagingPush/MessagingPushConfigOptions.swift
levibostian: Based on recent PR conversations we have had, instead of a mutating function, could we leverage constructors again? 

Here is a code sample that is similar to what is in this file:

```swift
public static func create(from newConfig: MessagingPushConfigOptions) -> MessagingPushConfigOptions {
            return MessagingPushConfigOptions(
                autoFetchDeviceToken: newConfig.autoFetchDeviceToken
                ...
            )
}
```

levibostian (approved):

Sources/MessagingPush/MessagingPush.swift
levibostian: I like the idea of creating a separate `configure()` function to allow the SDK to change it's config at runtime instead of config being required during SDK initialization. 

I have thought about how we could add this type of feature to our SDK but realized that it could cause scope creep or risk of runtime errors. 

Here is a scenario to describe this risk. 

1. `MessagingPush.initialize` gets called and sets the config to: 
```
autoTrackPushMetrics = true
```

2. Some code in the push module get executed. So, some dependencies in the DI graph get constructed. Here is an example constructor that you would see in the iOS codebase: 

```swift
class PushClickHandler {

  let autoTrackPushMetrics: Bool

  init(moduleConfig: ModulePushConfigOptions) {
    autoTrackPushMetrics = moduleConfig.autoTrackPushMetrics
  }
}
```

3. `MessagingPush.shared.configure` gets called and changes `autoTrackPushMetrics = false`. 

4. The existing `PushClickHandler` instance could still be in-memory and it contains the old config `autoTrackPushMetrics` value. Putting the SDK into a state where some dependencies are out-of-date. 

---

There are ways to avoid this scenario from happening. The push module is small in size so the changes of this scenario happening are much smaller then other modules such as Track or CDP. 

Even if we do not see this problem in the push module today, this is a easy risk to fall into. I think that the way the codebase is today, it's far too easy to accidentally write code that puts the SDK into this state. Therefore, I am nervous to add this feature at this time. 

If adding this feature is required to make the SDK function, a possible workaround is to only allow each module to be configured one time. No matter what function they call, `initialize` or `config`, we set module configuration one time and ignore future requests to change the config. 

I would like to add this feature to all of our modules. I would feel most confident if it was done when we could reduce the risk by making some refactors to the codebase first.

Apps/CocoaPods-FCM/src/AppDelegate.swift
```diff
@@ -28,7 +28,9 @@ class AppDelegate: NSObject, UIApplicationDelegate {
             // This line of code is internal to Customer.io for testing purposes. Do not add this code to your app.
             appSetSettings?.configureCioSdk(config: &config)
         }
-        MessagingInApp.initialize(eventListener: self)
+        MessagingInApp
+            .initialize(siteId: siteId, region: .US)
+            .setEventListener(eventListener: self)
```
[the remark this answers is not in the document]
mrehan27: [Updated](https://github.com/customerio/customerio-ios/pull/416/commits/1962e4dde1ec39e2779e757ccac5c1166b18574c) 👍🏻

Sources/MessagingPush/MessagingPush.swift
mrehan27: Thanks for sharing the use case, Levi. I agree that even if we address all current instances where the module config is cached, there's still a high chance of it being cached and misused in the future. I've [moved this discussion to Slack](https://customerio.slack.com/archives/C05UY5J4VLK/p1701251917423449) for detailed explanation of what I had in mind when adding this. Let’s discuss further there and decide on our next action items.

Sources/MessagingPush/MessagingPushConfigOptions.swift
mrehan27: I added this for `configure` method only, which is why I avoided recreating the instance. Since we're already [discussing removing it](https://github.com/customerio/customerio-ios/pull/416#discussion_r1408136656), once we finalize that decision, I'll also remove this method with that change.

PR #417, opened 2023-11-29
chore: data pipeline usage

Shahroz16:
Helps: [[CDP] iOS Native Refactor: data pipeline module usage#11647](https://github.com/customerio/issues/issues/11647)

Links the Datapipeline module with CustomerIOImplementation, it redirects the to this module and make sure whenever SDK gets initialized this, DataPipeline module gets initialized as well.


Sources/Tracking/CustomerIOImplementation.swift
```diff
@@ -83,6 +88,8 @@ class CustomerIOImplementation: CustomerIOInstance {
         identifier: String,
         body: RequestBody
     ) {
+        CIODataPipeline.analytics.identify(userId: identifier, traits: jsonAdapter.toJson(body))
```
levibostian, line 91: Should this line occur farther down in this function body? 

For example, the next line of the function is `if identifier.isBlankOrEmpty() { return }`. I was expecting that we only forward the identify request to the CDP module after we run logic such as this?

Sources/Tracking/CustomerIOImplementation.swift
```diff
@@ -191,6 +200,8 @@ class CustomerIOImplementation: CustomerIOInstance {
         name: String,
         data: RequestBody?
     ) {
+        CIODataPipeline.analytics.track(name: name, properties: jsonAdapter.toJson(data))
```
levibostian, line 203: There are multiple `track` functions in this file with different function parameters. All of the `track` variations call 1 common function, `trackEvent`. 

Shouldn't this line go into `trackEvent`? Otherwise, if a customer called, for example, `track(name: String, data: [String: Any])` function, then that request will not get forwarded to the CDP module. 

If the function `trackEvent` is confusing, feel free to make refactors you need. The goal of the `trackEvent` function is to be 1 function where all event and screen tracking logic exists. You can rename to `commonTrack` or something if that makes more sense to you.

Sources/Tracking/CustomerIOImplementation.swift
```diff
@@ -206,6 +217,8 @@ class CustomerIOImplementation: CustomerIOInstance {
         name: String,
         data: RequestBody
     ) {
+        CIODataPipeline.analytics.screen(title: name, properties: jsonAdapter.toJson(data))
```
levibostian, line 220: Same comment as above with expecting that this CDP call goes inside of the 1 common `trackEvent` function.

Sources/Tracking/CustomerIOImplementation.swift
```diff
@@ -220,12 +233,15 @@ class CustomerIOImplementation: CustomerIOInstance {
      is no active customer, this will fail to register the device
      */
     public func registerDeviceToken(_ deviceToken: String) {
+        CIODataPipeline.analytics.setDeviceToken(deviceToken)
```
levibostian, line 236: The function `addDeviceAttributes` is currently the common function used for when a device is registered with a profile and when adding attributes to a device.

Shouldn't this line go into that function instead?

levibostian (commented):
Not yet approving because of questions wondering if the CDP call is done in the correct function or in the correct place within the function body.

Sources/Tracking/CustomerIOImplementation.swift
Shahroz16:
```
public func track<RequestBody: Encodable>(
        name: String,
        data: RequestBody?
    ) {
        CIODataPipeline.analytics.track(name: name, properties: jsonAdapter.toJson(data))

        _ = trackEvent(type: .event, name: name, data: data)
    }

    public func track(name: String, data: [String: Any]) {
        track(name: name, data: StringAnyEncodable(logger: logger, data))
    }
```
The other track method seems to be calling the track method i added the CDP redirection in. I didn't add the code in the `trackEvent` because it contains logic for both screen and track, while we don't have to combine it as Segment had separate methods for both. 

And i don't see us keeping  method `track<RequestBody: Encodable>`  as it was supposed to be used internally and was helpful when we were using our BQ, if we are just redirect these calls, we don;t have use of it. Given track get removed from our SDK.

Sources/Tracking/CustomerIOImplementation.swift
Shahroz16: same comment as above.

Sources/Tracking/CustomerIOImplementation.swift
Shahroz16: `addDeviceAttributes` also takes in attributes that can be added, `setDeviceToken` doesn;t currently take in any attributes, we need to create that method ourself. I'll add a todo that this method will then move there.

Sources/Tracking/CustomerIOImplementation.swift
ami-aman: I second Levi. The following line of code (`if identifier.isBlankOrEmpty() ...`) seems to have no value if we identify the user as the first thing in the function. This could be confusing as the user could be identified anonymously and then see the error log in the console `profile cannot be identified: Identifier is empty. Please retry with a valid, non-empty identifier` in case identifier is blank or empty though a corner case but possible.

Either we re-think the implementation or we make sure if with segment do we need to check identifier for blank or empty ?

Sources/DataPipeline/CIODataPipelineConfig.swift
```diff
@@ -5,8 +5,12 @@ import Segment
...
-        let siteId = CustomerIO.shared.config?.siteId ?? ""
-        let apiKey = CustomerIO.shared.config?.apiKey ?? ""
+        configure(diGraph: CustomerIO.shared.diGraph)
+    }
+
+    static func configure(diGraph: DIGraph?) -> Configuration {
+        let siteId = diGraph?.sdkConfig.siteId ?? ""
+        let apiKey = diGraph?.sdkConfig.apiKey ?? ""
         let key = siteId + apiKey
         return Configuration(writeKey: key)
```
ami-aman, line 15: I'm wondering if there are plans to add support for `writeKey` in future tickets or pull requests. It seems like this implementation is designed for backward compatibility, but what happens if a user only provides `writeKey` and not `siteId` or `APIKey`?

Shahroz16: This is going to be replaced with `write key` the comment on the top,  

```
// TODO: Module Level configuration is going to be added here which will replace this config
```

`Module Level configuration` this module config would contain write key

Sources/Tracking/CustomerIOImplementation.swift
Shahroz16: Agreed, this line `if identifier.isBlankOrEmpty()` in question would ideally be removed with the removal of tracking code. So we might not even cater this case. I am not removing any code in these PRs to avoid confusion. 

In-case we decide to keep this check for some reason, because Segment doesn't validate for empty strings then we can put this line below.

ami-aman: Sounds good. Thanks for the clarification, Shahroz! Would you want to add a `TODO` here as well to not miss out removing this statement in the future tickets ?

Sources/DataPipeline/CIODataPipelineConfig.swift
ami-aman: Perfect, Thanks !

Sources/Tracking/CustomerIOImplementation.swift
levibostian: I am glad to hear that the CDP module call will be called by all versions of `track()`. 

I do still prefer to have all logic for tracking events inside of 1 function. For code readability and maintainability, I think it would be better to have all logic encapsulated together instead of having some code in a common function and some code not.

Shahroz16: I understand it would be easier, i can add a `todo` for that. My hesitation is, i don't wan't to unncessarily rely on the data classes we have created for our tracking and BQ because they might get removed. Simiarly the `trackEvent` method might not even exist and neither would the `track<RequestBody: Encodable>(name: String,data: RequestBody?`. 

The only method that would remain would be 

```
track(name: String, data: [String: Any])

and it would call 
        CIODataPipeline.analytics.track(name: name, properties: data)

```

Sources/Tracking/CustomerIOImplementation.swift
levibostian: I can understand that. 

I would be OK with returning early and ignoring the request as long as we also log a warning to help the customer understand they may have done something wrong.

Sources/Tracking/CustomerIOImplementation.swift
levibostian: Ah, ok. 

I would be OK with adding a todo and deciding after the fate of tracking code is determined. Good idea.

Sources/Tracking/CustomerIOImplementation.swift
```diff
@@ -165,6 +173,8 @@ class CustomerIOImplementation: CustomerIOInstance {
     }
 
     public func clearIdentify() {
+        CIODataPipeline.analytics.reset()
```
mrehan27, line 176: Shouldn't this be after log message?

mrehan27 (approved):
Thanks for the action items, looks good to me 👍🏻

PR #419, opened 2023-12-04
chore: removal of tracking specific methods

Shahroz16:
What this PR is doing is removing all the methods (public and internal) which take `RequestBody: Encodable`. 

Why?
- Segment doesn't support `Encodable` but rather `Codable` so we can't simply redirect the call from this method to analytics. 

What will the customers who were using these methods?
- We are going to ask them to update their classes from `Encodable` to `Codable`, they will just have to extend the classes from `Decodable` as `Codable = Encodable && Decodable` after doing so they can use the method exposed by the `Analytics` object directly. 

> Auto-generated files can be ignored


Sources/Common/CustomerIOInstance.swift
```diff
@@ -10,15 +10,6 @@ public protocol CustomerIOInstance: AutoMockable {
...
 
-    // sourcery:Name=identifyEncodable
-    // sourcery:DuplicateMethod=identify
-    func identify<RequestBody: Encodable>(
-        identifier: String,
-        // sourcery:Type=AnyEncodable
-        // sourcery:TypeCast="AnyEncodable(body)"
-        body: RequestBody
-    )
-
```
levibostian, line 21: (this comment can be applied to all code blocks removed in this PR, not just this one)

I don't understand the reason to remove these functions instead of keeping them and changing the function signature? 

Has this been considered as a solution? 

```swift
    // change Encodable to Codable
    func identify<RequestBody: Codable>(
```

levibostian: Update: I re-read the description a few times and what I think you're suggesting is that the CDP's extension methods that it has added will allow function signatures such as 
```swift
func identify<RequestBody: Codable>(
```

to be added. Therefore, the public API will still allow `func identify<RequestBody: Codable>(` to exist in the SDK, but you will need to import the CDP module in your Swift files to use it. 

Assuming that is all correct, I will continue to ask my original question of: has it been considered to keep all of these functions in the Common protocol? It doesn't feel right to me to have some forms of identify/track functions available in Common module while having more non-Segment specific identify/track functions available in another module.

Shahroz16: That is a very fair demand, for method that doesn't involve Segment oriented classes, customers shouldn't have to add external dependency for those. 

> keep all of these functions in the Common protocol?

But we won't want to keep *all* right? because first `common` module doesn't have any information about the `CDP` module, secondly, we won't want to bound `common` module with segment specific classes like `plugin` so we can easily deviate when needed.

Shahroz16: Another [PR](https://github.com/customerio/customerio-ios/pull/421) is made which updates the public method.

PR #422, opened 2023-12-05
refactor: introduced new method for sdk Initialization

mrehan27:
helps: https://github.com/customerio/issues/issues/11648

### Changes

- Introduced a new method for SDK initialization
- Updated APN UIKit to utilize the new initialization method
- Created `DataPipelineConfigOptions` for configuring CDP
- Added `DataPipeline` module to align with existing modules


Apps/APN-UIKit/APN UIKit/AppDelegate.swift
```diff
@@ -38,14 +38,15 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
             apiKey = storedApiKey
         }
-        CustomerIO.initialize(siteId: siteId, apiKey: apiKey, region: .US) { config in
-            config.logLevel = self.storage.isDebugModeEnabled ?? true ? .debug : .error
-            config.autoTrackDeviceAttributes = self.storage.isTrackDeviceAttrEnabled ?? true
-            config.backgroundQueueSecondsDelay = Double(self.storage.bgQDelay ?? "30") ?? 30
-            config.backgroundQueueMinNumberOfTasks = Int(self.storage.bgNumOfTasks ?? "10") ?? 10
-            config.autoTrackScreenViews = self.storage.isTrackScreenEnabled ?? true
+        let writeKey = "\(siteId):\(apiKey)"
+        CustomerIO.initialize(writeKey: writeKey) { sdkConfig, cdpConfig in
```
Shahroz16, line 42: we shouldn't be needing both `sdkConfig, cdpConfig`? old users come via sdkConfig and new ones should use cdpConfig?

Sources/DataPipeline/CustomerIO.swift
```diff
@@ -0,0 +1,31 @@
...
+    /**
+     Initialize the shared `instance` of `CustomerIO`.
+     Call this function when your app launches, before using `CustomerIO.instance`.
+     */
+    @available(iOSApplicationExtension, unavailable)
+    static func initialize(
+        writeKey: String,
+        configure configureHandler: ((inout SdkConfig, inout DataPipelineConfigOptions) -> Void)?
+    ) {
+        var sdkConfig = SdkConfig.Factory.create(siteId: "", apiKey: "", region: .US)
```
Shahroz16, line 13: we shouldnt be needing `sdkConfig` in the datapipeline?

Sources/DataPipeline/CustomerIO.swift
```diff
@@ -0,0 +1,31 @@
...
+        if let configureHandler = configureHandler {
+            configureHandler(&sdkConfig, &cdpConfig)
+        }
+
+        let newDiGraph = DIGraph(sdkConfig: sdkConfig)
+        let implementation = DataPipeline.initialize(moduleConfig: cdpConfig)
+
+        initializeSharedInstance(with: implementation, diGraph: newDiGraph, module: TrackingModuleHookProvider(), cleanupRepositoryImp: newDiGraph.cleanupRepository)
+
+        if sdkConfig.autoTrackScreenViews {
```
Shahroz16, line 25: intresting i think we should probably make autotraking screen views a plugin instead of config

Sources/DataPipeline/DataPipelineConfigOptions.swift
```diff
@@ -0,0 +1,73 @@
...
+    public var operatingMode: OperatingMode = .asynchronous
+
+    /// Tracking configurations
+    public var autoScreenViewBody: (() -> [String: Any])?
+    public var autoTrackDeviceAttributes: Bool = true
+    public var autoTrackScreenViews: Bool = true
+    #if canImport(UIKit)
+    public var filterAutoScreenViewEvents: ((UIViewController) -> Bool)?
+    #endif
+    public var trackApplicationLifecycleEvents: Bool = true
```
Shahroz16, line 64: interesting, we should look for ways to decouple these with data pipeline if possible

Sources/Common/Module/ModuleTopLevelObject.swift
```diff
@@ -8,7 +8,7 @@ import Foundation
 // There isn't a constructor populated via dependency injection. It's at the top node
 // of dependencies.
 open class ModuleTopLevelObject<ImplementationClass> {
-    private(set) var alreadyCreatedImplementation: ImplementationClass?
+    open var alreadyCreatedImplementation: ImplementationClass?
```
mrehan27, line 11: Updated only because of setters in `DataPipeline`

Sources/DataPipeline/CustomerIO+Events.swift
```diff
@@ -23,7 +23,7 @@ public extension CustomerIO {
     ///      https://segment.io/libraries/ios#ids
     /// - traits: A dictionary of traits you know about the user. Things like: email, name, plan, etc.
     func identify<T: Codable>(userId: String, traits: T?) {
-        CIODataPipeline.analytics.identify(userId: userId, traits: traits)
+        DataPipeline.shared.identify(identifier: userId, body: traits)
```
mrehan27, line 26: I just changed one call in this PR to demonstrate, future PRs will remove `CIODataPipeline` completely and all calls will look like this

Sources/DataPipeline/CustomerIO.swift
mrehan27: Yeah, added it for `DIGraph`. But will keep looking to remove the need of it in `DataPipeline`.

Sources/Tracking/CustomerIOImplementation.swift
```diff
@@ -48,6 +48,7 @@ class CustomerIOImplementation: CustomerIOInstance {
         // Initialiing the CIODataPipeline with the diGraph for now
         // TODO: initialize this module with its own config
         CIODataPipeline.initialize(diGraph: diGraph)
```
mrehan27, line 50: This will be removed in future PRs

Apps/APN-UIKit/APN UIKit/AppDelegate.swift
mrehan27: **Update:** We discussed this on call, and decided to remove `SdkConfig` for CDP users for now and update it from

```swift
static func initialize(
    writeKey: String,
    configure configureHandler: ((inout SdkConfig, inout DataPipelineConfigOptions) -> Void)?
)

CustomerIO.initialize(writeKey: writeKey) { sdkConfig, cdpConfig in
     sdkConfig.logLevel = .debug
     sdkConfig.autoTrackScreenViews = true

     cdpConfig.flushInterval = 30
     cdpConfig.flushAt = 10
}
```

to the following

```swift
static func initialize(
    writeKey: String,
    configure configureHandler: ((inout SdkConfig, inout DataPipelineConfigOptions) -> Void)?
)

CustomerIO.initialize(writeKey: writeKey, logLevel: .debug) { config in
    config.flushInterval = 30
    config.flushAt = 10
}
```

For `autoTrackScreenViews`, they should go with tracking module.

Shahroz16 (approved):
Great changes! thank you.

Sources/DataPipeline/DataPipelineConfigOptions.swift
```diff
@@ -0,0 +1,63 @@
+import CioInternalCommon
+import Foundation
+import Segment
+#if canImport(UIKit)
+import UIKit
+#endif
```
levibostian, line 6: What is this for? Is it needed?

Sources/DataPipeline/CustomerIO.swift
```diff
@@ -0,0 +1,28 @@
...
+
+public extension CustomerIO {
+    /**
+     Initialize the shared `instance` of `CustomerIO`.
+     Call this function when your app launches, before using `CustomerIO.instance`.
+     */
+    @available(iOSApplicationExtension, unavailable)
+    static func initialize(
+        writeKey: String,
+        logLevel: CioLogLevel = .error,
```
levibostian, line 11: I was not expecting log level to be a parameter but instead be set via config object? 

I think consistency of all optional parameters, such as log level, being set inside of a config object.

Sources/DataPipeline/DataPipeline.swift
```diff
@@ -0,0 +1,121 @@
...
+    @Atomic public private(set) static var moduleConfig: DataPipelineConfigOptions!
+    private static let moduleName = "DataPipeline"
+
+    private init() {
+        super.init(moduleName: Self.moduleName)
+    }
+
+    /**
+     Initialize the shared `instance` of `DataPipeline`.
+     Call this function when your app launches, before using `DataPipeline.shared`.
```
levibostian, line 14: > Call this function when your app launches, before using `DataPipeline.shared`.

This comment makes me think you're suggesting that customers should be calling `DataPipeline.initialize` instead of `CustomerIO.initialize` to initialize the SDK and use `DataPipeline.shared.identify()` instead of `CustomerIO.shared.identify()`. 

Maybe I am reading this comment wrong and it's not suggesting that customers call this function on app start but instead we need to when `CustomerIO.initialize` is called?

levibostian (approved):

Sources/DataPipeline/DataPipelineConfigOptions.swift
mrehan27: Forgot to remove it 🤦🏻 Thanks, removed

Sources/DataPipeline/CustomerIO.swift
levibostian: Thanks for that link. I didn't see that conversation. That does indeed answer my question. 

I am not the biggest fan of the public-API after that decision, but I understand why that decision was made. I would like to see more consistency in the public API of our SDK but truly making it consistent goes beyond the scope of the CDP SDK launch. So, I am OK with the public API we have here for now.

PR #427, opened 2023-12-07
chore: auto tracking screen view plugin

Shahroz16:
- Add Autotracking screen view as a plugin
- Cleanup of DataPipeline and its implementation
- Concept of dead instance to make sure we have instance of Analytics always available


levibostian (commented):
The code looks good. Only problem I have is I do not understand the reasoning for using a segment plugin to implement this solution. What problem needed to be solved by this PR? 

I suggest we add a comment to the `AutoTrackingScreenViews` plugin file explaining why it needs to be a plugin so we know for the future. 

Otherwise, I approve the code 👍

Shahroz16:
@levibostian sure thing, two reasons

- Moving towards plugin architecture, where optional things are plugins rather than classes we enable/disable 
- `autoScreenTracking` is tracking config and not available in CDP config that Datap pipeline users will only need
- For DataPipeline only users, if they wanted auto-tracking of screen it didn't make sense to ask them to add `tracking config` of autoScreenTracking and import of `tracking`

PR #430, opened 2023-12-11
feat: disable the feature, if a customer needs to

levibostian:
In a team call, we discussed the importance of customers being able to disable this feature if they need to. We do not anticipate a customer needing to do this, but because this feature does involve techniques such as swizzling, it's good practice to give customers the option to disable the functionality if they encounter issues in their app's edge case environment.

### DIFF
--- a/Sources/MessagingPush/MessagingPush.swift
+++ b/Sources/MessagingPush/MessagingPush.swift
@@ -66,7 +66,7 @@ public class MessagingPush: ModuleTopLevelObject<MessagingPushInstance>, Messagi
         let logger = diGraph.logger
         logger.debug("Setting up MessagingPush module...")
 
-        if Self.shared.moduleConfig.autoPushClickHandling {
+        if diGraph.sdkConfig.autoTrackPushEvents {
             diGraph.automaticPushClickHandling.start()
         }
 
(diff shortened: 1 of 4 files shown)


Sources/MessagingPush/MessagingPush.swift
```diff
@@ -66,7 +66,9 @@ public class MessagingPush: ModuleTopLevelObject<MessagingPushInstance>, Messagi
         let logger = diGraph.logger
         logger.debug("Setting up MessagingPush module...")
 
-        diGraph.automaticPushClickHandling.start()
+        if diGraph.sdkConfig.autoTrackPushEvents {
+            diGraph.automaticPushClickHandling.start()
+        }
```
levibostian, line 71: In a [previous commit in this PR](https://github.com/customerio/customerio-ios/pull/430/commits/f421379731d981565488c0d07b56c44a0f895346#diff-d322168af6d12b45ff1a57021d6a686396fe280ff72e00f2c1872bc9e4abff3fR69-R71) I added a new messaging push module sdk config option to enable/disable this feature. 

However, I later realized that our Sdk already comes with a config option that makes sense to re-use. I want to avoid confusing customers by having 2 config options in the SDK that have very similar names and use cases.

mrehan27 (approved):

Shahroz16 (approved):

PR #432, opened 2023-12-12
chore: data pipeline implementation using analytics

mrehan27:

### Changes

- Updated `DataPipelineImplementation` using `analytics` object
- Added `MetricEvent` event for push and in-app metrics
- Added `DeviceAttributes` plugin for adding custom device attributes

### DIFF
--- a/Sources/DataPipeline/DataPipelineImplementation.swift
+++ b/Sources/DataPipeline/DataPipelineImplementation.swift
@@ -32,6 +32,8 @@ class DataPipelineImplementation: DataPipelineInstance {
     }
 
     func clearIdentify() {
+        // TODO: [CDP] CustomerIOImplementation also call deleteDeviceToken from clearIdentify, but customers using DataPipeline only,
+        // we had to call this explicitly. Rethink on how can we make one call for both customers.
         deleteDeviceToken()
         analytics.reset()
     }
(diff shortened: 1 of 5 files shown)


ami-aman (approved):

Sources/DataPipeline/DataPipelineImplementation.swift
```diff
@@ -49,19 +54,52 @@ class DataPipelineImplementation: DataPipelineInstance {
...
+    }
 
     func registerDeviceToken(_ deviceToken: String) {
         analytics.setDeviceToken(deviceToken)
     }
 
     func deleteDeviceToken() {
-        fatalError("will be implemented later")
+        // Remove DeviceToken plugin to prevent attaching the token to every request
+        if let tokenPlugin = analytics.find(pluginType: DeviceToken.self) {
```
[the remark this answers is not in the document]
mrehan27: Yeah 😅 More token changes around Data pipeline are also coming soon so that might help too

Sources/DataPipeline/DataPipelineImplementation.swift
```diff
@@ -27,9 +27,14 @@ class DataPipelineImplementation: DataPipelineInstance {
         analytics.identify(userId: identifier, traits: body)
     }
 
-    var registeredDeviceToken: String?
+    var registeredDeviceToken: String? {
+        analytics.find(pluginType: DeviceToken.self)?.token
```
Shahroz16, line 31: hmm does it fetch it from segment storage?

Shahroz16 (approved):

Sources/DataPipeline/DataPipelineImplementation.swift
levibostian: This feels weird to me for the core CDP SDK to ask an optional plugin for a important piece of data: a device token. 

A device token is a really important piece of data our SDK relies on for a lot of feature. I would expect that the CDP SDK (or segment) keeps track of this piece of data and the device token plugin itself refers to the CDP/Segment SDK. Instead of the other way around, as shown here.

mrehan27: I think token isn't really important for segment because it is mainly for tracking. They just have [this small extension](https://github.com/customerio/cdp-analytics-swift/blob/main/Sources/Segment/Plugins/DeviceToken.swift#L33) to add token. But I can't see anywhere segment storing it. I'll be making another PR today to store/fetch token and utilize it, that might help make things more clear.

PR #433, opened 2023-12-12
fix: infinite loop with FCM SDK push delegate calling our delegate again

levibostian:
During QA testing on RN sample app with rn-firebase installed, I hit an infinite loop. 

* Push clicked. Push not sent from CIO. 
* Our push click handler called. Via the proxy, we forward request to rn-firebase push click handler. 
* rn-firebase push click handler handles the push. Then, it calls our push click handler again. 
* Repeat...

This PR fixes this infinite loop scenario by modifying the push history implementation. 

Code has been tested in RN FCM app.


Sources/MessagingPush/PushHandling/iOSPushEventListener.swift
```diff
@@ -52,6 +54,13 @@ class iOSPushEventListener: NSObject, PushEventListener, UNUserNotificationCente
     }
 
     func userNotificationCenter(_ center: UNUserNotificationCenter, didReceive response: UNNotificationResponse, withCompletionHandler completionHandler: @escaping () -> Void) {
+        guard !pushHistory.hasHandledPushDidReceive(pushId: response.pushId) else {
```
Shahroz16, line 57: Can different thread still cause this infinite loop due to a race condition?

levibostian: The only scenario that I can currently think of when a race condition could occur is if PushHistory was not thread-safe. 

Luckily, PushHistory is already thread-safe so I don't see that as a concern. 

Do you have a different scenario in mind?

Shahroz16: The non-atomicity between checking and marking a push notification as handled is the only thing that can pop into the mind. Maybe, if the method userNotificationCenter is called concurrently from different threads, there could still be a race condition between the time hasHandledPushDidReceive is called and didHandlePushDidReceive is called?

levibostian: Ah, that's a good point. 

I want to try and idea where we combine the 2 separate PushHistory functions into 1. Not only could that fix this issue that you brought up, but it could make using PushHistory more simple, too. I'll work on this tomorrow.

levibostian: Code has been modified to handle this edge case.

Shahroz16 (approved):

PR #434, opened 2023-12-13
refactor: device attributes using data pipeline

mrehan27:

### Changes

- Removed unused `initialize` from `DataPipeline`
- Updated `DataPipelineImplementation` to set device attributes using `analytics`
- Removed device attributes and token implementation from `CustomerIOImplementation`


Sources/DataPipeline/DataPipeline.swift
```diff
@@ -23,25 +23,6 @@ public class DataPipeline: ModuleTopLevelObject<DataPipelineInstance>, DataPipel
...
 
-    /**
-     Initialize the shared `instance` of `DataPipeline`.
-     Call this function when your app launches, before using `DataPipeline.shared`.
-     */
-    @discardableResult
-    public static func initialize(
-        writeKey: String,
-        configure configureHandler: ((inout DataPipelineConfigOptions) -> Void)? = nil
-    ) -> CustomerIOInstance {
```
mrehan27, line 34: Removed this as we are now using `DataPipeline.initialize(moduleConfig: DataPipelineConfigOptions)`

Sources/DataPipeline/DataPipelineImplementation.swift
```diff
@@ -2,14 +2,33 @@ import CioInternalCommon
...
+        self.dateUtil = diGraph.dateUtil
+        self.deviceInfo = diGraph.deviceInfo
+
+        initialize()
+    }
+
+    private func initialize() {
+        if let token = globalDataStore.pushDeviceToken {
+            // if the device token exists, pass it to the plugin to ensure device attributes are updated with each request
+            setDeviceToken(token)
```
mrehan27, line 30: Wasn't able to test complete flow because Journey implementation is not yet complete. But testing with assumptions and looking at code, it looks like [segment is storing token only in memory](https://github.com/customerio/cdp-analytics-swift/blob/main/Sources/Segment/Plugins/DeviceToken.swift#L33), so we apparently need to pass it on every app launch.

Sources/DataPipeline/DataPipelineImplementation.swift
```diff
@@ -96,6 +119,39 @@ class DataPipelineImplementation: DataPipelineInstance {
...
+        if deviceInfo.osName == nil {
+            logger.info("SDK being executed from unsupported OS. Ignoring request to register push token.")
+            return
+        }
+
+        // Consolidate all Apple platforms under iOS
+        let deviceOsName = "iOS"
+        // FIXME: [CDP] Fetch the right defaultDeviceAttributes here
+        // deviceAttributesProvider.getDefaultDeviceAttributes { defaultDeviceAttributes in
+        let defaultDeviceAttributes: [String: Any] = [:]
```
mrehan27, line 137: `DeviceAttributesProvider` relies on `SdkConfig` for SDK version and wrapper config. I'll refactor it in upcoming PRs and fix this.

Sources/Tracking/CustomerIOImplementation.swift
```diff
@@ -208,68 +202,7 @@ class CustomerIOImplementation: CustomerIOInstance {
...
      */
     public func registerDeviceToken(_ deviceToken: String) {
-        addDeviceAttributes(deviceToken: deviceToken)
-    }
-
-    /**
-     Adds device default and custom attributes and registers device token.
-     */
-    // TODO: Segment doesn't provide this method by default needs to get added
-    private func addDeviceAttributes(deviceToken: String, customAttributes: [String: Any] = [:]) {
```
mrehan27, line 218: With `DataPipelineImplementation` handling all this stuff, we may completely remove `CustomerIOImplementation`, but doing it in phases so it is easier for us to make sure we aren't ignoring anything unknowingly.

Sources/DataPipeline/DataPipelineImplementation.swift
```diff
@@ -2,14 +2,33 @@ import CioInternalCommon
...
+        self.globalDataStore = diGraph.globalDataStore
+        // self.deviceAttributesProvider = diGraph.deviceAttributesProvider
+        self.dateUtil = diGraph.dateUtil
+        self.deviceInfo = diGraph.deviceInfo
+
+        initialize()
+    }
+
+    private func initialize() {
+        if let token = globalDataStore.pushDeviceToken {
```
Shahroz16, line 28: we might not need do it anymore with journeys inclusion

Sources/DataPipeline/DataPipelineImplementation.swift
```diff
@@ -68,31 +87,77 @@ class DataPipelineImplementation: DataPipelineInstance {
...
+    /// Internal method for passing the device token to the plugin
+    private func setDeviceToken(_ deviceToken: String) {
+        logger.info("registering device token \(deviceToken)")
+        // Add device attributes first so they are updated for unsupported OS too
+        addDeviceAttributes()
+
+        // OS name might not be available if running on non-apple product. We currently only support iOS for the SDK
+        // and iOS should always be non-nil. Though, we are consolidating all Apple platforms under iOS but this check
+        // is required to prevent SDK execution for unsupported OS.
+        if deviceInfo.osName == nil {
```
Shahroz16, line 136: we were ignoring the attributes before as well, now we seem to be storing them

Sources/DataPipeline/DataPipelineImplementation.swift
```diff
@@ -68,31 +87,77 @@ class DataPipelineImplementation: DataPipelineInstance {
...
+            logger.info("SDK being executed from unsupported OS. Ignoring request to register push token.")
+            return
+        }
         analytics.setDeviceToken(deviceToken)
```
Shahroz16, line 140: can we use the same plugin for token and attributes?

Sources/DataPipeline/DataPipelineImplementation.swift
mrehan27: With moving token to DeviceAttributes updated this to same behavior as before

Sources/DataPipeline/plugins/DeviceAttributes.swift
```diff
@@ -6,24 +6,31 @@ class DeviceAttributes: Plugin {
     public let type = PluginType.before
     public weak var analytics: Analytics?
 
+    public var token: String?
```
mrehan27, line 9: We are now not using segment plugin for token and relying on our plugin instead

Sources/Tracking/CustomerIOImplementation.swift
```diff
@@ -208,28 +202,21 @@ class CustomerIOImplementation: CustomerIOInstance {
      is no active customer, this will fail to register the device
      */
     public func registerDeviceToken(_ deviceToken: String) {
-        addDeviceAttributes(deviceToken: deviceToken)
-    }
+        logger.info("initiating device token registration for \(deviceToken)")
```
mrehan27, line 205: I deliberately renamed this log message as customers using our tracking module would have seen same log twice, one from CustomerIOImplementation, one from DataPipeline

Sources/Tracking/CustomerIOImplementation.swift
```diff
@@ -208,28 +202,21 @@ class CustomerIOImplementation: CustomerIOInstance {
...
-        DataPipeline.shared.registerDeviceToken(deviceToken)
-
-        logger.info("registering device token \(deviceToken)")
-        logger.debug("storing device token to device storage \(deviceToken)")
-        // no matter what, save the device token for use later. if a customer is identified later,
-        // we can reference the token and register it to a new profile.
-        globalDataStore.pushDeviceToken = deviceToken
+        guard let identifier = DataPipeline.shared.analytics.userId else {
+            logger.debug("storing device token to device storage \(deviceToken)")
+            globalDataStore.pushDeviceToken = deviceToken
```
mrehan27, line 211: ` DataPipeline.shared.registerDeviceToken(deviceToken)` already does this, but couldn't think of better solution for early returns at the moment

Sources/Tracking/CustomerIOImplementation.swift
```diff
@@ -242,68 +229,29 @@ class CustomerIOImplementation: CustomerIOInstance {
...
-        guard let existingDeviceToken = globalDataStore.pushDeviceToken else {
+        guard (globalDataStore.pushDeviceToken) != nil else {
             logger.info("no device token exists so ignoring request to delete")
             return // no device token to delete, ignore request
         }
         // Do not delete push token from device storage. The token is valid
         // once given to SDK. We need it for future profile identifications.
 
-        guard let identifiedProfileId = profileStore.identifier else {
+        guard (DataPipeline.shared.analytics.userId) != nil else {
```
mrehan27, line 249: Tried to avoid relying on `profileStore`

Sources/Tracking/CustomerIOImplementation.swift
mrehan27: Update: We will be keeping `CustomerIOImplementation` as it will help us keep the same experience for existing users without impacting customer coming from Segment.

Sources/DataPipeline/DataPipelineImplementation.swift
mrehan27: Updated

Sources/DataPipeline/DataPipelineImplementation.swift
mrehan27: Fixed ✅

Sources/Tracking/CustomerIOImplementation.swift
mrehan27: Removed this as we decided to remove more checks

Sources/Tracking/CustomerIOImplementation.swift
mrehan27: Removed this as we decided to remove more checks

Sources/Tracking/CustomerIOImplementation.swift
mrehan27: Removed this as we decided to remove more checks

Sources/DataPipeline/DataPipelineImplementation.swift
```diff
@@ -57,49 +88,167 @@ class DataPipelineImplementation: DataPipelineInstance {
...
+            // token has been refreshed, delete old token to avoid registering same device multiple times
+            deleteDeviceToken()
+        }
+        deviceAttributesPlugin.token = deviceToken
+
+        // Consolidate all Apple platforms under iOS
+        deviceAttributesProvider.getDefaultDeviceAttributes { defaultDeviceAttributes in
+            let deviceAttributes: [String: Any] = defaultDeviceAttributes
+                .mergeWith([
+                    "last_used": self.dateUtil.now
```
mrehan27, line 173: Removed `platform` as it causes issues on server side, we no longer need it

mrehan27:
Device and metric events json have been verified. Merging this as the changes are required on other branches for testing and code changes.

Sources/DataPipeline/DataPipelineConfigOptions.swift
```diff
@@ -57,4 +57,7 @@ public struct DataPipelineConfigOptions {
     public var flushQueue: DispatchQueue = .init(label: "com.segment.operatingModeQueue", qos: .utility)
     public var operatingMode: OperatingMode = .asynchronous
     public var trackApplicationLifecycleEvents: Bool = true
+
+    /// Configuration options for users to easily add available plugins
+    public var autoTrackDeviceAttributes: Bool = true
```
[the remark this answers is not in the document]
levibostian: Could you ask this in the slack channel? I want to make sure it doesn't go unanswered.

Sources/DataPipeline/DataPipelineImplementation.swift
```diff
@@ -2,14 +2,40 @@ import CioInternalCommon
...
+        if moduleConfig.autoAddCustomerIODestination {
+            let customerIODestination = CustomerIODestination()
+            customerIODestination.analytics = analytics
+            analytics.add(plugin: customerIODestination)
+        }
+
+        if let existingDeviceToken = globalDataStore.pushDeviceToken {
+            // if the device token exists, pass it to the plugin and ensure device attributes are updated
+            addDeviceAttributes(token: existingDeviceToken)
+        }
```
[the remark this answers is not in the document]
levibostian: If this is true, could we make a `TODO` or something to document this change?

Sources/DataPipeline/plugins/DeviceAttributes.swift
```diff
@@ -6,24 +6,31 @@ class DeviceAttributes: Plugin {
...
     public var attributes: [String: Any]?
 
     public required init() {}
 
     public func execute<T: RawEvent>(event: T?) -> T? {
-        guard var workingEvent = event else { return event }
+        guard var workingEvent = event,
+              var context = workingEvent.context?.dictionaryValue
+        else { return event }
```
levibostian, line 17: What is this doing? From reading it, I don't understand what this code is for?

Sources/DataPipeline/plugins/DeviceAttributes.swift
```diff
@@ -6,24 +6,31 @@ class DeviceAttributes: Plugin {
...
                     context["device"] = device.mergeWith(attributes)
                 } else {
                     context["device"] = attributes
                 }
                 workingEvent.context = try JSON(context)
-            } catch {
-                analytics?.reportInternalError(error)
             }
+        } catch {
+            analytics?.reportInternalError(error)
```
levibostian, line 33: Since this file is code that we wrote, I was expecting to use the CIO logger, `logger.error()`. Could that work better?

Sources/DataPipeline/DataPipelineImplementation.swift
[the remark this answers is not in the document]
Shahroz16: I think we will need to see the use-case as its using a plugin, i probably transforming data on basis of event.

Sources/DataPipeline/plugins/DeviceAttributes.swift
Shahroz16: transforming the data, if you look into the `payload` expected by CDP and generated by SDK that might help in understand it more. 

`Raw Event` already has a json, which has `context` and other `json object` in it. This plugin updates the values of `context` json object.

PR #435, opened 2023-12-13
refactor: profile attributes using data pipeline

mrehan27:

### Changes

- Updated `DataPipelineImplementation` to set profile attributes using `analytics`
- Removed profile attributes and profile implementation from `CustomerIOImplementation`


Sources/Common/CustomerIOInstance.swift
```diff
@@ -20,6 +20,10 @@ public protocol CustomerIOInstance: AutoMockable {
         body: RequestBody
     )
 
+    // sourcery:Name=identifyAnonymousEncodable
+    // sourcery:DuplicateMethod=identify
+    func identify(body: Codable)
```
mrehan27, line 25: Used `Codable` directly because I was getting error with generics

```
Cannot find type 'RequestBody' in scope
```

while the implementation was

```
// sourcery:Name=identifyAnonymousEncodable
// sourcery:DuplicateMethod=identify
func identify<RequestBody: Codable>(
    // sourcery:Type=AnyEncodable
    body: RequestBody
)
```

Sources/DataPipeline/DataPipelineImplementation.swift
```diff
@@ -76,11 +81,49 @@ class DataPipelineImplementation: DataPipelineInstance {
     var profileAttributes: [String: Any] {
         get { analytics.traits() ?? [:] }
         set {
-            let userId = analytics.userId ?? analytics.anonymousId
-            analytics.identify(userId: userId, traits: newValue)
+            let userId = analytics.userId
+            guard let userId = userId else {
```
mrehan27, line 85: I have skipped empty/blank checks to match `analytics`

Sources/Tracking/CustomerIOImplementation.swift
```diff
@@ -72,93 +65,19 @@ class CustomerIOImplementation: CustomerIOInstance {
...
-                // this code assumes that the newly identified profile has been saved to device storage. only call this
-                // function until after the SDK stores the new profile identifier
-                registerDeviceToken(existingDeviceToken)
-            }
-
-            logger.debug("running hooks profile identified \(identifier)")
-            hooks.profileIdentifyHooks.forEach { hook in
-                hook.profileIdentified(identifier: identifier)
-            }
-        }
```
mrehan27, line 136: I have removed all these checks because it looks like they aren't needed for `analytics`

Sources/DataPipeline/DataPipelineImplementation.swift
```diff
@@ -39,22 +39,27 @@ class DataPipelineImplementation: DataPipelineInstance {
...
 
     func identify(identifier: String, body: [String: Any]) {
-        analytics.identify(userId: identifier, traits: body)
+        commonIdentifyProfile(userId: identifier, attributesDict: body)
     }
 
     func identify<RequestBody: Codable>(identifier: String, body: RequestBody) {
-        analytics.identify(userId: identifier, traits: body)
+        commonIdentifyProfile(userId: identifier, attributesCodable: body)
+    }
```
Shahroz16, line 47: Some `identify` calls are being sent directly by CustomerIO+Segment extensions, would we add these in here as well?

mrehan27: Discussed on call, will try using plugin or routing all through DataPipeline, whatever works

mrehan27:
Merging this as the changes are required on other branches for testing and code changes

Sources/Common/CustomerIOInstance.swift
levibostian: This might work for customers, but it also might not. I can't confirm unless we had it working on a sample app and so I saw some example code for reference. 

We can address this at a later time once we implement a `identify` call in the sample apps.

Sources/DataPipeline/DataPipelineImplementation.swift
levibostian: I think this is OK. If `analytics.userId` is empty or blank and we don't want that, it's too late at this point in time, setting profile attributes. We should have prevented `analytics.userId` being set with an empty/blank id in `identify`.

Sources/DataPipeline/DataPipelineImplementation.swift
```diff
@@ -76,11 +81,49 @@ class DataPipelineImplementation: DataPipelineInstance {
...
+        let currentlyIdentifiedProfile = analytics.userId
+        let isChangingIdentifiedProfile = currentlyIdentifiedProfile != nil && currentlyIdentifiedProfile != userId
+        let isFirstTimeIdentifying = currentlyIdentifiedProfile == nil
+
+        if isFirstTimeIdentifying || isChangingIdentifiedProfile {
+            // logger.debug("running hooks profile identified \(userId)")
+            // FIXME: [CDP] Request Journeys to invoke profile identify hooks
+            // hooks.profileIdentifyHooks.forEach { hook in
+            //     hook.profileIdentified(identifier: userId)
+            // }
```
levibostian, line 103: I *think* that this profile hook was only intended for having the messagingpush module register an existing device token to a profile. Now that registering devices tokens exists in the tracking module, we may not need a hook/journeys call at all.

Sources/DataPipeline/DataPipelineImplementation.swift
```diff
@@ -76,11 +81,49 @@ class DataPipelineImplementation: DataPipelineInstance {
...
+            let userId = analytics.userId
+            guard let userId = userId else {
+                logger.error("No user identified. If you don't have a userId but want to record traits, please pass traits using identify(body: Codable)")
+                return
+            }
+            commonIdentifyProfile(userId: userId, attributesDict: newValue)
         }
     }
 
+    private func commonIdentifyProfile(userId: String, attributesDict: [String: Any]? = nil, attributesCodable: Codable? = nil) {
```
levibostian, line 93: I don't see any checks preventing blank/empty userid in this function. 

Is that on purpose? 

I understand that maybe the segment SDK accepts blank/empty identifiers but maybe that will put our SDK into an invalid state as it has in the past?

levibostian: I am not sure what PR it was, but @Shahroz16, @ami-aman and I had this conversation. And I believe we decided it was best to prevent blank/empty identifiers by logging a warning and ignoring the identify request.

Sources/DataPipeline/DataPipelineImplementation.swift
```diff
@@ -76,11 +81,49 @@ class DataPipelineImplementation: DataPipelineInstance {
...
+
+    private func commonClearIdentify() {
+        // logger.debug("deleting device info from \(currentlyIdentifiedProfile) to stop sending push to a profile that is no longer identified")
+        // TODO: [CDP] Confirm how can we delete devices for CDP
+
+        // logger.debug("running hooks: profile stopped being identified \(currentlyIdentifiedProfile)")
+        // FIXME: [CDP] Request Journeys to invoke profile clearing hooks
+        // hooks.profileIdentifyHooks.forEach { hook in
+        //     hook.beforeProfileStoppedBeingIdentified(oldIdentifier: currentlyIdentifiedProfileIdentifier)
+        // }
```
levibostian, line 120: similar comment as with profile identified hooks --> registering device tokens. Maybe we don't need hooks/journeys at all here? Maybe this function can internally delete device token from profile?

Sources/Tracking/CustomerIOImplementation.swift
```diff
@@ -48,15 +48,8 @@ class CustomerIOImplementation: CustomerIOInstance {
...
-            [:]
-        }
-        set {
-            guard let existingProfileIdentifier = profileStore.identifier else {
-                return
-            }
-            identify(identifier: existingProfileIdentifier, body: newValue)
-        }
+        get { DataPipeline.shared.profileAttributes }
+        set { DataPipeline.shared.profileAttributes = newValue }
```
levibostian, line 52: This is OK for now, but I want to point out to possibly avoid more tech debt in future PRs. 

By the Tracking module using a singleton instance directly inside of function bodies like this will make this function body untestable. 

I faced a similar situation recently when the messagingpush module needed to call `CustomerIO.shared` in the track module. [I used `CustomerIOInstance` protocol](https://github.com/customerio/customerio-ios/blob/b61b6d52b87f02ccb5f406bd559f7ff17e0d3411/Sources/MessagingPush/PushHandling/PushClickHandler.swift#L17-L19) in the class so that I could mock it in automated tests. 

[I then added a DiGraph extension](https://github.com/customerio/customerio-ios/blob/fb7b75d93563e9ab97692a31d4365131b161306b/Sources/Tracking/CustomerIO.swift#L427-L436) so that `DataPipeline.shared` could be used in production when not mocking. 

This is a non-blocker. Sharing idea if it's helpful for future PRs.

PR #436, opened 2023-12-14
chore: track metric event using data pipeline

mrehan27:

### Changes

- Updated `DataPipelineImplementation` to track metric event for both push and in-app


Sources/DataPipeline/DataPipelineImplementation.swift
```diff
@@ -196,9 +196,32 @@ class DataPipelineImplementation: DataPipelineInstance {
     }
 
     func trackMetric(deliveryID: String, event: Metric, deviceToken: String) {
```
mrehan27, line 198: Should we rename it to `trackPushMetric` for clarity? 🤔

Sources/DataPipeline/DataPipelineImplementation.swift
```diff
@@ -196,9 +196,32 @@ class DataPipelineImplementation: DataPipelineInstance {
...
-        let properties = MetricEvent(event: name, metric: event, deliveryId: deliveryID, deliveryToken: deviceToken)
-        analytics.track(name: name, properties: properties)
+        logger.info("push metric \(event.rawValue)")
+
+        logger.debug("delivery id \(deliveryID) device token \(deviceToken)")
+
+        trackMetricEvent(deliveryID: deliveryID, event: event, deviceToken: deviceToken)
+    }
+
+    func trackInAppMetric(deliveryID: String, event: Metric, metaData: [String: Any]) {
```
mrehan27, line 206: This can be called by Journeys/EventBus consumer

Sources/DataPipeline/DataPipelineImplementation.swift
```diff
@@ -21,14 +21,11 @@ class DataPipelineImplementation: DataPipelineInstance {
...
+        initialize(diGraph: diGraph)
     }
 
-    private func initialize() {
-        if let token = globalDataStore.pushDeviceToken {
-            // if the device token exists, pass it to the plugin to ensure device attributes are updated with each request
-            setDeviceToken(token)
-        }
+    private func initialize(diGraph: DIGraphShared) {
+        analytics.add(plugin: ConsoleLogger(diGraph: diGraph))
```
Shahroz16, line 28: `ConsoleLogger` should be bound to debug mode no?

mrehan27: Thanks for the idea, updated it

mrehan27:
Merging this as the changes are required on other branches for testing and code changes

Sources/DataPipeline/DataPipelineImplementation.swift
levibostian: I am open to ideas. I like the idea of keeping `trackMetric` for backwards compatibility but I also see value in adding another function `trackPushMetric` that could simply perform an internal call to `trackMetric`? 

If you think we should perform a rename or function add, perhaps sending a message to the cdp slack channel? Since it's public api related, I think team input would be valuable.

Sources/DataPipeline/DataPipelineImplementation.swift
```diff
@@ -196,9 +196,33 @@ class DataPipelineImplementation: DataPipelineInstance {
...
+        logger.info("in-app metric \(event.rawValue)")
+
+        logger.debug("delivery id \(deliveryID) metaData \(metaData)")
+
+        trackMetricEvent(deliveryID: deliveryID, event: event, metaData: metaData)
+    }
+
+    /// Tracks metric events for push and in-app messages
+    private func trackMetricEvent(deliveryID: String, event: Metric, deviceToken: String? = nil, metaData: [String: Any] = [:]) {
+        // property keys should be camelCase
```
levibostian, line 216: I would like to ask why they should be camelCase?

Shahroz16: I think backend requirement.

PR #437, opened 2023-12-14
fix: prevent SDK's UNUserNotificationCenterDelegate from being garbage collected

levibostian:
While performing QA testing of [iOS more reliable `opened` metrics feature](https://github.com/customerio/customerio-ios/pull/403) in RN and Flutter, I encountered a bug that did not occur when I did QA testing on iOS. 

The bug occurs because of native iOS's `CustomerIO.initialize()` function getting called multiple times. I suggest reviewing [this comment](https://github.com/customerio/customerio-ios/pull/437/files#r1430398398) that tries to explain the bug and how it's encountered. 

This PR fixes the issue by introducing a workaround. I chose this implementation because: 
* It's low risk for our RN and Flutter SDKs. Compared to other solutions such as modifying the digraph or	`CustomerIO.initialize` behavior, this solution is much lower risk of introducing new bugs in the SDK. 
* I felt comfortable writing a workaround because of changes we have begun introducing to the native digraph with CDP work. A long-term non-workaround solution to this problem can be implemented as we continue to move our digraph's implementation to *create 1 instance of digraph that never gets re-created*. 

# Testing 

QA tests for this PR have been successful. 

My test case: 
* Run RN iOS sample app with this bug fix installed. Send device a push from CIO. Click on push, expect to see `opened` push metric reported. 
* Without modifying any code in sample app, revert this bug fix branch. Re-install the RN sample app on device. Send push again and click on push. Expect *not* to see `opened` push metric.

### DIFF
--- a/Apps/APN-UIKit/APN UIKit/AppDelegate.swift
+++ b/Apps/APN-UIKit/APN UIKit/AppDelegate.swift
@@ -13,8 +13,13 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
         initializeCioAndInAppListeners()
 
-        // Register a 2nd push click listener besides the Customer.io SDK.
-        // Tests that the SDK is able to handle push clicks when there are multiple click listeners.
+        /**
+         Registers the `AppDelegate` class to handle when a push notification gets clicked.
+         This line of code is optional and only required if you have custom code that needs to run when a push notification gets clicked on.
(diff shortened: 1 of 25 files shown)


Shahroz16 (approved):
Look okay for now, not the prettiest case to be handling but i believe with the decoupling for push/in-app modules and not having to initialize CustomerIO SDK from native would remove this problem.

PR #438, opened 2023-12-15
fix: manual push click handling may not track opened metric

levibostian:
Fixes a bug I discovered. See [comment](https://github.com/customerio/customerio-ios/pull/438/files#r1428173254) that explains bug. 

Added automated tests to verify bug fix.

### DIFF
--- a/Sources/MessagingPush/PushHandling/PushClickHandler.swift
+++ b/Sources/MessagingPush/PushHandling/PushClickHandler.swift
@@ -11,18 +11,17 @@ protocol PushClickHandler: AutoMockable {
 @available(iOSApplicationExtension, unavailable)
 // sourcery: InjectRegister = "PushClickHandler"
 class PushClickHandlerImpl: PushClickHandler {
-    private let sdkConfig: SdkConfig
     private let deepLinkUtil: DeepLinkUtil
     private let pushHistory: PushHistory
     private let customerIO: CustomerIOInstance
 
-    init(sdkConfig: SdkConfig, deepLinkUtil: DeepLinkUtil, pushHistory: PushHistory, customerIO: CustomerIOInstance) {
-        self.sdkConfig = sdkConfig
+    init(deepLinkUtil: DeepLinkUtil, pushHistory: PushHistory, customerIO: CustomerIOInstance) {
         self.deepLinkUtil = deepLinkUtil
         self.pushHistory = pushHistory
         self.customerIO = customerIO
     }
 
+    // Note: This function is called from automatic and manual push click handlers.
     func pushClicked(_ parsedPush: CustomerIOParsedPushPayload) {
         guard !pushHistory.hasHandledPushClick(deliveryId: parsedPush.deliveryId) else {
             // push has already been handled. exit early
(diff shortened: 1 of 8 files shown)


Sources/Tracking/CustomerIO.swift
```diff
@@ -427,6 +427,10 @@ public class CustomerIO: CustomerIOInstance {
 // Convenient way for other modules to access CustomerIOInstance as well as being able to mock CustomerIOInstance.
 public extension DIGraph {
     var customerIOInstance: CustomerIOInstance {
-        CustomerIO.shared
+        if let override: CustomerIOInstance = getOverriddenInstance() {
```
levibostian, line 430: modified this to allow mocking in integration tests

Tests/MessagingPush/IntegrationTest.swift
```diff
@@ -26,4 +27,18 @@ class IntegrationTest: SharedTests.IntegrationTest {
             MessagingPush.initialize()
         }
     }
+
+    func getPush(content: [AnyHashable: Any], deliveryId: String = .random, deviceToken: String = .random) -> CustomerIOParsedPushPayload {
```
levibostian, line 31: This is not a new function. It got moved into this file so that any integration test within MessagingPush module can use it.

Sources/MessagingPush/PushHandling/PushClickHandler.swift
```diff
@@ -32,10 +31,9 @@ class PushClickHandlerImpl: PushClickHandler {
 
         // Now we are ready to handle the push click.
 
-        // Track metrics
-        if sdkConfig.autoTrackPushEvents {
-            customerIO.trackMetric(deliveryID: parsedPush.deliveryId, event: .opened, deviceToken: parsedPush.deviceToken)
-        }
+        // Do not check SdkConfig if automatic push tracking is enabled. Always track opened metric.
+        // This function is also called by manual push click handling.
```
levibostian, line 35:
```suggestion
```

matt-frizzell (approved):
LGTM

Sources/MessagingPush/PushHandling/PushClickHandler.swift
```diff
@@ -32,10 +31,9 @@ class PushClickHandlerImpl: PushClickHandler {
 
         // Now we are ready to handle the push click.
 
-        // Track metrics
-        if sdkConfig.autoTrackPushEvents {
```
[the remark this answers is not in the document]
Shahroz16: Thanks for the very thorough explanation Levi, I believe this `if` was added to avoid the automatic tracking, by removing it do we track metric even when sets this config as `false` ?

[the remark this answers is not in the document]
levibostian: The `if` statement has been moved out of the `PushClickHandler` so that it can be used for both manual and automatic push click handing. Whatever a customer chooses to do. 

[Here](https://github.com/customerio/customerio-ios/blob/b61b6d52b87f02ccb5f406bd559f7ff17e0d3411/Sources/MessagingPush/MessagingPush.swift#L69C1-L69C1) is where the `if` statement has been moved to. Does that answer your question?

PR #439, opened 2023-12-15
test: fix flaky test

levibostian:
I have noticed this test function is a bit flaky when running on the CI. 

CI servers are less powerful machines so having the test take a little longer to pass makes sense. 

I don't recommend running the test function less < 100 times so I thought that bumping up the `wait()` function was the best solution. I don't like making a test suite take longer to execute (by bumping the wait() time), but a test suite that sometimes catches bugs and sometimes doesn't (by decreasing the number of times the test runs) sounds worse 🤷‍♀️.


matt-frizzell (approved):
I have issues with dealing with flaky tests with increased timeouts. In this scenario I'm ok and I'm happy that we have left the breadcrumb to understand this in the future,

PR #440, opened 2023-12-15
chore: added event bus with notification center

Shahroz16:
# EventBus

The EventBus is a component that allows different modules of your SDK to send and receive events. Events, which conform to the `EventRepresentable` protocol, enable actions to be triggered in response to specific occurrences in your app.

## Memory vs. File Storage

The EventBus manages event storage with two distinct mechanisms: memory storage and file storage.

### Memory Storage

- Memory storage is optimized for quick event delivery, especially for observers registered before events occur.
- All replay operations are done on memory storage
- Events in memory storage are not guaranteed to persist across application sessions.
- Ideal for events relevant only to the current session.

### File Storage

- File storage ensures the persistence of events, even across application sessions.
- Perfect for long-term persistence, such as metric tracking that happens via NSE.
- Supports event retrieval and storage.

## EventBusHandler

The `EventBusHandler` class manages event delivery and storage, facilitating interaction between the EventBus, memory storage, and file storage.

Key functionalities:

- Loads events from file storage during initialization, which were never recieved and add it in memory. 
- Registers observers for specific event types using `addObserver`.
- Replays events from memory storage for late-registered observers.
- Posts events using `postEvent`, storing them if no observers are registered.
- Deletes event from storage which are listened by atleast 1 observer but they stay in memory for later subscribers. 

## EventBus

The `SharedEventBus` class provides a shared EventBus instance for seamless communication across your app. It offers a range of functionalities:

- Posts events using the `post` method, delivering them to registered observers.
- Registers observers for specific event types with `addObserver`.
- Removes specific observers or all observers for an event type.
- Ensures thread safety during event posting and observer management.
- Automatically removes observers on deinitialization for proper cleanup.

### NotificationCenter Wrapper

Beneath its functionality, the `SharedEventBus` utilizes NotificationCenter but provides a wrapper for more control and tracking over its observers and publishers. This allows for efficient communication between different components of your application using events.

Why this is needed? Here is a slack [comment](https://customerio.slack.com/archives/C05UY5J4VLK/p1701952113139529) that might help. Its replacement of Journey module.


Sources/Common/Communication/EventBus.swift
```diff
@@ -1,140 +1,104 @@
...
+    private let queue = DispatchQueue(label: "com.eventbus.shared")
 
-    /// Initializes a new instance of `SharedEventBus`.
-    ///
-    /// - Parameter listenersRegistry: The registry used to manage listeners for different event types.
-    public init(listenersRegistry: EventListenersRegistry) {
-        self.listenersRegistry = listenersRegistry
+    deinit {
+        removeAllObservers()
     }
```
levibostian, line 25: I am wondering why this class has it's own observers data structure when the NotificationCenter already stores these observers for us? 

From reading the code, it looks like the reason is to remove all observers from the NotificationCenter when the eventbus `deinit` is called? 

If this is the reason, I don't think this will work. We run the risk of memory leaks by this implementation where classes that observe a NotificationCenter event are kept in memory instead of getting destroyed.

The only way that I can think of to prevent this scenario is to have each class remove itself as the observer in it's own `deinit` function: 

```swift
class Foo {

  ...
  eventBus.addObserver(self, ...)
  ...

  deinit {
    eventBus.removeObserver(self)
  }
}
```

levibostian: I don't currently understand the need for a `DispatchQueue` in this class.

From looking at the code, it seems like it's being used to create a mutex around the `observers` property. Is this correct?

Sources/Common/Communication/EventBus.swift
```diff
@@ -1,140 +1,104 @@
...
-    /// - Returns: A cancellable instance used to unsubscribe from the event.
-    @discardableResult
-    public func onReceive<E: EventRepresentable, S: Scheduler>(_ eventType: E.Type, performOn scheduler: S, action: @escaping (E) -> Void) -> AnyCancellable {
-        subscribeAndNotify(eventType: eventType, scheduler: scheduler, action: action)
+                // Posts the event asynchronously on the provided queue or immediately if no queue is provided.
+                if let queue = queue {
+                    queue.async(execute: postAction)
+                } else {
+                    postAction()
+                }
```
levibostian, line 52: Instead of using a `DispatchQueue` directly, could you instead use the SDK's `ThreadUtil` class? I think that could make this function's use less complex.

Sources/Common/Communication/EventBusHandler.swift
```diff
@@ -1,66 +1,87 @@
...
     private let eventBus: EventBus
     private let eventStorage: EventStorage
+    private var memoryStorage: [String: [any EventRepresentable]] = [:]
 
+    /// Initializes the EventBusHandler with an EventBus and EventStorage.
+    /// Automatically loads events from file-based storage into in-memory storage upon initialization.
     public init(eventBus: EventBus, eventStorage: EventStorage) {
         self.eventBus = eventBus
         self.eventStorage = eventStorage
+        loadEventsFromStorage()
```
levibostian, line 18: I wonder if calling this function directly in the constructor could cause lower performance of the SDK, if the constructor is called on the main thread? 

Perhaps, the `load` function is an async function that executes on a background thread? Or, it's a synchronous function but we add a comment suggesting it gets called from a background thread?

Sources/Common/Communication/EventBusHandler.swift
```diff
@@ -1,66 +1,87 @@
...
             do {
-                try eventStorage.store(event: event, forKey: key)
+                let key = eventType.key
+                let events: [any EventRepresentable] = try eventStorage.loadEvents(ofType: eventType)
+                memoryStorage[key, default: []].append(contentsOf: events)
             } catch {
-                print("Error storing event: \(error)")
+                // Handle the error, for example, log it or take some recovery actions
+                print("Error loading events for \(eventType): \(error)")
             }
```
levibostian, line 32: I agree, logging errors when a store encounters an error is a great idea. 

I do have 2 suggestions about this. 
1. Instead of `print`, could we use `logger.error` so that customers can more easily see these errors and report them to us? 
2. Could we modify `eventStorage.loadEvents(ofType: eventType)` to not `throw`? I prefer to [encapsulate all error catching in 1 place such as we do here in the SDK](https://github.com/customerio/customerio-ios/blob/94cbf686c3c2a405534cfe908f7166558a3e0b5d/Sources/Common/Util/FileStorage.swift#L93-L102). I like this approach because it requires less boilerplate code (having to use try/catch/log each time we call `eventStorage.loadEvents`.

Sources/Common/Communication/EventBusHandler.swift
```diff
@@ -1,66 +1,87 @@
...
     }
 
-    @discardableResult
-    public func onReceive<E: EventRepresentable>(_ eventType: E.Type, perform action: @escaping (E) -> Void) -> AnyCancellable {
-        subscribeAndReplay(eventType: eventType, action: action)
+    /// Adds an observer for a specific event type.
+    /// Replays any in-memory stored events of that type to the new observer.
+    public func addObserver<E: EventRepresentable>(_ eventType: E.Type, action: @escaping (E) -> Void) {
+        eventBus.addObserver(eventType, action: action)
+        replayEvents(forType: eventType)
```
levibostian, line 40: I am wondering if this implementation of the replay feature could cause issues 🤔. 

When thinking about the eventbus and the replay system, I was imagining that the eventbus stored an *ordered list* of events and replayed all events in order to the consumer once it attached. 

Order of events in our SDK have always mattered for the functionality of the SDK. Profile identify event has to come before tracking of events, for example. If those 2 events are delivered to consumers out of order, that could make the results going to the CDP be very different then intended. 

From reviewing the implementation of this class thus far, it seems as though the eventbus implementation does not store and replay an ordered list. 

Do you disagree that an ordered list of events is needed? 

As a bonus, I do think that 1 single list of events could make the implementation more simple in this class, too.

Sources/Common/Communication/EventBusHandler.swift
```diff
@@ -1,66 +1,87 @@
...
-        return subscription
+    /// Stores an event in file-based persistent storage.
+    /// This method is used when an event is posted but no observers are currently registered.
+    private func storeEvent<E: EventRepresentable>(_ event: E) {
+        do {
+            try eventStorage.store(event: event)
+        } catch {
+            print("Error storing event: \(error)")
+        }
     }
```
levibostian, line 76: I could be wrong here, but I think that the persistent event storage is only needed for `delivered` push metrics, correct? 

If this is correct, I think that before we commit to this implementation, we first need to go to the team for a decision. A decision on how `delivered` push events are to be implemented in the CDP SDK at launch. 

I say this because the persistent storage makes this eventbus implementation more complex to test and maintain. If we are only adding this complexity for `delivered` metrics (which has multiple implementation paths available), that is something good to bring to the team.

Sources/Common/Communication/EventStorage.swift
```diff
@@ -1,90 +1,69 @@
...
 
 // sourcery: InjectRegisterShared = "EventStorage"
 public class EventStorageManager: EventStorage {
-    private let fileManager: FileManager = .default
-    private let documentsDirectory: URL
+    private let fileManager = FileManager.default
+    private let baseDirectory: URL
+    private let encoder = JSONEncoder()
+    private let decoder = JSONDecoder()
+    private let ioQueue: DispatchQueue
```
levibostian, line 18: I was expecting to see dependencies such as `JsonAdapter`, `FileStorage`, and `ThreadUtil/Lock/LockManager` used here. 

Was there a reason to avoid using them here?

levibostian (commented):
I'm not yet done reviewing, but I am submitting my review thus far before my EOD.

Sources/Common/Communication/EventBusHandler.swift
Shahroz16: I think it's different then BQ, when does an event get replyed in Event bus case? 

if an observer, subscribes to an `event`, and if there are `events` of that type stored in memory, the Event bus replays those events for the subscriber. 

So the event as the `key` its order isn't important rather the value, as in how many `IdentifyProfile` events were stored as an array value to the key is important and is being stored in order too

```
private var memoryStorage: [String: [AnyEventRepresentable]] = [:]

The order of AnyEventRepresentable instances in these arrays is preserved as per their addition.
```

Sources/Common/Communication/EventBusHandler.swift
Shahroz16: In the call, we did decide to persist the events, especially in the case of `track metric/in-app` because since push/NSE/wrapper have lifecycles of its own, we can't rely on that there can't be a case where if we store events just in the memory, the user can't kill the app and those events that stayed in memory can get lost forever. 

Hence, we need to make sure, all the events have at least been sent to 1 subscriber that was listening for those events.

Sources/Common/Communication/EventStorage.swift
Shahroz16: I was unsure about future of `JsonAdapter` as it has alot of code added for BQ, if all of it goes away, would we still keep it? 

`FileStorage` simialrly, is oriented around BQ, example below, 

```
public enum FileType {
    case queueInventory
    case queueTask
    }
    func save(type: FileType, contents: Data, fileId: String?) -> Bool
```

Sources/DataPipeline/plugins/DataPipelinePublishedEvents.swift
```diff
@@ -0,0 +1,35 @@
+import CioInternalCommon
+import Foundation
+import Segment
+
+/// This class serves as a centralized hub for dispatching events generated from DataPipeline module.
+/// It is designed to reduce redundancy and enhance maintainability by funneling these operations through a single plugin.
+/// This plugin allows for decoupled modules to subscribe and react to these events.
```
levibostian, line 7: I do understand how a segment sdk plugin could reduce copy/pasted `eventbus.postEvent` calls, if the CDP module is calling `eventbus.postEvent` in a lot of places. 

I personally find de-coupled code, such as a plugin system, to sometimes make code harder to read, maintain, and debug. Depends on the use case. From reviewing the code thus far, I do not yet agree this is a good use case for a plugin. 

If we decided to not use this segment plugin, what are our other options for implementation?

Sources/MessagingInApp/MessagingInAppImplementation.swift
```diff
@@ -17,33 +17,33 @@ class MessagingInAppImplementation: MessagingInAppInstance {
...
 
     private func initialize() {
         inAppProvider.initialize(siteId: moduleConfig.siteId, region: moduleConfig.region, delegate: self)
 
-        busEventManager.onReceive(ProfileIdentifiedEvent.self) { event in
+        busEventManager.addObserver(ProfileIdentifiedEvent.self) { event in
             self.logger.debug("registering profile \(event.identifier) for in-app")
 
             self.inAppProvider.setProfileIdentifier(event.identifier)
-        }.store(in: &subscriptions)
```
levibostian, line 31: I believe that after removing this line of code, we can also remove other parts of the code? Such as:

```swift
import Combine

private var subscriptions: Set<AnyCancellable> = []
```

Sources/MessagingInApp/MessagingInAppImplementation.swift
```diff
@@ -73,12 +73,7 @@ extension MessagingInAppImplementation: GistDelegate {
         eventListener?.messageShown(message: InAppMessage(gistMessage: message))
 
         if let deliveryId = getDeliveryId(from: message) {
-            // the state of the SDK does not change if adding this queue task isn't successful so ignore result
-            // FIXME: [CDP] Pass to Journey
-            // _ = queue.addTrackInAppDeliveryTask(deliveryId: deliveryId, event: .opened)
-
-            // TODO: add the TrackInAppMetricEvent
-//            busEventManager.send(TrackInAppMetricEvent(deliveryID: deliveryId, event: "")
+            busEventManager.postEvent(TrackInAppMetricEvent(deliveryID: deliveryId, event: InAppMetric.opened.rawValue))
```
levibostian, line 76:
```suggestion
            busEventManager.postEvent(TrackInAppMetricEvent(deliveryID: deliveryId, event: .opened))
```
 
Could we make the `TrackInAppMetricEvent.event` parameter be a `InAppMetric` data type? It seems easier to use that way.

Sources/MessagingInApp/MessagingInAppImplementation.swift
```diff
@@ -100,9 +95,7 @@ extension MessagingInAppImplementation: GistDelegate {
         // a close action does not count as a clicked action.
         if action != "gist://close" {
             if let deliveryId = getDeliveryId(from: message) {
-                // the state of the SDK does not change if adding this queue task isn't successful so ignore result
-                // FIXME: [CDP] Pass to Journey
-                // _ = queue.addTrackInAppDeliveryTask(deliveryId: deliveryId, event: .clicked, metaData: ["action_name": name, "action_value": action])
+                busEventManager.postEvent(TrackInAppMetricEvent(deliveryID: deliveryId, event: InAppMetric.clicked.rawValue, params: ["action_name": name, "action_value": action]))
```
levibostian, line 98:
```suggestion
                busEventManager.postEvent(TrackInAppMetricEvent(deliveryID: deliveryId, event: InAppMetric.clicked.rawValue, actionName: name, actionValue: action)))
```

Could we avoid using hard-coded strings here?

Sources/MessagingInApp/hooks/HookProvider.swift
```diff
@@ -1,24 +1,24 @@
...
-    }
-
-    var screenTrackingHook: ScreenTrackingHook? {
-        // guard let diGraph = diGraph else { return nil }
-
-        // FIXME: [CDP] Find workaround by attaching hook to Journeys or reusing existing instance to utilize customer provided siteid
-        MessagingInApp.shared.implementation as? MessagingInAppImplementation
-    }
-}
+//// sourcery: InjectRegister = "ModuleHookProvider"
```
levibostian, line 4: Are we ready to delete the hooks code yet?

Sources/MessagingPush/MessagingPushImplementation.swift
```diff
@@ -50,24 +38,15 @@ class MessagingPushImplementation: MessagingPushInstance {
...
-        eventHandlingManager.sendOrSaveEvent(event: RegisterDeviceTokenEvent(token: deviceToken))
+        eventHandlingManager.postEvent(RegisterDeviceTokenEvent(token: deviceToken))
     }
 
     func trackMetric(deliveryID: String, event: Metric, deviceToken: String) {
-        // FIXME: [CDP] Pass to Journey
-        // customerIO?.trackMetric(deliveryID: deliveryID, event: event, deviceToken: deviceToken)
-
-        eventHandlingManager.sendOrSaveEvent(event: TrackMetricEvent(deliveryID: deliveryID, event: event.rawValue, deviceToken: deviceToken))
+        eventHandlingManager.postEvent(TrackMetricEvent(deliveryID: deliveryID, event: event.rawValue, deviceToken: deviceToken))
```
levibostian, line 49:
```suggestion
        eventHandlingManager.postEvent(TrackMetricEvent(deliveryID: deliveryID, event: event, deviceToken: deviceToken))
```

Similar comment about not needing to use `.rawValue`.

Sources/Common/Communication/EventBusHandler.swift
levibostian: The implementation in this PR is more simple. I prefer to not need to involve threading if we can help it. I agree with you that I also do not anticipate issues at this time. 

I think it would be good to add a comment reminding us of this decision. Here is a suggestion to try and help you out. 

```suggestion
        // We do not anticipate the number of events in storage to be a large number at this time. Therefore, we have decided to use this more simple approach of loading events in a synchronous way and not enforcing it be done on a different thread. 
        loadEventsFromStorage()
```

Sources/Common/Communication/EventBusHandler.swift
levibostian: Sorry if my original message was not descriptive enough in what I would like to ask. I would like to know why order of events does not matter in the eventbus?

Sources/Common/Communication/EventBusHandler.swift
levibostian: From my memory of the call, we agreed that a persistent event storage was needed only in the case of `delivered` events and a NSE. If we were to only talk about the host iOS app though, the only scenario where events could be lost is if the CIO SDK never gets initialized (which, I am not sure we need to support that use case?). 

If the above statement is true, then I do still believe that an in-memory store is all that we need in the eventbus.

Sources/Common/Communication/EventBus.swift
levibostian: > NotificationCenter, doesn't provide the information of hasObserver, so if we want to check if there any observers for any perticular event, we can't do so. Hence we are keeping a wrapper around notification center. This is needed for event storage, e.g if there is no observer around for tracking metric, we would know and hence persist the event.

Thanks that's helpful. 

I want to avoid keeping strong references to observers to avoid possible memory leaks. I prefer we let the NotificationCenter be the only one that keeps references to the observer objects. 

Could we change the implementation to keeping an `Int` count of the number of observers instead?

levibostian: > For thread-safety.

The implementation of the `DispatchQueue` looks to be the same as if the `@Atomic` property wrapper was used instead. Does `@Atomic` not work in this case?

Sources/Common/Communication/EventBus.swift
levibostian: The comment explains the reasoning for this `if/else` well. I don't have a suggestion on that. My suggestion was to see if we could make this function call easier to use by abstracting away the `DispatchQueue` class. 

```suggestion
func post<E>(_ event: E, onMainThread: Bool = false) -> Bool {
        ...
        if onMainThread {
           threadUtil.onMainThread {
             postAction()
           }
        } else {
          postAction() 
        }
```

levibostian: Non-blocker side note: I was expecting that the observer of the event was who would determine what thread to handle a eventbus event on. I don't think the code that posts the event needs to determine what thread it should be handled on?

Sources/Common/Communication/EventStorage.swift
levibostian: Good question. 

The `FileStorage` implementation could use a good update. It's quite bias towards the BQ at the moment. The `JsonAdapter`, I am not sure if we need to make any changes to it to be useful outside of the BQ. 

These classes are heavily used in the BQ today. I do however still see a lot of value in using these classes. Reducing boilerplate code, encapsulated error handling, and easy mocking to name a few reasons. 

Therefore, I do think it would be wise to continue using these classes. Even if we decide to change some of the implementation (such as the File storage)

Sources/DataPipeline/plugins/DataPipelinePublishedEvents.swift
Shahroz16: I would agree and disagree. 

Plugin architecture is made for situation like these where you don't care who is triggering the event, you want to ensure you are hooked to its processing timeline. 

If we don't do it via plugin, the alternative would be to then add all the `segment` methods in the CustomerIOInterface and then add implementation that routes to data pipeline and then have common methods that does it for you. 

So, if we have 3 identify methods, 4 track, 4 screen methods (deprecated, new ones) all of them would have their common methods, and inside we will need to `eventbus.post`. 

Its a lot of unnecessary steps, that make reading, and maintaining equally very hard?

Sources/MessagingInApp/MessagingInAppImplementation.swift
Shahroz16: Its in later PRs, where there is alot more removals.

Sources/MessagingInApp/MessagingInAppImplementation.swift
Shahroz16: We can, i wanted to keep these `events` as raw as possible because the notification center stores them as `key value` and I want parsing and storage to be as simple as possible.

Sources/MessagingInApp/MessagingInAppImplementation.swift
Shahroz16: I am not sure i follow, are you suggesting adding these as part of event? but they are meta data, and they can be more of these in future, we shouldn't ideally be adding all of the meta data attributes as parameters? 

hence `param` acts as optional meta data for event

Sources/MessagingInApp/hooks/HookProvider.swift
Shahroz16: Yes, in later  PRs, didn;t want to crowd this one.

Ahmed-CIO:
Where can I find more context about this effort aside from the code?
The PR describes pretty well what's the EventBus, but I need some pointers to learn why it matters now

Sources/Common/Communication/Event.swift
```diff
@@ -24,71 +24,120 @@ public extension EventRepresentable {
...
+            NewSubscriptionEvent.self
+        ]
+    }
+}
+
+// MARK: - Event Structs
+
 public struct ProfileIdentifiedEvent: EventRepresentable {
+    public var storageId: String
     public var params: [String: String] = [:]
```
Ahmed-CIO, line 50: public + mutable, not cool. quick path to threading issues

Sources/Common/Communication/Event.swift
```diff
@@ -24,71 +24,120 @@ public extension EventRepresentable {
...
-    public init(identifier: String, params: [String: String] = [:]) {
+    public init(storageId: String = UUID().uuidString, identifier: String, params: [String: String] = [:]) {
+        self.storageId = storageId
         self.identifier = identifier
         self.params = params
     }
 }
 
 public struct ScreenViewedEvent: EventRepresentable {
+    public var storageId: String
```
Ahmed-CIO, line 61: ditto

Sources/Common/Communication/Event.swift
```diff
@@ -24,71 +24,120 @@ public extension EventRepresentable {
...
     public var params: [String: String] = [:]
```
Ahmed-CIO, line 62: ditto

Sources/Common/Communication/Event.swift
```diff
@@ -24,71 +24,120 @@ public extension EventRepresentable {
...
     public let subscribedEventType: String
 
-    init<E: EventRepresentable>(subscribedEventType: E.Type, params: [String: String] = [:]) {
+    init<E: EventRepresentable>(storageId: String = UUID().uuidString, subscribedEventType: E.Type, params: [String: String] = [:]) {
+        self.storageId = storageId
         self.subscribedEventType = E.key
     }
 }
```
Ahmed-CIO, line 143: this comment applies for all structs above
* You don't need constructors to define default values
* `var` will fire back badly

Sources/Common/Communication/Event.swift
Shahroz16: The events are internal and usage of the events is done under dispatch queue only to make sure we are handling threading but i agree its still safe to make these immutable as we did in the protocol.

Sources/Common/Communication/Event.swift
```diff
@@ -24,71 +24,120 @@ public extension EventRepresentable {
...
 
 public struct DeleteDeviceTokenEvent: EventRepresentable {
+    public var storageId: String
     public var params: [String: String] = [:]
 
-    public init(params: [String: String] = [:]) {
+    public init(storageId: String = UUID().uuidString, params: [String: String] = [:]) {
+        self.storageId = storageId
         self.params = params
```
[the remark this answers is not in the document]
Shahroz16: Since `EventRepresentable ` implements `Codable` we need to have the constructor if we use `let` instead of `var`

From what I have read, since i am new to iOS and swift 😄 

When you declare a property in a Codable conforming struct with both let (making it immutable) and a default value, Swift's Decodable implementation won't be able to overwrite that default value during the decoding process. This is because immutable properties (let) can only be set once, and in your case, they are already set with a default value.

Sources/Common/Communication/Event.swift
Shahroz16: Added response in previous comment

Shahroz16:
@Ahmed-CIO good call out, added the reasoning in the description. But here is the slack [comment](https://customerio.slack.com/archives/C05UY5J4VLK/p1701952113139529) that might help understand the reasoning. 

It basically is the [replacement of the `Journey module`](https://github.com/customerio/issues/issues/11651)

Sources/Common/Communication/Event.swift
Ahmed-CIO: even internal, if it smells it smells and going to cause pain sooner or later :)

Ahmed-CIO:
Thanks a lot Shahroz for the reference, really helpful.
I couldn't get to the rest of the PR as it is pretty packed.
I would really push for:
- The PR can be split in separate smaller PRs (one example is all the EventRepresentable implementations), let's split it into a more manageable smaller ones. Otherwise the iteration over the comments can drag and things can fall between the cracks
- Let's make sure we don't merge anything without appropriate test coverage
- Aside from the comment on the mutability, I already saw an example of race condidion and another of memory leak. Let's make use of `actor`s, async and await instead of GCD

These means this work will take bit longer than anticipated, but I think it is worth it

Shahroz16:
@Ahmed-CIO it's pretty packed because it's made after merging other broken down smaller already reviewed PRs for this feature. 

This PR was created to pivot from `callbacks` mechanism to `notification` center. I did make a [PR](https://github.com/customerio/customerio-ios/pull/445) to add automated tests for event bus. 

Would be really interested to know more about the
> I already saw an example of race condidion and another of memory leak

> Let's make use of actors, async and await instead of GCD
I don't have any specific preference, we can move towards actors for better support. 

Regarding splitting,

>The PR can be split in separate smaller PRs (one example is all the EventRepresentable implementations), let's split it into a more manageable smaller ones. Otherwise the iteration over the comments can drag and things can fall between the cracks

I think that can be very painful and maintaining the PRs might get even more painful as i have got two more PRs on top of this one, i would rather let this one drag 😄 But if its getting hard to review since you didn't get to see the PRs before that, i can try breaking it down.

Ahmed-CIO:
> This PR was created to pivot from callbacks mechanism to notification center. I did make a https://github.com/customerio/customerio-ios/pull/445 to add automated tests for event bus.

Since this is already done this way, I am happy to keep it as is for this particular PR as it has other things worth prioritizing.

> Would be really interested to know more about the

Alright, this is a bit subtle. Check out [this chunk](https://github.com/customerio/customerio-ios/blob/c45d97bb56db689b6fefe3b1a0362511a7cf7ac8/Sources/Common/Communication/EventBus.swift#L42-L52)


```swift
               hasObservers = true
                let postAction = {
                    self.notificationCenter.post(name: NSNotification.Name(E.key), object: event)
                }

                // Posts the event asynchronously on the provided queue or immediately if no queue is provided.
                if let queue = queue {
                    queue.async(execute: postAction)
                } else {
                    postAction()
                }
```
`postAction` which is an escaping closure, has strong reference to `self`. Even if no evidence this will cause memory leak, it is source of troubles. But here is more. We then `queue.async(execute: postAction)` we pass it to an arbitrary queue that we have no idea or control of what it does or when it does it. So now imagine the handler of that event, ends up calling `EventBus.post` in the same queue. Well, bad things happen. 
Some would say: "But we know this internal API and we know how to use it". You know now, while you are writing it. Few weeks later when you are only on one end of the code you wouldn't remember what the other end internals look like.

"The solution is easy, we use weak reference" that's absolutely true. The bigger issue is: if we have an alternative that prevents us from falling into these pitfalls, it is worth considering. I am talking about async/await, actors, tasks, etc.

For the data race: in the [same code chunk](https://github.com/customerio/customerio-ios/blob/c45d97bb56db689b6fefe3b1a0362511a7cf7ac8/Sources/Common/Communication/EventBus.swift#L38-L55)

The hasObserver will sometime return as a true and other as false even if it is in all cases it has observer. That's because there is no telling whether the method will return first the queue will execute first and propagate its memory changes to the calling thread

> I don't have any specific preference, we can move towards actors for better support.
From lot's of experience with multi-threaded codebase, I promise you anything that gives the tiniest possibility to help avoid threading issue is worth taking seriously. actors with tasks, async and await enforce better code flow and thread safety constraints that can take get used to, but they are best hope to minimize threading issues.

> I think that can be very painful and maintaining the PRs might get even more painful as i have got two more PRs on top of this one,

It is well worth it added work, yet I wouldn't like for it to be painful for you. Will try to help out find things that can help make this easier

> i would rather let this one drag
Almost everything I asked for here is going to cause some sort of dragging, so there will be dragging either way. Yet, I am afraid that given the scope of the change, it will eventually be impossible to cover everything properly. Given the impact of this piece, I would err on the side that makes it easier for us to get it right

Shahroz16:
Broken down and reworked PRs can be found here:

- https://github.com/customerio/customerio-ios/pull/449
- https://github.com/customerio/customerio-ios/pull/450
- https://github.com/customerio/customerio-ios/pull/451
- https://github.com/customerio/customerio-ios/pull/452
- https://github.com/customerio/customerio-ios/pull/453
- https://github.com/customerio/customerio-ios/pull/454 

Keeping this PR open because it has a lot of conversations for reference and understanding.

Shahroz16:
merged via broken down PRs.

PR #444, opened 2023-12-19
fix: add support for local notifications

levibostian:
This PR [as well as APN sample app setup PR](https://github.com/customerio/customerio-ios/pull/443) adds support for local push notifications [to the reliable opened push metrics feature](https://github.com/customerio/issues/issues/11289). 

The current implementation has a bug that could make handling local push notifications not work. 

### Expected behavior of the SDK 

* Local push notification with a hard-coded/non-unique identifier is displayed on the device. 
* Push gets clicked. 
* Host app is able to handle that local push getting clicked. 

### Actual behavior 

* Local push notification with a hard-coded/non-unique identifier is displayed on the device. 
* Push gets clicked. 
* Host app is able to handle that local push getting clicked *one time*. 
* If you try and display the local push 2+ times and click on it, the host app will not get a callback to handle the push getting clicked. 

# The solution  

This PR fixes this bug by storing the push ID *and* the date that the push request was sent to the device. Previously, the SDK was only using the push ID which may be hard-coded and not unique. By also using the date, we can feel more confident that a local push with the same push ID can be handled 2+ times. 

I also switched the push history implementation to in-memory instead of persistent. Having an in-memory store is easier to maintain in the future since it can get re-created when app killed. The push history store's purpose is to prevent infinite loops from occurring. I cannot think of a scenario where we need to persist the pushes handled between app launches.

### DIFF
--- a/Apps/APN-UIKit/APN UIKit/AppDelegate.swift
+++ b/Apps/APN-UIKit/APN UIKit/AppDelegate.swift
@@ -99,6 +99,8 @@ extension AppDelegate: UNUserNotificationCenterDelegate {
             name: "push clicked",
             data: ["push": response.notification.request.content.userInfo]
         )
+
+        completionHandler()
     }
 
     // To test sending of local notifications, display the push while app in foreground. So when you press the button to display local push in the app, you are able to see it and click on it.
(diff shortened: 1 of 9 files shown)


Shahroz16 (approved):
Looks okay to me but generally can we add a test with larger number of notifications and multiple threads as well?

PR #446, opened 2023-12-20
test: create test cases covering all use cases we manually test today

levibostian:
This is the start of writing automated tests around the automatic push click handling feature. 

This PR is a set of (failing) test cases. Future PRs will be to get the tests compiling and also passing. 

The focus on this PR is not necessarily the implementation of the test functions. The focus is trying to make sure we are testing all known use cases that we manually test today. The idea is that if we can eventually get all of these test functions in this PR compiling and passing, we can feel confident that this feature will work in production without needing to run all of these cases manually.

### DIFF
--- a/Sources/MessagingPush/PushHandling/iOSPushEventListener.swift
+++ b/Sources/MessagingPush/PushHandling/iOSPushEventListener.swift
@@ -128,15 +128,6 @@ class iOSPushEventListener: NSObject, PushEventListener, UNUserNotificationCente
 
         guard !pushHistory.hasHandledPush(pushEvent: .didReceive, pushId: response.pushId, pushDeliveryDate: response.pushDeliveryDate) else {
             // push has already been handled. exit early
-
-            /*
-             Some 3rd party SDKs (such as FCM SDK) have an implementation of swizzling that can create an infinite loop with our notification center proxy. We keep a history of push notifications that have been handled to prevent this infinite loop.
-
-             Example scenario of infinite loop:
-             - This `userNotificationCenter(didReceive:)` function gets called by OS when a push is clicked.
-             - Our notification center proxy calls the FCM SDK’s `userNotificationCenter(didReceive:)` function. FCM's swizzling implementation involves making a call back to the host app's current UserNotificationCenter.delegate (which is our SDK). See code: https://github.com/firebase/firebase-ios-sdk/blob/5890db966963fd76cfd020d68c0067a7741bef06/FirebaseMessaging/Sources/FIRMessagingRemoteNotificationsProxy.m#L498-L504
-             ... this call to the host app's current delegate means that this function is called again. Once this function is called again, we have gotten into an infinite loop.
-             */
             return
         }
 
--- a/Tests/MessagingPush/PushHandling/AutomaticPushClickHandlingIntegrationTests.swift
+++ b/Tests/MessagingPush/PushHandling/AutomaticPushClickHandlingIntegrationTests.swift
@@ -168,6 +168,10 @@ class AutomaticPushClickHandlingIntegrationTest: IntegrationTest {
 
         XCTAssertEqusl(pushClickHandlerMock.pushClickedCount, 2) // should be handled again, even though we have already handled push with that push ID before.
     }
+    
+    // TODO: the test class here tests only when a push is clicked. Add test functions for when a push is received on the device and we need to determine if we should show the push while app in foreground or not. 
+    // Another way to put it, we need to test our SDK's logic of: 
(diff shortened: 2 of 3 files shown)


Tests/MessagingPush/PushHandling/AutomaticPushClickHandlingIntegrationTests.swift
```diff
@@ -0,0 +1,210 @@
...
+        let givenLocalNotification = getLocalNotification(pushId: "hard-coded-non-unique-id")
+
+        clickOnPushNotification(givenLocalNotification)
+
+        XCTAssertEqusl(pushClickHandlerMock.pushClickedCount, 1) // should only be handled 1 time
+
+        clickOnPushNotification(givenLocalNotification)
+
+        XCTAssertEqusl(pushClickHandlerMock.pushClickedCount, 2) // should be handled again, even though we have already handled push with that push ID before.
+    }
```
levibostian, line 170:
```suggestion
    }
    
    // TODO: the test class here tests only when a push is clicked. Add test functions for when a push is received on the device and we need to determine if we should show the push while app in foreground or not. 
    // Another way to put it, we need to test our SDK's logic of: 
    // userNotificationCenter(willPresent:)
```

Sources/MessagingPush/PushHandling/iOSPushEventListener.swift
```diff
@@ -128,15 +128,6 @@ class iOSPushEventListener: NSObject, PushEventListener, UNUserNotificationCente
 
         guard !pushHistory.hasHandledPush(pushEvent: .didReceive, pushId: response.pushId, pushDeliveryDate: response.pushDeliveryDate) else {
             // push has already been handled. exit early
-
-            /*
```
levibostian, line 132: This comment got moved into the test class. I felt that it made more sense to have this comment explained in the test function that tests against this use case. So, if the test fails, you can read the comment that explains why it could be failing.

Tests/MessagingPush/PushHandling/AutomaticPushClickHandlingIntegrationTests.swift
```diff
@@ -0,0 +1,214 @@
...
+    
+    // TODO: the test class here tests only when a push is clicked. Add test functions for when a push is received on the device and we need to determine if we should show the push while app in foreground or not. 
+    // Another way to put it, we need to test our SDK's logic of: 
+    // userNotificationCenter(willPresent:)
+}
+
+extension AutomaticPushClickHandlingInsetegrationTest {
+    func setupAutomaticPushClickHandlingFeature() {
+        // TODO: we might need to change this function body because we may not be able to run the swizzling code in the test function. Running this line below will setup the swizzling code.
```
levibostian, line 179: Bringing attention to this important note. 

I have not yet been able to create automated tests around Swift swizzling that we can rely on. 

Therefore, expect in a future PR that our automated tests do not actually executed the swizzling code in our SDK for this feature. Instead, we will *mock/stub* the swizzling code by instead directly calling the functions in our SDK that the swizzling code calls. This means that we will still be able to have automated tests around the logic of this feature, but some parts of this feature (such as swizzling) will still need to be tested manually. 

In the future, we can decide how we test swizzling code.

Tests/MessagingPush/PushHandling/AutomaticPushClickHandlingIntegrationTests.swift
```diff
@@ -0,0 +1,214 @@
...
+
+        // Then, iOS will call this function on the delegate.
+        // Note: It's important that we test that the `withContentHandler` callback function gets called either by our SDK (when we handle it), or the 3rd party handler.
+        //       We add an expectation to verify that 1 push click handler calls it.
+        let expectCompletionHandlerCalled = expectation(description: "Expect completion handler called by a click handler")
+        expectCompletionHandlerCalled.expectedFulfillmentCount = 1 // Test will fail if called 2+ times which could indicate a bug because only 1 push click handler should be calling it.
+
+        // TODO: We may need to change the syntax of this line below.
+        // It's difficult to create an instance of UNNotificationRequest because it's initializer is internal.
+        // Therefore, we may need to call a different function then the one below. However, we will make sure that we are still testing the logic of our SDK when this function is called in production.
```
levibostian, line 197: Bringing attention to this important note for what to expect in future PRs on the implementation of this project.

Ahmed-CIO (changes requested):
let's keep it a draft till it is ready for reviewing

levibostian:
I will introduce the tests inside of future PRs. Trying to keep all PRs in stack able to run and pass all CI checks.

PR #447, opened 2023-12-21
refactor: abstract UserNotifications framework from SDK 

levibostian:
In order for us to write automated tests for [this feature](https://github.com/customerio/issues/issues/11289), we need to be able to mock the `UserNotifications` iOS framework. Treating the framework like any other dependency that our SDK has. 

This PR is part of this work by abstracting the framework out of our SDK. Making most of our SDK code unaware that the framework even exists. The part of the SDK code that does interact with the `UserNotifications` framework will only be executed in production, not in tests.


Sources/MessagingPush/PushHandling/NotificationCenterDelegate.swift
```diff
@@ -0,0 +1,50 @@
...
+}
+
+struct PushNotification {
+    let pushId: String
+    let deliveryDate: Date
+    let title: String
+    let message: String
+    let data: [AnyHashable: Any]
+
+    init(notification: UNNotification) {
```
levibostian, line 21: Parse classes inside of `UserNotifications` framework into classes that we can create instances of.

Sources/MessagingPush/PushHandling/NotificationCenterDelegate.swift
```diff
@@ -0,0 +1,50 @@
...
+}
+
+class BaseSdkPushEventListener: NSObject, UNUserNotificationCenterDelegate, PushEventListener {
+    var delegate: UNUserNotificationCenterDelegate {
+        self
+    }
+
+    func userNotificationCenter(_ center: UNUserNotificationCenter, didReceive response: UNNotificationResponse, withCompletionHandler completionHandler: @escaping () -> Void) {
+        onPushClicked(PushNotification(notification: response.notification), completionHandler: completionHandler)
+    }
```
levibostian, line 37: This is the main focus on this fix. 

Because we cannot directly call `func userNotificationCenter(_ center: UNUserNotificationCenter, didReceive response: UNNotificationResponse, withCompletionHandler completionHandler: @escaping () -> Void) {` in our tests, we instead...
1. have our tests call `onPushClicked(PushNotification(notification: response.notification), completionHandler: completionHandler)` 
2. in production, when iOS performs a call to `func userNotificationCenter(_ center: UNUserNotificationCenter, didReceive response: UNNotificationResponse, withCompletionHandler completionHandler: @escaping () -> Void) {`, it will simply make a request to `onPushClicked(PushNotification(notification: response.notification), completionHandler: completionHandler)`, which is a function that we have covered in tests.

Sources/MessagingPush/PushHandling/iOSPushEventListener.swift
```diff
@@ -34,7 +28,7 @@ protocol PushEventListener: AutoMockable {
  - During the native iOS SDK's initialization, the SDK's digraph instance is re-created. All objects (and singletons) in that old digraph instance are deleted from memory.
  - That's bad! If the PushEventListener singleton instance was stored in the digraph, it would be deleted from memory. The OS would no longer be able to send push notification callbacks to the SDK.
  */
-class iOSPushEventListener: NSObject, PushEventListener, UNUserNotificationCenterDelegate {
+class iOSPushEventListener: BaseSdkPushEventListener {
```
levibostian, line 31: This class now no longer has knowledge of the `UserNotifications` framework. This class contains all logic to handle push click events for our SDK but is also fully testable! All by inheriting to our protocol, `PushEventListener`.

Tests/MessagingPush/PushHandling/AutomaticPushClickHandlingIntegrationTests.swift
```diff
@@ -180,20 +180,19 @@ extension AutomaticPushClickHandlingInsetegrationTest {
...
-        let hostAppDelegate = notificationCenterMock.currentDelegate
+        let hostAppDelegate: iOSPushEventListener = notificationCenterMock.currentDelegate
 
         // Then, iOS will call this function on the delegate.
         // Note: It's important that we test that the `withContentHandler` callback function gets called either by our SDK (when we handle it), or the 3rd party handler.
         //       We add an expectation to verify that 1 push click handler calls it.
         let expectCompletionHandlerCalled = expectation(description: "Expect completion handler called by a click handler")
         expectCompletionHandlerCalled.expectedFulfillmentCount = 1 // Test will fail if called 2+ times which could indicate a bug because only 1 push click handler should be calling it.
 
-        // TODO: We may need to change the syntax of this line below.
```
levibostian, line 191: This PR will solve this TODO placeholder.

Tests/MessagingPush/PushHandling/AutomaticPushClickHandlingIntegrationTests.swift
```diff
@@ -180,20 +180,19 @@ extension AutomaticPushClickHandlingInsetegrationTest {
...
         // It's difficult to create an instance of UNNotificationRequest because it's initializer is internal.
         // Therefore, we may need to call a different function then the one below. However, we will make sure that we are still testing the logic of our SDK when this function is called in production.
-        hostAppDelegate?.userNotificationCenter?(notificationCenterMock, didReceive: UNNotificationRequestMock(), withContentHandler: { _ in
+        hostAppDelegate?.pushClicked(PushNotification(pushContent)) { _ in
```
levibostian, line 193: Our tests can now test when a push is clicked 🎊!

Sources/MessagingPush/PushHandling/NotificationCenterDelegate.swift
```diff
@@ -0,0 +1,50 @@
+import Foundation
+import UserNotifications
+
+protocol PushEventListener {
```
levibostian, line 4: This protocol is our SDK's equivalent to `UNUserNotificationCenterDelegate`. By having our own version of `UNUserNotificationCenterDelegate` with functions and parameters that our tests can call, this can make our SDK's `UNUserNotificationCenterDelegate` functions testable.

Sources/MessagingPush/PushHandling/NotificationCenterFrameworkAdapter.swift
```diff
@@ -0,0 +1,87 @@
...
+// an instance of this class, first.
+//
+// This allows us to write automated tests around our SDK's push handling logic because classes inside of `UserUnotifications` internal and not mockable.
+public struct PushNotification {
+    let pushId: String
+    let deliveryDate: Date
+    let title: String
+    let message: String
+    let data: [AnyHashable: Any]
+    let rawNotification: UNNotification
```
Shahroz16, line 77: I am a little confused here, will this be `nullable`? because I thought the reason behind creating this class was (not being able to/have to) create a fake instance of `UNNotification`.

levibostian: I agree, this is confusing. I was hoping that this struct could be a true replacement to `UNNotification`. 

I added it because `CustomerIOParsedPushPayload.swift` depends on `UNNotification` instances. [This line of code](https://github.com/customerio/customerio-ios/pull/447/files#diff-8e556516d6bc8528639b392515340069010e7c0cf0499f60002cc6ef72cfe3acR137) calls `CustomerIOParsedPushPayload.parse`. So we need a `UNNotification` reference to make that `parse()` call. 

I agree, having `UNNotification` in here is not ideal. There are consequences to it. 

🤔 We are using `CustomerIOParsedPushPayload` for 2 different use cases today. Composing rich push notifications as well as parsing a push payload. I am thinking, it could be a value add to split `CustomerIOParsedPushPayload` into 2 separate classes: 1 class that just parses pushes and 1 class that composes a push. 

That could remove the need for `UNNotification`.

levibostian: This PR's focus is on the UserNotification adapter. The change ^^^ will come in a future PR that improves testability of the code. I am glad you asked about this.

Shahroz16: Glad to hear it's coming in upcoming PRs and I think that change will indeed improve this.

Shahroz16 (approved):

PR #450, opened 2023-12-24
chore: shared event bus pr2

Shahroz16:

More Details about the previous implementation can be found in the previous [PR](https://github.com/customerio/customerio-ios/pull/440)

PR 2: Holds the implementation of the Shared event bus using the notification center and actors.


Sources/Common/Communication/EventBus.swift
```diff
@@ -0,0 +1,106 @@
...
+actor SharedEventBus: EventBus {
+    private var notificationCenter: NotificationCenter = .default
+    private var observers: [String: [NSObjectProtocol]] = [:]
+
+    deinit {
+        // Clean up by removing all observers when the EventBus is deinitialized.
+        // Capture a weak reference to self to avoid retain cycles
+        Task { [weak self] in
+            await self?.removeAllObservers()
+        }
```
Ahmed-CIO, line 45: This bit challenged me a bit and I believe it won't do what you want it to do.
The problem here is that there is no guarantee that self won't be deallocated before the task is executed.
Hence, you will end up with `self == nil` before you get a chance to call `removeAllObservers`

I was thinking about not using weak; however, that's not great either as I have no idea what's the behavior of retaining `self` after `deinit` finish executing. All I can say is since `self` isn't expected to outlive `deinit` then nothing good will come out of retaining it. So we need an alternative approach here for the cleanup bit.
I will give it a bit of thoughts

Sources/Common/Communication/EventBus.swift
```diff
@@ -0,0 +1,106 @@
...
+    /// - Parameter eventType: The event type for which to remove all observers.
+    func removeObserver(for eventType: String) async {
+        if let observerList = observers[eventType] {
+            for observer in observerList {
+                notificationCenter.removeObserver(observer)
+            }
+            observers[eventType] = nil
+        }
+    }
+}
```
Ahmed-CIO, line 106: Due to the issue with removing the observers from the notification center on deinit, here is a suggestion.
In this one, we took out the observers dictionary and notification center in their own non-protected class (this need to be non-accessible outside the SharedEventBus), and we do the cleanup in its own deinit.
`EventBusObserversHolder` is not thread safe, hence it is only okay to use from the actor and should not expose it any further.

```suggestion
class EventBusObserversHolder {
  let notificationCenter: NotificationCenter = .default
  var observers: [String: [NSObjectProtocol]] = [:]
  
  func removeAllObservers() {
    observers.forEach { _, observerList in
      observerList.forEach(notificationCenter.removeObserver)
    }
    observers.removeAll()
  }
  
  deinit {
    self.removeAllObservers();
  }
}

actor SharedEventBus: EventBus {
  let holder = EventBusObserversHolder();

  init() {
    DIGraphShared.shared.logger.debug("SharedEventBus initialized")
  }
  
  /// Posts an event to all registered observers of its type.
  ///
  /// - Parameter event: The event to be posted.
  /// - Returns: True if the event has been posted to any observers, false otherwise.
  @discardableResult
  func post(_ event: AnyEventRepresentable) -> Bool {
    let key = event.key
    if let observerList = holder.observers[key], !observerList.isEmpty {
      holder.notificationCenter.post(name: NSNotification.Name(key), object: event)
      return true
    }
    return false
  }
  
  /// Registers an observer for a specific event type.
  ///
  /// - Parameters:
  ///   - eventType: The type of the event to observe.
  ///   - action: The action to be executed when the event is received.
  func addObserver(_ eventType: String, action: @escaping (AnyEventRepresentable) -> Void) {
    let observer = holder.notificationCenter.addObserver(forName: NSNotification.Name(eventType), object: nil, queue: nil) { notification in
      if let event = notification.object as? AnyEventRepresentable {
        action(event)
      }
    }
    // Store the observer reference for later management.
    if holder.observers[eventType] != nil {
      holder.observers[eventType]?.append(observer)
    } else {
      holder.observers[eventType] = [observer]
    }
  }
  
  /// Removes all observers from the EventBus.
  ///
  /// This function is used for cleanup or resetting the event handling system.
  func removeAllObservers() {
    holder.removeAllObservers();
  }
  
  /// Removes all observers for a specific event type.
  ///
  /// - Parameter eventType: The event type for which to remove all observers.
  func removeObserver(for eventType: String) {
    if let observerList = holder.observers[eventType] {
      for observer in observerList {
        holder.notificationCenter.removeObserver(observer)
      }
      holder.observers[eventType] = nil
    }
  }
}
```

Sources/Common/Communication/EventBus.swift
```diff
@@ -0,0 +1,106 @@
...
+            observers[eventType]?.append(observer)
+        } else {
+            observers[eventType] = [observer]
+        }
+    }
+
+    /// Removes all observers from the EventBus.
+    ///
+    /// This function is used for cleanup or resetting the event handling system.
+    func removeAllObservers() async {
```
Ahmed-CIO, line 88: no need to specify `async` explicitly in the actor implementation. Actor methods are all async unless you mark them as `nonisolated`

Ahmed-CIO (changes requested):
Thanks a lot @Shahroz16 , that's much cleaner and easier to review than before!
I only have one significant comment that I mentioned in regard to the `deinit`. I added a quickly made-up suggestion, but you can definitely find something better.

Once addressed, alongside anything addressable from the github checks, all should be good to merge

Sources/Common/Communication/EventBus.swift
Shahroz16: From the basic use-case testing, it did seem to deinit, but I agree there is no certainty that self will be present, but it won't crash but it wont do the job either. An alternative approach could be us not holding on to the reference of the notification center and letting components using it remove it before their `deinit`. But that brings an uncertainty that we won't know for sure if the observers have been removed/unsubscribed, because we will be at mercy of users of the events bus. 

I am thinking about an alternative too.

Sources/Common/Communication/EventBus.swift
Shahroz16: Indeed it's not needed in the implementation, but it's needed in the protocol since its a `nonisolated protocol requirement`.

If an actor now implements a protocol, automatically added method stubs will all include `async` keyword, you will need to remove them intentionally, hence left these keywords there for symmetry.

Sources/Common/Communication/EventBus.swift
Shahroz16: This can indeed be a good alternate, let me think about this.

Sources/Common/Communication/EventBus.swift
Ahmed-CIO: > But that brings an uncertainty that we won't know for sure if the observers have been removed/unsubscribed

Yes, that will leave it for chance. I think extracting a tiny piece outside the actor to do the cleanup on its own deinit is a viable approach. We just need to make sure it doesn't include any sort of logic beyond the cleanup

Sources/Common/Communication/EventBus.swift
Ahmed-CIO: it is non-ideal that auto-complete adds them for an actor conforming to the protocol. Including them have a semantic impact which shouldn't be added unless needed.

By default, all actor methods are `async` from the outside world. So if the protocol define `async`, they will be async without adding `async` in their actor counterpart.
However, internally it is different story. By default actor methods can call each other without `await`, which means your execution flow goes line by line as you would expect. However, if one method marked explicitly with `async` then even actor methods will have to `await` for it even if it is not needed. And it is cascading effect.

The only exception, is if the actor is calling an external api that's marked as async, then the actor calling function needs to be async in order to be able to await for the external api call

Shahroz16: > However, if one method marked explicitly with async then even actor methods will have to await for it even if it is not needed. And it is cascading effect.

This is a good point you raised, The actor ensures thread safety for both synchronous and asynchronous methods. The key difference is in how they are called and whether they can perform suspension points (await) and adding async would enforce us to wait for them.

Ahmed-CIO (approved):
image

PR #451, opened 2023-12-24
chore: event bus storage pr3

Shahroz16:

PR 3: Holds the implementation of the persistence storage of events. 
Why do we specifically need this class, here is the [comment](https://github.com/customerio/customerio-ios/pull/440#discussion_r1432024727) where I explained it.


Sources/Common/Communication/EventStorage.swift
```diff
@@ -0,0 +1,100 @@
...
+        self.jsonAdapter = jsonAdapter
+    }
+
+    /// Updates the base directory for event storage, primarily used for testing.
+    /// - Parameter baseDirectory: The new base directory for storing events.
+    func updateBaseDirectory(baseDirectory: URL) {
+        self.baseDirectory = baseDirectory
+    }
+
+    func store(event: AnyEventRepresentable) async throws {
```
Ahmed-CIO, line 55: nit
```suggestion
    func store(event: AnyEventRepresentable) throws {
```

Sources/Common/Communication/EventStorage.swift
```diff
@@ -0,0 +1,100 @@
...
+        do {
+            try fileManager.removeItem(at: eventFileURL)
+        } catch {
+            logger.error("Warning: Could not remove event file at \(eventFileURL). Error: \(error)")
+        }
+    }
+
+    /// Creates a directory at the specified URL if it does not already exist.
+    /// - Parameter directory: The directory URL to create.
+    private func createDirectoryIfNeeded(_ directory: URL) async throws {
```
Ahmed-CIO, line 96: nit
```suggestion
    private func createDirectoryIfNeeded(_ directory: URL) throws {
```

Sources/Common/Communication/EventStorage.swift
```diff
@@ -0,0 +1,100 @@
...
+    func store(event: AnyEventRepresentable) async throws {
+        let eventTypeDirectory = baseDirectory.appendingPathComponent(event.key)
+        try await createDirectoryIfNeeded(eventTypeDirectory)
+
+        let eventFileURL = eventTypeDirectory.appendingPathComponent("\(event.storageId).json")
+        let eventData = try jsonAdapter.encoder.encode(event)
+        try eventData.write(to: eventFileURL)
+    }
+
+    func loadEvents(ofType eventType: String) async throws -> [AnyEventRepresentable] {
```
Ahmed-CIO, line 64:
```suggestion
    func loadEvents(ofType eventType: String) throws -> [AnyEventRepresentable] {
```

Sources/Common/Communication/EventStorage.swift
```diff
@@ -0,0 +1,100 @@
...
+                events.append(event)
+            } catch {
+                logger.error("Warning: Could not decode event at \(url). Error: \(error)")
+            }
+        }
+        return events
+    }
+
+    // Removes a specific event file
+    func remove(ofType eventType: String, withStorageId storageId: String) async {
```
Ahmed-CIO, line 85:
```suggestion
    func remove(ofType eventType: String, withStorageId storageId: String) {
```

Ahmed-CIO (approved):
all looks good. Please address these comments before merging:
* remove `async` from the actor method declaration. By default all actor methods won't be called from outside the actor context without `await`ing them.
* address the GH checks

image

Sources/Common/Communication/EventStorage.swift
Shahroz16: since this method is performing file operation, i think it should be async

Sources/Common/Communication/Event.swift
```diff
@@ -49,6 +49,29 @@ public enum EventTypesRegistry {
...
+        case RegisterDeviceTokenEvent.key:
+            return RegisterDeviceTokenEvent.self
+        case DeleteDeviceTokenEvent.key:
+            return DeleteDeviceTokenEvent.self
+        case NewSubscriptionEvent.key:
+            return NewSubscriptionEvent.self
+        default:
+            throw EventBusError.invalidEventType
+        }
+    }
```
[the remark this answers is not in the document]
Shahroz16: It absolutely would be 😄, I think I started with it being an enum but pivoted because of the challenges it brought with inheritance, being codable and few other challenges

Shahroz16:
We decided to go with the actor method without async for the first iteration and after performance evaluation we can add it back if needed.

PR #452, opened 2023-12-24
chore: event bus handler pr4

Shahroz16:

PR 4: Holds the implementation of the event bus handler, the interface exposed and used to all components of SDKs that would interact with Eventbus.


Sources/Common/Communication/EventBusHandler.swift
```diff
@@ -0,0 +1,113 @@
...
+    /// Replays events of a specific type to any new observers, ensuring they receive past events.
+    /// - Parameter eventType: The event type for which to replay events.
+    private func replayEvents<E: EventRepresentable>(forType eventType: E.Type) async {
+        let key = eventType.key
+        let storedEvents = await memoryStorage.eventsForKey(key)
+
+        for event in storedEvents {
+            if let specificEvent = event as? E {
+                logger.debug("EventBusHandler: Replaying event type - \(specificEvent)")
+                let isSent = await eventBus.post(specificEvent)
```
Ahmed-CIO, line 76: how do we ensure the event is sent only to the new observers?
I mean if there is an already existing observer, and then another one gets added, it seems we will call them both. Which will be duplicate call for the existing observer

Sources/Common/Communication/EventBusHandler.swift
```diff
@@ -0,0 +1,113 @@
...
+            try await eventStorage.store(event: event)
+        } catch {
+            logger.debug("Error storing event: \(error)")
+        }
+    }
+
+    /// Removes an event from persistent storage.
+    /// - Parameter event: The event to remove.
+    public func removeFromStorage<E: EventRepresentable>(_ event: E) async {
+        await eventStorage.remove(ofType: event.key, withStorageId: event.storageId)
```
Ahmed-CIO, line 111: when do we clean up the in memory storage? so far I see we are only appending to it. Which at one point will need a clean up

Ahmed-CIO (changes requested):
Review summary:
* The reply functionality needs clarity around the intended semantics for which observer receives which events.
* In memory storage cleanup mechanism is missing
* Lastly, let's make sure we understand the reasoning behind chosing to mark a method as `async` or use `Task` inside its implementation

Sources/Common/Communication/EventBusHandler.swift
```diff
@@ -0,0 +1,113 @@
...
+                action(specificEvent)
+            } else {
+                self.logger.debug("Error: Event type did not match")
+            }
+        }
+
+        Task {
+            await eventBus.addObserver(eventType.key, action: adaptedAction)
+            await replayEvents(forType: eventType)
+        }
```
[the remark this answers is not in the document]
Shahroz16: Doing so, `addObserver` remains synchronous from the caller's perspective. The async work is managed internally without requiring the caller to handle it or even be aware of it. Makes the public API easier to adapt, every module subscribing won't have to add `Task` on their end. 

Since registering an observer is a setup process that doesn't need to be tightly synchronized with other operations, using a Task made sense. For instance, our modules simply need to start listening for events and don’t have subsequent actions that depend on the successful registration hence a fire-and-forget approach. 

It also ensures consistency of experience for all modules subscribing. Similar to what NotificationCenter API offers.

Sources/Common/Communication/EventBusHandler.swift
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    /// Removes an observer for a specific event type.
+    /// - Parameter eventType: The event type for which to remove the observer.
+    public func removeObserver<E: EventRepresentable>(for eventType: E.Type) {
+        Task { await eventBus.removeObserver(for: E.key) }
+    }
```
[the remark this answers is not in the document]
Shahroz16: same response as above

Sources/Common/Communication/EventBusHandler.swift
```diff
@@ -0,0 +1,113 @@
...
+
+    /// Replays events of a specific type to any new observers, ensuring they receive past events.
+    /// - Parameter eventType: The event type for which to replay events.
+    private func replayEvents<E: EventRepresentable>(forType eventType: E.Type) async {
+        let key = eventType.key
+        let storedEvents = await memoryStorage.eventsForKey(key)
```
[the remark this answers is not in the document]
Shahroz16: Currently no there isn't a way and we probably don't need it right now either. But can definitely add it in future iteration.

There are two kind of events components subscribe to

- Internal events, for example when `DataPipeline` is processing an event called screen, we need to inform `InApp` module as well to set its current `page/screen` accordingly 

- Events for DataPipeline, when push/in-app wants a metric event to be tracked, they publish it and the subscriber is `DataPipeline` which when recieve these events adds a timestamp to it and then send it. So old/new all events it recieves it timestamp them. So a campaign that is time based would get an event either way.

Sources/Common/Communication/EventBusHandler.swift
Shahroz16: This is a good question, i do have the implementation for it but didn't add it in the first iteration to keep the implementation easier to understand. Its will be added in later PR :D

Sources/Common/Communication/EventBusHandler.swift
Shahroz16: I want to think we don't clean up the memory, it would be cleaned on SDK/App restart, and we could just limit the number of items that get added to the memory. 

Even though there is only a limited number of events that gets sent via eventbus in SDK lifecycle, if an anomaly happen and users starts bombarding with those events we only keep X amount of latest events to be replayed to new consumers.

Ahmed-CIO (approved):
Given the complexity of addressing some of the comments, we agreed to proceed with this PR in a follow up PRs we will address:
* Validate whether the reply functionality will actually work as intended cross an app and its extension
* Ensuring observers know the real time the event was posted (i.e timestamp the events)
* Ensuring observers don't receive notifications more times than the times the event was actually posted
* Limit the in-memory storage by capping how many events we can store in it

Shahroz16:
Follow up [PR](https://github.com/customerio/customerio-ios/pull/455) and [task](https://github.com/customerio/issues/issues/11953) created.

PR #453, opened 2023-12-24
chore: event bus usage pr5

Shahroz16:

PR 5: Holds the usage of the event bus handler and its working.

### DIFF
--- a/Sources/Common/autogenerated/AutoDependencyInjection.generated.swift
+++ b/Sources/Common/autogenerated/AutoDependencyInjection.generated.swift
@@ -531,13 +531,25 @@ extension DIGraphShared {
         EventBusHandler(eventBus: eventBus, eventStorage: eventStorage, logger: logger)
     }
 
-    // EventStorage
+    // EventStorage (singleton)
     var eventStorage: EventStorage {
         getOverriddenInstance() ??
-            newEventStorage
+            sharedEventStorage
     }
 
-    private var newEventStorage: EventStorage {
+    var sharedEventStorage: EventStorage {
+        DispatchQueue(label: "DIGraphShared_EventStorage_singleton_access").sync {
+            if let overridenDep: EventStorage = getOverriddenInstance() {
+                return overridenDep
+            }
+            let existingSingletonInstance = self.singletons[String(describing: EventStorage.self)] as? EventStorage
+            let instance = existingSingletonInstance ?? _get_eventStorage()
+            self.singletons[String(describing: EventStorage.self)] = instance
+            return instance
+        }
+    }
+
+    private func _get_eventStorage() -> EventStorage {
         EventStorageManager(logger: logger, jsonAdapter: jsonAdapter)
     }
 
--- a/Sources/Common/autogenerated/AutoDependencyInjection.generated.swift
+++ b/Sources/Common/autogenerated/AutoDependencyInjection.generated.swift
@@ -509,13 +509,25 @@ extension DIGraphShared {
         DeviceMetricsGrabberImpl()
     }
 
-    // EventBusHandler
+    // EventBusHandler (singleton)
     public var eventBusHandler: EventBusHandler {
         getOverriddenInstance() ??
-            newEventBusHandler
+            sharedEventBusHandler
     }
 
-    private var newEventBusHandler: EventBusHandler {
+    public var sharedEventBusHandler: EventBusHandler {
+        DispatchQueue(label: "DIGraphShared_EventBusHandler_singleton_access").sync {
+            if let overridenDep: EventBusHandler = getOverriddenInstance() {
+                return overridenDep
+            }
+            let existingSingletonInstance = self.singletons[String(describing: EventBusHandler.self)] as? EventBusHandler
+            let instance = existingSingletonInstance ?? _get_eventBusHandler()
+            self.singletons[String(describing: EventBusHandler.self)] = instance
+            return instance
+        }
+    }
+
+    private func _get_eventBusHandler() -> EventBusHandler {
         EventBusHandler(eventBus: eventBus, eventStorage: eventStorage, logger: logger)
     }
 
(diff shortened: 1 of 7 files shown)


Sources/DataPipeline/DataPipelineImplementation.swift
```diff
@@ -36,6 +38,26 @@ class DataPipelineImplementation: DataPipelineInstance {
             // if the device token exists, pass it to the plugin and ensure device attributes are updated
             addDeviceAttributes(token: existingDeviceToken)
         }
+
+        // plugin to publish data pipeline events
+        analytics.add(plugin: DataPipelinePublishedEvents(diGraph: diGraph))
+
+        // subscribe to journey events
+        subscribeToJourneyEvents()
```
levibostian, line 46: I don't find comments like this valuable in the code. From reading the lines of code, I can see that a plugin is being added and journeys events are being subscribed to. 

If a comment is added, I would find them valuable if the comments mention why we are adding a plugin and why we are subscribing to journeys events.

Sources/DataPipeline/plugins/DataPipelinePublishedEvents.swift
```diff
@@ -0,0 +1,35 @@
+import CioInternalCommon
+import Foundation
+import Segment
+
+/// This class serves as a centralized hub for dispatching events generated from DataPipeline module.
+/// It is designed to reduce redundancy and enhance maintainability by funneling these operations through a single plugin.
+/// This plugin allows for decoupled modules to subscribe and react to these events.
+class DataPipelinePublishedEvents: EventPlugin {
```
levibostian, line 8: I do not understand why this class is needed? 

From my knowledge of segment plugins, here is an example flow for how I believe this plugin would work: 

1. A profile is identified in the Segment SDK. 
2. Segment SDK calls this plugin's `identify(event: IdentifyEvent)` function
3. EventBus posts new `ProfileIdentifiedEvent` event. <--- why do this need to happen? 

My biggest question is: Why do we need our EventBus to post a new `ProfileIdentifiedEvent` event?

I was expecting that once a profile is identified in the Segment SDK, our EventBus should have already posted an `ProfileIdentifiedEvent` event. 

I was expecting a flow like this: 

1. Profile identified in the CIO SDK. 
2. From the CIO SDK, EventBus posts new `ProfileIdentifiedEvent` event. 
3. A profile is identified in the Segment SDK. 

From the flow above, I don't see a need for this plugin. 

---

If there is a use case that I am not aware of that this plugin solves, I would appreciate editing the comment in this class explaining why it's needed.

levibostian (changes requested):
PR looks good. I do have 1 blocker: https://github.com/customerio/customerio-ios/pull/453/files#r1440665476 that I want to better understand before approval.

Sources/DataPipeline/plugins/DataPipelinePublishedEvents.swift
Shahroz16: I am not sure if i fully understand the question, but I would like to explain the flow to see if it helps resolve any confusion.

Before I respond, I would like to point to one aspect of plugins in the Segment. 

> Plugins are run through a timeline, which executes in order of insertion based on their types

So the very first line, explain at what point this plugin would execute  
`var type: Segment.PluginType = .before`

> Why do we need our EventBus to post a new ProfileIdentifiedEvent event?

I tried to answer this with this comment

>/// This plugin allows for decoupled modules to subscribe and react to these events.

When a profile is identified, we want to let `InApp` module know what `user id` to set, similarly screen, and the same goes for reset. Similarly, Push can subscribe to its events. 

>I do not understand why this class is needed.

I tried to answer that with
> /// This class serves as a centralized hub for dispatching events generated from DataPipeline module.
/// It is designed to reduce redundancy and enhance maintainability by funneling these operations through a single plugin.

We have multiple `identify` methods, some accept `identifier` some `user id`, similarly some would accept, `map` or `codable`. Similarly, we would have multiple screen methods and the same for others. 

Instead of posting in the `event bus` from all of those methods or trying to create a common method you then call from all of those methods. This plugin offers a centralised way of handling it. 

No matter if you call,
`CustomerIO.identify(idenfity) or CustomerIO.identify(userId)` etc, it will end up at this plugin which will then publish that event for any other module that might be subscribing to it.

Sources/Common/autogenerated/AutoDependencyInjection.generated.swift
```diff
@@ -509,23 +509,47 @@ extension DIGraphShared {
...
+    // EventBusHandler (singleton)
     public var eventBusHandler: EventBusHandler {
         getOverriddenInstance() ??
-            newEventBusHandler
+            sharedEventBusHandler
     }
 
-    private var newEventBusHandler: EventBusHandler {
+    public var sharedEventBusHandler: EventBusHandler {
+        DispatchQueue(label: "DIGraphShared_EventBusHandler_singleton_access").sync {
```
Ahmed-CIO, line 519: I was going to comment on this part and totally forgot it is autogenerated -_-

Sources/DataPipeline/plugins/DataPipelinePublishedEvents.swift
Ahmed-CIO: looking at the code and this conversation, I believe this conversation doesn't immediately result in anything actionable in the code itself. Maybe a chat followed by a conclusion or a documentation can help instead of having this conversation over the PR and going through longer feedback loop

Sources/DataPipeline/DataPipelineImplementation.swift
Ahmed-CIO: +1, please clean up the comments before merging

Ahmed-CIO (approved):
I am unblocking this one on the basis that:
- The code itself looks good to go.
- The open questions can be addressed without blocking the PR

If addressing the questions resulted in a need to change anything, a follow up PR will be published to address these comments

PR #454, opened 2023-12-24
chore: event bus unit tests pr6

Shahroz16:

PR 6: Holds the automated test for storage, event bus, and handlers. Also, the changes that were required to make code more testable. 

General instructions: 
- `EventBusTest` holds tests for the EventBus class. 
- `EventStorageTest` holds tests for Event storage in file persistence
- `EventBusHandler` has mocks of `EventBus` and `EventStorage` so it relies on mocks. 

The flow of operations:
- When  we are trying to test the `posting` of an event, we are reliant on `postClosure` of mockedEventBus
- Replay doesn't use event bus, but we rather just trigger the callbacks, so it can easily be tested by that.


levibostian (changes requested):
I am having a difficult time reviewing the PR. 

I would be better able to review it if it was broken down into smaller, more focused PRs.

Sources/Common/Communication/EventStorage.swift
```diff
@@ -79,7 +79,8 @@ actor EventStorageManager: EventStorage {
                 logger.error("Warning: Could not decode event at \(url). Error: \(error)")
             }
         }
-        return events
+        // Sort events based on their storageId (which includes a timestamp).
```
Ahmed-CIO, line 82: this comment does help big time. I still did not go to the tests yet, but usually I pick this things by looking at the tests too

Tests/Common/Communication/EventBusHandlerTests.swift
```diff
@@ -0,0 +1,565 @@
...
+
+        // Expectations
+        let firstObserverReceivedExpectation = XCTestExpectation(description: "First observer should receive event only once")
+        firstObserverReceivedExpectation.expectedFulfillmentCount = 1
+        let secondObserverReceivedExpectation = XCTestExpectation(description: "Second observer should receive event only once")
+        secondObserverReceivedExpectation.expectedFulfillmentCount = 1
+
+        mockEventBus.postClosure = {
+            _ in
+            firstObserverReceivedExpectation.fulfill()
```
Ahmed-CIO, line 154: why is this fulfilled here?

Tests/Common/Communication/EventBusHandlerTests.swift
```diff
@@ -0,0 +1,565 @@
...
+            // it will be recieved via notification center hence the mocks
+        }
+
+        // Post the event
+        eventBusHandler.postEvent(event)
+
+        // Register second observer
+        eventBusHandler.addObserver(RegisterDeviceTokenEvent.self) { _ in
+            // Second observer action, reply is supposed to happen
+            secondObserverReceivedExpectation.fulfill()
```
Ahmed-CIO, line 169: why is this handled differently than the `addObserver` above?

Tests/Tracking/CustomerIOImplementation+ScreenViewsTest.swift
Ahmed-CIO, line 1: let's either update the tests probably in these files or delete the files altogether. Leaving files with commented out code isn't beautiful

Tests/Common/Util/RingBufferTest.swift
```diff
@@ -0,0 +1,59 @@
...
+        XCTAssertEqual(ringBuffer.dequeue(), 1, "Dequeued element should be 1")
+        XCTAssertEqual(ringBuffer.dequeue(), 2, "Dequeued element should be 2")
+        XCTAssertNil(ringBuffer.dequeue(), "Buffer should be empty")
+    }
+
+    func test_EnqueueBeyondCapacity_ShouldOverwriteOldestElement() {
+        var ringBuffer = RingBuffer<Int>(capacity: 2)
+        ringBuffer.enqueue(1)
+        ringBuffer.enqueue(2)
+        ringBuffer.enqueue(3) // This should overwrite the first element (1)
```
Ahmed-CIO, line 21: the comment confused me but the test assertion below matched what I would expect from a queue

Ahmed-CIO (changes requested):
* Supper pleased to see the efforts in the tests
* The PR title is event bus unit tests, so confused to see changes like adding the `getStorageID` in this PR. Let's keep it in its PR to discuss the rational behind it
* Test classes can be stacked separately and reviewed separately, let's do this please
* Let's put the deleted tests in their own PR as well, if they are general cleanup, or in the PR where we put their replacement if they happen to be replaced by one of the new tests

Tests/Common/Util/RingBufferTest.swift
Shahroz16: whats its saying is since the `limit is 2`, the last element when `enqued` should kick the 1st element out, but i understand what you mean.

Tests/Common/Communication/EventBusHandlerTests.swift
Shahroz16: since the real posting of events is done by the Notification center, the best we can do is make sure is `event bus` is calling `post` method.

Tests/Common/Communication/EventBusHandlerTests.swift
Shahroz16: as discussed earlier, since replay is not a feature of Notification center, so we don't utilize that for replay, we just trigger the lambda passed. 

So the mock won't be called because eventbus won't be posting the event but it would be rather us triggering the callback on lambda passed.

Shahroz16:
@Ahmed-CIO @levibostian 

Thank you folks for the initial pass on it, I tried to make it lighter. Now the PR has the following files

4 Test classes:
- EventBus tests 
- EventBus storage test
- EventBusHandler test
- RingBuffer test

Smaller changes:
- Some changes like adding `AutoMockable` to the interface of these classes to mock them in test or exposing new public method as read-only to validate tests. 
- Refactoring of internal method names
 
The reason I decided to have all of them together is to give a reviewer reference points, for example, while reviewing `EvenbusHandler` a review can say, well since a method is heavily operated under mocks wonder the `EventBus` that is mocked here internally have a test to verify that, so they can jump to `EventBusTest` class and verify that. 

For example you want to verify the order of events, you can check both `EventBusTest` and `EventBusStorage` to verify both have tests that check for the order of events. 

If it is still easier to review separate PRs for each test class, I am happy to do that.

Shahroz16:
closing as the broken down PRs are merged

PR #458, opened 2024-01-02
fix: app crash on identify method

ami-aman:
This PR fixes the following scenario: 
Sample app was crashing in a scenario where the user sends `nil` value for body in `identify` method. The specific use case is when a user clicks `Generate Random Login` and attempts to login without a body. 

### DIFF
--- a/Sources/Common/CustomerIOInstance.swift
+++ b/Sources/Common/CustomerIOInstance.swift
@@ -232,6 +232,18 @@ public class CustomerIO: CustomerIOInstance {
         identifier: String,
         body: RequestBody
     ) {
+        // This code handles a specific scenario where a user
+        // unexpectedly provides a `nil` value for the RequestBody
+        // parameter, which the method isn't designed to handle directly.
+        // To prevent errors and ensure compatibility, this condition checks
+        // if the `body` is `nil`. If it is, it substitutes it with an
+        // EmptyRequestBody() object before passing it along to the
+        // identify(). This replacement safeguards against possible
+        // issues that could arise(eg. app crash). While the `identify`
+        // method itself provide polymorphic capabilities to handle `nil`,
+        // this specific check within this method offers an additional
+        // layer of protection and clarity for this case.
+        // (refer https://github.com/customerio/customerio-ios/blob/94cbf686c3c2a405534cfe908f7166558a3e0b5d/Sources/Tracking/CustomerIO.swift#L71-L75)
         if let optionalValue = body as? Any?, optionalValue == nil {
             implementation?.identify(identifier: identifier, body: EmptyRequestBody())
             return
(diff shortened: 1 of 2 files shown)


Sources/Common/CustomerIOInstance.swift
```diff
@@ -232,6 +232,22 @@ public class CustomerIO: CustomerIOInstance {
...
+        // EmptyRequestBody() object before passing it along to the
+        // identify(). This replacement safeguards against possible
+        // issues that could arise(eg. app crash). While the `identify`
+        // method itself provide polymorphic capabilities to handle `nil`,
+        // this specific check within this method offers an additional
+        // layer of protection and clarity for this case.
+        // (refer https://github.com/customerio/customerio-ios/blob/94cbf686c3c2a405534cfe908f7166558a3e0b5d/Sources/Tracking/CustomerIO.swift#L71-L75)
+        if let optionalValue = body as? Any?, optionalValue == nil {
+            implementation?.identify(identifier: identifier, body: EmptyRequestBody())
+            return
```
Shahroz16, line 249: personal preference opinion only,

if and else, over early return.

Shahroz16 (approved):

Sources/Common/CustomerIOInstance.swift
mrehan27: Personal preference again but I would prefer guard which I think is more common in Swift and our codebase 😄

mrehan27 (approved):

ami-aman:
@mrehan27 @Shahroz16 directly converting a non-optional property to optional isn't possible in Swift. This is one of the ways we can indirectly attain what's said above. I do agree that there could be personal preferences but swift doesn't allow the conversion and hence the alternative approach has been followed here for comparison of a non-nil property with `nil`.

Sources/Common/CustomerIOInstance.swift
ami-aman: [Response](https://github.com/customerio/customerio-ios/pull/458#issuecomment-1875471780)

Shahroz16: i meant this, think this would have worked. 
```
if let optionalValue = body as? Any?, optionalValue == nil {
            implementation?.identify(identifier: identifier, body: EmptyRequestBody())
        } else {
            implementation?.identify(identifier: identifier, body: body)
        }
```

ami-aman:
> I would like to request automated tests for this fix. Was there a reason for not adding them in the PR I am unaware of?

We are updating the test cases separately as a part of clean up.

Sources/Common/CustomerIOInstance.swift
Shahroz16: No it doesn't and hence, didn't say so either. 😄 

I was just responding to this reply, that the personal preference I was mentioning was utilizing the same logic for comparison. We choose to go with it or not thats another thing. 

>I do agree that there could be personal preferences but swift doesn't allow the conversion and hence the alternative approach has been followed here for comparison of a non-nil property with nil.

PR #459, opened 2024-01-02
test: fix device attribute provider test cases

ami-aman:

### DIFF
--- a/Tests/Tracking/Util/DeviceAttributesProviderTest.swift
+++ b/Tests/Tracking/Util/DeviceAttributesProviderTest.swift
@@ -1,3 +1,4 @@
+@testable import CioDataPipelines
 @testable import CioInternalCommon
 @testable import CioTracking
 import Foundation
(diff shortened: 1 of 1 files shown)


Shahroz16 (approved):
Looks good, since we are not updating device attributes differently, is that going to be a separate PR?

Tests/Tracking/Util/DeviceAttributesProviderTest.swift
```diff
@@ -55,23 +56,6 @@ class DeviceAttributesProviderTest: UnitTest {
 
         waitForExpectations()
     }
-
-    func test_getSdkVersionAttribute_givenNotUsingSdkWrapper_expectGetSDKVersion() {
```
mrehan27, line 59: Shouldn't we still have test to verify correct SDK version being attached? Or do we plan add more tests for `DeviceAttributesProvider` in later PR?

ami-aman: I would add the test cases in another PR. This PR is just a fix to the broken test cases that was being a blocker to run other tests cases.

ami-aman:
@Shahroz16 
> since we are not updating device attributes differently, is that going to be a separate PR
Apologies but I think I am not sure what exactly you are referring to, to be done in a separate PR? 🤔

Shahroz16:
> @Shahroz16
> 
> > since we are ~~not~~ now updating device attributes differently, is that going to be a separate PR
> 
> Apologies but I think I am not sure what exactly you are referring to, to be done in a separate PR? 🤔

I meant do you plan to add tests for Device attributes? since now we relying on Datapipeline for device attributes, but i talked offline with Rehan about it, we will setup a framework for it first.

PR #462, opened 2024-01-04
test: ringbuffer for event memory cache

Shahroz16:
Broken down PR for https://github.com/customerio/customerio-ios/pull/454/files#top

Please go through the details in the parent ticket for further context.

### DIFF
--- a/Sources/Common/Util/RingBuffer.swift
+++ b/Sources/Common/Util/RingBuffer.swift
@@ -12,6 +12,11 @@ struct RingBuffer<T> {
     // Flag to indicate if the buffer is full.
     private var isFull: Bool = false
 
+    // Public read-only accessor for isFul
+    public func isBufferFull() -> Bool {
+        isFull
+    }
+
     /// Initializes the ring buffer with a specified capacity.
     /// - Parameter capacity: The maximum number of elements the buffer can hold.
     init(capacity: Int) {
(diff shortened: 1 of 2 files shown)


Tests/Common/Util/RingBufferTest.swift
```diff
@@ -0,0 +1,59 @@
...
+        var ringBuffer = RingBuffer<Int>(capacity: 3)
+        ringBuffer.enqueue(1)
+        ringBuffer.enqueue(2)
+
+        XCTAssertEqual(ringBuffer.dequeue(), 1, "Dequeued element should be 1")
+        XCTAssertEqual(ringBuffer.dequeue(), 2, "Dequeued element should be 2")
+        XCTAssertNil(ringBuffer.dequeue(), "Buffer should be empty")
+    }
+
+    func test_EnqueueBeyondCapacity_ShouldPushOldestElement() {
```
ami-aman, line 17:
```suggestion
    func test_EnqueueBeyondCapacity_ShouldRemoveOldestElement() {
```
The method name is messing with my brain. The name provides an impression that it should "push the oldest element" which makes me think that the oldest element should be pushed out as output when dequeued but the method removes the first element and pops out in FIFO order when dequeued. 🤔

ami-aman (approved):

Sources/Common/Util/RingBuffer.swift
```diff
@@ -12,6 +12,11 @@ struct RingBuffer<T> {
     // Flag to indicate if the buffer is full.
     private var isFull: Bool = false
 
+    // Public read-only accessor for isFul
```
ami-aman, line 15:
```suggestion
    // Public read-only accessor for isFull
```

ami-aman (approved):

PR #465, opened 2024-01-04
chore: data pipeline tests base setup

mrehan27:

###  Changes

- Setting up tests environment for `DataPipeline`
- Updated `DataPipeline` so it can work with fake instances
- Added support methods to make testing `DataPipeline` easier
- Changed the name from `CustomerIOImplementationTest` to `DataPipelineInteractionTests` so it is easier to track updated tests

_**Note:** This is the main PR where we'll add more test changes later on_

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -81,6 +81,9 @@ let package = Package(
         .target(name: "CioDataPipelines",
                 dependencies: ["CioInternalCommon", .product(name: "Segment", package: "Segment")],
                 path: "Sources/DataPipeline"),
+        .testTarget(name: "DataPipelineTests",
+                    dependencies: ["SharedTests"],
+                    path: "Tests/DataPipeline"),
 
         // APN
         .target(name: "CioMessagingPushAPN",
--- a/Tests/DataPipeline/Support/TestUtilities.swift
+++ b/Tests/DataPipeline/Support/TestUtilities.swift
@@ -1,0 +1,46 @@
+import Foundation
+@testable import Segment
+import XCTest
+
+// MARK: - Helper Plugins
+
+class OutputReaderPlugin: Plugin {
+    let type: PluginType
+    var analytics: Analytics?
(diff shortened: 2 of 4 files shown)


Package.swift
```diff
@@ -81,6 +81,9 @@ let package = Package(
         .target(name: "CioDataPipelines",
                 dependencies: ["CioInternalCommon", .product(name: "Segment", package: "Segment")],
                 path: "Sources/DataPipeline"),
+        .testTarget(name: "DataPipelineTests",
```
ami-aman, line 84: Wondering if we name it as `CioDataPipelineTests` to maintain consistency ? Just thinking out loud !

Tests/DataPipeline/Support/TestUtilities.swift
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    func execute<T>(event: T?) -> T? where T: RawEvent {
+        lastEvent = event
+        var eventSummary = ""
+
+        if let event = event {
+            eventSummary += " Event type: \(event.type ?? "nil")"
+            events.append(event)
+            if let event = event as? TrackEvent {
```
Shahroz16, line 25: We just want to print events of type `TrackEvent`?

Shahroz16 (approved):

Package.swift
mrehan27: None of our test package has `Cio` prefix. It is only for main packages, probably to keep them unique for public apis. So no changes needed here.

Tests/DataPipeline/Support/TestUtilities.swift
mrehan27: Just printed event name, which is available in `TrackEvent` only. For rest of the events e.g. `identify`, `type` is generally helpful.

ami-aman (approved):

PR #467, opened 2024-01-05
test: event bus storage 

Shahroz16:
Broken down PR for https://github.com/customerio/customerio-ios/pull/454/files#top

Please go through the details in the parent ticket for further context.


Tests/Common/Communication/EventStorageTests.swift
```diff
@@ -0,0 +1,165 @@
...
+
+class EventStorageTest: UnitTest {
+    var eventStorageManager: EventStorageManager!
+
+    override func setUp() {
+        super.setUp()
+        if let tempDirectory = createTemporaryDirectory() {
+            eventStorageManager = EventStorageManager(logger: log, jsonAdapter: jsonAdapter)
+            Task {
+                await eventStorageManager.updateBaseDirectory(baseDirectory: tempDirectory)
```
Ahmed-CIO, line 13: A better option would be to make the setup function async: `override func setUp() async`

Tests/Common/Communication/EventStorageTests.swift
```diff
@@ -0,0 +1,165 @@
...
+    // MARK: - Event Storage Tests
+
+    func test_storeEvent_givenValidEvent_expectFileExists() async throws {
+        let event = ProfileIdentifiedEvent(identifier: "testID")
+        try await eventStorageManager.store(event: event)
+
+        let eventFileURL = await eventStorageManager.baseDirectory
+            .appendingPathComponent(ProfileIdentifiedEvent.key)
+            .appendingPathComponent("\(event.storageId).json")
+        let fileExists = FileManager.default.fileExists(atPath: eventFileURL.path)
```
Ahmed-CIO, line 46: while I understand what we are validating, this is a very fragile test as it would break even if the storage behavior is not breaking. i.e if the internal implementation changes which path it stores the data, the users of the storage won't be impacted, yet the test will break. Can we do something to avoid needed to replicate or know the internal implementation in order to validate the behavior?

Edit: one idea is to have multiple instances of the event storage, and see if writing using one instance and reading the value through the other instance will give you the correct data? this way you validated the persistency, but not the internal details

Tests/Common/Communication/EventStorageTests.swift
```diff
@@ -0,0 +1,165 @@
...
+        XCTAssertTrue(fileExists, "Event file should exist after storing")
+    }
+
+    func test_storeEvent_givenNewEventType_expectDirectoryCreated() async throws {
+        let event = ResetEvent() // Assume this is a new event type
+        try await eventStorageManager.store(event: event)
+
+        let eventTypeDirectory = await eventStorageManager.baseDirectory.appendingPathComponent(ResetEvent.key)
+        let directoryExists = FileManager.default.fileExists(atPath: eventTypeDirectory.path)
```
Ahmed-CIO, line 55: same as above.

Tests/Common/Communication/EventStorageTests.swift
```diff
@@ -0,0 +1,165 @@
...
+
+    func test_loadEvents_givenCorruptDataInFile_expectValidEventsLoaded() async throws {
+        // Ensure the directory for the event type exists
+        let eventTypeDirectory = await eventStorageManager.baseDirectory.appendingPathComponent(ProfileIdentifiedEvent.key)
+        try FileManager.default.createDirectory(at: eventTypeDirectory, withIntermediateDirectories: true)
+
+        // Store a corrupt event file manually
+        let corruptData = Data("corrupt data".utf8)
+        let corruptFileURL = eventTypeDirectory.appendingPathComponent("corruptEvent.json")
+        try corruptData.write(to: corruptFileURL)
```
Ahmed-CIO, line 89: trying to think of ways to avoid "knowing" the path in order to test. One idea maybe a custom event that when "encoded" serializes bad data. Did not validate this idea though, but something worth to be done for the same reason as mentioned in the few tests above

Tests/Common/Communication/EventStorageTests.swift
```diff
@@ -0,0 +1,165 @@
...
+    // MARK: - Event Removal Tests
+
+    func test_removeEvent_givenStoredEvent_expectEventRemoved() async throws {
+        let event = ProfileIdentifiedEvent(identifier: "testID")
+        try await eventStorageManager.store(event: event)
+
+        let eventFileURL = await eventStorageManager.baseDirectory
+            .appendingPathComponent(ProfileIdentifiedEvent.key)
+            .appendingPathComponent("\(event.storageId).json")
+        await eventStorageManager.remove(ofType: ProfileIdentifiedEvent.key, withStorageId: event.storageId)
```
Ahmed-CIO, line 110: same

Tests/Common/Communication/EventStorageTests.swift
```diff
@@ -0,0 +1,165 @@
...
+        await withThrowingTaskGroup(of: Void.self) { group in
+            for event in events {
+                group.addTask {
+                    try await self.eventStorageManager.store(event: event)
+                }
+            }
+        }
+
+        // Load events and verify
+        let loadedEvents = try await eventStorageManager.loadEvents(ofType: ProfileIdentifiedEvent.key)
```
Ahmed-CIO, line 139: on the surface that might look like useless test. However, it is right now because we are using actors. If for whatever reason that change, this test can add value.

The reality though, you can't deterministically validate thread safety. In their definition their produce inconsistent behavior :s

Ahmed-CIO (approved):
While I have observation on accessing implementation details in a test.
I won't block this PR as it has been idle for several days.

But I would request having a follow up task (doesn't have to be done immediately), to reflect the review comments here so we have it under our radar and pick it up when we do cleanup activities

Tests/Common/Communication/EventStorageTests.swift
Shahroz16: I am not sure I fully get it, we are overriding the 
```
/*!
 * @method -setUp
 * Setup method called before the invocation of each test method in the class.
 */
- (void)setUp;
```
from XCAbstractTest, there isn't an async one that we can use. We can probably create another async function, and then call that from this method inside `Task`, but I am not sure how if that would be better or same.

Tests/Common/Communication/EventStorageTests.swift
Shahroz16: that's a fair point, updating the test

Tests/Common/Communication/EventStorageTests.swift
Shahroz16: good idea, I'll incorporate in later PR

PR #468, opened 2024-01-08
fix: ensure messages are only shown once

stephen-pope-customer-io:
The idea is to keep track of the shown messages by queueId (which is unique per message) and skip the shown ones

This is based on the same solution applied for gist-web: https://github.com/customerio/gist-web/pull/26

Task: https://github.com/customerio/issues/issues/11967

- [x] [Assign members of your team](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/requesting-a-pull-request-review) to review the pull request. 
- [x] Wait for pull request status checks to complete. If there are problems, fix them until you see that [all status checks are passing](https://external-content.duckduckgo.com/iu/?u=https%3A%2F%2Fsymfony.com%2Fdoc%2F4.3%2F_images%2Fdocs-pull-request-symfonycloud.png&f=1&nofb=1). 

### DIFF
--- a/Sources/MessagingInApp/Gist/Gist.swift
+++ b/Sources/MessagingInApp/Gist/Gist.swift
@@ -94,8 +94,7 @@ public class Gist: GistDelegate {
         let queueId = message.queueId
 
         // Skip shown messages
-        if queueId != nil &&
-            shownMessageQueueIds.contains(queueId!) {
+        if queueId != nil && shownMessageQueueIds.contains(queueId!) {
             Logger.instance.info(message: "Message \(queueId!) already shown, skipping.")
             return
         }
--- a/Sources/MessagingInApp/Gist/Gist.swift
+++ b/Sources/MessagingInApp/Gist/Gist.swift
@@ -57,22 +57,12 @@ public class Gist: GistDelegate {
     // MARK: Message Actions
 
     public func showMessage(_ message: Message, position: MessagePosition = .center) -> Bool {
-        guard let queueId = message.queueId else {
-            Logger.instance.error(message: "Message does not have a queueId.")
-            return false
-        }
 
-        if shownMessageQueueIds.contains(queueId) {
-            Logger.instance.info(message: "Message \(queueId) already shown, skipping.")
-            return false
-        }
-        
         if let messageManager = getModalMessageManager() {
             Logger.instance.info(message: "Message cannot be displayed, \(messageManager.currentMessage.messageId) is being displayed.")
         } else {
             let messageManager = createMessageManager(siteId: siteId, message: message)
             messageManager.showMessage(position: position)
-            shownMessageQueueIds.insert(queueId)
             return true
         }
         return false
@@ -95,6 +85,13 @@ public class Gist: GistDelegate {
     // MARK: Events
 
     public func messageShown(message: Message) {
+        let queueId = message.queueId
+        
+        if queueId != nil && shownMessageQueueIds.contains(queueId!) {
+            Logger.instance.info(message: "Message \(queueId!) already shown, skipping.")
+            return
+        }
+
         Logger.instance.debug(message: "Message with route: \(message.messageId) shown")
         if message.gistProperties.persistent != true {
             logMessageView(message: message)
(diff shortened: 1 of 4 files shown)


Sources/MessagingInApp/Gist/Gist.swift
```diff
@@ -56,11 +57,22 @@ public class Gist: GistDelegate {
     // MARK: Message Actions
 
     public func showMessage(_ message: Message, position: MessagePosition = .center) -> Bool {
+        guard let queueId = message.queueId else {
```
BernardGatt, line 60: This check would prevent anyone from launching a message programmatically.

Shahroz16 (changes requested):
the only problem I forsee is that we insert the message queue id in the `set` and then webview failing to load message. 

so maybe, we need to insert it either we are certain that message is displayed, or we need to remove from the queue, if the `MessageManager` return an error.? 

@BernardGatt what do you think?

BernardGatt:
> the only problem I forsee is that we insert the message queue id in the `set` and then webview failing to load message.
> 
> so maybe, we need to insert it either we are certain that message is displayed, or we need to remove from the queue, if the `MessageManager` return an error.?
> 
> @BernardGatt what do you think?

Yeah, for web we're inserting it when the message is shown (right after first load). I think the best place for this is inside: `public func messageShown(message: Message) {`

Sources/MessagingInApp/Gist/Gist.swift
```diff
@@ -56,11 +57,22 @@ public class Gist: GistDelegate {
...
+            Logger.instance.error(message: "Message does not have a queueId.")
+            return false
+        }
+
+        if shownMessageQueueIds.contains(queueId) {
```
BernardGatt, line 65: The same check needs to be included in embedMessage as well.

Sources/MessagingInApp/Gist/Gist.swift
```diff
@@ -83,13 +85,24 @@ public class Gist: GistDelegate {
     // MARK: Events
 
     public func messageShown(message: Message) {
+        let queueId = message.queueId
+        
+        if queueId != nil && shownMessageQueueIds.contains(queueId!) {
```
Shahroz16, line 90: I think we probably need to check for if the message is `persistent` or not, @BernardGatt correct me if i am wrong here. 

Because with persistent we load the same message until dismissed?

BernardGatt: Good catch Shahroz!!

To avoid this we could check if the message is persistent and if it is we don't add it to the list. We only add it when dismissMessage is called and the message is persistent.

What do you think?

Shahroz16: Even better !!

Sources/MessagingInApp/Gist/Gist.swift
```diff
@@ -83,13 +90,26 @@ public class Gist: GistDelegate {
     // MARK: Events
 
     public func messageShown(message: Message) {
+        let queueId = message.queueId
+
+        // Skip shown messages
+        if queueId != nil, shownMessageQueueIds.contains(queueId!) {
+            Logger.instance.info(message: "Message \(queueId!) already shown, skipping.")
+            return
```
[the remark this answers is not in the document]
BernardGatt: Thank you Shahroz, Steve's away and I'm picking up this change. I will ping you in the channel once the changes are implemented.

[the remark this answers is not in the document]
BernardGatt: I moved some calls around. `logMessageView` only gets called when the message has been dismissed so in the case of persistent messages, they won't be added to the `shownMessageQueueIds` list until properly dismissed.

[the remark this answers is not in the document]
Shahroz16: awesome, I think we can probably modularize the queuing part a little so this logic of when to display/when to log, etc gets encapsulated. 

We don't have to do it now but maybe sometime later before it this class gets too big.

Shahroz16 (commented):
My two reservations with this, we can ignore them if they are intentional. 

First up, are we ensuring "messages are only shown once" as the PR title suggests or are we ensuring "messages are logged only once"? because this PR is catering latter. 

Secondly, should we clear the queue at some point? i am not sure what might be the right time and place for that would be. Maybe when we fetch for messages and it returns empty?

Shahroz16 (approved):
Looks good. 🙌

PR #472, opened 2024-01-09
test: data pipeline implementation interaction

mrehan27:

###  Changes

- Fixed tests for `DataPipelineImplementation`
- Added tests for verifying `DataPipeline` communication with `analytics`


Sources/Common/Communication/EventBusHandler.swift
```diff
@@ -1,11 +1,19 @@
 import Combine
 import Foundation
 
+public protocol EventBusHandler {
```
mrehan27, line 4: Added interface so calls to EventBus can be tested

Sources/DataPipeline/DataPipelineImplementation.swift
mrehan27, line 1: The file contains changes to fix bugs/behavior changes identified while fixing and running these tests

Tests/Shared/Mocks/EventBusHandler.swift
```diff
@@ -0,0 +1,62 @@
+import CioInternalCommon
+
+/// Mock of the EventBusHandler class, designed to mimic AutoMockable
+/// Once the class is generated using AutoMockable, it should seamlessly replace the current implementation without any issues
+public class EventBusHandlerMock: EventBusHandler, Mock {
```
mrehan27, line 5: Due to generics, I couldn't generate it using `AutoMockable`, but I tried to keep the implementation similar to it. If we can generate this using `AutoMockable`, that should hopefully work without making any change in tests.

Tests/Tracking/migration/DataPipelineMigrationAssistantTests.swift
```diff
@@ -0,0 +1,76 @@
+@testable import CioDataPipelines
+@testable import CioInternalCommon
+import Foundation
+@testable import Segment
+import SharedTests
+import XCTest
+
+class DataPipelineMigrationAssistantTests: UnitTest {
```
mrehan27, line 8: I just moved them from [`CustomerIOImplementationTests`](https://github.com/customerio/customerio-ios/blob/main-replica-for-cdp/Tests/DataPipeline/DataPipelineInteractionTests.swift#L413-L452) to a separate file.

Sources/DataPipeline/DataPipelineImplementation.swift
mrehan27, line 1: The file contains changes to fix bugs/behavior changes identified while fixing and running these tests

mrehan27: Resolving as shown outdated

mrehan27, line 1: The file contains changes to fix bugs/behavior changes identified while fixing and running these tests

Tests/Shared/Mocks/EventBusHandler.swift
Shahroz16: i think it's a fair call, using generic here really helps with the casting of events when mapped from map

Shahroz16 (approved):

Tests/Shared/Mocks/EventBusHandler.swift
mrehan27: Agreed

ami-aman (approved):

PR #474, opened 2024-01-10
ci: fix lint assertions 

levibostian:
Right now, our iOS SDK CI server runs binny on Linux OS instead of macOS. I wanted to try to make the iOS SDK CI run on Linux instead of macOS because Linux is faster, but [it doesn't work](https://github.com/customerio/customerio-ios/actions/runs/7452320766/job/20275317836?pr=470). I believe this is because the tools that binny executes (lint, formatter) are iOS tools and not may work on Linux. 

To try and fix our iOS SDK CI, I [ran an experiment](https://github.com/customerio/customerio-ios/actions/runs/7477383566/job/20349926338?pr=474) of seeing if we run binny on macOS machines instead of Linux, it would fix the issue. It did fix it!

This PR changes the CI server to run lint/formatter on macos instead of Linux to fix the issue. 

# Dependencies before merging

- https://github.com/customerio/binny/pull/9

### DIFF
--- a/.github/workflows/build-sample-apps.yml
+++ b/.github/workflows/build-sample-apps.yml
@@ -1,146 +1,146 @@
-name: Build sample apps 
+# name: Build sample apps 
 
-on: 
-  pull_request: # build sample apps for every commit pushed to an open pull request (including drafts)
-  push: 
-    branches: [main]
-  workflow_dispatch: # allow manual run     
+# on: 
+#   pull_request: # build sample apps for every commit pushed to an open pull request (including drafts)
+#   push: 
+#     branches: [main]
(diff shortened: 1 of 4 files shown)


Shahroz16 (approved):

ami-aman (approved):
👍

levibostian:
There is still a false positive when running the linter on the CI server at the moment. There could be a lint check that fails, but the CI server does not catch this. The CI status check will succeed. 

This is a binny bug. [This bug fix](https://github.com/customerio/binny/pull/10) will take care of it. 

I'll make a new PR if that binny bug doesn't fix this CI server's issue.

PR #477, opened 2024-01-11
fix: removed last_used from properties 

Shahroz16:
more details can be found here:
https://customerio.slack.com/archives/C05UY5J4VLK/p1704830320326159?thread_ts=1704811102.597619&cid=C05UY5J4VLK

### DIFF
--- a/Sources/DataPipeline/DataPipelineImplementation.swift
+++ b/Sources/DataPipeline/DataPipelineImplementation.swift
@@ -190,9 +190,6 @@ class DataPipelineImplementation: DataPipelineInstance {
         // Consolidate all Apple platforms under iOS
         deviceAttributesProvider.getDefaultDeviceAttributes { defaultDeviceAttributes in
             let deviceAttributes: [String: Any] = defaultDeviceAttributes
-                .mergeWith([
-                    "last_used": self.dateUtil.now
-                ])
                 .mergeWith(customAttributes)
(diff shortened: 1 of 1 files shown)


Sources/DataPipeline/DataPipelineImplementation.swift
```diff
@@ -190,9 +190,6 @@ class DataPipelineImplementation: DataPipelineInstance {
         // Consolidate all Apple platforms under iOS
         deviceAttributesProvider.getDefaultDeviceAttributes { defaultDeviceAttributes in
             let deviceAttributes: [String: Any] = defaultDeviceAttributes
-                .mergeWith([
-                    "last_used": self.dateUtil.now
-                ])
```
ami-aman, line 195: Migration question: Do we also need to remove `last_used` from the unprocessed tasks we get from background queue during migration ? We do get this property in certain json payloads and we send them as is to CDP.

ami-aman (approved):

Sources/DataPipeline/DataPipelineImplementation.swift
ami-aman: If yes, then we can make the change in another PR, no need to hold this PR.

Shahroz16: Yup, we will need to remove it, along with that check the `timestamp` in general.

PR #478, opened 2024-01-15
chore: data pipeline tests base improvements

mrehan27:

###  Changes

- Added helper methods for setting up fake/mocked instances for `DataPipeline` tests
- Fixes and improvements in base `UnitTest` and `IntegrationTest` classes
- Fixed broken test for `last_used` attribute

### DIFF
--- a/Sources/Tracking/CustomerIO.swift
+++ b/Sources/Tracking/CustomerIO.swift
@@ -78,11 +78,4 @@ public extension CustomerIO {
 
         initializeSharedInstance(with: implementation, diGraph: newDiGraph)
     }
-
-    // Initialize for Integration Tests
-    static func initializeIntegrationTests(diGraph: DIGraph) {
-        // FIXME: [CDP] Fix tests using DataPipeline
-        // let implementation = CustomerIOImplementation(diGraph: diGraph)
-        // initializeSharedInstance(with: implementation, diGraph: diGraph)
-    }
 }
--- a/Tests/DataPipeline/DataPipelineInteractionTests.swift
+++ b/Tests/DataPipeline/DataPipelineInteractionTests.swift
@@ -33,8 +33,7 @@ class DataPipelineInteractionTests: UnitTest {
         // setting up analytics for testing
         analytics = implementation.analytics
         // OutputReaderPlugin helps validating interactions with analytics
-        outputReader = OutputReaderPlugin()
-        analytics.add(plugin: outputReader)
+        outputReader = attachPlugin(analytics: analytics, plugin: OutputReaderPlugin())
         // wait for analytics queue to start emitting events
         waitUntilStarted(analytics: analytics)
     }
--- a/Tests/DataPipeline/DataPipelineInteractionTests.swift
+++ b/Tests/DataPipeline/DataPipelineInteractionTests.swift
@@ -427,10 +427,10 @@ class DataPipelineInteractionTests: UnitTest {
     func test_registerDeviceToken_givenProfileIdentified_expectStoreAndRegisterDevice() {
         let givenDeviceToken = String.random
         let givenIdentifier = String.random
-        let givenDefaultAttributes: [String: Any] = ["foo": "bar"]
-        let expectedAttributes = givenDefaultAttributes.mergeWith([
-            "last_used": dateUtilStub.givenNow
-        ])
+        let givenDefaultAttributes: [String: Any] = [
+            "cio_sdk_version": "3.0.0",
+            "push_enabled": true
+        ]
 
         mockDeviceTokenDependencies(defaultAttributes: givenDefaultAttributes)
         customerIO.identify(identifier: givenIdentifier)
@@ -445,14 +445,12 @@ class DataPipelineInteractionTests: UnitTest {
         XCTAssertEqual(deviceUpdatedEvent?.deviceToken, givenDeviceToken)
         XCTAssertEqual(globalDataStoreMock.pushDeviceToken, givenDeviceToken)
 
-        assertDictionariesEqual(expectedAttributes, deviceUpdatedEvent?.properties) { key, expected, actual in
+        assertDictionariesEqual(givenDefaultAttributes, deviceUpdatedEvent?.properties) { key, expected, actual in
             switch key {
-            case "foo":
+            case "cio_sdk_version":
                 XCTAssertEqual((expected[key] as! String), actual[key] as? String)
-            case "last_used":
-                // analytics SDK wraps date in double quotes
-                let actualValue = ((actual[key] as? Encodable)?.toString())?.trimmingCharacters(in: CharacterSet.punctuationCharacters)
-                XCTAssertEqual((expected[key] as! Date).iso8601(), actualValue)
+            case "push_enabled":
+                XCTAssertEqual((expected[key] as! Bool), actual[key] as? Bool)
             default:
                 XCTFail("unexpected key received: '\(key)'")
             }
--- a/Tests/DataPipeline/DataPipelineInteractionTests.swift
+++ b/Tests/DataPipeline/DataPipelineInteractionTests.swift
@@ -27,10 +27,7 @@ class DataPipelineInteractionTests: UnitTest {
         diGraphShared.override(value: eventBusHandlerMock, forType: EventBusHandler.self)
         diGraphShared.override(value: globalDataStoreMock, forType: GlobalDataStore.self)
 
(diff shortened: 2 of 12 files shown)


Sources/Tracking/CustomerIO.swift
```diff
@@ -78,11 +78,4 @@ public extension CustomerIO {
 
         initializeSharedInstance(with: implementation, diGraph: newDiGraph)
     }
-
-    // Initialize for Integration Tests
-    static func initializeIntegrationTests(diGraph: DIGraph) {
```
mrehan27, line 83: Added similar method to `DataPipeline` module

Tests/DataPipeline/DataPipelineInteractionTests.swift
```diff
@@ -427,10 +423,10 @@ class DataPipelineInteractionTests: UnitTest {
     func test_registerDeviceToken_givenProfileIdentified_expectStoreAndRegisterDevice() {
         let givenDeviceToken = String.random
         let givenIdentifier = String.random
-        let givenDefaultAttributes: [String: Any] = ["foo": "bar"]
-        let expectedAttributes = givenDefaultAttributes.mergeWith([
-            "last_used": dateUtilStub.givenNow
```
mrehan27, line 432: The test was broken after recent PR merge due to change in expectations from server side, removed `last_used` to fix test as server no longer requires it.

Tests/Tracking/migration/DataPipelineMigrationAssistantTests.swift
```diff
@@ -23,10 +24,7 @@ class DataPipelineMigrationAssistantTests: UnitTest {
         diGraph.override(value: dateUtilStub, forType: DateUtil.self)
         diGraph.override(value: backgroundQueueMock, forType: Queue.self)
 
-        let moduleConfig = DataPipelineConfigOptions.Factory.create(writeKey: "test")
-        let implementation = DataPipelineImplementation(diGraph: diGraphShared, moduleConfig: moduleConfig)
-
-        DataPipeline.setupSharedTestInstance(implementation: implementation, config: moduleConfig)
+        let implementation = DataPipeline.createAndSetSharedTestInstance(diGraphShared: diGraphShared, config: dataPipelineModuleConfig)
         customerIO = CustomerIO(implementation: implementation, diGraph: diGraph)
```
levibostian, line 28: Directly calling `DataPipeline.createAndSetSharedTestInstance` and `CustomerIO(implementation: implementation, diGraph: diGraph)` feels like boilerplate code to me. I was expecting that these 2 lines of code in this file would instead call a super method to initialize the `CustomerIO` instance. 

Is there a reason why this test class cannot do this?

Tests/DataPipeline/Support/TestUtilities.swift
```diff
@@ -76,12 +77,24 @@ extension RawEvent {
...
     if let startupQueue = analytics.find(pluginType: StartupQueue.self) {
         while startupQueue.running != true {
             RunLoop.main.run(until: Date.distantPast)
         }
     }
 }
+
+/// Attaches and returns plugin only if it wasn't attached previously
+func attachPlugin<P: Plugin>(analytics: Analytics? = DataPipeline.shared.analytics, plugin: P) -> P {
+    guard let analytics = analytics else { return plugin }
```
levibostian, line 93:
```suggestion
    guard let analytics = analytics else { fatalError() }
```

I was not expecting to see this line of code exit early and ignore the request. Is there a reason to do this instead of fail?  

If a developer calls this function to attach a plugin, the function runs without throwing an error, they would expect the plugin got attached but maybe it did not. I would expect that by ignoring the request, this could make writing and debugging tests more difficult.

Tests/DataPipeline/Support/TestUtilities.swift
```diff
@@ -76,12 +77,24 @@ extension RawEvent {
...
+
+    if analytics.find(pluginType: P.self) != nil {
+        fatalError("Plugin \(P.self) is already attached")
```
levibostian, line 96:
```suggestion
        throw new Error()
```

Because these utility functions are executed in tests, I suggest that we throw errors instead of call `fatalError`. 

I suggest this because when you use `fatalError`, the entire test suite will exit early and not finish if it ever executes. But if we use throw, that 1 test function will fail and the remainder of the suite continues running.

Tests/DataPipeline/DataPipelineInteractionTests.swift
```diff
@@ -27,17 +27,13 @@ class DataPipelineInteractionTests: UnitTest {
         diGraphShared.override(value: eventBusHandlerMock, forType: EventBusHandler.self)
         diGraphShared.override(value: globalDataStoreMock, forType: GlobalDataStore.self)
 
-        let moduleConfig = DataPipelineConfigOptions.Factory.create(writeKey: "test")
-        let implementation = DataPipelineImplementation(diGraph: diGraphShared, moduleConfig: moduleConfig)
-
-        DataPipeline.setupSharedTestInstance(implementation: implementation, config: moduleConfig)
+        let implementation = DataPipeline.createAndSetSharedTestInstance(diGraphShared: diGraphShared, config: dataPipelineModuleConfig)
         customerIO = CustomerIO(implementation: implementation, diGraph: diGraph)
```
levibostian, line 31: Similar comment about boilerplate code constructing `CustomerIO` instances.

Tests/DataPipeline/DataPipelineInteractionTests.swift
levibostian: Thanks for adding a comment here explaining this. 

> as server no longer requires it.

I understand the API may not require it, but I believe our SDK still sends this parameter. Therefore, I think that we should continue to test it if our SDK does send it. 

Although, looking at the test function here, the payload of `givenDefaultAttributes` looks small in size. Making me think that this test is not meant to test the accuracy of what the SDK payload sends, but is instead testing something else?

levibostian (approved):
All my suggestions are non-blocking.

Tests/Shared/IntegrationTest.swift
```diff
@@ -16,11 +17,11 @@ open class IntegrationTest: UnitTest {
     public private(set) var sampleDataFilesUtil: SampleDataFilesUtil!
 
     override open func setUp() {
-        setUp(modifySdkConfig: nil)
+        setUp(modifyModuleConfig: nil)
     }
 
-    open func setUp(modifySdkConfig: ((inout SdkConfig) -> Void)? = nil) {
-        super.setUp(modifySdkConfig: modifySdkConfig)
+    open func setUp(modifyModuleConfig: ((inout DataPipelineConfigOptions) -> Void)? = nil) {
```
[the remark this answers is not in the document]
mrehan27: I deliberately removed it as our code is now mostly relying on params from `DataPipelineConfigOptions` instead of `SdkConfig`. So if any test is using `SdkConfig`, it should be updated to use `DataPipelineConfigOptions`. But so far it apparently hasn't broken anything, if it will, we should refactor the test to utilize `DataPipelineConfigOptions`.

Tests/Tracking/migration/DataPipelineMigrationAssistantTests.swift
```diff
@@ -1,6 +1,7 @@
 @testable import CioDataPipelines
 @testable import CioInternalCommon
 @testable import CioTracking
+@testable import DataPipelineTests
```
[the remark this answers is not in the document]
mrehan27: Strangely, it works the other way locally 🤷🏻‍♂️ But thanks for pointing this out as I realized we longer need this import and have removed it.

Tests/Tracking/migration/DataPipelineMigrationAssistantTests.swift
mrehan27: Thanks for the idea, added extension methods for this.

Tests/DataPipeline/Support/TestUtilities.swift
mrehan27: Updated 👍🏻

Tests/DataPipeline/Support/TestUtilities.swift
mrehan27: Throwing error will force handling them and might make things more complex. Ideally, this should only execute while writing new tests when plugin is reattached mistakenly. So I think it is better to keep `fatalError` here to keep calling this method simple.

Tests/DataPipeline/DataPipelineInteractionTests.swift
mrehan27: Updated

PR #481, opened 2024-01-18
chore: http common refactoring and changes for tracking 

Shahroz16:
helps https://linear.app/customerio/issue/MOB-13/http-layer-for-nse
root ticket: https://github.com/customerio/issues/issues/11953

**TLDR** (work required): We need to add an HTTP layer that exists in the `Messaging Push` module when a push is received in NSE, we can just use that HTTP client to track the delivery for us. HTTP client is going to use the CIO data pipeline API endpoint. 

## Problem: 
Currently, HTTP client implementation was meant to stay in the `common` layer, but not anymore, if we are only utilizing the http client rich push notification, downloading URL/ tracking metric. Then implementation should be `Messaging` module specific as well.  

## Solution:
This PR does following:
- Update http client and other related things like error, etc to be module specific and can be used in modules
- Decouple these modules from tracking (as in SDK config etc) so the other modules can utilized those methods as well. 
- Changes in HTTP clients such as the CDP tracking endpoint needed to track metrics 

What this PR isn't doing:
- Clean up unused http client code implementation, usage in tracking etc. The clean of tracking code and unused methods of http client etc would be in another PR, otherwise they would make the review harder.


Sources/Common/Service/CIOApiEndpoint.swift
```diff
@@ -33,19 +35,23 @@ public enum CIOApiEndpoint {
...
 
 public extension CIOApiEndpoint {
-    func getUrl(baseUrls: HttpBaseUrls) -> URL? {
-        URL(string: getUrlString(baseUrls: baseUrls))
+    func getUrl(baseUrl: String) -> URL? {
+        URL(string: getUrlString(baseUrl: baseUrl))
     }
 
-    func getUrlString(baseUrls: HttpBaseUrls) -> String {
+    func getUrl(baseUrls: HttpBaseUrls) -> URL? {
```
Shahroz16, line 48: this is going to go away in later PRs, as we don't use it anymore. keeping in this PR so the build remains compilable if anyone wants to run it.

Sources/Common/Service/HttpRequestRunner.swift
```diff
@@ -6,7 +6,7 @@ import FoundationNetworking
 /**
  Exists to be able to mock http requests so we can test our HttpClient's response handling logic.
  */
-protocol HttpRequestRunner: AutoMockable {
+public protocol HttpRequestRunner: AutoMockable {
```
Shahroz16, line 9: to be available in other modules

Sources/Common/Service/CIOApiEndpoint.swift
```diff
@@ -8,6 +8,7 @@ public enum CIOApiEndpoint {
     case trackCustomerEvent(identifier: String)
     case pushMetrics
     case trackDeliveryMetrics
+    case trackNSEPushDeliveryMetrics
```
levibostian, line 11: Do we need to introduce a new case here? I see `trackDeliveryMetrics` in the line above that I was expecting we would use instead.

levibostian (commented):
Could you provide a description with more information behind this change? The code looks OK, but I do not understand why we are making it. That would be helpful to make a higher quality review.

Sources/Common/Service/CIOApiEndpoint.swift
Shahroz16: Wanted to keep the current implementation for CIO tracking APIs separated from CIO data pipeline APIs. And all other cases would get removed because we don't use track API endpoints anymore.

levibostian: Ok, I understand why adding another case makes sense. 

```suggestion
    case trackPushMetricsCdp
```

I think it would be good to use a name that describes the endpoint (track push metrics), instead of a name that describes the caller (NSE).

levibostian (approved):

PR #482, opened 2024-01-18
chore: added rich push delivery tracking

Shahroz16:
helps: https://linear.app/customerio/issue/MOB-13/http-layer-for-nse
root ticket: https://github.com/customerio/issues/issues/11953

TLDR (work required): We need to add an HTTP layer that exists in the `Messaging Push` module when a push is received in NSE, we can just use that HTTP client to track the delivery for us. HTTP client is going to use the CIO data pipeline API endpoint. 

## Problem: 
Currently, HTTP client implementation was meant to stay in the `common` layer, but not anymore, if we are only utilizing the HTTP client rich push notification, downloading URL/ tracking metric. The implementation should be `Messaging` module-specific as well.  

## Solution:
This PR does the following:
- Add HTTP client implementation in the Messaging Push since it's the only module that utilizes it. 
- Add logic for tracking metrics for push

What this PR isn't doing:
- Clean up unused HTTP client code implementation, usage in tracking, etc. The cleaning of tracking code and unused methods of HTTP client etc would be in another PR, otherwise, they would make the review harder.

### DIFF
--- a/Sources/MessagingPush/MessagingPush.swift
+++ b/Sources/MessagingPush/MessagingPush.swift
@@ -36,14 +36,29 @@ public class MessagingPush: ModuleTopLevelObject<MessagingPushInstance>, Messagi
     public static func initialize(
         configure configureHandler: ((inout MessagingPushConfigOptions) -> Void)? = nil
     ) -> MessagingPushInstance {
-        var configOptions = moduleConfig
-
         if let configureHandler = configureHandler {
-            configureHandler(&configOptions)
+            // pass current config reference to update it without needing to recreate
+            configureHandler(&moduleConfig)
         }
 
         shared.initializeModule()
         return shared
+    }
+
+    /// MessagingPush initializer for Notification Service Extension
+    @available(iOS, unavailable)
+    @available(iOSApplicationExtension, introduced: 13.0)
+    @discardableResult
+    public static func initialize(
+        writeKey: String,
+        configure configureHandler: ((inout MessagingPushConfigOptions) -> Void)? = nil
+    ) -> MessagingPushInstance {
+        initialize { config in
+            config.writeKey = writeKey
+            if let configureHandler = configureHandler {
+                configureHandler(&config)
+            }
+        }
     }
 
     private func initializeModule() {
--- a/Sources/MessagingPush/MessagingPushImplementation.swift
+++ b/Sources/MessagingPush/MessagingPushImplementation.swift
@@ -10,18 +10,21 @@ class MessagingPushImplementation: MessagingPushInstance {
     let logger: Logger
     let jsonAdapter: JsonAdapter
     let eventBusHandler: EventBusHandler
+    let richPushDeliveryTracker: RichPushDeliveryTracker
 
     /// testing init
     init(
         moduleConfig: MessagingPushConfigOptions,
         logger: Logger,
         jsonAdapter: JsonAdapter,
-        eventBusHandler: EventBusHandler
+        eventBusHandler: EventBusHandler,
+        richPushDeliveryTracker: RichPushDeliveryTracker
     ) {
         self.moduleConfig = moduleConfig
         self.logger = logger
         self.jsonAdapter = jsonAdapter
         self.eventBusHandler = eventBusHandler
+        self.richPushDeliveryTracker = richPushDeliveryTracker
     }
 
     init(diGraph: DIGraphShared, moduleConfig: MessagingPushConfigOptions) {
@@ -29,6 +32,7 @@ class MessagingPushImplementation: MessagingPushInstance {
         self.logger = diGraph.logger
         self.jsonAdapter = diGraph.jsonAdapter
         self.eventBusHandler = diGraph.eventBusHandler
+        self.richPushDeliveryTracker = diGraph.richPushDeliveryTracker
     }
 
     func deleteDeviceToken() {
@@ -41,6 +45,14 @@ class MessagingPushImplementation: MessagingPushInstance {
 
     func trackMetric(deliveryID: String, event: Metric, deviceToken: String) {
         eventBusHandler.postEvent(TrackMetricEvent(deliveryID: deliveryID, event: event.rawValue, deviceToken: deviceToken))
+    }
+
+    func trackMetricFromNSE(
+        deliveryID: String,
+        event: Metric,
+        deviceToken: String
+    ) {
+        richPushDeliveryTracker.trackMetric(token: deviceToken, event: event, deliveryId: deliveryID) { _ in }
     }
 
     #if canImport(UserNotifications)
--- a/Sources/MessagingPush/RichPush/RichPushDeliveryTracker.swift
+++ b/Sources/MessagingPush/RichPush/RichPushDeliveryTracker.swift
@@ -1,0 +1,43 @@
+import CioInternalCommon
+import Foundation
+
+// sourcery: InjectRegisterShared = "RichPushDeliveryTracker"
+class RichPushDeliveryTracker {
+    let httpClient: HttpClient
+    let logger: Logger
+
+    init(httpClient: HttpClient, logger: Logger) {
+        self.httpClient = httpClient
+        self.logger = logger
+    }
+
+    func trackMetric(token: String, event: Metric, deliveryId: String, timestamp: String? = nil, onComplete: @escaping (Result<Void, HttpRequestError>) -> Void) {
+        let properties: [String: Any] = [
+            "anonymousId": deliveryId,
+            "properties": [
+                "recipient": token,
+                "metric": event.rawValue,
+                "deliveryId": deliveryId
+            ],
+            "event": "Report Delivery Event"
+        ]
+
+        let endpoint: CIOApiEndpoint = .trackNSEPushDeliveryMetrics
+        guard let httpParams = HttpRequestParams(
+            endpoint: endpoint,
+            baseUrl: RichPushHttpClient.defaultAPIHost,
+            headers: nil,
+            body: try? JSONSerialization.data(withJSONObject: properties)
+        ) else {
+            logger.error("Error constructing HTTP request. Endpoint: \(endpoint)")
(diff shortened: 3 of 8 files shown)


Sources/MessagingPush/MessagingPush.swift
```diff
@@ -36,16 +36,31 @@ public class MessagingPush: ModuleTopLevelObject<MessagingPushInstance>, Messagi
     public static func initialize(
         configure configureHandler: ((inout MessagingPushConfigOptions) -> Void)? = nil
     ) -> MessagingPushInstance {
-        var configOptions = moduleConfig
-
         if let configureHandler = configureHandler {
-            configureHandler(&configOptions)
+            // pass current config reference to update it without needing to recreate
+            configureHandler(&moduleConfig)
```
Shahroz16, line 41: `moduleConfig` is declared at the top and is atomic. Its being utilized everywhere.

Sources/MessagingPush/RichPush/RichPushHttpClient.swift
```diff
@@ -0,0 +1,192 @@
+import CioInternalCommon
+import Foundation
+#if canImport(FoundationNetworking)
+import FoundationNetworking
+#endif
+
+// sourcery: InjectRegisterShared = "HttpClient"
+public class RichPushHttpClient: HttpClient {
```
Shahroz16, line 8: Class is very similar to the `HttPClient` implementation in the `Tracking` module, but stripped down to only contain things needed and decoupled from any `Queue` logic.

Sources/MessagingPush/RichPush/RichPushDeliveryTracker.swift
```diff
@@ -0,0 +1,43 @@
...
+    let httpClient: HttpClient
+    let logger: Logger
+
+    init(httpClient: HttpClient, logger: Logger) {
+        self.httpClient = httpClient
+        self.logger = logger
+    }
+
+    func trackMetric(token: String, event: Metric, deliveryId: String, timestamp: String? = nil, onComplete: @escaping (Result<Void, HttpRequestError>) -> Void) {
+        let properties: [String: Any] = [
```
Shahroz16, line 15: since we don't plan on reusing this, kept as a map rather than a struct/class.

Sources/MessagingPush/MessagingPush.swift
```diff
@@ -36,16 +36,31 @@ public class MessagingPush: ModuleTopLevelObject<MessagingPushInstance>, Messagi
...
         }
 
         shared.initializeModule()
         return shared
     }
 
+    /// MessagingPush initializer for Notification Service Extension
```
levibostian, line 48: From my understanding, customers are asked to call `CustomerIO.initialize()` to provide the `writeKey` in the host app. Then, the `writeKey` can be referenced by other modules in the code. 

This change looks like we are introducing `writeKey` into the push module. This `writeKey` would only be used by the NSE for delivered push metrics. 

Assuming this is correct, could we avoid introducing `writeKey` to the push module? Having to manage multiple `writeKeys` in the SDK, depending on the context, would not be ideal. I would consider this change as introducing technical debt and I would prefer we avoid it if possible.

Sources/MessagingPush/RichPush/RichPushDeliveryTracker.swift
```diff
@@ -0,0 +1,43 @@
...
+class RichPushDeliveryTracker {
+    let httpClient: HttpClient
+    let logger: Logger
+
+    init(httpClient: HttpClient, logger: Logger) {
+        self.httpClient = httpClient
+        self.logger = logger
+    }
+
+    func trackMetric(token: String, event: Metric, deliveryId: String, timestamp: String? = nil, onComplete: @escaping (Result<Void, HttpRequestError>) -> Void) {
```
levibostian, line 14:
```suggestion
    func trackMetric(token: String, deliveryId: String, timestamp: String? = nil, onComplete: @escaping (Result<Void, HttpRequestError>) -> Void) {
```

The class name and use case seems to only be for delivered tracking. I don't think that we need to have an `event` parameter?

Sources/MessagingPush/RichPush/RichPushDeliveryTracker.swift
```diff
@@ -0,0 +1,43 @@
...
+            ],
+            "event": "Report Delivery Event"
+        ]
+
+        let endpoint: CIOApiEndpoint = .trackNSEPushDeliveryMetrics
+        guard let httpParams = HttpRequestParams(
+            endpoint: endpoint,
+            baseUrl: RichPushHttpClient.defaultAPIHost,
+            headers: nil,
+            body: try? JSONSerialization.data(withJSONObject: properties)
```
levibostian, line 30: Is there a reason for not using `jsonAdapter` here?

Sources/MessagingPush/RichPush/RichPushHttpClient.swift
levibostian: Instead of introducing a new `HttpClient` implementation inside of the push module, could we re-use the `HttpClient` inside of the Common module? 

I think that adding this file (that looks identical to the one in Common) to the push module would introduce technical debt to the project and I would like to avoid that if possible.

Sources/MessagingPush/MessagingPushImplementation.swift
```diff
@@ -43,6 +47,14 @@ class MessagingPushImplementation: MessagingPushInstance {
         eventBusHandler.postEvent(TrackMetricEvent(deliveryID: deliveryID, event: event.rawValue, deviceToken: deviceToken))
     }
 
+    func trackMetricFromNSE(
```
levibostian, line 50:
```suggestion
    func trackMetricOrFail(
```

From a code maintenance standpoint, I suggest we name functions by what they do (attempt or track metric, or fail) instead of where they are used (NSE). 

I think it could be ever better to isolate the NSE logic from the host app logic since the use case is different. 

Example: 
```swift
// Used by host app
class MessagingPushImplementation {
  trackMetric()
    eventBus...
}

// Used by NSE
class MessagingPushNSEImplementation: MessagingPushImplementation {
  override trackMetric()
    httpClient...
}
```

levibostian (changes requested):
Rejecting due to concerns I have about introducing technical debt into the codebase from the PR. 

I have a feeling I am missing context behind the implementation choice. If that is true, some of these questions could be resolved by me better understanding the decisions in this PR.

Sources/MessagingPush/RichPush/RichPushDeliveryTracker.swift
Shahroz16: I think I tried answering this in previous PRs too, so let me explain with an example. `jsonAdapter` is not that agnostic is meant to work as an adapter for json that our `tracking API` expected and accepted. 

For example in this case, if were to use the method of `jsonAdapter` 

```
 public func fromDictionary<T: Decodable>(_ dictionary: [AnyHashable: Any]) -> T? {
        do {
            let jsonData = try JSONSerialization.data(withJSONObject: dictionary)

            return fromJson(jsonData)
        } catch {
            log.error("\(error.localizedDescription), dictionary: \(dictionary)")
        }

        return nil
    }
```

`let jsonData = try JSONSerialization.data(withJSONObject: dictionary)` it's great and solves our problem, but later you see the very next line, `fromJson(jsonData)`, it starts decoding this `json` and breaks it. 

Now I can go ahead and start creating these new methods in `jsonAdapter` but again `jsonAdapter` is biased, you see the decode being used, now that wouldn't work too because to be compatible with Segment we need to use it `iso8601` as date decoding strategy. 

```
var decoder: JSONDecoder {
        let decoder = JSONDecoder()
        decoder.dateDecodingStrategy = .secondsSince1970
        return decoder
    }
```


Hence, I avoided using the `jsonAdapter` too aggressively or any other class for the utility that's not agnostic, for cleanup later on we can create the utility class but currently, the usage is too low to need one imho.

Sources/MessagingPush/RichPush/RichPushHttpClient.swift
Shahroz16: `CIOHttpClient` the implementation HttpClient in the `common` would get deleted, since its not being used anymore and was reliant on CIO api/headers etc.

levibostian: I understand that the current `CIOHttpClient` is specific to the Track API. That client is going to go away since we will no longer call the Track API. 

Because the common module contains general classes that can be used by any module in the SDK, it makes the most sense to me for a general `CIOCDPHttpClient` to be contained inside of the common module.

Sources/MessagingPush/RichPush/RichPushDeliveryTracker.swift
levibostian: If `JsonAdapter` is not being used in multiple places, then I am OK with not making a change in this current PR. 


The implementation of `JsonAdapter` today could be bias towards Track.

Sources/MessagingPush/MessagingPushImplementation.swift
levibostian: Although I don't like having 2 `trackMetrics` functions with different bodies, we know we have more refactors coming for the NSE target. Therefore, I am OK with holding off on change to then.

levibostian (approved):

Sources/MessagingPush/RichPush/RichPushHttpClient.swift
Shahroz16: `HttpClient` lives in common, if we were to add `CIOCDPHttpClient` in the common as well, we are going to make a decision to bind `common` module with `datapipeline` which doesn't make much sense, since we have multiple modules `push/in-app` and more in future that might just be feature modules independent of datapipeline. 

For example, Gist is using only its own http client for gist services, it doesn't want to utilize `CIOCDPHttpClient`. 

Additionally, `CIOCDPHttpClient` requires, `writeKey` and we won't have that in `common` module.

PR #485, opened 2024-01-23
test: data pipeline json compatibility

mrehan27:
helps: https://linear.app/customerio/issue/MOB-12/improve-cdp-test-coverage

###  Changes

- Added json compatibility tests

### DIFF
--- a/Tests/DataPipeline/DataPipelineCompatibilityTests.swift
+++ b/Tests/DataPipeline/DataPipelineCompatibilityTests.swift
@@ -1,0 +1,438 @@
+@testable import CioDataPipelines
+@testable import CioInternalCommon
+import Foundation
+@testable import Segment
+import SharedTests
+import XCTest
+
+private typealias SavedEvent = [String: Any]
+
+class DataPipelineCompatibilityTests: UnitTest {
+    private var customerIO: CustomerIO!
+    private var analytics: Analytics!
+    private var storage: Storage!
+
+    private let deviceInfoMock = DeviceInfoMock()
+    private let eventBusHandlerMock = EventBusHandlerMock()
+    private let globalDataStoreMock = GlobalDataStoreMock()
+
+    override func setUp() {
+        super.setUp()
+
+        diGraphShared.override(value: dateUtilStub, forType: DateUtil.self)
+        diGraphShared.override(value: deviceInfoMock, forType: DeviceInfo.self)
+        diGraphShared.override(value: eventBusHandlerMock, forType: EventBusHandler.self)
+        diGraphShared.override(value: globalDataStoreMock, forType: GlobalDataStore.self)
+
+        customerIO = createCustomerIOInstance()
+
+        // setting up analytics for testing
+        analytics = customerIO.analytics
+        guard let analytics = analytics else {
+            fatalError("Analytics instance is nil. The SDK has been set up incorrectly.")
+        }
+        // wait for analytics queue to start emitting events
+        analytics.waitUntilStarted()
+        // get storage instance so we can read final events
+        storage = analytics.storage
+        storage.hardReset(doYouKnowHowToUseThis: true)
+    }
+
+    override func tearDown() {
+        customerIO.clearIdentify()
+        super.tearDown()
+    }
+
+    // MARK: profile
+
+    func test_identifyWithoutAttributes_expectFinalJSONHasCorrectKeysAndValues() {
+        let givenIdentifier = String.random
+
+        customerIO.identify(identifier: givenIdentifier)
+
+        let allEvents = readTypeFromStorage(key: Storage.Constants.events)
+        let filteredEvents = allEvents.filter { $0.eventType == "identify" }
+        XCTAssertEqual(filteredEvents.count, 1, "too many events received")
+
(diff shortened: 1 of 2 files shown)


Tests/DataPipeline/DataPipelineCompatibilityTests.swift
```diff
@@ -0,0 +1,438 @@
...
+    // MARK: profile
+
+    func test_identifyWithoutAttributes_expectFinalJSONHasCorrectKeysAndValues() {
+        let givenIdentifier = String.random
+
+        customerIO.identify(identifier: givenIdentifier)
+
+        let allEvents = readTypeFromStorage(key: Storage.Constants.events)
+        let filteredEvents = allEvents.filter { $0.eventType == "identify" }
+        XCTAssertEqual(filteredEvents.count, 1, "too many events received")
```
ami-aman, line 55: "too many events received" Is this the right message for this test ? 🤔

Tests/DataPipeline/DataPipelineCompatibilityTests.swift
```diff
@@ -0,0 +1,438 @@
...
+
+        guard let savedEvent = filteredEvents.last else {
+            XCTFail("saved event must not be nil")
```
Shahroz16, line 58: bit confused here, the check before here verifies there is only `1` event, and this checks if it's not empty, wouldn't it be verified in the check before that?

Tests/DataPipeline/DataPipelineCompatibilityTests.swift
```diff
@@ -0,0 +1,438 @@
...
+
+        XCTAssertEqual(savedEvent[keyPath: "userId"] as? String, givenIdentifier)
+        XCTAssertMatches(
+            savedEvent[mapKeyPath: "traits"],
+            givenBody,
+            withTypeMap: [["age"]: Int.self]
+        )
+    }
+
+    func test_eventBeforeIdentify_expectFinalJSONHasNoUserId() {
```
Shahroz16, line 89: should we verify if there is an anonymous id?

Tests/DataPipeline/DataPipelineCompatibilityTests.swift
```diff
@@ -0,0 +1,438 @@
...
+        guard let savedEvent = filteredEvents.first else {
+            XCTFail("saved event must not be nil")
+            return
+        }
+
+        XCTAssertEqual(savedEvent[keyPath: "event"] as? String, "Device Created or Updated")
+        XCTAssertEqual(savedEvent[keyPath: "userId"] as? String, givenIdentifier)
+        XCTAssertEqual(savedEvent[keyPath: "context.device.token"] as? String, givenToken)
+        XCTAssertFalse(savedEvent.containsKey("last_used"))
+        XCTAssertFalse(savedEvent.containsKey("platform"))
```
Shahroz16, line 148: are these all the required attributes? if not i think we should add all the required attributes to add a new device?

Tests/DataPipeline/DataPipelineCompatibilityTests.swift
```diff
@@ -0,0 +1,438 @@
...
+        }
+
+        XCTAssertEqual(savedEvent[keyPath: "event"] as? String, "Device Created or Updated")
+        XCTAssertEqual(savedEvent[keyPath: "userId"] as? String, givenIdentifier)
+        XCTAssertEqual(savedEvent[keyPath: "context.device.token"] as? String, givenToken)
+
+        XCTAssertMatches(
+            savedEvent[mapKeyPath: "properties"],
+            expectedData,
+            withTypeMap: [["debugMode"]: Bool.self]
```
Shahroz16, line 182: `debugMode` what is this for?

Shahroz16 (approved):
Looks majorly good, added some suggestions.

Tests/DataPipeline/DataPipelineCompatibilityTests.swift
mrehan27: I couldn't think of better than this 😄 If you have ideas, please feel free to share.

Tests/DataPipeline/DataPipelineCompatibilityTests.swift
mrehan27: Unfortunately not exactly, because when the number is `0`, it may crash the tests. My [explanation in base PR](https://github.com/customerio/customerio-ios/pull/484#discussion_r1464606826) might help explain this better.

Tests/DataPipeline/DataPipelineCompatibilityTests.swift
mrehan27: Updated

Tests/DataPipeline/DataPipelineCompatibilityTests.swift
mrehan27: Added `type` as well. If you have more mind, please share so I can add them too

Tests/DataPipeline/DataPipelineCompatibilityTests.swift
mrehan27: Just dummy values for testing custom attributes.

PR #486, opened 2024-01-23
chore: track clean-up 

Shahroz16:
PR for https://linear.app/customerio/issue/MOB-41/track-cleanup

All this PR contains is the removal of the classes, and code that is not being used anymore. It is either being replaced by 
- CDP
- Event Bus
- HTTP layer in NSE

Note:
- There are still some classes that aren't being utilized but i left because they might be utilized later for prep for production, e.g UserAgentUtil 
- Also, the classes needed for migration are left too.


Sources/Common/Background Queue/Queue.swift
```diff
@@ -63,39 +32,6 @@ public protocol Queue: AutoMockable {
 }
 
 public extension Queue {
```
Shahroz16, line 34: only methods remaining are the ones needed for migration

Sources/Common/Service/CIOApiEndpoint.swift
```diff
@@ -1,40 +1,16 @@
 import Foundation
 
 public enum CIOApiEndpoint {
-    case findAccountRegion
-    case identifyCustomer(identifier: String)
-    case registerDevice(identifier: String)
-    case deleteDevice(identifier: String, deviceToken: String)
-    case trackCustomerEvent(identifier: String)
-    case pushMetrics
-    case trackDeliveryMetrics
```
Shahroz16, line 10: used track API, which we don't use anymore

PR #487, opened 2024-01-25
chore: test clean up and refactor

Shahroz16:
PR helps cleanup tests.

What is being removed/updated 
- Tests that aren't used anymore
- HTTP test has been updated for common code and HTTP client test has been moved to Push messaging modules
- Queue test has been removed and code that is only being used for test now, is now in test directory rather than source.


Sources/Common/Background Queue/QueueStorage.swift
```diff
@@ -34,13 +34,13 @@ public protocol QueueStorage: AutoMockable {
 public class FileManagerQueueStorage: QueueStorage {
     private let fileStorage: FileStorage
     private let jsonAdapter: JsonAdapter
-    private let siteId: String
+    let siteId: String
```
Shahroz16, line 37: made public for testing as this is just used for testing and migration

Tests/MessagingPush/HttpClientTest.swift
```diff
@@ -0,0 +1,179 @@
+@testable import CioInternalCommon
+@testable import CioMessagingPush
+import Foundation
+@testable import SharedTests
+#if canImport(FoundationNetworking)
+import FoundationNetworking
+#endif
+import XCTest
+
+class HttpClientTest: UnitTest {
```
Shahroz16, line 10: moved the client here as HTTP client that its testing is MessagingPush specific

Tests/Tracking/migration/QueueExtension.swift
```diff
@@ -0,0 +1,73 @@
+@testable import CioInternalCommon
+import Foundation
+
+public enum QueueTaskGroup {
```
Shahroz16, line 4: removed it from common and added it in test as its only used in testing

Tests/Tracking/CustomerIOIntegrationTests.swift
```diff
@@ -1,166 +0,0 @@
-@testable import CioInternalCommon
-@testable import CioTracking
-import Foundation
-import SharedTests
-import XCTest
-
-class CustomerIOIntegrationTests: IntegrationTest {
```
mrehan27, line 7: Why is this removed? 🤔

Tests/MessagingPush/HttpTest.swift
```diff
@@ -18,24 +19,23 @@ import XCTest
...
-        if let siteId = getEnvironmentVariable("SITE_ID"), let apiKey = getEnvironmentVariable("API_KEY") {
+        if let writeKey = getEnvironmentVariable("WRITE_KEY") {
             runner = UrlRequestHttpRequestRunner()
-            session = CIOHttpClient.getCIOApiSession(
-                siteId: siteId,
-                apiKey: apiKey,
-                userAgentHeaderValue: userAgentUtil.getUserAgentHeaderValue()
+            session = RichPushHttpClient.getCIOApiSession(
+                key: writeKey,
+                userAgentHeaderValue: deviceInfo.getUserAgentHeaderValue()
```
mrehan27, line 38: Aren't we using `userAgentUtil` anymore? Because looking at the code, `userAgentHeaderValue: userAgentUtil.getUserAgentHeaderValue()` made more sense to me than `userAgentHeaderValue: deviceInfo.getUserAgentHeaderValue()` 😅

mrehan27 (approved):
Just a few comments, looks good me otherwise 👍🏻

Tests/MessagingPush/HttpTest.swift
Shahroz16: We still have user agent class, but what the useragent class was doing was combining device info information with the client information. But since we don't have any client information in the NSE. We just need device info for this purpose

PR #490, opened 2024-01-31
chore: handle logged in user migration use case

ami-aman:
closes: https://linear.app/customerio/issue/MOB-60/update-all-migration-methods-to-use-timestamp-and-originaltimestamp

### About the PR
This PR covers a use case scenario where a user migrates from the Journeys module to the CDP module while already logged in. This ensures the CDP module is informed about the currently logged-in user for seamless processing of events further.


Sources/Tracking/CustomerIO.swift
```diff
@@ -35,15 +35,25 @@ public extension CustomerIO {
...
+        let implementation = DataPipeline.initialize(moduleConfig: DataPipelineConfigOptions.Factory.create(sdkConfig: sdkConfig))
+        let profileStore = newDiGraph.profileStore
+
+        // This code handles the scenario where a user migrates
+        // from the Journeys module to the CDP module while already logged in.
+        // This ensures the CDP module is informed about the
+        // currently logged-in user for seamless processing of events.
+        if DataPipeline.shared.analytics.userId == nil {
+            if let identifier = profileStore.identifier {
+                DataPipeline.shared.identify(identifier: identifier, body: [:])
```
mrehan27, line 48: We should clear `profileStore.identifier` after setting it for `analytics` to make sure the code only run once.

Use case in my that could break things:

- Identify user using old SDK
- Update SDK to use `DataPipeline` -> The profile should be identified in `DataPipeline`
- Clear identify using new SDK
- Relaunch the app so SDK initialization is called again -> The will re-identify with old id in `DataPipeline` even though it was cleared before relaunch

Sources/Tracking/CustomerIO.swift
```diff
@@ -35,15 +35,25 @@ public extension CustomerIO {
...
         var sdkConfig = SdkConfig.Factory.create(siteId: siteId, apiKey: apiKey, region: region)
+        let newDiGraph = DIGraph(sdkConfig: sdkConfig)
+        let implementation = DataPipeline.initialize(moduleConfig: DataPipelineConfigOptions.Factory.create(sdkConfig: sdkConfig))
+        let profileStore = newDiGraph.profileStore
+
+        // This code handles the scenario where a user migrates
+        // from the Journeys module to the CDP module while already logged in.
+        // This ensures the CDP module is informed about the
+        // currently logged-in user for seamless processing of events.
+        if DataPipeline.shared.analytics.userId == nil {
```
mrehan27, line 46: It would be great if we can write a test case for it too

mrehan27 (changes requested):
Requesting changes as I think we need to fix potential bug before merging

Sources/Tracking/CustomerIO.swift
ami-aman: Great catch on the use case, Rehan ! I have made the necessary updates and you can find the changes i this [commit](https://github.com/customerio/customerio-ios/pull/490/commits/4aee0591e4b9c4624a4f100484de0d7a9ca0b62e).

Sources/Tracking/CustomerIO.swift
levibostian: For readability and testability, I think it would be good to move this code into it's own encapsulated migration class.

Sources/Tracking/CustomerIO.swift
levibostian: I think a test should be written to confirm this use case is handled.

levibostian (changes requested):
I would like to request automated tests before the PR gets merged.

ami-aman:
> I would like to request automated tests before the PR gets merged.

Updated. Requested for re-review.

Tests/Tracking/migration/DataPipelineMigrationAssistantTests.swift
```diff
@@ -71,4 +74,25 @@ class DataPipelineMigrationAssistantTests: UnitTest {
...
+        XCTAssertNotNil(migrationAssistant.handleAlreadyIdentifiedMigratedUser())
+        XCTAssertNil(DataPipeline.shared.analytics.userId)
+    }
+
+    func test_givenUserOnCDPIdentified_expectNoUpdate() {
+        let givenIdentifier = String.random
+        DataPipeline.shared.identify(identifier: givenIdentifier)
+        XCTAssertNotNil(migrationAssistant.handleAlreadyIdentifiedMigratedUser())
+        XCTAssertEqual(DataPipeline.shared.analytics.userId, givenIdentifier)
+        XCTAssertNil(profileStoreMock.identifier)
```
mrehan27, line 96: Shouldn't it be like this?

```suggestion
        let givenIdentifier = String.random

        profileStoreMock.identifier = String.random
        DataPipeline.shared.identify(identifier: givenIdentifier)
        
        XCTAssertNotNil(migrationAssistant.handleAlreadyIdentifiedMigratedUser())
        XCTAssertEqual(DataPipeline.shared.analytics.userId, givenIdentifier)
        XCTAssertNil(profileStoreMock.identifier)
```

Sources/Tracking/CustomerIO.swift
```diff
@@ -35,16 +35,17 @@ public extension CustomerIO {
...
+        let migrationAssistant = newDiGraph.dataPipelineMigrationAssistant
 
         if let configureHandler = configureHandler {
             configureHandler(&sdkConfig)
         }
 
-        let implementation = DataPipeline.initialize(moduleConfig: DataPipelineConfigOptions.Factory.create(sdkConfig: sdkConfig))
-
+        // Handle logged-in user from Journeys to CDP
+        migrationAssistant.handleAlreadyIdentifiedMigratedUser()
```
mrehan27, line 47: I'm okay with this, but I would have preferred making only 1 call to `migrationAssistant` here, where it could further break things inside the class and do what it wants to do. e.g.

```swift
migrationAssistant.handleMigration()
```

and inside

```swift
func handleMigration() {
    handleAlreadyIdentifiedMigratedUser()
    handleQueueBacklog()
}
```

mrehan27 (approved):
Non-blocking comments, looks good to me otherwise 👍🏻

Tests/Tracking/migration/DataPipelineMigrationAssistantTests.swift
ami-aman: The method is `test_givenUserOnCDPIdentified_expectNoUpdate` i.e if CDP module has already identified a user then we do not expect the method to update anything. In your suggestion, you expect `profileStoreMock.identifier` to be `nil` which does not matter in this case. Does this help ?

Sources/Tracking/CustomerIO.swift
ami-aman: Thanks for the suggestion, find changes [here](https://github.com/customerio/customerio-ios/pull/490/commits/05ddf03e2b45379c67de1cac6347930e7618e6b6).

Tests/Tracking/migration/DataPipelineMigrationAssistantTests.swift
mrehan27: I actually only wanted to make sure that `profileStore` has some identifier added to it, so added `profileStoreMock.identifier = String.random` only.

But looking at this again, I think there should never be a case where a customer is identified on CDP before `profileStore` was cleared, so we can ignore this.

Sources/Tracking/migration/DataPipelineMigrationAssistant.swift
```diff
@@ -2,14 +2,21 @@ import CioDataPipelines
 import CioInternalCommon
 import Foundation
 
+public protocol DataPipelineMigration: AutoMockable {
+    func handleAlreadyIdentifiedMigratedUser()
+    func handleQueueBacklog()
+    func getAndProcessTask(for task: QueueTaskMetadata)
```
levibostian, line 8:
```suggestion
```

Nit-picky, but this function feels like an internal/private function only needed inside of `DataPipelineMigrationAssistant` implementation. So, not sure why we need to expose it in a protocol.

Sources/Tracking/migration/DataPipelineMigrationAssistant.swift
```diff
@@ -20,12 +27,34 @@ class DataPipelineMigrationAssistant {
...
         self.threadUtil = threadUtil
+        self.profileStore = profileStore
+    }
+
+    func handleMigration() {
+        handleAlreadyIdentifiedMigratedUser()
+        handleQueueBacklog()
+    }
+
+    func handleAlreadyIdentifiedMigratedUser() {
```
levibostian, line 45: Besides migrating the profile ID, does our SDK also need to migrate the currently registered device token? 

If so, it can be in a separate PR.

Tests/Tracking/migration/DataPipelineMigrationAssistantTests.swift
```diff
@@ -11,17 +11,20 @@ class DataPipelineMigrationAssistantTests: UnitTest {
...
     private var migrationAssistant: DataPipelineMigrationAssistant { diGraph.dataPipelineMigrationAssistant }
     private var queueStorage: QueueStorage { diGraph.queueStorage }
 
     override func setUp() {
         super.setUp()
 
+        diGraph.override(value: profileStoreMock, forType: ProfileStore.self)
         diGraph.override(value: dateUtilStub, forType: DateUtil.self)
         diGraph.override(value: backgroundQueueMock, forType: Queue.self)
+        diGraph.override(value: migrationAssistantMock, forType: DataPipelineMigration.self)
```
levibostian, line 27: I am curious why this test class is creating an instance of `migrationAssistant` via `diGraph.dataPipelineMigrationAssistant` instead of [what is commonly found in other test classes by using the class constructor](https://github.com/customerio/customerio-ios/blob/56584e39684da0b1ee4d642e43fcbaaee879d213/Tests/Common/Background%20Queue/QueueRunRequestTest.swift#L20-L27).

levibostian, line 27: Why does `diGraph.override(value: migrationAssistantMock, forType: DataPipelineMigration.self)` exist? 

This class is testing the `DataPipelineMigration` implementation, so it shouldn't be mocked?

Tests/Tracking/migration/DataPipelineMigrationAssistantTests.swift
```diff
@@ -71,4 +74,25 @@ class DataPipelineMigrationAssistantTests: UnitTest {
         XCTAssertNotNil(migrationAssistant.handleQueueBacklog())
         XCTAssertEqual(backgroundQueueMock.deleteProcessedTaskCallsCount, 0)
     }
+
+    func test_givenAlreadyIdentifiedProfile_expectUpdateUserId() {
+        let givenProfileIdentifiedInJourneys = String.random
+        profileStoreMock.identifier = givenProfileIdentifiedInJourneys
+        XCTAssertNotNil(migrationAssistant.handleAlreadyIdentifiedMigratedUser())
+        XCTAssertEqual(DataPipeline.shared.analytics.userId, givenProfileIdentifiedInJourneys)
+        XCTAssertNil(profileStoreMock.identifier)
```
levibostian, line 83:
```suggestion
        let givenProfileIdentifiedInJourneys = String.random
        profileStoreMock.identifier = givenProfileIdentifiedInJourneys        
        XCTAssertNil(DataPipeline.shared.analytics.userId)
        
        migrationAssistant.handleAlreadyIdentifiedMigratedUser()
        
        XCTAssertEqual(DataPipeline.shared.analytics.userId, givenProfileIdentifiedInJourneys)
        XCTAssertNil(profileStoreMock.identifier)
```

I think this test could be improved by adding the `XCTAssertNil(DataPipeline.shared.analytics.userId)` line to make sure the state of the SDK is what we expect before and after calling `migrationAssistant.handleAlreadyIdentifiedMigratedUser()`.

Tests/Tracking/migration/DataPipelineMigrationAssistantTests.swift
levibostian, line 1: I think we need another test that calls `handleAlreadyIdentifiedMigratedUser` multiple times and we expect a migration to only occur the first time we call it and not future times that we call it.

levibostian (changes requested):
I added suggestions to make the tests cover more use cases.

Sources/Tracking/migration/DataPipelineMigrationAssistant.swift
ami-aman: Device token should be updated on [initialising the SDK](https://github.com/customerio/customerio-ios/blob/10168a3e6807c52404d66e96aeb49ec881f3956d/Sources/Common/CustomerIOInstance.swift#L128) from stored value in globalDataStore.

ami-aman: If this make sense to you then please feel free to resolve the conversation.

Tests/Tracking/migration/DataPipelineMigrationAssistantTests.swift
ami-aman: I am removing `migrationAssistantMock` as it is not in use and was added as a part of a dev test.

Tests/Common/Extensions/DateExtensionTests.swift
```diff
@@ -59,4 +59,18 @@ class DateExtensionTest: UnitTest {
...
+    // MARK: formatToIso8601WithMilliseconds
+
+    func test_givenDate_formatToIso8601WithMilliseconds_expectString() {
+        let date = dateUtilStub.now
+        let formattedDate = date.string(format: .iso8601WithMilliseconds)
+
+        let formatter = ISO8601DateFormatter()
+        formatter.formatOptions = [.withInternetDateTime, .withFractionalSeconds]
+        formatter.timeZone = TimeZone(identifier: "UTC")
+        let expectedDate = formatter.string(from: date)
```
levibostian, line 72: This test might accurately test the use case, but unless I fully understand this highlighted block of code, I cannot be certain. 

Also, is this highlighted block of code has a bug, then we cannot be sure if the code under test is working or not. 

As a suggestion, I suggest using a hard-coded string as the expected value. Something like: 

```swift
    func test_givenDate_formatToIso8601WithMilliseconds_expectString() {
        let givenDate = Date(...)
        let expectedString = "2023-12-04 ..."

        let actualString = givenDate.string(format: .iso8601WithMilliseconds)

        XCTAssertEqual(expectedString, actualString)
     }
```

Tests/Tracking/migration/DataPipelineMigrationAssistantTests.swift
levibostian: Thanks for writing a test for this. 

Looking at the test function body, it doesn't test the use case that I was referring to. 

The use case I had in mind was: 
* A profile is identified in Journeys SDK
* The SDK gets initialized, so the CDP migration occurs. 
* The app gets killed. 
* The SDK get initialized the next time the app opens. 
* Because the SDK has been migrated to use the CDP SDK, we expect the CDP migration to not happen again. 

I was expecting code similar to: 
```swift
        profileStore.identifier = givenIdentifier 

        XCTAssertNil(DataPipeline.shared.analytics.userId)
        let didMigrationRun = migrationAssistant.handleAlreadyIdentifiedMigratedUser()
        AssertTrue(didMigrationRun)
        XCTAssertEqual(DataPipeline.shared.analytics.userId, givenIdentifier)

        // The next time the SDK gets initialized and the migration runs again: 
        let didMigrationRunAgain = migrationAssistant.handleAlreadyIdentifiedMigratedUser()
        AssertFalse(didMigrationRunAgain)
    }
```

ami-aman: @levibostian `handleAlreadyIdentifiedMigratedUser` does not return anything. I have written a similar test case. The userId gets updated only once. If it has already been set then we do not expect any change. Request you to recheck the test case and let me know if there's a miss ?

Tests/Common/Extensions/DateExtensionTests.swift
ami-aman: Updated https://github.com/customerio/customerio-ios/pull/490/commits/fa796e28a0e2e8150c301e4afce7af98a3f389c8

Tests/Tracking/migration/DataPipelineMigrationAssistantTests.swift
```diff
@@ -59,4 +64,45 @@ class DataPipelineMigrationAssistantTests: UnitTest {
...
+
+        // Update the user identifier and re-call handleAlreadyIdentifiedMigratedUser
+        // to ensure the user does not undergo the migration process again
+        // after being identified on the CDP
+        let updatedIdentifier = String.random
+        profileStoreMock.identifier = updatedIdentifier
+        XCTAssertNotNil(migrationAssistant.handleAlreadyIdentifiedMigratedUser())
+        XCTAssertEqual(DataPipeline.shared.analytics.userId, givenIdentifier)
+        XCTAssertEqual(profileStoreMock.identifier, updatedIdentifier)
+    }
```
levibostian, line 107: I think that this test function is trying to test [this use case](https://github.com/customerio/customerio-ios/pull/490/files#r1476464247) suggested in code review. 

If that is true, then here is a suggestion I have for this test body. 

```suggestion
    func test_migrateProfileId_expectMigrationCodeRunOnce() {
        // profile previously identified in SDK, before CDP migration
        let givenIdentifier = String.random
        profileStoreMock.identifier = updatedIdentifier
        XCTAssertNil(DataPipeline.shared.analytics.userId)
        
        migrationAssistant.handleAlreadyIdentifiedMigratedUser()
        
        // CDP migration is performed for the first time in the SDK. 
        migrationAssistant.handleAlreadyIdentifiedMigratedUser()
        // Check that the migration was successful:
        XCTAssertEqual(DataPipeline.shared.analytics.userId, givenIdentifier)

        // Update the user identifier and re-call handleAlreadyIdentifiedMigratedUser
        // to ensure the user does not undergo the migration process again
        // after being identified on the CDP
        let updatedIdentifier = String.random
        profileStoreMock.identifier = updatedIdentifier
        
        migrationAssistant.handleAlreadyIdentifiedMigratedUser()
        
        // We expect the CDP profile ID is the same value from the 1st migration done. 
        XCTAssertEqual(DataPipeline.shared.analytics.userId, givenIdentifier)
    }
```

Tests/Tracking/migration/DataPipelineMigrationAssistantTests.swift
levibostian: Left a suggestion: https://github.com/customerio/customerio-ios/pull/490/files#r1479932261

Tests/Tracking/migration/DataPipelineMigrationAssistantTests.swift
ami-aman: Thanks for the suggestion. The suggested code breaks, hence I am fixing it and making the update

levibostian (approved):

PR #492, opened 2024-01-31
refactor: parse a push notification not coupled to UserNotifications framework

levibostian:
https://linear.app/customerio/issue/MBL-68/parse-a-clicked-push-notification-de-coupled-from-usernotifications

The current class that parses push notifications, CustomerIOParsedPushPayload, is tightly coupled with the UserNotifications framework. In order to write automated tests for SDK push features, we need to parse push notifications in a de-coupled way.

This commit updates the de-coupled PushNotifications data type to have feature parity with the CustomerIOParsedPushPayload data type.


Sources/MessagingPush/CustomerIOParsedPushPayload.swift
levibostian, line 1: This file is getting deleted in next PR. So, this diff doesn't matter very much here.

Sources/MessagingPush/UserNotificationsFramework/Wrappers.swift
levibostian, line 1: This file got updated because the `PushNotification` protocol got changed.

Tests/MessagingPush/CustomerIOParsedPushPayloadTest.swift
levibostian, line 1: Deleted file because it's no longer needed. The test cases are covered now in `Tests/MessagingPush/Type/PushNotificationTest.swift`

Sources/MessagingPush/Type/PushNotification.swift
levibostian, line 1: This is the main focus of this PR. Updating the de-coupled `PushNotification` protocol to have feature parity with `CustomerIOParsedPushPayloadTest`. 

I also added some comments for future maintenance.

Sources/MessagingPush/Type/PushNotification.swift
```diff
@@ -3,9 +3,115 @@ import Foundation
...
     var pushId: String { get }
-    var deliveryDate: Date { get }
-    var title: String { get }
-    var message: String { get }
-    var data: [AnyHashable: Any] { get }
+
+    // Date when the push was delivered to the device.
+    // Some subclasses of UserNotifications framework does not include a delivery date so we have to keep this optional.
+    // Depending on the request that the framework makes to our SDK will determine if we get a date or not.
+    var deliveryDate: Date? { get }
```
Shahroz16, line 14: good call out. Are we using it, if so what for, and if not, why are we adding it?

Sources/MessagingPush/Type/PushNotification.swift
```diff
@@ -3,9 +3,115 @@ import Foundation
...
+            data["CIO"] as? [AnyHashable: Any]
+        }
+        set {
+            data["CIO"] = newValue
+        }
+    }
+
+    var cioPushPayload: [AnyHashable: Any]? {
+        get {
+            cioPayload?["push"] as? [AnyHashable: Any]
```
Shahroz16, line 60: is there a way to make it more type-safe? not just here but below too?

Sources/MessagingPush/Type/PushNotification.swift
levibostian: > Are we using it, if so what for

Yes, [it's used to make sure we do not handle the same push 2+ times](https://github.com/customerio/customerio-ios/blob/2f8dddfbc20a24ae1c56abf5d6ed501106f4201a/Sources/MessagingPush/Store/PushHistory.swift#L20).

Sources/MessagingPush/Type/PushNotification.swift
levibostian: I believe you're suggestion is in regards to the return value. Looking for something like this: 

`var cioPushPayload: CioPushPayload?` instead of the generic `[AnyHashable: Any]?`? 

I did consider that, but I do think that it makes this class more complex and harder to read without adding much value. 

I think type-safety would be really valuable for `cioPayload` and `cioPushPayload` properties, but only if they were `public` and meant to be used outside of this class. Which, these 2 properties are not designed to be used outside of this class. 

As an alternative solution, if `cioPayload` and `cioPushPayload` were set to `private`, I no longer think type-safety would be needed. 

Does that sound good? If not, have a suggestion?

Shahroz16 (approved):

Sources/MessagingPush/Type/PushNotification.swift
Shahroz16: That does sound reasonable, additionally, I have not tried it but could we initialize these values on `init`? as in instead of 
first,  `cioPayload used as [AnyHashable: Any]?` and then `cioPushPayload`
retrieved via `cioPayload?["push"] as? [AnyHashable: Any]`

could we just use `data` and initialize all these variables?

levibostian: > could we initialize these values on init? 
Doing this might not be possible because `PushNotification` is a protocol and not a struct/class. We would need to do a refactor to do this idea. 

Also, I am not sure if I see value in doing this. I like how `data` is the payload's single-source-of-truth. If we were to make `cioPayload`, for example, it's own `var` variable on the `PushNotification` object instead of this `getter/setter` dynamic property, then we would need to maintain multiple variables instead of 1, `data`.

levibostian: I made the fields `private` https://github.com/customerio/customerio-ios/pull/492/commits/406143aca7caeddb10902fe219dbb9abe1aa9726

PR #493, opened 2024-01-31
refactor: replace CustomerIOParsedPushPayload with de-coupled alternative

levibostian:
Part of de-coupling CustomerIOParsedPushPayload from UserNotifications framework.

This PR deletes the CustomerIOParsedPushPayload class and then gets the code to compile again by instead using the de-coupled PushNotification protocol.

Because the CustomerIOParsedPushPayload object is part of the public-API of the SDK, a backwards compatible alias is created to not break the API.


Sources/MessagingPush/Type/CustomerIOParsedPushPayload.swift
levibostian, line 1: The SDK's functions called in a NSE target to compose a rich push notification currently returns the `CustomerIOParsedPushPayload` data type. Therefore, this data type is part of the public-API of our SDK. 

This file tries to keep backwards compatibility while also not trying to add a lot of work for us in the long-term to maintain this code.

Sources/MessagingPush/RichPush/MessagingPush+RichPush.swift
```diff
@@ -51,10 +51,9 @@ extension MessagingPushImplementation {
         logger.info("did receive notification request. Checking if message was a push sent from Customer.io...")
         logger.debug("notification request: \(request.content.userInfo)")
 
-        guard let parsedPush = CustomerIOParsedPushPayload.parse(
-            notificationContent: request.content,
-            jsonAdapter: jsonAdapter
-        ) else {
+        let push = UNNotificationWrapper(notificationRequest: request)
+
+        guard let pushCioDeliveryInfo = push.cioDelivery else {
```
Shahroz16, line 56: hmm `isPushSentFromCio` would have made more sense here, maybe we get this value below?

Sources/MessagingPush/RichPush/RichPushRequest.swift
```diff
@@ -29,7 +26,7 @@ class RichPushRequest {
             guard let self = self else { return }
 
             if let localFilePath = localFilePath {
-                self.pushContent.addImage(localFilePath: localFilePath)
+                self.push.cioRichPushImageFile = localFilePath
```
Shahroz16, line 29: directly mutating push within the same class can lead to issues, especially in a multi-threaded environment as `start()` could potentially be called from different threads.

should we ensure modifications to push are thread-safe.?

levibostian: Glad you are asking questions about threading. Always important to ask. 

This PR is meant to be a full refactor, where logic is not modified. So with this line of code, the original implementation was also performing a mutation. 

Since this code is stable in production today, and my understanding that this class will not be called by multiple threads in a NSE target, I do not think we need to consider this use case in this class.

Sources/MessagingPush/RichPush/MessagingPush+RichPush.swift
levibostian: I had the same thought when I wrote the code. I would also prefer to use `isPushSentFromCio`, but I was also trying to avoid any force optionals: 

```swift 
let push = UNNotificationWrapper(notificationRequest: request)

guard push.isPushSentFromCio else {
  return 
}

let pushCioDeliveryInfo = push.cioDelivery! <--- what I want to avoid
```

Perhaps, this would be more readable and avoid a force optional? 

```swift 
let push = UNNotificationWrapper(notificationRequest: request)

guard push.isPushSentFromCio, let pushCioDeliveryInfo = push.cioDelivery else {
  return 
}
```

Sources/MessagingPush/RichPush/RichPushRequest.swift
Shahroz16: Good points, unfortunately didn't review or missed in initial implementation. 

Following things are in my mind
- Previously `self.push` alternative was a `immutable` its mutable now.

> class will not be called by multiple threads in a NSE target
I am not sure how certain can we be on this? because i haven't found any documentation on it. Even this is true we still need to make sure we are catering for multi threading, while making http requests that might run on background thread while NSE execution might be being done on main. 

> This PR is meant to be a full refactor, where logic is not modified. So with this line of code, the original implementation was also performing a mutation.
I am okay, if this is not taken care of in this PR, that would make sense. But can we cater in another PR and have ticket for it?

Shahroz16 (approved):
Taking the conversation offline as it's non blocker for this PR.

PR #495, opened 2024-02-01
fix: keep UNUserNotificationCenterDelegate instance in-memory after re-initialize SDK

levibostian:
https://linear.app/customerio/issue/MBL-69/assert-the-sdks-usernotification-objects-do-not-get-deleted-from

The SDK comes with a `UNUserNotificationCenterDelegate` instance that is provided to the OS. The OS keeps a reference to it and will call it to handle push notification events, such as when a push is clicked.

It's very important that the object that the OS has in-memory, stays in-memory otherwise the SDK will not be called to handle push events.


Sources/MessagingPush/PushHandling/iOSPushEventListener.swift
```diff
@@ -3,94 +3,32 @@ import CioTracking
...
-    private var overridePushClickHandler: PushClickHandler?
-    private var overridePushHistory: PushHistory?
-
-    // Below is a set of getters for all dependencies of this class.
-    // Each getter will first check if a test override exists. If so, return that. Otherwise, return an instance from the digraph.
-    // It's important that we use do not keep a strong reference to any dependencies or to the digraph. Otherwise, the SDK would crash if the SDK's digraph gets re-initialized and
-    // this class tries to access old instances of dependencies.
-    private var jsonAdapter: JsonAdapter? {
-        overrideJsonAdapter ?? diGraph?.jsonAdapter
-    }
```
levibostian, line 46: All of this code that enabled this class to be created and maintained in-memory outside of the digraph still is required, but has been moved into `Sources/MessagingPush/UserNotificationsFramework/UserNotificationsFrameworkAdapter.swift` because that is the SDK's new `UNUserNotificationCenterDelegate` instance. 

This class used to be the delegate, but got changed after we de-coupled this class from the `UserNotifications` framework.

Tests/MessagingPush/UserNotificationsFramework/UserNotificationsFrameworkAdapterTest.swift
```diff
@@ -0,0 +1,32 @@
...
+    // MARK: UNUserNotificationCenterDelegate
+
+    /**
+     The SDK comes with a `UNUserNotificationCenterDelegate` instance that is provided to the OS. The OS keeps a reference to it and will call it to handle push notification events, such as when a push is clicked.
+
+     It's very important that the object that the OS has in-memory, stays in-memory otherwise the SDK will not be called to handle push events.
+
+     This test makes sure that the `UNUserNotificationCenterDelegate` instance stays in-memory, even after other SDK dependencies may be destroyed.
+     */
+    func test_expectSdkUNUserNotificationCenterDelegateInMemoryAfterSdkReinitialized() {
```
levibostian, line 16: New test that makes sure that the `UNUserNotificationCenterDelegate` delegate stays in-memory.

Sources/MessagingPush/UserNotificationsFramework/UserNotificationsFrameworkAdapter.swift
```diff
@@ -22,19 +23,37 @@ protocol UserNotificationsFrameworkAdapter {
...
+     This class is the SDK's `UNUserNotificationCenterDelegate` instance.  Meaning, the CIO SDK registers this class with the OS in order to
+     receive callbacks when push notifications are interacted with.
+
+     It's important that the instance of this class provided to the OS stays in memory so our SDK can receive those OS callbacks.
+
+     In order to promise this class's singleton instance stays in memory, the singleton instance is *not* stored inside of the digraph (like all other singletons in our SDK is).
+     */
+    static let shared = UserNotificationsFrameworkAdapterImpl()
+
+    // It's important that we use do not keep a strong reference to any dependencies or to the digraph. Otherwise, the SDK would crash if the SDK's digraph gets re-initialized and
```
Shahroz16, line 40: That is a good call out but it is still something that if not managed correctly or by mistake/negligence can cause unexpected results. Can we avoid it? as in not using `PushEventHandler` at all for `DiGraph` just relying on stub/actual instance even for testing.

Shahroz16: Ah, I see your point, unfortunately with the current `digraph` implementation doesn't seem possible but I can easily someone holding a reference here by mistake. Another could be to take all these classes out of `digraph` but that would mean way more changes. 

Can we please a ticket about this so we don't lose sight of it? and we can probabaly try to create a way in our Sourcery template to have some objects as `App lifecycle` available.

Shahroz16 (approved):

Sources/MessagingPush/UserNotificationsFramework/UserNotificationsFrameworkAdapter.swift
levibostian: [ticket created](https://linear.app/customerio/issue/MBL-73/di-graph-gets-re-created-while-sdk-in-memory).

levibostian:
Tests passed on CI, but failed to upload codecov report. 

Because this branch is merging into another feature branch, I am going to unblock this PR by merging as-is.

PR #499, opened 2024-02-03
chore: added support for cocoapods

Shahroz16:


CustomerIODataPipelines.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |spec|
   spec.name         = "CustomerIODataPipelines"
-  spec.version      = "2.10.1" # Don't modify this line - it's automatically updated
+  spec.version      = "2.10.2" # Don't modify this line - it's automatically updated
```
mrehan27, line 3: Isn't this auto updated? 🤔

mrehan27 (approved):
LGTM 👍🏻

CustomerIODataPipelines.podspec
Shahroz16: it would have been if this was part of `main` but unless its merged in main, it doesn't get updated with new version being rolled out.

PR #500, opened 2024-02-05
chore: fix tests after dependency updates

mrehan27:
helps: [MBL-50](https://linear.app/customerio/issue/MBL-50/resolve-test-failures-caused-by-dependency-updates)

###  Changes

- Fixed test imports
- Updated unit tests to utilize `setUpSharedInstanceForUnitTest`
- Resolved compilation issues in following tests due to dependency updates:
  - `DataPipelineCompatibilityTests`
  - `DataPipelineInteractionTests`
  - `DeviceAttributesProviderTest`
  - `MessagingPush` API tests


Sources/DataPipeline/DataPipelineConfigOptions.swift
```diff
@@ -26,6 +26,7 @@ public struct DataPipelineConfigOptions {
         public static func create(sdkConfig: SdkConfig) -> DataPipelineConfigOptions {
             let writeKey = "\(sdkConfig.siteId):\(sdkConfig.apiKey)"
             var result = DataPipelineConfigOptions(writeKey: writeKey)
+            result.autoTrackDeviceAttributes = sdkConfig.autoTrackDeviceAttributes
```
mrehan27, line 29: The test device attributes was breaking we now use `DataPipelineConfigOptions` and it was not copied from `SdkConfig` mistakenly

Tests/DataPipeline/DataPipelineCompatibilityTests.swift
```diff
@@ -7,23 +7,29 @@ import XCTest
 
 private typealias SavedEvent = [String: Any]
 
-class DataPipelineCompatibilityTests: UnitTest {
+class DataPipelineCompatibilityTests: IntegrationTest {
```
mrehan27, line 10: Updated the test to inherit from `IntegrationTest` instead for two reasons: 1) `DeviceInfoStub` is already available in `IntegrationTest`, and 2) since we are testing multiple aspects, such as the attachment of the device token and the emission of events, `IntegrationTest` seemed like the more suitable option for this test.

Tests/DataPipeline/DataPipelineCompatibilityTests.swift
```diff
@@ -34,7 +40,7 @@ class DataPipelineCompatibilityTests: UnitTest {
     func test_identifyWithoutAttributes_expectFinalJSONHasCorrectKeysAndValues() {
         let givenIdentifier = String.random
 
-        customerIO.identify(identifier: givenIdentifier)
+        customerIO.identify(userId: givenIdentifier)
```
mrehan27, line 43: Because we can no longer inherit `Tracking` package in `DataPipeline` tests, we can only use what we have exposed in `DataPipeline` module

Tests/DataPipeline/DataPipelineInteractionTests.swift
```diff
@@ -5,19 +5,17 @@ import Foundation
 @testable import SharedTests
 import XCTest
 
-class DataPipelineInteractionTests: UnitTest {
+class DataPipelineInteractionTests: IntegrationTest {
```
mrehan27, line 8: Same as above.

Updated the test to inherit from `IntegrationTest` instead for two reasons: 1) `DeviceInfoStub` is already available in `IntegrationTest`, and 2) since we are testing multiple aspects, such as the attachment of the device token and the emission of events, `IntegrationTest` seemed like the more suitable option for this test.

Tests/MessagingPushAPN/APITest.swift
```diff
@@ -27,13 +27,11 @@ class MessagingPushAPNAPITest: UnitTest {
 
         MessagingPush.shared.application(
             "",
-            didFailToRegisterForRemoteNotificationsWithError: CustomerIOError
-                .notInitialized
+            didFailToRegisterForRemoteNotificationsWithError: MessagingPushError.registrationFailed
```
mrehan27, line 29: Replaced with generic error as `CustomerIOError` is no longer accessible by `MessagingPush`

Tests/MessagingPushFCM/APITest.swift
```diff
@@ -30,10 +30,9 @@ class MessagingPushFCMAPITest: UnitTest {
 
         MessagingPush.shared.application(
             "",
-            didFailToRegisterForRemoteNotificationsWithError: CustomerIOError
-                .notInitialized
+            didFailToRegisterForRemoteNotificationsWithError: MessagingPushError.registrationFailed
```
mrehan27, line 32: Replaced with generic error as `CustomerIOError` is no longer accessible by `MessagingPush`

Tests/Shared/SampleData/MessagingPushError.swift
```diff
@@ -0,0 +1,7 @@
+import Foundation
+
+/// Class to represent generic errors related to push notification operations, specifically for testing MessagingPush API.
+public enum MessagingPushError: Error {
+    /// Indicates a failure in the registration process for push notifications.
+    case registrationFailed
```
mrehan27, line 6: Random error to test `MessagingPush` APIs

Sources/DataPipeline/DataPipelineConfigOptions.swift
levibostian: Do we have a test for this `DataPipelineConfigOptions.create(sdkConfig)` function? Looks like one could be valuable to catch bugs like this where we forget to add a `result.N = sdkConfig.N` line.

Tests/Shared/SampleData/MessagingPushError.swift
levibostian: This class has the comment: 

> /// Class to represent generic errors related to push notification operations, specifically for testing MessagingPush API.

I was expecting a class name that matched that description. Maybe `GenericError` as class name? `MessagingPushError.registrationFailed` sounds specific and sounds like a real error that could happen. If this error case was a real error that could be thrown, I would expect that error class to be in the `Sources/` module instead of `Tests/`. 

This is a non-blocking suggestion.

levibostian (approved):

Tests/Shared/SampleData/MessagingPushError.swift
mrehan27: Updated to `GenericError`

Sources/DataPipeline/DataPipelineConfigOptions.swift
mrehan27: Unfortunately, we don't have a test for this function yet. I would love to add one though, but I'm unsure about the type of test that would ensure its reliability and fail every time we add a property to `DataPipelineConfigOptions` unless we specify that it's not from `SdkConfig`. If you have any ideas, please do share, and I can then work on adding the necessary tests.

PR #501, opened 2024-02-05
chore: fix tests after dependency updates

mrehan27:
helps: [MBL-50](https://linear.app/customerio/issue/MBL-50/resolve-test-failures-caused-by-dependency-updates)

###  Changes

- Fixed `DIGraph` tests to inherit `UnitTest` instead of `IntegrationTest`
- Updated stencil so `DIGraphShared` can also be tested for resolving dependencies

### DIFF
--- a/Tests/Common/DIGraphTests.swift
+++ b/Tests/Common/DIGraphTests.swift
@@ -3,7 +3,7 @@ import Foundation
 import SharedTests
 import XCTest
 
-class DIGraphTests: IntegrationTest {
+class DIGraphTests: UnitTest {
     func testDependencyGraphComplete() {
         diGraph.testDependenciesAbleToResolve() // test will fail if an exception occurs while running this function
     }
--- a/Tests/Common/DIGraphTests.swift
+++ b/Tests/Common/DIGraphTests.swift
@@ -5,6 +5,7 @@ import XCTest
 
 class DIGraphTests: UnitTest {
     func testDependencyGraphComplete() {
+        diGraphShared.testDependenciesAbleToResolve() // test will fail if an exception occurs while running this function
         diGraph.testDependenciesAbleToResolve() // test will fail if an exception occurs while running this function
     }
 }
--- a/Tests/Shared/DIGraphTest.swift
+++ b/Tests/Shared/DIGraphTest.swift
@@ -18,7 +18,7 @@ import XCTest
  }
  ```
  */
-open class BaseDIGraphTest: IntegrationTest {
+open class BaseDIGraphTest: UnitTest {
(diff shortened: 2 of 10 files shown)


Tests/Common/DIGraphTests.swift
```diff
@@ -3,8 +3,9 @@ import Foundation
 import SharedTests
 import XCTest
 
-class DIGraphTests: IntegrationTest {
+class DIGraphTests: UnitTest {
```
mrehan27, line 6: Updated the test to inherit from `UnitTest` instead because `Common` module has no longer access to any `IntegrationTest` and we can neither really initialize the SDK or any module in it. However, the test should work fine using `UnitTest` too.

Tests/Shared/DIGraphTest.swift
```diff
@@ -18,13 +18,19 @@ import XCTest
  }
  ```
  */
-open class BaseDIGraphTest: IntegrationTest {
+open class BaseDIGraphTest: UnitTest {
```
mrehan27, line 21: Same as above.

Updated the test to inherit from `UnitTest` instead because `Shared` module also has no longer access to any `IntegrationTest` and we can neither really initialize the SDK or any module in it. However, the test should work fine using `UnitTest` too.

Tests/Common/DIGraphTests.swift
levibostian: This might be beyond the scope of this PR, but I wonder...

I see this test class is in `Common` module. This PR includes modifications to digraph in other modules such as `Sources/MessagingPush/autogenerated/AutoDependencyInjection.generated.swift`. Because of that, I was expecting this PR to also be modifying a test class such as: `Tests/MessagingPush/DIGraphTests.swift`, but it looks like that file doesn't exist today. Even in `main`. 

So, I don't think that the file `Sources/MessagingPush/autogenerated/AutoDependencyInjection.generated.swift` that got modified in this file is getting tested?

levibostian (approved):

Tests/Common/DIGraphTests.swift
mrehan27: Sorry, couldn't get your point completely. I can see we have [`DIGraphTests` in `MessagingPush`](https://github.com/customerio/customerio-ios/blob/main-replica-for-cdp/Tests/MessagingPush/DITests.swift). Isn't it enough to test changes for `Sources/MessagingPush/autogenerated/AutoDependencyInjection.generated.swift`?

Also, after reading your comment, I do realize it was missing for `DataPipeline` so [added them](https://github.com/customerio/customerio-ios/pull/501/commits/e3564ea0112ea312ab4226559f684b7c2c4d3adb).

PR #506, opened 2024-02-07
chore: fix in-app tests

mrehan27:
helps: [MBL-14](https://linear.app/customerio/issue/MBL-14/fix-tests-broken-due-to-datapipeline-changes)

###  Changes

- Updated `MessagingInAppTest` and `MessagingInAppImplementationTest` to compile and utilize latest changes
- Matched with `main` to make sure we are not removing any test unknowingly


Tests/MessagingInApp/MessagingInAppTest.swift
```diff
@@ -66,36 +49,18 @@ class MessagingInAppTest: UnitTest {
...
-
-    // MARK: initialize functions with SDK not initialized
-
-    func test_initialize_givenSdkNotInitialized_expectModuleIsInitialized() {
-        sdkInitializedUtilMock.underlyingIsInitlaized = false
-
-        MessagingInApp.initialize(siteId: .random, region: .US)
-
-        assertModuleInitialized(isInitialized: true, givenEventListener: nil)
-    }
```
mrehan27, line 86: Because modules are now independent of SDK initialization, we don't need these tests at the moment.

Tests/Shared/Core/UnitTestBase.swift
```diff
@@ -14,7 +14,10 @@ import XCTest
 /// For SDK-wide tests, child classes can conveniently inherit from `UnitTest`, designed specifically for testing only SDK APIs.
 open class UnitTestBase<Component>: XCTestCase {
     public let testWriteKey = "test"
-    public var diGraphShared: DIGraphShared!
+    // Initialize DIGraphShared instantly to closely copy production behavior.
+    // The graph in production is initialized statically and can be accessed using .shared.
+    // This also allows us to override any dependency without waiting for SDK/modules to initialize and setup.
+    public var diGraphShared: DIGraphShared = .init()
```
mrehan27, line 20: Updated mainly for `test_initialize_givenExistingIdentifier_expectGistSetProfileIdentifier`, but the comment above should help understand why I think it is good to initialize `DIGraphShared` early.

Tests/MessagingInApp/MessagingInAppImplementationTest.swift
```diff
@@ -237,3 +250,18 @@ class MessagingInAppImplementationTest: UnitTest {
...
+        waitForEventBus(timeoutInSeconds: timeoutInSeconds)
+    }
+
+    func waitForEventBus(timeoutInSeconds: Double = 1.0) {
+        let expectation = XCTestExpectation(description: "wait for \(timeoutInSeconds) seconds")
+        DispatchQueue.main.asyncAfter(deadline: .now() + timeoutInSeconds) {
+            expectation.fulfill()
+        }
+        wait(for: [expectation], timeout: timeoutInSeconds)
+    }
```
levibostian, line 266: This eventbus wait function looks like a sleep function (wait X amount of time, hope that the operation completes in that time). 

I strongly want to avoid introducing a sleep pattern in our test suite to avoid our test suite running longer then necessary to run. 

Is there a way that we can have the `XCTestExpectation` only wait for the eventbus to finish? I am assuming I would see code like this: 

```swift
    func postEventAndWait<E: EventRepresentable>(event: E, timeoutInSeconds: Double = 1.0) {
        eventBusHandler.postEvent(event)
        let expectEventBusToFinishPosting = XCTestExpectation
        eventBusHandler.postAllEvents(onComplete: {
          expectEventBusToFinishPosting.fulfill()
        })

        waitfor(expectEventBusToFinishPosting)
    }
```

Is this possible? If so, is it within the scope of this PR or do we need a new ticket for it?

Tests/Shared/Core/UnitTestBase.swift
levibostian: If I understand the impact of this change, I disagree that we should make this change. 

I think this change would initialize the graph instance 1 time in the test class, instead of being re-initialized before every test function. 

If this change was to fix the `test_initialize_givenExistingIdentifier_expectGistSetProfileIdentifier` test, I would like to propose an alternative solution to make that test function correctly.

levibostian (changes requested):
I consider only 1 comment to be blocking: https://github.com/customerio/customerio-ios/pull/506#discussion_r1481900552

Otherwise, rest are non-blocking suggestions. Looks good!

Tests/Shared/Core/UnitTestBase.swift
mrehan27: Yes, it won't be re-initialized before every test function. But we do `reset` it on every `tearDown`. So far, this hasn't break anything. But if this can give us false results in future, we can surely try another approach to fix this specific test.

Tests/MessagingInApp/MessagingInAppImplementationTest.swift
```diff
@@ -38,42 +43,45 @@ class MessagingInAppImplementationTest: UnitTest {
...
-        messagingInApp.profileIdentified(identifier: given)
+        postEventAndWait(event: ProfileIdentifiedEvent(identifier: given))
 
         XCTAssertEqual(inAppProviderMock.setProfileIdentifierCallsCount, 1)
         XCTAssertEqual(inAppProviderMock.setProfileIdentifierReceivedArguments, given)
     }
 
     func test_givenProfileNoLongerIdentified_expectRemoveFromInApp() {
-        messagingInApp.beforeProfileStoppedBeingIdentified(oldIdentifier: String.random)
+        super.setUp()
```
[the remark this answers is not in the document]
mrehan27: Makes sense. We can call `postEventAndWait(event: ProfileIdentifiedEvent(identifier: String.random))` first to improve the test, I'll update the code.

Tests/MessagingInApp/MessagingInAppImplementationTest.swift
mrehan27: Unfortunately, no. [`EventBus`](https://github.com/customerio/customerio-ios/blob/main-replica-for-cdp/Sources/Common/Communication/EventBusHandler.swift#L100-L110) does not have anything today that we can utilize to confirm the event emission. However, if we want to update it, we can create separate ticket for it and update other EventBus tests in it too.

Sources/Tracking/autogenerated/AutoMockable.generated.swift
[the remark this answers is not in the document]
mrehan27: Merged changes from base branch

Tests/MessagingInApp/MessagingInAppImplementationTest.swift
levibostian: I had a feeling that we may need to add this functionality to the EventBus. 

Thanks for pointing out that the EventBus already has some other tests today that would also need changed. 

Because of this scope, I agree, creating a separate ticket would be a good move. [I made one here](https://linear.app/customerio/issue/MBL-95/reduce-time-it-takes-for-test-suite-to-execute-by-replacing-sleep).

Tests/Shared/Core/UnitTestBase.swift
levibostian: Ah, I was only looking at `setup()` and did not see we are calling `reset` in `teardown`. 

Because we are calling `reset` between each test function running, I think my call-out wouldn't be an issue. 

If performing this `.init()` here causes side effects in our tests, I don't think that this is a good idea. With the context that you have with the test suite changes, if you can't think of any side effects of doing this change, then I am OK with resolving this.

levibostian (approved):
Rest of my suggestions are non-blocking

Tests/MessagingInApp/MessagingInAppImplementationTest.swift
mrehan27: Thanks for creating the ticket

Tests/Shared/Core/UnitTestBase.swift
mrehan27: Sounds good. I also [slightly adjusted `tearDown`](https://github.com/customerio/customerio-ios/pull/506/commits/689e1e50d46cb7da6eeed0b03cb14e5757ac0e5c) to ensure graphs are reset to their exact initial state. This way, if any test fails, it's more likely due to an issue with the test itself or something that needs correction. Thanks for highlighting.

PR #514, opened 2024-02-09
fix: one of the manual push tracking functions doesnt track metrics

levibostian:
While reviewing the diff for the new automatic push click handling feature, I noticed that one of the changes I made would result in push metrics not being reported, for one of the manual push tracking functions in our public API.

I reverted this function back to logic that is similar to what you see in `main` to fix the issue.

Now that our push module can have automated tests written against it, we can catch bugs like this in the future. Writing these tests is out of scope for this bug fix and can be introduced in the future.

### DIFF
--- a/Sources/MessagingPush/PushHandling/ManualPushHandling+UserNotifications.swift
+++ b/Sources/MessagingPush/PushHandling/ManualPushHandling+UserNotifications.swift
@@ -59,11 +59,12 @@ extension MessagingPushImplementation {
         didReceive response: UNNotificationResponse,
         withCompletionHandler completionHandler: @escaping () -> Void
     ) -> Bool {
-        let push = UNNotificationWrapper(notification: response.notification)
-
-        guard push.isPushSentFromCio else {
-            // Exit early without calling completionHandler so that customer calls it instead.
-
+        // to keep this code DRY, forward the request to another function to perform all the logic:
+        guard let pushContent = userNotificationCenter(center, didReceive: response) else {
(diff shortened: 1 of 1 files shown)


Ahmed-CIO (approved):
The code itself looks fine to me.
I would also appreciate updating the PR description to include how did you made sure this code still works? this specially important because the code alone won't be enough to understand this change.
Alternatively, if a test automation isn't existing to ensure that, writing one in this PR can also help verifying that.

Sources/MessagingPush/PushHandling/ManualPushHandling+UserNotifications.swift
```diff
@@ -59,11 +59,12 @@ extension MessagingPushImplementation {
         didReceive response: UNNotificationResponse,
         withCompletionHandler completionHandler: @escaping () -> Void
     ) -> Bool {
-        let push = UNNotificationWrapper(notification: response.notification)
-
-        guard push.isPushSentFromCio else {
-            // Exit early without calling completionHandler so that customer calls it instead.
-
+        // to keep this code DRY, forward the request to another function to perform all the logic:
+        guard let pushContent = userNotificationCenter(center, didReceive: response) else {
```
Shahroz16, line 63: had to think hard as to why might not it work, but i think its because of this, this was missed? 

```
if response.didClickOnPush {
            manualPushClickHandling(push: push)
        }
```

Shahroz16 (approved):

PR #515, opened 2024-02-09
refactor: fix all lint errors

levibostian:
Fixing all lint errors for automatic push click handling feature changes.

See comments made in the PR for this commit to understand some of the decisions made.

### DIFF
--- a/Sources/MessagingPush/PushHandling/PushEventHandlerProxy.swift
+++ b/Sources/MessagingPush/PushHandling/PushEventHandlerProxy.swift
@@ -29,9 +29,9 @@ class PushEventHandlerProxyImpl: PushEventHandlerProxy {
     private var nestedDelegates: [String: PushEventHandler] = [:]
 
     func addPushEventHandler(_ newHandler: PushEventHandler) {
-        // TODO: this line below seems fragile. If we change the class name, this could break.
+        // this line below seems fragile. If we change the class name, this could break.
         // could digraph inject instance of the SDK's intance before setting singleton?
-        let doesDelegateBelongToCio = newHandler is iOSPushEventListener
+        let doesDelegateBelongToCio = newHandler is IOSPushEventListener
 
         guard !doesDelegateBelongToCio else {
             return
--- a/Sources/MessagingPush/PushHandling/iOSPushEventListener.swift
+++ b/Sources/MessagingPush/PushHandling/iOSPushEventListener.swift
@@ -4,7 +4,7 @@ import Foundation
 
 @available(iOSApplicationExtension, unavailable)
 // sourcery: InjectRegister = "PushEventHandler"
-class iOSPushEventListener: PushEventHandler {
+class IOSPushEventListener: PushEventHandler {
     private let jsonAdapter: JsonAdapter
     private var pushEventHandlerProxy: PushEventHandlerProxy
     private let moduleConfig: MessagingPushConfigOptions
--- a/Tests/.swiftlint.yml
+++ b/Tests/.swiftlint.yml
@@ -23,3 +23,7 @@ disabled_rules:
   # for tests. 
   - type_body_length
   - file_length 
+  # It's recommended to not use force casts in test suite because if a cast fails, the test suite will immediately stop executing. 
+  # We want the test suite to always finish executing all tests so we can see all the failures and fix them faster. 
+  # Instead of force casting, consider using XCTFail() to fail the test function if a cast is not successful. 
+  # - force_cast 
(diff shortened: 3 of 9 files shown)


Sources/MessagingPush/PushHandling/PushEventHandlerProxy.swift
```diff
@@ -29,9 +29,9 @@ class PushEventHandlerProxyImpl: PushEventHandlerProxy {
     private var nestedDelegates: [String: PushEventHandler] = [:]
 
     func addPushEventHandler(_ newHandler: PushEventHandler) {
-        // TODO: this line below seems fragile. If we change the class name, this could break.
+        // this line below seems fragile. If we change the class name, this could break.
```
levibostian, line 32: TODO is removed because [we have a ticket](https://linear.app/customerio/issue/MBL-70/assert-that-sdk-does-not-forward-push-click-events-to-itself) that will take care of this.

Sources/MessagingPush/PushHandling/iOSPushEventListener.swift
```diff
@@ -4,7 +4,7 @@ import Foundation
 
 @available(iOSApplicationExtension, unavailable)
 // sourcery: InjectRegister = "PushEventHandler"
-class iOSPushEventListener: PushEventHandler {
+class IOSPushEventListener: PushEventHandler {
```
levibostian, line 7: There is a lint error requiring that all classes start with an uppercase letter. 

If this class was public-facing I may suggest we disable this rule. But being internal, not the biggest deal to use `IOS` 🤷‍♀️

Tests/MessagingPush/PushHandling/AutomaticPushClickedIntegrationTest.swift
```diff
@@ -4,7 +4,9 @@ import Foundation
 import SharedTests
 import XCTest
 
-class AutomaticPushClickHandlingIntegrationTest: IntegrationTest {
+// Integration tests that simulate a push being clicked on a device and
+// how our SDK handles that event.
+class AutomaticPushClickedIntegrationTest: IntegrationTest {
```
levibostian, line 9: Change class name because lint said the class name was too long (> 40 characters).

Tests/MessagingPush/PushHandling/ManualPushHandlingTest.swift
```diff
@@ -8,8 +8,8 @@ class ManualPushHandlingIntegrationTests: IntegrationTest {
     private let customerIOMock = CustomerIOInstanceMock()
 
     // The manual push click handling functions are currently housed in `MessagingPushImplementation`. Get instance for integration test.
-    private var messagingPush: MessagingPushImplementation {
-        MessagingPush.shared.implementation as! MessagingPushImplementation
+    private var messagingPush: MessagingPushImplementation? {
+        notNilOrFail(MessagingPush.shared.implementation as? MessagingPushImplementation)
```
levibostian, line 12: See: https://github.com/customerio/customerio-ios/pull/515/files#diff-1d3434a790c3ca7bb3569d46f740490dd83bd1dbe1db453a46501d91d15d8a06R52-R55

Tests/.swiftlint.yml
```diff
@@ -23,3 +23,7 @@ disabled_rules:
   # for tests. 
   - type_body_length
   - file_length 
+  # It's recommended to not use force casts in test suite because if a cast fails, the test suite will immediately stop executing.
```
levibostian, line 26: Multiple times now, I have gone into this file to disable `force_cast` rule, thinking it isn't a big deal to allow this rule in our test classes. But, then I remember that there are consequences to it. 

So, adding this comment block to remind others of the consequences to disabling force_cast lint rule.

Sources/MessagingPush/PushHandling/PushEventHandlerProxy.swift
Ahmed-CIO: this is one of those instance where I see good value in having a GH issue referenced in the code to make it clear that we have it on our radar.

Sources/MessagingPush/PushHandling/iOSPushEventListener.swift
Ahmed-CIO: Apart from the linter, maybe the prefix iOS isn't needed after all. I mean, it would make sense if we have IOSPushEvenListner, MacOSEventListner, etc. But we don't.

Ahmed-CIO (approved):

Shahroz16 (approved):

PR #516, opened 2024-02-12
test: prevent infinite loop if sdk forwards push event back to itself

levibostian:
Closes: https://linear.app/customerio/issue/MBL-70/assert-that-sdk-does-not-forward-push-click-events-to-itself

The CIO SDK push event handler forwards events to other handlers in the host app, allowing multiple push SDKs to be installed simultaneously.

However, forwarding events back to the CIO SDK push event handler could cause an infinite loop. A previous solution to prevent this was removed as it was fragile and not a good long-term fix. This commit adds a test to simulate the scenario to make sure an infinite loop will not occur.

Luckily, the push notification history feature prevents an infinite loop! No additional fix is needed.

Link to code that prevents infinite loop: https://github.com/customerio/customerio-ios/blob/55df69c3200c61bf2abaaa136a70d5de009ea7f0/Sources/MessagingPush/PushHandling/iOSPushEventListener.swift#L31-L34

### DIFF
--- a/Sources/MessagingPush/PushHandling/PushEventHandlerProxy.swift
+++ b/Sources/MessagingPush/PushHandling/PushEventHandlerProxy.swift
@@ -29,14 +29,6 @@ class PushEventHandlerProxyImpl: PushEventHandlerProxy {
     private var nestedDelegates: [String: PushEventHandler] = [:]
 
     func addPushEventHandler(_ newHandler: PushEventHandler) {
-        // this line below seems fragile. If we change the class name, this could break.
-        // could digraph inject instance of the SDK's intance before setting singleton?
-        let doesDelegateBelongToCio = newHandler is IOSPushEventListener
-
-        guard !doesDelegateBelongToCio else {
-            return
-        }
-
         nestedDelegates[String(describing: newHandler)] = newHandler
     }
 
--- a/Tests/MessagingPush/PushHandling/PushEventHandlerTest.swift
+++ b/Tests/MessagingPush/PushHandling/PushEventHandlerTest.swift
@@ -1,0 +1,41 @@
+@testable import CioMessagingPush
+import Foundation
+import SharedTests
+import XCTest
+
+class PushEventHandlerTest: UnitTest {
+    private var pushEventHandler: PushEventHandler!
+
+    private var pushEventHandlerProxy: PushEventHandlerProxy {
+        diGraph.pushEventHandlerProxy
+    }
+
+    private var pushClickHandler = PushClickHandlerMock()
+
+    override func setUp() {
+        super.setUp()
+
+        pushEventHandler = IOSPushEventListener(jsonAdapter: diGraph.jsonAdapter, pushEventHandlerProxy: pushEventHandlerProxy, moduleConfig: MessagingPushConfigOptions(), pushClickHandler: pushClickHandler, pushHistory: diGraph.pushHistory, logger: log)
+    }
+
+    // MARK: onPushAction
+
+    /*
+     The CIO SDK push event handler forwards push notifications to other push event handlers in the host app.
+
+     But, what if the CIO SDK event handler forwards a push event back to the CIO SDK event handler? This could cause an infinite loop.
+
+     This test simulates that scenario to make sure that an inifinite loop would not happen.
+     */
+    func test_onPushAction_expectNoInfiniteLoopIfSdkForwardsPushEventBacktoSdkAgain() {
+        // The push event handler proxy is what forwards push events to other push event handlers in the app.
+        // To test if an infinite loop would happen, add ourself as a push event handler to get events forwarded to.
+        pushEventHandlerProxy.addPushEventHandler(pushEventHandler)
+
+        // Make sure push is not from CIO otherwise event does not get forwarded.
+        let givenPush = PushNotificationStub.getPushNotSentFromCIO()
+
+        // Send the CIO SDK push event handler an event. If an infinite loop occurs, this test will timeout or crash.
+        pushEventHandler.onPushAction(PushNotificationActionStub(push: givenPush, didClickOnPush: true)) {}
+    }
+}


Tests/MessagingPush/PushHandling/PushEventHandlerTest.swift
```diff
@@ -0,0 +1,41 @@
...
+    func test_onPushAction_expectNoInfiniteLoopIfSdkForwardsPushEventBacktoSdkAgain() {
+        // The push event handler proxy is what forwards push events to other push event handlers in the app.
+        // To test if an infinite loop would happen, add ourself as a push event handler to get events forwarded to.
+        pushEventHandlerProxy.addPushEventHandler(pushEventHandler)
+
+        // Make sure push is not from CIO otherwise event does not get forwarded.
+        let givenPush = PushNotificationStub.getPushNotSentFromCIO()
+
+        // Send the CIO SDK push event handler an event. If an infinite loop occurs, this test will timeout or crash.
+        pushEventHandler.onPushAction(PushNotificationActionStub(push: givenPush, didClickOnPush: true)) {}
```
ami-aman, line 39: Do we expect a timeout or a crash in any case? What I am focussing on is the expected use case for timeout or a crash.

Sources/MessagingPush/PushHandling/PushEventHandlerProxy.swift
```diff
@@ -29,14 +29,6 @@ class PushEventHandlerProxyImpl: PushEventHandlerProxy {
...
 
     func addPushEventHandler(_ newHandler: PushEventHandler) {
-        // this line below seems fragile. If we change the class name, this could break.
-        // could digraph inject instance of the SDK's intance before setting singleton?
-        let doesDelegateBelongToCio = newHandler is IOSPushEventListener
-
-        guard !doesDelegateBelongToCio else {
-            return
-        }
-
```
mrehan27, line 39: Is this the fragile fix you mentioned in the PR description? From what I understand, this was repetition to avoid an infinite loop. The test added now confirms that push history already prevents the infinite loop, so this fix is no longer required, right?

Tests/MessagingPush/PushHandling/PushEventHandlerTest.swift
```diff
@@ -0,0 +1,41 @@
...
+     */
+    func test_onPushAction_expectNoInfiniteLoopIfSdkForwardsPushEventBacktoSdkAgain() {
+        // The push event handler proxy is what forwards push events to other push event handlers in the app.
+        // To test if an infinite loop would happen, add ourself as a push event handler to get events forwarded to.
+        pushEventHandlerProxy.addPushEventHandler(pushEventHandler)
+
+        // Make sure push is not from CIO otherwise event does not get forwarded.
+        let givenPush = PushNotificationStub.getPushNotSentFromCIO()
+
+        // Send the CIO SDK push event handler an event. If an infinite loop occurs, this test will timeout or crash.
```
mrehan27, line 38: Is it possible to mock push history and induce a timeout/crash to more effectively validate our assumptions?

Shahroz16: Same, should we timeout it something like this? 

```
let expectation = expectation(description: "Completion handler should be called")

    // ... Set up your asynchronous operation ...

    // This is crucial for timeout assertions
    waitForExpectations(timeout: 5.0) { error in
        if let error = error {
            XCTFail("Timeout error: \(error)")
        } 
    }
```

Sources/MessagingPush/PushHandling/PushEventHandlerProxy.swift
levibostian: > Is this the fragile fix you mentioned in the PR description?
The `is IOSPushEventListener` line of code above is the most fragile part. 

> so this fix is no longer required, right?
Correct. 

The idea behind this PR is: create a test that will verify that an infinite loop will not occur in the future. Also, remove this fragile code because it's not a sustainable solution and it also does not work. I was able to create an automated test that failed and proved that this fragile code is already outdated.

Tests/MessagingPush/PushHandling/PushEventHandlerTest.swift
levibostian: I was wondering the same thing when I wrote this test. But, I currently cannot think of any way to do this. None of the code under test is async and I don't see how mocking the push history will help, especially when the push history is the reason why an infinite loop does not exist. Therefore, I think it's important to run the real implementation of push history. 

The test function asserts that an infinite loop will not occur. If an infinite loop does occur, then a stackoverflow will occur and cause the test to crash. I think that is enough to validate the test?

Sources/MessagingPush/PushHandling/PushEventHandlerProxy.swift
mrehan27: Makes sense. I just wanted to confirm that the code changes aren't really changing anything and that the added test helps us verify this. All good.

Tests/MessagingPush/PushHandling/PushEventHandlerTest.swift
mrehan27: Since the code that prevents the infinite loop is:

```swift
guard !pushHistory.hasHandledPush(pushEvent: .didReceive, pushId: push.pushId, pushDeliveryDate: dateWhenPushDelivered)
```

And considering `PushHistory` is our class too, if `pushHistory.hasHandledPush` keeps returning `false`, shouldn't that trigger timeout/crash validating our assumptions about this behavior?

mrehan27 (approved):
Approving since none of my comment are blockers 👍🏻

Tests/MessagingPush/PushHandling/PushEventHandlerTest.swift
levibostian: If we wanted to assert that the push history feature is specifically the reason why an infinite loop does not happen, then yes, I can see how that idea could work. 

However, I don't think that is valuable to the test suite. We do not care how the infinite loop is prevented, we only care that an infinite loop does not occur. 

Same goes for [this test](https://github.com/customerio/customerio-ios/blob/216341c932d29caf334da53df3b4ab37b97bc94a/Tests/MessagingPush/PushHandling/AutomaticPushClickedIntegrationTest.swift#L88-L121) that is also testing an infinite loop does not occur. The test sets up a scenario that could create an infinite loop. If this test ever fails, then we have to fix the code and prevent the infinite loop from happening.

levibostian:
Merging, but feel free to keep code review conversations going until they are resolved.

PR #518, opened 2024-02-13
fix: app crash fix on random login

ami-aman:
Fixes app crash on generating a random login since it sends a `nil` as the function parameter. Previously this fix was made in Journeys module. 

 fixes a point from https://linear.app/customerio/issue/MBL-109/findings-from-sample-app-cleanup-and-using-ios-cdp-sdk

### DIFF
--- a/Sources/DataPipeline/CustomerIO+Events.swift
+++ b/Sources/DataPipeline/CustomerIO+Events.swift
@@ -23,6 +23,10 @@ public extension CustomerIO {
     ///      https://segment.io/libraries/ios#ids
     /// - traits: A dictionary of traits you know about the user. Things like: email, name, plan, etc.
     func identify<T: Codable>(userId: String, traits: T?) {
+        guard let traits = traits else {
+            DataPipeline.shared.identify(identifier: userId, body: EmptyRequestBody())
+            return
+        }
         DataPipeline.shared.identify(identifier: userId, body: traits)
     }
 


levibostian (changes requested):
Could you add tests to this PR?

Sources/DataPipeline/CustomerIO+Events.swift
```diff
@@ -23,6 +23,10 @@ public extension CustomerIO {
     ///      https://segment.io/libraries/ios#ids
     /// - traits: A dictionary of traits you know about the user. Things like: email, name, plan, etc.
     func identify<T: Codable>(userId: String, traits: T?) {
+        guard let traits = traits else {
+            DataPipeline.shared.identify(identifier: userId, body: EmptyRequestBody())
```
Shahroz16, line 27: I am a little confused about the error, analytics do accept `traits` as `nil`. At what point does it fail? additionally these methods would be updated because they are based on tracking so we need to find the route cause of why its failing.

ami-aman: Identify method accept `traits` as `nil` but `DataPipeline.shared.identify('user', body)` expects a non-nil value of body. [Check out this method from DataPipeline](https://github.com/customerio/customerio-ios/blob/d474c2ef6ceca186850aabe2cd2bc98d2eed4545/Sources/DataPipeline/DataPipeline.swift#L93).

Shahroz16: ah! i think its okay for now, but ideally they all need to change to use the ones Analytics provide. But thats a separate ticket.

ami-aman: > additionally these methods would be updated because they are based on tracking 

Does this mean that we are going to deprecate/remove this method? If this is the case then I would close this PR.

The root cause is [this](https://github.com/customerio/customerio-ios/blob/d474c2ef6ceca186850aabe2cd2bc98d2eed4545/Sources/DataPipeline/CustomerIO%2BEvents.swift#L25) method being called from the sample app (and I assume the same from customer's app too) which sends a `nil` in the body i.e. traits. The point where the app crashes is in Segment's Event [class here](https://github.com/customerio/cdp-analytics-swift/blob/f700c4c2c052eff3748a85eae64649509725f8af/Sources/Segment/Events.swift#L45) which throws an exception while parsing the nil value.

mrehan27: We'll be removing methods from `Tracking` i.e. `identify(identifier)`, but this one will stay and will be calling 
[`commonIdentifyProfile`](https://github.com/customerio/customerio-ios/blob/ae215bd86ef8a155aeb0457a9a056dae19b57ede/Sources/DataPipeline/DataPipelineImplementation.swift#L122) either directly or using some other existing methods. So if we need to merge the fix, we should fix it calling new methods, or just close it for now and fix later when these methods have been updated.

mrehan27 (changes requested):

ami-aman:
I will close this PR to avoid confusion and uncheck the pointer in [this ticket](https://linear.app/customerio/issue/MBL-109/findings-from-sample-app-cleanup-and-using-ios-cdp-sdk). We can cater when we update the methods.

PR #519, opened 2024-02-13
feat: use new header to set polling interval

BernardGatt:
Updates the polling interval remotely through the new X-Gist-Queue-Polling-Interval header.

Part of: [INAPP-12243](https://linear.app/customerio/issue/INAPP-12243/update-the-in-app-sdks-to-configure-polling-based-on-remote)


levibostian (approved):

levibostian:
I see the lint/formatter CI check is failing. I checked [the CI job](https://github.com/customerio/customerio-ios/actions/runs/7890346304/job/21532308468?pr=519) and I think the lint/formatter is failing in the diff check. Indicating that the formatting you ran on your development machine is different from what the CI generated. 

Perhaps your development environment is not up-to-date with the latest tooling this code base uses? If you could double [check this doc](https://github.com/customerio/customerio-ios/blob/main/docs/dev-notes/DEVELOPMENT.md#development-tools) to see if you have the `binny` tool installed and updated to the latest version, then run `make format` on your machine and push up the code then I think the lint errors will go away.

levibostian:
In regards to the sample apps not compiling, the CI job fails with error message, "Could not install WWDR certificate". 

This is a known issue with the tools that we use to compile our apps. It doesn't happen too often so retrying the CI job takes care of it most of the time.

Sources/MessagingInApp/Gist/Network/BaseNetwork.swift
```diff
@@ -31,7 +32,7 @@ enum BaseNetwork {
         URLSession.shared.dataTask(with: urlRequest, completionHandler: { data, response, error in
             if let error = error { completionHandler(.failure(error)) }
             guard let data = data, let response = response as? HTTPURLResponse,
-                  (200 ... 299).contains(response.statusCode)
+                  (200 ... 304).contains(response.statusCode)
```
Shahroz16, line 35: `304` is content not modified, I know we were relying on it before but were returning `failure` apparently, so just wanted to check nothing breaks from before right?

Sources/MessagingInApp/Gist/Managers/QueueManager.swift
```diff
@@ -46,4 +51,16 @@ class QueueManager {
             completionHandler(.failure(error))
         }
     }
+
+    private func updatePollingInterval(headers: [AnyHashable: Any]) {
+        if let newPollingIntervalString = headers["x-gist-queue-polling-interval"] as? String,
+           let newPollingInterval = Double(newPollingIntervalString),
+           newPollingInterval != Gist.shared.messageQueueManager.interval {
+            DispatchQueue.main.async {
```
Shahroz16, line 59: why do we need `DispatchQueue.main` here? because the method `setup` is also using `DispatchQueue.main.asyncAfter`

Sources/MessagingInApp/Gist/Network/BaseNetwork.swift
BernardGatt: Before this change, we used to rely on the default `NSURLSession`'s behavior to cache any 304 responses, it used to take 304s and convert them to the latest 200 responses automatically, we can't have that anymore because otherwise, the new header would be part of that cached response.

Sources/MessagingInApp/Gist/Managers/QueueManager.swift
BernardGatt: We're using that because of the polling timer being created in the `setup` call.

ami-aman (approved):

PR #520, opened 2024-02-14
chore: removed logger that was dependent on sdkconfig

Shahroz16:
In order to enable logging all module-wise, we need to get rid of a logger that is dependent on `SdkConfig` and only is being used in `Tracking`. 

helps: https://linear.app/customerio/issue/MBL-105/logging-in-modules


mrehan27 (changes requested):
I think we can minimize the changes by adding the following extension to `DIGraph`, which seems like a better option since it doesn't directly access the shared `DIGraph`, and also be beneficial for testing. Do you still see any reason to proceed with all the current changes, or would it be okay to revert and implement extensions instead?

```swift
public extension DIGraph {
    var logger: Logger {
        DIGraphShared.shared.logger
    }
}
```

mrehan27 (approved):
Looks good to me 👍🏻

PR #521, opened 2024-02-14
chore: update loglevel in all modules

Shahroz16:
helps: https://linear.app/customerio/issue/MBL-105/logging-in-modules

Share logger and config among all modules and dependencies including Segment and Gist

### DIFF
--- a/Sources/DataPipeline/CustomerIO.swift
+++ b/Sources/DataPipeline/CustomerIO.swift
@@ -19,9 +19,14 @@ public extension CustomerIO {
 
         let implementation = DataPipeline.initialize(moduleConfig: cdpConfig)
 
-        var sdkConfig = SdkConfig.Factory.create(siteId: "", apiKey: "", region: .US)
-        sdkConfig.logLevel = logLevel
+        let sdkConfig = SdkConfig.Factory.create(siteId: "", apiKey: "", region: .US)
         let newDiGraph = DIGraph(sdkConfig: sdkConfig)
+
+        // set the logLevel for ConsoleLogger
+        DIGraphShared.shared.logger.setLogLevel(logLevel)
+        // enable Analytics logs accordingly to logLevel
+        CustomerIO.shared.setDebugLogsEnabled(logLevel == .debug)
+
         initialize(implementation: implementation, diGraph: newDiGraph)
     }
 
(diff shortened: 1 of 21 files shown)


Ahmed-CIO (approved):
LGTM... SHIP IT

Sources/DataPipeline/CustomerIO.swift
```diff
@@ -21,6 +21,12 @@ public extension CustomerIO {
 
         let sdkConfig = SdkConfig.Factory.create(siteId: "", apiKey: "", region: .US)
         let newDiGraph = DIGraph(sdkConfig: sdkConfig)
+
+        // set the logLevel for ConsoleLogger
+        DIGraphShared.shared.logger.setLogLevel(logLevel)
+        // enable Analytics logs accordingly to logLevel
+        CustomerIO.shared.setDebugLogsEnabled(logLevel == .debug)
```
mrehan27, line 28: Is there a reason for not putting this into `initialize` method? Keeping this outside `initialize` will result in `logLevel` not being set in `logger` for tests.

mrehan27 (approved):
Just one question, looks good otherwise 👍🏻

Sources/DataPipeline/CustomerIO.swift
ami-aman: [This seems to be in `initialize` method only. ](https://github.com/customerio/customerio-ios/blob/5d04ed4ef58736e54ab4b406992d4132df27044c/Sources/DataPipeline/CustomerIO.swift#L26)

Sources/MessagingInApp/Provider/InAppProvider.swift
```diff
@@ -3,7 +3,7 @@ import Foundation
 
 // wrapper around Gist SDK to make it mockable
 protocol InAppProvider: AutoMockable {
-    func initialize(siteId: String, region: Region, delegate: GistDelegate)
+    func initialize(siteId: String, region: Region, delegate: GistDelegate, logging: Bool)
```
ami-aman, line 6: Do you plan to update the test cases for this change in a separate PR ?

ami-aman (approved):

Sources/DataPipeline/CustomerIO.swift
Shahroz16: I didn't want to rely on `DiGraph` for this as well, as i would assume we decouple away from it.

Sources/MessagingInApp/Provider/InAppProvider.swift
Shahroz16: We don't have a test for initialization for `MessagingInApp`, I assume when we add that then we can check for logging too.

PR #522, opened 2024-02-14
chore: OS conditions for removed and deprecated APIs

Ahmed-CIO:
## Context
This PR does not introduce any functional changes. It only aimed at handling warnings and compile error due to deprecated APIs and one removed API in VisionOS 1.0

## How did I make sure this did not break anything?
- Manually build APN UIKit app to make sure the compiled code is not stripped out for iOS and device locale new API works as expected
- Same on VisionOS which uses SwiftUI. `keyWindow` call is stripped out


ami-aman (approved):
👍

PR #523, opened 2024-02-14
chore: minor lint warnings removed

Ahmed-CIO:
No functional changes in this PR.
Fixed 3 lint warnings
- Removed 2 unused variables
- Removed non-existing import which is reported as ignored

## How did I make sure this code works?
Built and ran APN UIKit sample app.
I am not sure if the non-existing import removal have impact on other non-native wrappers though. If it is, then I will revert this line from this PR as it would need to be conditioned appropriately

### DIFF
--- a/Sources/MessagingPushAPN/autogenerated/AutoDependencyInjection.generated.swift
+++ b/Sources/MessagingPushAPN/autogenerated/AutoDependencyInjection.generated.swift
@@ -54,7 +54,9 @@ extension DIGraph {
     // internal scope so each module can provide their own version of the function with the same name.
     @available(iOSApplicationExtension, unavailable) // some properties could be unavailable to app extensions so this function must also.
     func testDependenciesAbleToResolve() -> Int {
-        return 0
+        var countDependenciesResolved = 0
+
+        return countDependenciesResolved
     }
 }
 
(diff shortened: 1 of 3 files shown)


Sources/MessagingPushAPN/autogenerated/AutoDependencyInjection.generated.swift
```diff
@@ -54,9 +54,7 @@ extension DIGraph {
     // internal scope so each module can provide their own version of the function with the same name.
     @available(iOSApplicationExtension, unavailable) // some properties could be unavailable to app extensions so this function must also.
     func testDependenciesAbleToResolve() -> Int {
-        var countDependenciesResolved = 0
-
-        return countDependenciesResolved
+        return 0
```
Shahroz16, line 57: makes sense, but it's an autogenerated file, so this will be reverted probably when the next time `make generator` runs.

levibostian: Correct. That's why the CI status check is failing on this PR at the moment. We do a `git diff` on the CI to make sure that `make format` was run *before* pushing the code to the remote git repo.

Ahmed-CIO: ah, yes. That is very true. Hard to differentiate till you look at the file path :D

PR #525, opened 2024-02-14
chore: removed old methods for initializing SDK

mrehan27:
helps: [MBL-103](https://linear.app/customerio/issue/MBL-103/public-api-remove-existing-initialization-method-using-siteid-and)

###  Changes

- Removed old `initialize` methods from `Tracking` package
- Removed `NotificationServiceExtensionSdkConfig` as it is no longer needed
- Removed `IntegrationTest.swift` from `Tracking` package as it was breaking and not needed now either


ami-aman (approved):

Sources/Tracking/CustomerIO.swift
```diff
@@ -21,92 +21,3 @@ public extension CustomerIOInstance {
...
-        var newSdkConfig = NotificationServiceExtensionSdkConfig.Factory.create(siteId: siteId, apiKey: apiKey, region: region)
-        if let configureHandler = configureHandler {
-            configureHandler(&newSdkConfig)
-        }
-
-        let sdkConfig = newSdkConfig.toSdkConfig()
-        let newDiGraph = DIGraph(sdkConfig: sdkConfig)
-        let implementation = DataPipeline.initialize(moduleConfig: DataPipelineConfigOptions.Factory.create(sdkConfig: sdkConfig))
-
-        initializeSharedInstance(with: implementation, diGraph: newDiGraph)
```
Shahroz16, line 68: `initializeSharedInstance` can also be removed now, no? if so would be part of next PR?

mrehan27: That's still being [used by `DataPipeline`](https://github.com/customerio/customerio-ios/blob/39e05f1cd02727d0d4ad67444f05c2c86e9ca31d/Sources/DataPipeline/CustomerIO.swift#L38), but yes there will be more changes in upcoming PRs, and it might either be removed or updated with new requirements.

PR #526, opened 2024-02-14
chore: fixed api tests for old initialization

mrehan27:
helps: [MBL-103](https://linear.app/customerio/issue/MBL-103/public-api-remove-existing-initialization-method-using-siteid-and)

###  Changes

- Added alias for `MessagingPushConfigOptions` to `FCM` and `APN`
- Moved old `APITest` from `Tracking` to appropriate modules including `DataPipeline`, `MessagingPushAPN` and `MessagingPushFCM`

### DIFF
--- a/Sources/MessagingPushAPN/Aliases.swift
+++ b/Sources/MessagingPushAPN/Aliases.swift
@@ -6,6 +6,7 @@ import Foundation
 // implementation that thier app uses (example: CioMessagingPushAPN).
 
 public typealias MessagingPush = CioMessagingPush.MessagingPush
+public typealias MessagingPushConfigOptions = CioMessagingPush.MessagingPushConfigOptions
 #if canImport(UserNotifications)
 public typealias CustomerIOParsedPushPayload = CioMessagingPush.CustomerIOParsedPushPayload
 #endif
--- a/Tests/MessagingPushAPN/APITest.swift
+++ b/Tests/MessagingPushAPN/APITest.swift
@@ -18,6 +18,12 @@ class MessagingPushAPNAPITest: UnitTest {
     func test_allPublicFunctions() throws {
         try skipRunningTest()
 
+        // This is the `initialize()` function that's available to Notification Service Extension and not available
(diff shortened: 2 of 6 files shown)


Sources/MessagingPushAPN/Aliases.swift
```diff
@@ -6,6 +6,7 @@ import Foundation
 // implementation that thier app uses (example: CioMessagingPushAPN).
 
 public typealias MessagingPush = CioMessagingPush.MessagingPush
+public typealias MessagingPushConfigOptions = CioMessagingPush.MessagingPushConfigOptions
```
mrehan27, line 9: `APITest` helped me identify this was missing from `Aliases`

Tests/MessagingPushAPN/APITest.swift
```diff
@@ -18,6 +18,12 @@ class MessagingPushAPNAPITest: UnitTest {
     func test_allPublicFunctions() throws {
         try skipRunningTest()
 
+        // This is the `initialize()` function that's available to Notification Service Extension and not available
+        // to other targets (such as iOS).
+        // You should be able to uncomment the initialize() function below and should get compile errors saying that the
+        // function is not available to iOS.
+        // MessagingPush.initialize(writeKey: "") { (config: inout MessagingPushConfigOptions) in }
```
Shahroz16, line 25: can we have `available` check on it instead of commenting it out?

Shahroz16 (approved):

Tests/MessagingPushAPN/APITest.swift
mrehan27: I don't think XCode will allow calling because it is only available for `iOSApplicationExtension`. And that's probably the reason why we have it the [same way in `main` today](https://github.com/customerio/customerio-ios/blob/216341c932d29caf334da53df3b4ab37b97bc94a/Tests/Tracking/APITest.swift#L43). Do you have any other ideas that you would like to share?

PR #528, opened 2024-02-14
chore: merge main into cdp branch

levibostian:
Closes: https://linear.app/customerio/issue/MBL-90/merge-automatic-click-handling-feature-into-cdp-branch

This PR is the result of merging `main` into `main-replica-for-cdp` and fixing everything. I got the code to compile, test suite to pass, and sample apps to compile. 

# How to review 

Most of the code in this PR is a boring `git merge` diff. All the code in this PR has already been reviewed since all code is in `main` or `main-replica-for-cdp`. 

So, to make the review process fast, I left comments in this PR for files that I find important to get a review on. All other files without a comment are considered boilerplate. 

# Not in scope of this PR 

* CI status checks for Lint and Cocoapods sample app. This is because `main-replica-for-cdp` also fails these 2 checks at the moment. 

# Notes

After this PR gets approved, I will *merge* (not squash) this PR. It's important that all of the git commits in this PR are merged into `main-replica-for-cdp` to keep the `main` commit history. If you squash this PR, we lose some of those commits.


Sources/MessagingPush/MessagingPush.swift
```diff
@@ -69,19 +85,29 @@ public class MessagingPush: ModuleTopLevelObject<MessagingPushInstance>, Messagi
...
         }
+        moduleConfig.writeKey = writeKey
+
+        shared.initializeModuleIfNotAlready()
+
+        return shared
     }
 
-    private func initializeModule() {
-        guard getImplementationInstance() == nil else {
```
levibostian, line 81: `main` contains new integration tests that verifies the push module only gets initialized once. That test failed with the current implementation. I replaced the implementation with the one in `main` and the test passed.

Sources/MessagingPush/MessagingPush.swift
```diff
@@ -129,3 +155,14 @@ public class MessagingPush: ModuleTopLevelObject<MessagingPushInstance>, Messagi
...
+// Convenient way for other modules to access CustomerIOInstance as well as being able to mock CustomerIOInstance.
+public extension DIGraphShared {
+    var messagingPushInstance: MessagingPushInstance {
+        if let override: MessagingPushInstance = getOverriddenInstance() {
+            return override
+        }
+
+        return MessagingPush.shared
+    }
+}
```
levibostian, line 168: This is added to write automated tests for `opened` tracking. This is required because currently, `opened` metrics are reported in the `MessagingPush` class. 

I am not the biggest fan of this code, but I didn't want to expand scope of this PR by changing implementation logic. This code is the same implementation being used in `main` today.

Sources/MessagingPush/MessagingPushConfigOptions.swift
levibostian, line 1: Added a new config option.

Sources/MessagingPush/UserNotificationsFramework/UserNotificationsFrameworkAdapter.swift
```diff
@@ -0,0 +1,127 @@
...
+    // Called when a new `UNUserNotificationCenterDelegate` is set on the host app. Our Swizzling is what calls this function.
+    func newNotificationCenterDelegateSet(_ newDelegate: UNUserNotificationCenterDelegate?)
+}
+
+/**
+ Keep this class small and simple because it is only able to be tested in QA testing. All logic for handling push events should be in the rest of the code base that has automated tests around it.
+ */
+@available(iOSApplicationExtension, unavailable)
+// sourcery: InjectRegisterShared = "UserNotificationsFrameworkAdapter"
+// sourcery: InjectSingleton
```
levibostian, line 27: The big call-out in this file is that I did modify it a little from what is inside of `main`. 

Because `DiGraphShared` is always in-memory, even if the SDK is initialized multiple times, we no longer [need the workarounds](https://github.com/customerio/customerio-ios/blob/cdf2f1fc45f1dadf0966a81c907f0ee1cb244f4f/Sources/MessagingPush/UserNotificationsFramework/UserNotificationsFrameworkAdapter.swift#L28-L52) that were previously needed to keep this class always in-memory. 

So, this PR uses the code generation to add this class to the digraph!

Sources/MessagingPush/PushHandling/PushEventHandlerProxy.swift
levibostian, line 1: Same comment as, https://github.com/customerio/customerio-ios/pull/528/files#r1490048117

Sources/Templates/AutoDependencyInjection.stencil
levibostian, line 1: The cocoapods sample app was not compiling until I made these changes. 

Swift annotations were not being added to `DigraphShared` like they were doing to `DiGraph`.

Sources/Tracking/CustomerIO.swift
```diff
@@ -87,7 +87,8 @@ public extension CustomerIO {
         }
         // automatically add the AutoTrackingScreenViews plugin if autoTrackScreenViews is enabled
         if sdkConfig.autoTrackScreenViews {
-            DataPipeline.shared.analytics.add(plugin: AutoTrackingScreenViews(filterAutoScreenViewEvents: sdkConfig.filterAutoScreenViewEvents, autoScreenViewBody: sdkConfig.autoScreenViewBody))
+            // FIXME: [CDP] cocoapods sample app not able to compile because of this line
```
levibostian, line 90: Added a FIXME comment to try and get the cocoapods sample app to compile.

levibostian:
After the latest merge of `main-replica-cdp` into this branch, the automated tests are now failing. Looking at `main-replica-cdp`, they are failing, too.

mrehan27 (approved):
Added a few comments, looks good overall 👍🏻

Sources/MessagingPush/MessagingPush.swift
mrehan27: Do you know the cause of its failure? Is the behavior change related to `lock` or `@Atomic` variable check? I’m asking because we have similar implementation in other modules, and if there’s a common issue, we should fix it before final release.

No action needed for this PR though, we can address this in a separate ticket.

Sources/MessagingPush/MessagingPush.swift
Shahroz16: hmm thinking out load, `MessagingPush` shouldn't ideally need `CustomerIOInstance`... We should be utilizing the event bus and the implementation class of  CustomerIOInstance should be handling those.

Sources/MessagingPush/Util/DeepLinkUtil.swift
```diff
@@ -1,15 +1,12 @@
 import CioInternalCommon
 import Foundation
-#if canImport(UIKit)
```
Shahroz16, line 3: was this cleanup? or change of logic ?

Shahroz16 (approved):
left some feedback, not blocking and can be catered in separate PR if needed

Sources/MessagingPush/MessagingPush.swift
levibostian: Made a [ticket](https://linear.app/customerio/issue/MBL-129/sdk-modules-can-only-be-initialized-1-time) for this. 

I do not remember why `guard getImplementationInstance() == nil else {` does not work and a mutex lock does. Luckily, I was able to write an automated test that reproduced the bug and the mutex lock made the test pass. So, I feel more confident that we have a working solution after being able to reproduce the behavior.

Sources/MessagingPush/MessagingPush.swift
levibostian: I think you're referring to the comment in the code snippet that includes `CustomerIOInstance`. If so, that is an outdated comment. 

```suggestion
// Convenient way for other modules to access instance as well as being able to mock instance in tests.
public extension DIGraphShared {
    var messagingPushInstance: MessagingPushInstance {
        if let override: MessagingPushInstance = getOverriddenInstance() {
            return override
        }

        return MessagingPush.shared
    }
}
```

I'll update the comment to remove the confusion.

Tests/Shared/Core/UnitTestBase.swift
```diff
@@ -120,7 +120,6 @@ open class UnitTestBase<Component>: XCTestCase {
         diGraphShared.reset()
         diGraphShared = .init()
         diGraph.reset()
-        diGraph = nil
```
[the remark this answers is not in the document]
levibostian: This was removed on purpose. 

With this line of code present, I was getting flaky test results. Removing this line was the only way that I was able to get the test suite running consistently.

Apps/APN-UIKit/APN UIKit/AppDelegate.swift
```diff
@@ -95,31 +99,20 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+            name: "push clicked",
+            data: ["push": response.notification.request.content.userInfo]
         )
 
-        // If the Customer.io SDK does not handle the push, it's up to you to handle it and call the
-        // completion handler. If the SDK did handle it, it called the completion handler for you.
-        if !handled {
-            completionHandler()
-        }
+        completionHandler()
```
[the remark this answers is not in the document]
levibostian: I am not 100% sure for this particular method. For other methods in the `UserNotification` framework, if you do not call the completionhandler, you will have consequences such as a long delay for push notifications to be displayed on the device, for example. 

Because this method's job is only to handle when a push getting clicked, I am not sure if there are any consequences? At least, the iPhone user may not notice any weird behavior like they would if a push had a long delay to be displayed.

PR #529, opened 2024-02-15
fix: fixes an issue with multiple timers being scheduled

BernardGatt:
While testing on standalone I noticed that timers weren't being stopped when a new one was scheduled, this should fix the issue.

### DIFF
--- a/Sources/MessagingInApp/Gist/Managers/MessageQueueManager.swift
+++ b/Sources/MessagingInApp/Gist/Managers/MessageQueueManager.swift
@@ -3,11 +3,16 @@ import UIKit
 
 class MessageQueueManager {
     var interval: Double = 600
-    private var queueTimer: Timer!
+    private var queueTimer: Timer?
     // The local message store is used to keep messages that can't be displayed because the route rule doesnt match.
     private var localMessageStore: [String: Message] = [:]
 
     func setup(skipQueueCheck: Bool = false) {
+        if let queueTimer {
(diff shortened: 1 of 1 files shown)


Sources/MessagingInApp/Gist/Managers/MessageQueueManager.swift
```diff
@@ -3,11 +3,16 @@ import UIKit
...
-    private var queueTimer: Timer!
+    private var queueTimer: Timer?
     // The local message store is used to keep messages that can't be displayed because the route rule doesnt match.
     private var localMessageStore: [String: Message] = [:]
 
     func setup(skipQueueCheck: Bool = false) {
+        if let queueTimer {
+            queueTimer.invalidate()
+            self.queueTimer = nil
+        }
```
levibostian, line 14: Should we also add this to `deinit` of this class?

levibostian:
```suggestion
            self.queueTimer?.invalidate()
            self.queueTimer = nil
```

Nit-picky suggestion. I think this code reads easier without the `if` because we are re-constructing the `queueTimer` in the line below.

BernardGatt: Given the instance is created and maintained from a static class, I don't think its required.

levibostian (approved):

PR #531, opened 2024-02-15
refactor: improve debug logs for UserNotification framework classes

levibostian:
When debugging the sample apps, you will see SDK debug statements when a push event occurs:

(siteid:xyz) Push event: willPresent. push: CioMessagingPush.UNNotificationWrapper
(siteid:xyz) On push action event. push action: CioMessagingPush.UNNotificationResponseWrapper

These log statements are not very useful because they do not actually tell us what the contents are of `UNNotificationWrapper`, for example.

Swift `structs` when they are printed include values of the properties inside of them. So, the debug logs are much more valuable if we change the data types to structs.

I tested this change with sample apps.

commit-id:4aeb187e


Shahroz16 (approved):
Hoping in logs it shows the content and not the object location in memory 😄

PR #533, opened 2024-02-15
refactor: remove screenview tracking config options from SdkConfig

levibostian:
Part of: https://linear.app/customerio/issue/MBL-120/public-api-remove-unused-properties-from-sdkconfig

None of the code was referencing any of these config options. Therefore, I did not need to make any more modifications to the code besides removing the properties from the Common module.

---

**Stack**:
- #547
- #546
- #540
- #539
- #534
- #533 ⬅
- #532

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #534, opened 2024-02-15
refactor: remove background queue config options from SdkConfig

levibostian:
Part of: https://linear.app/customerio/issue/MBL-120/public-api-remove-unused-properties-from-sdkconfig

The BQ feature that deletes expired tasks will no longer work after this change. Because most of the BQ implementation remains inside of the codebase and has not yet been removed, I commented out the code that is no longer working so the code compiles. The commented out code could be deleted or fixed when the rest of the BQ code is taken care of in the codebase.

---

**Stack**:
- #547
- #546
- #540
- #539
- #534 ⬅
- #533
- #532

### DIFF
--- a/Sources/Common/Store/SdkConfig.swift
+++ b/Sources/Common/Store/SdkConfig.swift
@@ -27,9 +27,6 @@ public struct SdkConfig {
                 region: region,
                 trackingApiUrl: region.productionTrackingUrl,
                 autoTrackPushEvents: true,
-                backgroundQueueMinNumberOfTasks: 10,
-                backgroundQueueSecondsDelay: 30,
-                backgroundQueueExpiredSeconds: Seconds.secondsFromDays(3),
                 logLevel: CioLogLevel.error
             )
         }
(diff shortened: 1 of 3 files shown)


Shahroz16 (approved):
lgtm

Sources/Common/Background Queue/QueueStorage.swift
```diff
@@ -147,42 +147,49 @@ public class FileManagerQueueStorage: QueueStorage {
...
-            // if deleting a task is not successful, we can ignore that
-            // because it doesn't negatively effect the state of the SDK or the queue.
-            _ = self.delete(storageId: taskToDelete.taskPersistedId)
-        }
-
-        return Array(tasksToDelete)
+        // When SdkConfig was modified to remove unused config options, this function could no longer run
+        // since the expired config option was removed.
+
+        // Commenting this code out for now, in case it's decided we still need it. Otherwise, this function can be deleted.
```
mrehan27, line 153: If the code is not required by any class for running the migrations, I would vote for removing it completely.

mrehan27 (approved):

Sources/Common/Background Queue/QueueStorage.swift
levibostian: Done.

levibostian:
✓ Commit merged in pull request [#540](https://github.com/customerio/customerio-ios/pull/540)

PR #537, opened 2024-02-16
refactor: deprecate tracking apis

mrehan27:
helps: [MBL-102](https://linear.app/customerio/issue/MBL-102/public-api-deprecate-tracking-apis-to-trigger-compile-time-errors)

### Changes

- Deprecated duplicate tracking APIs
- Relocated deprecated APIs to `Common` so they are re-exported with all modules and trigger compile-time warnings
- Added new public APIs to `CustomerIOInstance`
- Updated relevant classes to implement new public apis
- Removed redundant methods

###  What's not covered

- Updating documentation for public APIs, it requires a detailed review and will be addressed in a separate ticket
- Fixing related tests and updating auto-generated files (covered in #538)


Sources/Common/CustomerIOInstance.swift
```diff
@@ -135,7 +71,7 @@ public class CustomerIO: CustomerIOInstance {
         // arising from lifecycle differences between wrapper SDKs and native SDK.
         let globalDataStore = diGraph.globalDataStore
         if let token = globalDataStore.pushDeviceToken {
-            registerDeviceToken(token)
+            // registerDeviceToken(token)
```
mrehan27, line 74: `postInitialize` is already being removed in `MBL-123` so this shouldn't be a problem

Sources/Common/CustomerIOInstance.swift
```diff
@@ -144,185 +80,7 @@ public class CustomerIO: CustomerIOInstance {
...
-        // identify(). This replacement safeguards against possible
-        // issues that could arise(eg. app crash). While the `identify`
-        // method itself provide polymorphic capabilities to handle `nil`,
-        // this specific check within this method offers an additional
-        // layer of protection and clarity for this case.
-        // (refer https://github.com/customerio/customerio-ios/blob/94cbf686c3c2a405534cfe908f7166558a3e0b5d/Sources/Tracking/CustomerIO.swift#L71-L75)
-        if let optionalValue = body as? Any?, optionalValue == nil {
-            implementation?.identify(identifier: identifier, body: EmptyRequestBody())
-            return
-        }
```
mrehan27, line 228: Do we still need this workaround? I remember it caused sample app to crash on Random Login, but it so far is working without it. Any other way to test and confirm this?

Sources/Common/Deprecated/CustomerIO+Tracking.swift
mrehan27, line 1: This file contains all those public apis that are now deprecated from `Tracking`

Sources/DataPipeline/CustomerIO+DataPipeline.swift
```diff
@@ -0,0 +1,227 @@
+import CioInternalCommon
+import Foundation
+
+public protocol DataPipelinePublicAPI: AutoMockable {
```
mrehan27, line 4: The interface is doing almost the same thing as `CustomerIOInstance`, but it just move away from `Common` to `DataPipeline` because we no longer need those methods in `Common`.

Sources/DataPipeline/DataPipeline.swift
mrehan27, line 1: The class has mostly method name changes only

Sources/DataPipeline/DataPipelineImplementation.swift
mrehan27, line 1: The class too has mostly method name changes only

Sources/DataPipeline/CustomerIO+DataPipeline.swift
mrehan27: Removed for now

Sources/Common/CustomerIOInstance.swift
ami-aman: It is crashing on using Datapipeline methods too. This will be addressed in **MBL-109** !

ami-aman: Yes, but we may not need tracking's identify method now, so this solution is not needed here in this case.

Shahroz16 (approved):

Sources/Common/CustomerIOInstance.swift
mrehan27: I too think the same, this might not be needed now. However, if still required, we can update it in `MBL-109`

Sources/Common/Deprecated/CustomerIO+Tracking.swift
```diff
@@ -0,0 +1,36 @@
+import Foundation
+
+public extension CustomerIOInstance {
+    @available(*, unavailable, message: "Tracking module is no longer supported. Use 'identify(userId:traits:)' from 'DataPipeline' for improved features. See documentation for migration details.")
```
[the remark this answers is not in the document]
mrehan27: Good idea, updated to following for consistency

```
CioTracking module is no longer supported. Use 'identify(userId:traits:)' from 'CioDataPipelines' for improved features. See documentation for migration details.
```

PR #530, opened 2024-02-15
feat: migration module to cater to all migration tasks 

ami-aman:
part of: https://linear.app/customerio/issue/MBL-72/tracking-migration-without-tracking-module

Pending:
- Test cases
- Config for siteId

The pending tasks will be done in a separate PR.
Complete each step to get your pull request merged in. [Learn more about the workflow this project uses](https://github.com/customerio/customerio-ios/blob/develop/docs/dev-notes/GIT-WORKFLOW.md). 
- [ ] [Assign members of your team](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/requesting-a-pull-request-review) to review the pull request. 
- [ ] Wait for pull request status checks to complete. If there are problems, fix them until you see that [all status checks are passing](https://external-content.duckduckgo.com/iu/?u=https%3A%2F%2Fsymfony.com%2Fdoc%2F4.3%2F_images%2Fdocs-pull-request-symfonycloud.png&f=1&nofb=1). 
- [ ] Wait until the pull request has been reviewed *and approved* by a teammate
- [ ] After code reviews are approved and you determine this PR is ready to merge, select *Squash and Merge* button on this screen, leave the title and description to the default values, then merge the PR.

PR #538, opened 2024-02-16
chore: fix tests for deprecated tracking apis

mrehan27:
helps: [MBL-102](https://linear.app/customerio/issue/MBL-102/public-api-deprecate-tracking-apis-to-trigger-compile-time-errors)

### Changes

- Auto generated file updates from #537 
- Fixed tests broken due to changes in mentioned PR
- Updated public api tests to match the expectations

### DIFF
--- a/Sources/Tracking/migration/DataPipelineMigrationAssistant.swift
+++ b/Sources/Tracking/migration/DataPipelineMigrationAssistant.swift
@@ -48,7 +48,7 @@ class DataPipelineMigrationAssistant: DataPipelineMigration {
         // currently logged-in user for seamless processing of events.
         if DataPipeline.shared.analytics.userId == nil {
             if let identifier = profileStore.identifier {
-                DataPipeline.shared.identify(identifier: identifier)
+                CustomerIO.shared.identify(userId: identifier)
                 // Remove identifier from storage
                 // so same profile can not be re-identifed
                 profileStore.identifier = nil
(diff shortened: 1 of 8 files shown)


Tests/DataPipeline/APITest.swift
```diff
@@ -31,6 +36,47 @@ class DataPipelineAPITest: UnitTest {
             config.autoAddCustomerIODestination = false
         }
 
+        // Identify
+        CustomerIO.shared.identify(userId: "")
+        mock.identify(userId: "", traits: nil)
```
mrehan27, line 41: We have to pass `nil` explicitly for mock, I don't think that's a problem as actual public methods are callable due to default values. Let me know if someone thinks we need to add methods explicitly for this.

Sources/Tracking/migration/DataPipelineMigrationAssistant.swift
```diff
@@ -48,7 +48,7 @@ class DataPipelineMigrationAssistant: DataPipelineMigration {
         // currently logged-in user for seamless processing of events.
         if DataPipeline.shared.analytics.userId == nil {
             if let identifier = profileStore.identifier {
-                DataPipeline.shared.identify(identifier: identifier)
+                CustomerIO.shared.identify(userId: identifier)
```
ami-aman, line 51: This file is now removed in [this PR](https://github.com/customerio/customerio-ios/pull/530)

Shahroz16 (approved):

Sources/Tracking/migration/DataPipelineMigrationAssistant.swift
mrehan27: Thanks for highlighting Aman, but without this change, `main-replica-for-cdp` will break and not compile. So I'll keep it for now and we can resolve it later in the mentioned PR during merge.

PR #541, opened 2024-02-16
refactor: decouple BQ from SdkConfig

levibostian:
Closes: https://linear.app/customerio/issue/MBL-126/de-couple-bq-migration-from-sdkconfig

`SdkConfig.siteId` parameter is going away. So, we need to decouple classes `SdkConfig` when they need a `siteId`. Instead, these classes will be provided a `siteId` through function parameters.

Starting at the DataPipe migration class, I added a `siteId` parameter. Then, modified all dependencies of the migration class until the code compiled. Most of the code that changed was the BQ file storage and all of it's tests.

commit-id:8ca21f6c

### DIFF
--- a/Sources/Common/Background Queue/Queue.swift
+++ b/Sources/Common/Background Queue/Queue.swift
@@ -22,25 +22,9 @@ public typealias ModifyQueueResult = (success: Bool, queueStatus: QueueStatus)
  before the task runs.
  */
 public protocol Queue: AutoMockable {
-    func deleteExpiredTasks()
-
-    func getAllStoredTasks() -> [QueueTaskMetadata]
-
-    func getTaskDetail(_ task: QueueTaskMetadata) -> TaskDetail?
-
-    func deleteProcessedTask(_ task: QueueTaskMetadata)
-}
-
-public extension Queue {
-    // Get list of all unprocessed tasks in background queue
-    func getAllStoredTasks() -> [QueueTaskMetadata] {
-        getAllStoredTasks()
-    }
-
-    // Delete already processed task from the background queue
-    func deleteProcessedTask(_ task: QueueTaskMetadata) {
-        deleteProcessedTask(task)
-    }
+    func getAllStoredTasks(siteId: String) -> [QueueTaskMetadata]
+    func getTaskDetail(_ task: QueueTaskMetadata, siteId: String) -> TaskDetail?
+    func deleteProcessedTask(_ task: QueueTaskMetadata, siteId: String)
 }
 
 public struct TaskDetail {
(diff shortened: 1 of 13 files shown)


Tests/Shared/SampleData/SampleDataFilesUtil.swift
levibostian, line 1: I had to modify this file because of `FileStore` needing `siteid` parameters now. 

I noticed the codebase is no longer calling `saveSdkV1QueueFiles` anywhere so I removed the function.

Sources/Common/Background Queue/Queue.swift
```diff
@@ -52,57 +36,44 @@ public struct TaskDetail {
...
         let storageId = task.taskPersistedId
-        if !storage.delete(storageId: storageId) {
+        if !storage.delete(storageId: storageId, siteId: siteId) {
             logger.error("Failed to delete task with storage id: \(storageId).")
         }
     }
-
-    public func deleteExpiredTasks() {
-        _ = storage.deleteExpired()
-    }
```
levibostian, line 107: [In another open PR](https://github.com/customerio/customerio-ios/pull/534#discussion_r1492494884), we discussed removing the BQ delete expired code from the codebase since we're no longer using it. 

I removed it from this PR as well to get the SDK code to compile successfully.

Sources/Common/Background Queue/QueueStorage.swift
```diff
@@ -24,99 +24,76 @@ import Foundation
...
     private let jsonAdapter: JsonAdapter
-    let siteId: String
-    private let sdkConfig: SdkConfig
     let logger: Logger
     let dateUtil: DateUtil
     private var inventoryStore: QueueInventoryMemoryStore
 
     let lock: Lock
 
-    private var inventory: [QueueTaskMetadata]? {
```
levibostian, line 45: All of the `inventory` get/set code moved into `getInventory` and `saveInventory` functions. Since we now need a `siteId` to get and save inventory.

Sources/Common/Background Queue/QueueStorage.swift
```diff
@@ -24,99 +24,76 @@ import Foundation
...
         self.inventoryStore = inventoryStore
     }
 
-    public func getInventory() -> [QueueTaskMetadata] {
-        inventory ?? []
+    public func getInventory(siteId: String) -> [QueueTaskMetadata] {
+        lock.lock()
+        defer { lock.unlock() }
+
+        if let inventoryCache = inventoryStore.inventory {
```
Shahroz16, line 62: maybe we can get rid memory inventory as well? as we just need to do this operation once?

Shahroz16 (approved):

mrehan27 (approved):

Sources/Common/Background Queue/QueueStorage.swift
levibostian: I considered that, but I don't think that we should because the BQ migration code would take a performance hit. Since the BQ migration code does both reading and writing to the BQ storage, the in-memory cache is providing value.

PR #542, opened 2024-02-16
chore: siteId configuration and migration test cases

ami-aman:
part of: https://linear.app/customerio/issue/MBL-72/tracking-migration-without-tracking-module

Merges into:  https://github.com/customerio/customerio-ios/pull/530

Includes:
- SiteId configuration
- Migration Test cases updates 


Tests/Migration/DataPipelineMigrationAssistantTests.swift
```diff
@@ -1,8 +1,7 @@
-@testable import CioDataPipelines
+// @testable import CioDataPipelines
```
Shahroz16, line 1: why is this commented out? should be removed right?

Tests/Migration/DataPipelineMigrationAssistantTests.swift
```diff
@@ -12,16 +11,24 @@ class DataPipelineMigrationAssistantTests: UnitTest {
     // that have gone undiscovered in the code when `CustomerIO` passes a request to `DataPipelineImplementation`.
     private let profileStoreMock = ProfileStoreMock()
     private let backgroundQueueMock = QueueMock()
-
-    private var migrationAssistant: DataPipelineMigrationAssistant { diGraph.dataPipelineMigrationAssistant }
+    private let migrationHandler = DataPipelineMigrationActionMock()
     private var queueStorage: QueueStorage { diGraph.queueStorage }
-
+    public var migrationAssistant: DataPipelineMigrationAssistant!
```
Shahroz16, line 16: can we avoid making it nullable? we have been trying to remove it from test cases as they tend to introduce flakiness in tests.

Sources/DataPipeline/DataPipelineConfigOptions.swift
```diff
@@ -39,6 +39,9 @@ public struct DataPipelineConfigOptions {
         self.writeKey = writeKey
     }
 
+    /// Site Id that will be used for tracking data migration
+    public let siteId: String = ""
```
Shahroz16, line 43: how is this being set?

ami-aman: Will be set by the customer, have added a default value though. Is there something else that you expect?

mrehan27: I think Shahroz is correct here and customers might not be able to set it because the constructor above only takes `writeKey`?

Tests/Migration/DataPipelineMigrationAssistantTests.swift
```diff
@@ -12,16 +11,24 @@ class DataPipelineMigrationAssistantTests: UnitTest {
...
         diGraph.override(value: backgroundQueueMock, forType: Queue.self)
+        migrationAssistant = DataPipelineMigrationAssistant(handler: migrationHandler, diGraph: diGraph)
+    }
+
+    // MARK: performMigration
+
+    func test_performMigration_WithAndWithoutUserId() {
+        backgroundQueueMock.getAllStoredTasksReturnValue = []
+        XCTAssertNotNil(migrationAssistant.performMigration(for: nil))
+        XCTAssertNotNil(migrationAssistant.performMigration(for: String.random))
```
mrehan27, line 31: Can you please add comments to explain what is being tested and how? `migrationAssistant.performMigration` seems to return nothing significant, are we just assuming if this doesn't crash, it works? If yes, can you pleas explain it using comments?

Tests/Migration/DataPipelineMigrationAssistantTests.swift
```diff
@@ -86,69 +93,54 @@ class DataPipelineMigrationAssistantTests: UnitTest {
...
+    }
+
     func test_givenAlreadyIdentifiedProfile_expectUpdateUserId() {
         let givenProfileIdentifiedInJourneys = String.random
         profileStoreMock.identifier = givenProfileIdentifiedInJourneys
-        DataPipeline.shared.analytics.reset()
-        XCTAssertNil(DataPipeline.shared.analytics.userId)
-
-        migrationAssistant.handleAlreadyIdentifiedMigratedUser()
+        migrationAssistant.handleAlreadyIdentifiedMigratedUser(for: nil)
```
mrehan27, line 119: We probably need to update the test name here?

Tests/Migration/DataPipelineMigrationAssistantTests.swift
```diff
@@ -86,69 +93,54 @@ class DataPipelineMigrationAssistantTests: UnitTest {
...
 
-        XCTAssertEqual(DataPipeline.shared.analytics.userId, givenProfileIdentifiedInJourneys)
         XCTAssertNil(profileStoreMock.identifier)
     }
 
-    func test_givenNoIdentifiedProfile_expectNoUpdateInUserId() {
```
mrehan27, line 101: Is this test not needed now?

Tests/Migration/DataPipelineMigrationAssistantTests.swift
```diff
@@ -86,69 +93,54 @@ class DataPipelineMigrationAssistantTests: UnitTest {
...
-        // after being identified on the CDP
-        let updatedIdentifier = String.random
-        profileStoreMock.identifier = updatedIdentifier
-
-        migrationAssistant.handleAlreadyIdentifiedMigratedUser()
-
-        // We expect the CDP profile ID is the same value from the 1st migration done.
-        XCTAssertEqual(DataPipeline.shared.analytics.userId, givenIdentifier)
-    }
+//
```
mrehan27, line 145: Need to remove this typo

Sources/DataPipeline/DataPipelineConfigOptions.swift
ami-aman: Ah! Yes, right! 🤦

ami-aman: My understanding is that we expect the customer to set the value of siteId using config like

```
CustomerIO.initialize(writeKey: writeKey, logLevel: logLevel) { config in
            config.siteId = "site_id_value_from_customer"
 }
```

Yes, the other way is to get the value as a parameter but I am not sure if we want that ? Let me know if we are not on same page ! I have no preferences and can make the changes if required.

Shahroz16: but it is `let` ?

ami-aman: Ah ! Yes, you are right. Fixed in next commit

Tests/Migration/DataPipelineMigrationAssistantTests.swift
ami-aman: Now that we can not access `DataPipeline` in Migration so it is not possible to find if user Id gets updated or not, hence this test was removed.

Shahroz16 (commented):
there are 29 files being shown here, is it not updated from `main-replica`? can we please update that first to make sure we have latest code and we are not missing anything here?

Sources/DataPipeline/DataPipelineConfigOptions.swift
```diff
@@ -35,6 +30,9 @@ public struct DataPipelineConfigOptions {
         self.writeKey = writeKey
     }
 
+    /// Site Id that will be used for tracking data migration
+    public var siteId: String = ""
```
mrehan27, line 34: Can we avoid storing `siteId` here? We recently removed it from `SdkConfig`, if there is an option to avoid storing it, I would prefer not storing it here again.

ami-aman:
@Shahroz16 

> there are 29 files being shown here, is it not updated from main-replica? can we please update that first to make sure we have latest code and we are not missing anything here?

This PR is updated with `main-replica-for-cdp`. The base branch for this PR is `aman/migration-without-tracking-module`.

Screenshot 2024-02-22 at 2 22 25 PM

Sources/DataPipeline/DataPipelineConfigOptions.swift
ami-aman: We need `siteId` for migration specifically to fetch the `profileIdentifier` and in `fileStorage`. We would not be using `siteId` directly through config but would be iterating down to the methods as and where required. Since we discussed the same 1:1 on slack and now that we both are on same page, I am resolving this comment.

PR #544, opened 2024-02-20
refactor: sdk config builder

mrehan27:

### Changes

- Added `SDKConfigBuilder` for building SDK config when initializing the SDK
- Added `SDKConfigBuilderResult` for storing and passing config builder results
- Simplified `DataPipelineConfigOptions` struct

### What's not covered

- Fixing tests broken due to changes made in this PR (fixed in #545)
- Updating sample app and writing tests for `SDKConfigBuilder`, will open PR soon
- Simplifying `SdkConfig` (we'll cover this later)

### DIFF
--- a/Tests/Common/Service/UserAgentUtilTest.swift
+++ b/Tests/Common/Service/UserAgentUtilTest.swift
@@ -23,9 +23,7 @@ class UserAgentUtilTest: UnitTest {
     }
 
     override func setUp() {
-        super.setUp(modifySdkConfig: { config in
-            config._sdkWrapperConfig = self.sdkWrapperConfig
-        })
+        super.setUp()
 
         userAgentUtil = UserAgentUtilImpl(deviceInfo: deviceInfoMock, sdkConfig: sdkConfig)
     }
--- a/Tests/Common/Util/FileStorageTest.swift
+++ b/Tests/Common/Util/FileStorageTest.swift
@@ -33,7 +33,7 @@ class FileStorageTest: UnitTest {
     private var siteId: String!
 
     override func setUp() {
-        super.setUp(siteId: siteId, modifySdkConfig: nil)
+        super.setUp()
 
         siteId = String.random()
 
--- a/Tests/DataPipeline/APITest.swift
+++ b/Tests/DataPipeline/APITest.swift
@@ -29,12 +29,7 @@ class DataPipelineAPITest: UnitTest {
         try skipRunningTest()
 
         // Initialize
-        CustomerIO.initialize(writeKey: "") { (config: inout DataPipelineConfigOptions) in
-            config.autoAddCustomerIODestination = false
-        }
-        CustomerIO.initialize(writeKey: "", logLevel: .debug) { (config: inout DataPipelineConfigOptions) in
-            config.autoAddCustomerIODestination = false
-        }
+        CustomerIO.initialize(withConfig: SDKConfigBuilder(writeKey: "").build())
 
         // Identify
         CustomerIO.shared.identify(userId: "")
@@ -93,19 +88,19 @@ class DataPipelineAPITest: UnitTest {
     func test_allPublicModuleConfigOptions() throws {
         try skipRunningTest()
 
-        CustomerIO.initialize(writeKey: "", logLevel: .error) { config in
-            config.apiHost = ""
-            config.cdnHost = ""
-            config.flushAt = 10
-            config.flushInterval = 10.0
-            config.autoAddCustomerIODestination = false
-            config.defaultSettings = Settings(writeKey: "")
-            config.flushPolicies = [CountBasedFlushPolicy(), IntervalBasedFlushPolicy()]
-            config.flushQueue = DispatchQueue(label: "")
-            config.operatingMode = OperatingMode.asynchronous
-            config.trackApplicationLifecycleEvents = true
-            config.autoTrackDeviceAttributes = true
-        }
+        SDKConfigBuilder(writeKey: "")
+            .apiHost("")
+            .cdnHost("")
+            .flushAt(10)
+            .flushInterval(10.0)
+            .autoAddCustomerIODestination(false)
+            .defaultSettings(Settings(writeKey: ""))
+            .flushPolicies([])
+            .flushQueue(DispatchQueue(label: ""))
+            .operatingMode(OperatingMode.asynchronous)
+            .trackApplicationLifecycleEvents(true)
+            .autoTrackDeviceAttributes(true)
+            .siteId("")
     }
 
     func test_autoTrackingScreenViewsPluginOptions() throws {
--- a/Tests/DataPipeline/Core/UnitTest.swift
+++ b/Tests/DataPipeline/Core/UnitTest.swift
@@ -8,7 +8,6 @@ import SharedTests
 /// specific to module components. Ideal for isolated tests of individual functions and classes.
 open class UnitTest: SharedTests.UnitTestBase<CustomerIO> {
     public var dataPipelineConfigOptions: DataPipelineConfigOptions!
-    public var enableLogs: Bool = false
 
     // Use this `CustomerIO` instance when invoking `CustomerIOInstance` functions in unit tests.
     // This ensures convenience and consistency across unit tests, and guarantees the correct instance is used for testing.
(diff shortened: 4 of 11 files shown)


Sources/DataPipeline/Config/SDKConfigBuilder.swift
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    @discardableResult
+    public func flushPolicies(_ policies: [FlushPolicy]) -> SDKConfigBuilder {
+        flushPolicies = policies
+        return self
+    }
+
+    @discardableResult
+    public func addFlushPolicies(_ policies: [FlushPolicy]) -> SDKConfigBuilder {
```
mrehan27, line 95: This one is additional. I added so customers can append to our flush policies if needed.

Sources/DataPipeline/Config/SDKConfigBuilder.swift
```diff
@@ -0,0 +1,147 @@
...
+/// Usage Example:
+/// ```
+/// let config = SDKConfigBuilder(writeKey: "your_write_key")
+///   .apiHost("your_api_host")
+///   .flushAt(30)
+///   // additional configuration as needed...
+///   .build()
+/// // Use `config` for initializing the SDK...
+/// ```
+public class SDKConfigBuilder {
```
levibostian, line 19:
```suggestion
public struct SDKConfigBuilder {
```

Because this class's job is mostly to hold onto properties and modify them, can we make this a `struct`? 

A swift struct vs class is similar to Kotlin's class vs data class conversation. data classes/structs provide some benefits that could be useful for us in the future.

Sources/DataPipeline/Config/SDKConfigBuilder.swift
```diff
@@ -0,0 +1,147 @@
...
+    // configuration options for DataPipelineConfigOptions
+    private let writeKey: String
+    private var apiHost: String = SDKConfigBuilder.defaultAPIHost
+    private var cdnHost: String = SDKConfigBuilder.defaultCDNHost
+    private var flushAt: Int = 20
+    private var flushInterval: Seconds = 30
+    private var autoAddCustomerIODestination: Bool = true
+    private var defaultSettings: Settings?
+    private var flushPolicies: [FlushPolicy] = [CountBasedFlushPolicy(), IntervalBasedFlushPolicy()]
+    private var flushQueue: DispatchQueue = .init(label: "com.segment.operatingModeQueue", qos: .utility)
```
levibostian, line 36: Instead of having hard-coded values here in our SDK that match the Segment default, could we reference the Segment default values? 

```suggestion
    private var flushPolicies: [FlushPolicy] = SegmentSdk.defaultFlushPolicy
```

Sources/DataPipeline/Config/SDKConfigBuilder.swift
levibostian: I wonder if this is confusing to have a `flushPolicies` and a `addFlushPolicies`? I wouldn't want a customer to use the wrong one and they experience weird behavior in their app. 

Makes me want to consider not adding this method to avoid customer confusion. 

I like your idea though.

Sources/DataPipeline/Config/SDKConfigBuilderResult.swift
levibostian, line 1: I would be OK to have this 1 line of code added to the `Sources/DataPipeline/Config/SDKConfigBuilder.swift` file. Since the code in the 2 files are so similar, I don't see a big reason to move into separate file.

Sources/DataPipeline/CustomerIO.swift
```diff
@@ -6,22 +6,15 @@ public extension CustomerIO {
...
-        logLevel: CioLogLevel = .error,
-        configure configureHandler: ((inout DataPipelineConfigOptions) -> Void)?
-    ) {
-        var cdpConfig = DataPipelineConfigOptions.Factory.create(writeKey: writeKey)
-
-        if let configureHandler = configureHandler {
-            configureHandler(&cdpConfig)
-        }
+    static func initialize(withConfig config: SDKConfigBuilderResult) {
+        // SdkConfig is currently not stored anywhere, we'll later add option to store and access it.
```
levibostian, line 10:
```suggestion
```

Unless there is a problem caused by not storing the SdkConfig anywhere, I am OK with not storing it somewhere and removing this comment.

Sources/DataPipeline/DataPipelineConfigOptions.swift
levibostian, line 1: I see the default values got moved into the Builder. 

I like the idea of the default config values being defined in this file. It makes more sense to me. But, not a requirement. 

Is there an easy way to keep the default values in here? Maybe by making all of the properties in the Builder be optional by default? 

If this makes the code messier, let's leave it as it is now.

levibostian (approved):

Sources/DataPipeline/Config/SDKConfigBuilder.swift
mrehan27: I had the same thoughts while adding this, so I completely agree with you. The idea was to help customers who simply want to append their policy to default values.

But since we both had same concern, I have removed `addFlushPolicies` for now. We can later improve this to help customers if needed.

Sources/DataPipeline/Config/SDKConfigBuilder.swift
mrehan27: Thanks for sharing, updated.

Sources/DataPipeline/Config/SDKConfigBuilder.swift
mrehan27: Unfortunately, Segment also [hardcodes default values](https://github.com/customerio/cdp-analytics-swift/blob/main/Sources/Segment/Configuration.swift) in a similar fashion. So, unless we make changes to the Segment SDK, there's no direct way to reference these values.

I'm currently writing tests to ensure that our values match Segment's, will open a PR for it soon.

Sources/DataPipeline/Config/SDKConfigBuilderResult.swift
mrehan27: Moved to same

Sources/DataPipeline/CustomerIO.swift
mrehan27: I've updated the comment, but I want to keep it as a reminder that this decision was made deliberately.

Sources/DataPipeline/DataPipelineConfigOptions.swift
mrehan27: I agree with you. I also preferred not to include them in the builder, as having default values in the same class seemed more logical to me too.

However, it poses some challenges currently. Particularly, for any property that is nullable and has a non-null default value, it becomes difficult for the builder class to determine whether `nil` was provided by the customer or by the builder. Keeping all defaults in the builder ensures a single source of truth and gives priority to customer-provided values. So, let's keep it as it is for now.

Sources/DataPipeline/Config/DataPipelineConfigOptions.swift
```diff
@@ -0,0 +1,28 @@
+import CioInternalCommon
+import Foundation
+import Segment
+
+/// Configuration options for the Customer.io Data Pipeline module
+public struct DataPipelineConfigOptions {
+    /// Server key
+    public let writeKey: String
```
ami-aman, line 8: I think we might need to add `siteId` here too for  migration ! 🤔

Sources/DataPipeline/Config/SDKConfigBuilder.swift
mrehan27: I've to revert this, because it fails when using chaining. The code in `APITest` looks like

```
SDKConfigBuilder(writeKey: "")
  .apiHost("")
  .cdnHost("")
  .flushAt(10)
  .flushInterval(10.0)
  .autoAddCustomerIODestination(false)
  .defaultSettings(Settings(writeKey: ""))
  .flushPolicies([])
  .addFlushPolicies([CountBasedFlushPolicy(), IntervalBasedFlushPolicy()])
  .flushQueue(DispatchQueue(label: ""))
  .operatingMode(OperatingMode.asynchronous)
  .trackApplicationLifecycleEvents(true)
  .autoTrackDeviceAttributes(true)
```
and fails with following error with struct

```
Cannot use mutating member on immutable value: function call returns immutable value
```

Sources/DataPipeline/Config/DataPipelineConfigOptions.swift
mrehan27: Thanks for highlighting, I'll add so you can reuse it.

PR #546, opened 2024-02-20
refactor: remove siteId, apiKey, and region from SdkConfig

levibostian:
Part of: https://linear.app/customerio/issue/MBL-120/public-api-remove-unused-properties-from-sdkconfig

Some dependencies made this change difficult because they are tightly coupled to SdkConfig. KeyValueStore being the biggest problem.

The `CioProfileStore` is currently broken with this change because now it uses a global data store but it needs to have a siteid sandboxed store. Will be fixing this in next commit.

---

**Stack**:
- #547
- #546 ⬅

### DIFF
--- a/Sources/Common/Store/GlobalDataStore.swift
+++ b/Sources/Common/Store/GlobalDataStore.swift
@@ -12,48 +12,6 @@ public protocol GlobalDataStore: AutoMockable {
     func deleteAll()
 }
 
-// sourcery: InjectRegister = "GlobalDataStore"
-public class CioGlobalDataStore: GlobalDataStore {
-    private let keyValueStorage: KeyValueStorage
-
-    public var pushDeviceToken: String? {
-        get {
-            keyValueStorage.string(.pushDeviceToken)
-        }
-        set {
-            keyValueStorage.setString(newValue, forKey: .pushDeviceToken)
-        }
-    }
-
-    public var httpRequestsPauseEnds: Date? {
-        get {
-            keyValueStorage.date(.httpRequestsPauseEnds)
-        }
-        set {
-            keyValueStorage.setDate(newValue, forKey: .httpRequestsPauseEnds)
-        }
-    }
-
-    public init(keyValueStorage: KeyValueStorage) {
-        self.keyValueStorage = keyValueStorage
-
-        self.keyValueStorage.switchToGlobalDataStore()
-    }
-
-    // How to get instance before DI graph is constructed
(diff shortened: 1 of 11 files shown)


Sources/Common/Store/GlobalDataStore.swift
```diff
@@ -80,14 +38,6 @@ public class CioSharedDataStore: GlobalDataStore {
         self.keyValueStorage = keyValueStorage
     }
 
-    // How to get instance before DI graph is constructed
-    public static func getInstance() -> GlobalDataStore {
-        let newInstance = UserDefaultsSharedKeyValueStorage(
-            deviceMetricsGrabber: DeviceMetricsGrabberImpl()
-        )
-        return CioSharedDataStore(keyValueStorage: newInstance)
-    }
```
levibostian, line 89: The `getInstance` function was previously only needed because of `DIGraph` not getting constructed until a `siteId` was provided by customers. 

But, with `DIGraphShared`, we can resolve all dependencies immediately. So, a function like this is not needed. I deleted it so we no longer need to maintain it.

Sources/Common/Util/KeyValueStorage.swift
```diff
@@ -37,25 +35,21 @@ public protocol KeyValueStorage: SharedKeyValueStorage {
  */
 // sourcery: InjectRegister = "KeyValueStorage"
 public class UserDefaultsKeyValueStorage: KeyValueStorage {
-    private var siteId: String?
+    // Set this value before calling any other function in this class, to make sure you get/set values in a sandboxed key/value store.
+    // This solution is fragile because if you forget to set the siteid, you will get/set from the global data store.
+    // This temporary solution will be replaced with a better solution.
+    var siteId: String?
```
levibostian, line 41: [The next PR](https://github.com/customerio/customerio-ios/pull/547) in the stack replaces this fragile implementation with something more reliable.

Tests/Common/Util/KeyValueStorageTest.swift
levibostian, line 1: Changes made in this file are only to get the tests compiling and passing. 

[The next PR](https://github.com/customerio/customerio-ios/pull/547) in the stack changes this file a lot because we reduce the number of key/value stores down to 1.

mrehan27 (approved):
Looks good 👍🏻

Shahroz16 (approved):

levibostian:
✓ Commit merged in pull request [#547](https://github.com/customerio/customerio-ios/pull/547)

PR #549, opened 2024-02-21
fix: all sdk modules can only be initialized once

levibostian:
Closes: https://linear.app/customerio/issue/MBL-129/sdk-modules-can-only-be-initialized-1-time

All SDK modules have a top-level `ModuleName.initialize()` function. We have a SDK requirement that this function is only executed one time and ignore all future calls.

This change brings a few improvements to this requirement:
1. Move implementation into base class, `ModuleTopLevelObject`, to encapsulate the logic and allow re-use by all SDK modules.
2. Adds automated tests for `ModuleTopLevelObjectTest` super class to assert that our code only allows initializing modules one time. Note: These tests were originally written in `MessagingPushTest` and were moved to new `ModuleTopLevelObjectTest` class.

commit-id:63854a04

### DIFF
--- a/Sources/Common/Module/ModuleTopLevelObject.swift
+++ b/Sources/Common/Module/ModuleTopLevelObject.swift
@@ -8,17 +8,28 @@ import Foundation
 // There isn't a constructor populated via dependency injection. It's at the top node
 // of dependencies.
 open class ModuleTopLevelObject<ImplementationClass> {
-    open var alreadyCreatedImplementation: ImplementationClass?
+    /*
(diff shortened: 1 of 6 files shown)


Shahroz16 (approved):
Do you think this is going to break the sample app, because it probably updates config from settings by initializing again?

levibostian:
I wouldn't say the sample apps are broken (non-functional), but this change could indeed make it so you must kill and relaunch the app for settings to take effect. 

The APN app required kill/relaunch already so the only app this impacts is FCM app.

PR #551, opened 2024-02-22
fix: sample app issues

Shahroz16:
closes: https://linear.app/customerio/issue/MBL-109/findings-from-sample-app-cleanup-and-using-ios-cdp-sdk

more context: https://www.loom.com/share/971cb87f0216460faa0cefbf4fc9591e

solves a couple of issues:
- incorrect header 
- nse image not showing at times


Sources/MessagingPush/RichPush/MessagingPush+RichPush.swift
```diff
@@ -73,30 +73,19 @@ extension MessagingPushImplementation {
...
             // This conditional will only work in production and not in automated tests. But this file cannot be in automated tests so this conditional is OK for now.
             if let composedRichPush = composedRichPush as? UNNotificationWrapper {
-                self.finishTasksThenReturn(contentHandler: contentHandler, notificationContent: composedRichPush.notificationContent)
+                self.logger
+                    .debug(
+                        "running all background queue tasks and waiting until complete to prevent OS from killing notification service extension before all HTTP requests have been performed"
+                    )
+
+                self.logger.info("Customer.io push processing is done!")
+                contentHandler(composedRichPush.notificationContent)
```
levibostian, line 82: When we call the `contentHandler()`, we are telling the OS that our NSE is completely done doing what it needs to do. 

This might be out of scope of this PR, but I think that we need to make sure that before we call the contentHandler, we confirm the `delivered` metric HTTP request has finished, correct? 

I see above this line of code `trackMetricFromNSE(deliveryID: pushCioDeliveryInfo.id, event: .delivered, deviceToken: pushCioDeliveryInfo.token)` and because there is no async callback in that line of code, we are not informed when the HTTP request concludes, correct?

Sources/Common/Service/HttpRequestRunner.swift
```diff
@@ -56,10 +56,11 @@ public class UrlRequestHttpRequestRunner: HttpRequestRunner {
         let directoryURL = fileType.directoryToSaveFiles(fileManager: FileManager.default)
 
         session.downloadTask(with: url) { tempLocation, response, _ in
-            guard let tempLocation = tempLocation, let uniqueFileName = response?.suggestedFilename else {
+            guard let tempLocation = tempLocation, let suggestedFileName = response?.suggestedFilename else {
                 return onComplete(nil)
             }
 
+            let uniqueFileName = UUID().uuidString + "_" + suggestedFileName
```
levibostian, line 63: I think we should have a comment in the code explaining why having a unique file name is important. Or, why this UUID is needed beyond "suggest file name"

Sources/MessagingPush/RichPush/RichPushHttpClient.swift
```diff
@@ -160,8 +160,10 @@ extension RichPushHttpClient {
 
     static func getBasicSession() -> URLSession {
         let configuration = URLSessionConfiguration.ephemeral
-        configuration.httpMaximumConnectionsPerHost = 2
         configuration.allowsCellularAccess = true
+        configuration.timeoutIntervalForResource = 30
+        configuration.timeoutIntervalForRequest = 60
+        configuration.httpAdditionalHeaders = [:]
```
levibostian, line 166: Why were these lines added?

Tests/MessagingPush/HttpTest.swift
```diff
@@ -33,10 +33,29 @@ open class HttpTest: UnitTest {
...
+            cioSession = RichPushHttpClient.getCIOApiSession(
                 key: writeKey,
                 userAgentHeaderValue: deviceInfo.getUserAgentHeaderValue()
             )
+            session = RichPushHttpClient.getBasicSession()
+        }
+    }
+
+    func testDownloadFileCreatesExpectedFile() {
+        if let session = session {
```
levibostian, line 45: I like this test and think we should get it running in the CI, at least in the short-term. 

Assuming you agree, could we modify `session` to be non-nil? So we don't need this `if let` condition? I would like us to feel more confident that this test will execute and having this `if let` makes me think sometimes it might not run. 

Also, what do you think about performing a 2nd HTTP request in this test function, downloading the same image URL, and expect that the 2 local file paths are unique?

levibostian (approved):
Should this PR be pointing to `main`? 

The PR title says this is a sample app issue, but it's a NSE rich push image issue. I suggest modifying the PR title to indicate that this is a bug fix for rich push images being more reliable.

Sources/MessagingPush/RichPush/RichPushHttpClient.swift
Shahroz16: they are in `main` so I think they were adding values in there.

Sources/Common/Service/HttpRequestRunner.swift
Shahroz16: i would have hoped naming would have done the trick, as in `suggestedFileName` instead of `uniqueFileName` which was before, because suggested file name even from the name doesn't say its going to be unique. 

and then adding `UUID().uuidString` and referring it as `uniqueFileName` to mention we would have a unique path. 

So from a presepective of someone who is looking for the first time, they would understand whats happening just by looking into names.
 But i can add comment on it, if it helps.

Sources/MessagingPush/RichPush/MessagingPush+RichPush.swift
Shahroz16: yup, that is correct, i'll create a separate PR for it. But good callout.

Tests/MessagingPush/HttpTest.swift
Shahroz16: We could run it in CI, if needed. `if let` was there to ensure it doesn't run on CI but we can update that.

Shahroz16 (commented):
Will create a separate PR directed at `main` with this fix. This was for CDP sample app to be able to pass all tests.

PR #555, opened 2024-02-22
ci: disable CI status checks that we are letting fail during CDP development

levibostian:
During CDP development, we have been OK with letting some of the CI status checks fail before we merge a PR. Instead of merging with some status checks be red (fail), this change disables CI running the checks that we currently let fail. This way, we can get back in the habit of not merging a PR until all CI status checks are green. Making it less likely we will break the main-cdp branch by accident.

### DIFF
--- a/.github/workflows/build-sample-apps.yml
+++ b/.github/workflows/build-sample-apps.yml
@@ -54,13 +54,17 @@ jobs:
         sample-app: 
         # List all sample apps you want to have compiled. 
         # List item is name of directory inside of "Apps" directory for the corresponding app to compile. 
-        - "CocoaPods-FCM"
+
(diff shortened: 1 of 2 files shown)


Shahroz16 (commented):
I would rather have these checks fail just for us to know what things are still breaking rather than suppressing them. 

For example, on every link fail i check if they were caused by `todo/fixme` or some other linting thing that i could have fixed. 

Linting ticket is being worked on as we review this ticket as well and cocoapods one would be right after that. But its my take, if others find suppressing these helpful then go-ahead.

levibostian:
Since multiple team members prefer to have failed status checks, I'll close this PR.

PR #556, opened 2024-02-22
test: fix broken test on main-cdp branch

levibostian:
Fixes failing test on main-cdp branch that was introduced in recent PR merge.

### DIFF
--- a/Tests/Common/Service/UserAgentUtilTest.swift
+++ b/Tests/Common/Service/UserAgentUtilTest.swift
@@ -25,7 +25,10 @@ class UserAgentUtilTest: UnitTest {
     override func setUp() {
         super.setUp()
 
-        userAgentUtil = UserAgentUtilImpl(deviceInfo: deviceInfoMock, sdkConfig: sdkConfig)
+        var sdkConfigWithWrapperConfig = sdkConfig
+        sdkConfigWithWrapperConfig._sdkWrapperConfig = sdkWrapperConfig
+
+        userAgentUtil = UserAgentUtilImpl(deviceInfo: deviceInfoMock, sdkConfig: sdkConfigWithWrapperConfig)
     }
 
     // MARK: getUserAgentHeaderValue


Tests/Common/Service/UserAgentUtilTest.swift
```diff
@@ -25,7 +25,10 @@ class UserAgentUtilTest: UnitTest {
     override func setUp() {
         super.setUp()
 
-        userAgentUtil = UserAgentUtilImpl(deviceInfo: deviceInfoMock, sdkConfig: sdkConfig)
+        var sdkConfigWithWrapperConfig = sdkConfig
+        sdkConfigWithWrapperConfig._sdkWrapperConfig = sdkWrapperConfig
```
mrehan27, line 29: Do we plan to keep `sdkWrapperConfig`? I think it is better to reassess the need of both `sdkWrapperConfig` and `UserAgentUtil`. If they aren't required anymore, I would prefer removing all related code instead of implementing temporary fixes to address false positives in test results. In my opinion, these are more prone to errors. Failing test signals an issue that needs to be fixed. However, once it begins producing false positives, there's high risk it might be ignored unless we're actively tracking it.

mrehan27 (changes requested):

Tests/Common/Service/UserAgentUtilTest.swift
Shahroz16: would second rehan we dont need it anymore

Shahroz16 (approved):
Approving to unblock we can revisit conversation in another pr

Tests/Common/Service/UserAgentUtilTest.swift
levibostian: [Created ticket](https://linear.app/customerio/issue/MBL-137/remove-unneeded-sdkwrapperconfig-and-useragentutil) into backlog.

PR #558, opened 2024-02-22
fix: prevent duplicate automatic screenview events from being tracked

levibostian:
Closes: https://linear.app/customerio/issue/MBL-130/automatic-screen-view-feature-causing-multiple-events-tracked

The SDK keeps track of the last screen automatically tracked. If the new screenview event is equal to this last screen tracked, ignore the request.

The linear ticket gives context on why this solution was chosen over alternative approaches.

commit-id:336b0aff


Tests/DataPipeline/DataPipelineImplementation+ScreenViewsTest.swift
Shahroz16, line 1: can we please add a test for revisit as well? 
- visit a screen 
- visit another screen
- revisit the first screen

Shahroz16 (approved):
Changes look good once the test is added.

Tests/DataPipeline/DataPipelineImplementation+ScreenViewsTest.swift
levibostian: Added.

PR #559, opened 2024-02-23
refactor: inapp config builder

mrehan27:

### Changes

- Added `MessagingInAppConfigBuilder` for building `MessagingInAppConfigOptions` when initializing the module
- Added tests for `MessagingInAppConfigBuilder`
- Simplified `MessagingInAppConfigOptions` struct
- Updated `MessagingInApp` initialization to use builder
- Updated `APITest` for `MessagingInApp`
- Updated APN UIKit sample app use new initialization for `MessagingInApp`
- Fixed test with new changes


Sources/MessagingInApp/Config/MessagingInAppConfigBuilder.swift
```diff
@@ -0,0 +1,65 @@
...
+        guard let siteId = dictionary[Keys.siteId.rawValue] as? String else {
+            throw MessagingInAppConfigBuilderError.missingSiteId
+        }
+        guard let region = (dictionary[Keys.region.rawValue] as? String).map(Region.getRegion) else {
+            throw MessagingInAppConfigBuilderError.missingRegion
+        }
+
+        return MessagingInAppConfigBuilder(siteId: siteId, region: region).build()
+    }
+}
```
levibostian, line 65: I think that this is for SDK wrappers, correct? 

Assuming that's true, I would like to see if there is a more maintainable, less boilerplate solution to building a config object from a dictionary. 

A couple ideas I can think of now would be to generate the code like we generate mocks and digraph today. Or, [use swift property wrappers](https://www.avanderlee.com/swift/property-wrappers/). 

Or, maybe we have a different solution altogether? Should we wait to implement this code when we build the RN SDK?

levibostian (approved):

Sources/MessagingInApp/Config/MessagingInAppConfigBuilder.swift
mrehan27: Yes, you're right, this is for SDK wrappers.

I agree on preferring a more maintainable with less boilerplate solution for building a config object from a dictionary. However, that's outside the scope of this ticket. I would avoid removing this functionality since our native SDKs support it today, and its removal can become blocker for wrappers. So I plan to keep it for the time being. Given that it's utilized internally by our SDKs, we can modify the implementation later without impacting any customer-facing APIs.

Also, can you please create a ticket (or proposal) for this and add your idea (and any additional ideas if you have) to it? So we can consider in near future. May be when we start working on wrapper SDKs? 🤔

PR #560, opened 2024-02-23
refactor: push config builder

mrehan27:

### Changes

- Added `MessagingPushConfigBuilder` for building `MessagingPushConfigOptions` when initializing the module
- Added tests for `MessagingPushConfigBuilder`
- Simplified `MessagingPushConfigOptions` struct
- Updated `MessagingPush` initialization to use builder
- Updated `APITest` for `MessagingPush`
- Added initialization for NSE in `MessagingPushAPN` and `MessagingPushFCM`
- Updated aliases for `MessagingPushAPN` and `MessagingPushFCM`
- Updated APN UIKit sample app use new initialization for `MessagingPushAPN`
- Updated `moduleConfig` to non-static variable retrieving from `implementation`
- Fixed test with new changes

### DIFF
--- a/Tests/MessagingPushFCM/APITest.swift
+++ b/Tests/MessagingPushFCM/APITest.swift
@@ -23,11 +23,16 @@ class MessagingPushFCMAPITest: UnitTest {
         MessagingPush.initialize()
         MessagingPush.initialize(withConfig: MessagingPushConfigBuilder().build())
 
+        // MessagingPushFCM should be able to be initialized with the same initialize() function as MessagingPush.
+        MessagingPushFCM.initialize()
+        MessagingPushFCM.initialize(withConfig: MessagingPushConfigBuilder().build())
+
         // This is the `initialize()` function that's available to Notification Service Extension and not available
         // to other targets (such as iOS).
         // You should be able to uncomment the initialize() function below and should get compile errors saying that the
         // function is not available to iOS.
         // MessagingPush.initialize(withConfig: MessagingPushConfigBuilder(cdpApiKey: "").build())
+        // MessagingPushFCM.initialize(withConfig: MessagingPushConfigBuilder(cdpApiKey: "").build())
 
         // `moduleConfig` is not really meant to be accessed by customers, so it is okay to not have it in the mock.
         // However, it is public so we should make sure it does not change.
--- a/Tests/MessagingPushAPN/APITest.swift
+++ b/Tests/MessagingPushAPN/APITest.swift
@@ -23,11 +23,16 @@ class MessagingPushAPNAPITest: UnitTest {
         MessagingPush.initialize()
         MessagingPush.initialize(withConfig: MessagingPushConfigBuilder().build())
 
+        // MessagingPushAPN should be able to be initialized with the same initialize() function as MessagingPush.
+        MessagingPushAPN.initialize()
+        MessagingPushAPN.initialize(withConfig: MessagingPushConfigBuilder().build())
+
         // This is the `initialize()` function that's available to Notification Service Extension and not available
         // to other targets (such as iOS).
         // You should be able to uncomment the initialize() function below and should get compile errors saying that the
         // function is not available to iOS.
         // MessagingPush.initialize(withConfig: MessagingPushConfigBuilder(cdpApiKey: "").build())
+        // MessagingPushAPN.initialize(withConfig: MessagingPushConfigBuilder(cdpApiKey: "").build())
 
         // `moduleConfig` is not really meant to be accessed by customers, so it is okay to not have it in the mock.
         // However, it is public so we should make sure it does not change.


Sources/MessagingPush/MessagingPush.swift
```diff
@@ -7,10 +7,10 @@ import Foundation
   */
 public class MessagingPush: ModuleTopLevelObject<MessagingPushInstance>, MessagingPushInstance {
     @Atomic public private(set) static var shared = MessagingPush()
-    @Atomic public private(set) static var moduleConfig: MessagingPushConfigOptions = .Factory.create()
     private static let moduleName = "MessagingPush"
 
     private var globalDataStore: GlobalDataStore
+    public var moduleConfig: MessagingPushConfigOptions { implementation?.moduleConfig ?? MessagingPushConfigBuilder().build() }
```
mrehan27, line 13: Previously, we were just modifying the static config. Since it is updated with builder now, I thought it would be better to avoid static and reuse implementation instead.

Sources/MessagingPush/MessagingPush.swift
```diff
@@ -85,22 +76,13 @@ public class MessagingPush: ModuleTopLevelObject<MessagingPushInstance>, Messagi
...
-        configure configureHandler: ((inout MessagingPushConfigOptions) -> Void)? = nil
-    ) -> MessagingPushInstance {
-        if let configureHandler = configureHandler {
-            configureHandler(&moduleConfig)
-        }
-        moduleConfig.cdpApiKey = cdpApiKey
-
-        shared.initializeModuleIfNotAlready()
-
+    public static func initializeForExtension(withConfig config: MessagingPushConfigOptions) -> MessagingPushInstance {
```
mrehan27, line 79: Because two `initialize` with same name and params aren't possible, I modified the name. Open for suggestions though.

Sources/MessagingPush/MessagingPush.swift
```diff
@@ -58,23 +56,16 @@ public class MessagingPush: ModuleTopLevelObject<MessagingPushInstance>, Messagi
...
 
         guard moduleInitializedFirstTime else {
             return shared
         }
 
         // Some part of the initialize is specific only to non-NSE targets.
         // Put those parts in this non-NSE initialize method.
-        if Self.moduleConfig.autoTrackPushEvents {
+        if config.autoTrackPushEvents {
             DIGraphShared.shared.automaticPushClickHandling.start()
```
mrehan27, line 69: If we didn't have this code, we could probably merge `initialize` and `initializeForExtension`. But I think we can't merge them yet because of iOS restrictions for NSE.

Sources/MessagingPush/Config/MessagingPushConfigBuilder.swift
```diff
@@ -0,0 +1,112 @@
...
+        if let autoTrackPushEvents = dictionary[Keys.autoTrackPushEvents.rawValue] as? Bool {
+            builder.autoTrackPushEvents(autoTrackPushEvents)
+        }
+        if let showPushAppInForeground = dictionary[Keys.showPushAppInForeground.rawValue] as? Bool {
+            builder.showPushAppInForeground(showPushAppInForeground)
+        }
+
+        return builder.build()
+    }
+}
```
levibostian, line 112: [Same comment](https://github.com/customerio/customerio-ios/pull/559#discussion_r1501143564)

Sources/MessagingPush/Config/MessagingPushConfigOptions.swift
```diff
@@ -0,0 +1,20 @@
+import CioInternalCommon
+
+/// Defines configuration options for the Customer.io Push Messaging module.
+///
+/// Use `MessagingPushConfigBuilder` for constructing its instances. For detailed usage, see builder class documentation.
+public struct MessagingPushConfigOptions {
+    let cdpApiKey: String
```
levibostian, line 7: Why not public?

Sources/MessagingPush/MessagingPush.swift
levibostian: [Since the config can be injected into classes via digraph](https://github.com/customerio/customerio-ios/pull/560/files#diff-01f1b1edfc257ed559f8b272e971182e1daf2be00dc39a0c343bdbcc45dc5148R16-R20), what do you think about continuing to use the previous `static` implementation and update `MessagingPushImplementation(diGraph: diGraphShared, moduleConfig: config)` to `MessagingPushImplementation(diGraph: diGraphShared)`? 

The static solution is easier for me to read and confirm it's atomic compared to the solution in this PR.

Sources/MessagingPush/MessagingPush.swift
```diff
@@ -51,22 +48,15 @@ public class MessagingPush: ModuleTopLevelObject<MessagingPushInstance>, Messagi
      */
     @discardableResult
     @available(iOSApplicationExtension, unavailable)
-    public static func initialize(
-        configure configureHandler: ((inout MessagingPushConfigOptions) -> Void)? = nil
-    ) -> MessagingPushInstance {
+    public static func initialize(withConfig config: MessagingPushConfigOptions = MessagingPushConfigBuilder().build()) -> MessagingPushInstance {
```
levibostian, line 51: I am assuming there is a default value provided because the builder has no required params? 

If that's true, would a code comment be valuable to remember this decision?

Sources/MessagingPush/MessagingPushInstance.swift
```diff
@@ -10,6 +10,7 @@ import UserNotifications
 // 1. Cannot be used in an App Extension so therefore, we cannot guarantee that the function
 //    exists in the SDK code at compile time.
 public protocol MessagingPushInstance: AutoMockable {
+    var moduleConfig: MessagingPushConfigOptions { get }
```
levibostian, line 13: By adding this line here, does it mean that it's officially part of the public API? 

Assuming it's part of the public API, could we have it added to APITest?

Tests/MessagingPushFCM/APITest.swift
levibostian, line 1: Could you also modify the API tests for APN module? 
Also, add the `initialize()` functions to these tests?

levibostian (approved):

Sources/MessagingPush/Config/MessagingPushConfigBuilder.swift
mrehan27: [Responded](https://github.com/customerio/customerio-ios/pull/559#discussion_r1502207125). We will be keeping it for now.

Sources/MessagingPush/Config/MessagingPushConfigOptions.swift
mrehan27: Previously we didn’t expose it for customers so they don’t get confused when setting config. But with builder pattern, I see that’s not required, so updated to `public`.

Sources/MessagingPush/MessagingPushInstance.swift
mrehan27: Not officially though. I don't think we'll be recommending customers ever to use config from here. But I have added to `APITest` so we know whenever this changes.

Tests/MessagingPushFCM/APITest.swift
mrehan27: Oh yes, missed it. Thanks for reminding, added now.

PR #566, opened 2024-02-27
refactor: remove all fixme's from the source code

ami-aman:
part of: https://linear.app/customerio/issue/MBL-128/ios-cdp-removefix-todos-and-fixmes-in-codebase

Remove/Fix all FIXME markers from the source code that were added while working on CDP. 

### DIFF
--- a/Sources/Common/Communication/Event.swift
+++ b/Sources/Common/Communication/Event.swift
@@ -82,6 +82,20 @@ public enum EventTypesRegistry {
 // Each event struct should have properties relevant to its specific use case.
 // They must include `storageId`, `params`, and any other relevant information.
 
+public struct ProfileAlreadyIdentifiedEvent: EventRepresentable {
+    public let storageId: String
+    public let params: [String: String]
+    public let identifier: String
+    public let timestamp: Date
+
+    public init(storageId: String = UUID().uuidString, identifier: String, timestamp: Date = Date(), params: [String: String] = [:]) {
+        self.storageId = storageId
+        self.identifier = identifier
+        self.timestamp = timestamp
+        self.params = params
+    }
+}
+
 public struct ProfileIdentifiedEvent: EventRepresentable {
     public let storageId: String
     public let params: [String: String]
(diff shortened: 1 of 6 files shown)


Sources/Common/Communication/Event.swift
```diff
@@ -82,6 +82,20 @@ public enum EventTypesRegistry {
 // Each event struct should have properties relevant to its specific use case.
 // They must include `storageId`, `params`, and any other relevant information.
 
+public struct ProfileAlreadyIdentifiedEvent: EventRepresentable {
```
ami-aman, line 85: I created a new event to keep things clear but was wondering if we could use the already existing event struct `ProfileIdentifiedEvent` for the already identified users too since the functioning is same i.e to call the `setProfileIdentifier` method

`self.inAppProvider.setProfileIdentifier(event.identifier)` 🤔

Shahroz16: Yup, you are right, it can be resused.

ami-aman: Done

Sources/DataPipeline/DataPipelineImplementation.swift
```diff
@@ -45,6 +45,13 @@ class DataPipelineImplementation: DataPipelineInstance {
 
         // subscribe to journey events emmitted from push/in-app module to send them via datapipelines
         subscribeToJourneyEvents()
+        postProfileAlreadyIdentifiedToInAppModule()
+    }
+
+    private func postProfileAlreadyIdentifiedToInAppModule() {
+        if let identifier = analytics.userId {
```
Shahroz16, line 52: have we thought about the case, of an `Existing user` who updated the SDK from tracking to CDP runs the App, at this time `analytics.userId` is `nil` and this method gets called. Later we run the migration and set the `analytics.user` from `profileStore` but by that time this method had already been passed.

ami-aman: You're right! I have covered this case in the [latest commit](https://github.com/customerio/customerio-ios/pull/566/commits/7c998765d046036f7a5edf4d126d521198b14429)

Shahroz16 (approved):

PR #567, opened 2024-02-27
chore: user agent updates

Shahroz16:
closes: https://linear.app/customerio/issue/MBL-106/user-agent

- Adds user-agent in payload
- Unify plugin to update context properties changes
- Added test


Sources/Common/Service/UserAgentUtil.swift
```diff
@@ -4,14 +4,12 @@ public protocol UserAgentUtil: AutoMockable {
     func getUserAgentHeaderValue() -> String
 }
 
-// sourcery: InjectRegister = "UserAgentUtil"
+// sourcery: InjectRegisterShared = "UserAgentUtil"
 public class UserAgentUtilImpl: UserAgentUtil {
     private let deviceInfo: DeviceInfo
-    private let sdkWrapperConfig: SdkWrapperConfig?
```
mrehan27, line 10: If `sdkWrapperConfig` is not being used now, can we remove it completely? I'm okay if we do it in another PR or ticket, but just wanted to check if that simply needs to be removed, we can probably do that.

Sources/DataPipeline/Plugins/Context.swift
```diff
@@ -1,14 +1,18 @@
...
     public let type = PluginType.before
     public weak var analytics: Analytics?
 
-    public var token: String?
+    public var deviceToken: String?
     public var attributes: [String: Any] = [:]
 
+    var userAgentUtil: UserAgentUtil {
+        DIGraphShared.shared.userAgentUtil
+    }
```
mrehan27, line 14: I would prefer it as `let` and initializing in constructor so it is assigned only once 🤓

ami-aman (approved):

mrehan27 (approved):
Looks good, just a few comments 👍🏻

Sources/Common/Service/UserAgentUtil.swift
Shahroz16: i wanted to do it a separate PR but just doing it in this PR might be easier 😆

PR #568, opened 2024-02-27
test: update config tests

mrehan27:
helps: [MBL-121](https://linear.app/customerio/issue/MBL-121/write-tests-to-validate-datapipelineconfigoptions-behavior)

### Changes

- Simplified `SdkConfig`
- Updated docs for all configs
- Updated a few config test names for clarity
- Added and improved tests for `SdkConfigBuilder` and `SdkConfig`

### DIFF
--- a/Sources/Common/Store/SdkConfig.swift
+++ b/Sources/Common/Store/SdkConfig.swift
@@ -37,8 +37,6 @@ public struct SdkConfig {
         case logLevel
     }
 
-    /// To help you get setup with the SDK or debug SDK, change the log level of logs you
-    /// wish to view from the SDK.
     public let logLevel: CioLogLevel
 
     // private init to ensure `SdkConfig` can either be created using `SDKConfigBuilder` or `SdkConfig.Factory` only.
--- a/Sources/Common/Store/SdkConfig.swift
+++ b/Sources/Common/Store/SdkConfig.swift
@@ -9,7 +9,7 @@ public struct SdkConfig {
         #if DEBUG
         // Only used for testing purposes.
         public static func create(logLevel: CioLogLevel? = nil) -> SdkConfig {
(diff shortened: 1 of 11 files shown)


Sources/Common/Store/SdkConfig.swift
```diff
@@ -1,51 +1,44 @@
...
         // config features
         case logLevel
     }
 
-    /// To help you get setup with the SDK or debug SDK, change the log level of logs you
-    /// wish to view from the SDK.
-    public var logLevel: CioLogLevel
+    public let logLevel: CioLogLevel
+
+    // private init to ensure `SdkConfig` can either be created using `SdkConfig.Factory` only.
```
Shahroz16, line 40:
```suggestion
    // private init to ensure `SdkConfig` can be created using `SdkConfig.Factory` only.
```

Sources/DataPipeline/Config/SDKConfigBuilder.swift
```diff
@@ -7,10 +7,11 @@ import Segment
 /// the SDK, ensuring all required and optional settings are appropriately configured before the
 /// SDK is initialized.
 ///
-/// Usage Example:
+/// **Usage Example:**
+///
 /// ```
 /// let config = SDKConfigBuilder(cdpApiKey: "your_cdp_api_key")
-///   .apiHost("your_api_host")
```
Shahroz16, line 13: wonder why we removed this? 🤔

Shahroz16 (approved):

Sources/DataPipeline/Config/SDKConfigBuilder.swift
mrehan27: I just tried updating example with more common usage 😄

PR #569, opened 2024-02-28
fix: decoupled autoTrackPushEvents config

Shahroz16:
- Provide users with the flexibility to only disable opened.
- Deprecation of  autoTrackPushEvents
- Handle case for legacy user autoTrackPushEvents with false

closes: https://linear.app/customerio/issue/MBL-138/[bug]-decouple-autotrack-config

### DIFF
--- a/Sources/Common/Store/SdkConfig.swift
+++ b/Sources/Common/Store/SdkConfig.swift
@@ -13,7 +13,7 @@ import UIKit
  let sdkConfigInstance = SdkConfig.Factory.create(region: .US)
  // now, you can modify it
  sdkConfigInstance.trackingApiUrl = "https..."
- sdkConfigInstance.autoTrackPushEvents = false
+ sdkConfigInstance.autoTrackPushMetricEvents = [.delivered,.opened]
  ```
  */
 public struct SdkConfig {
@@ -27,6 +27,7 @@ public struct SdkConfig {
                 region: region,
                 trackingApiUrl: region.productionTrackingUrl,
                 autoTrackPushEvents: true,
+                autoTrackPushMetricEvents: [.delivered, .opened],
                 backgroundQueueMinNumberOfTasks: 10,
                 backgroundQueueSecondsDelay: 30,
                 backgroundQueueExpiredSeconds: Seconds.secondsFromDays(3),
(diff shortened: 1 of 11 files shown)


Sources/Common/Store/SdkConfig.swift
```diff
@@ -27,6 +27,7 @@ public struct SdkConfig {
                 region: region,
                 trackingApiUrl: region.productionTrackingUrl,
                 autoTrackPushEvents: true,
```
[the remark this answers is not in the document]
ami-aman: Or it is still there because you added it as a deprecated property ?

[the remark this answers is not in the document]
Shahroz16: yup its deprecated, we dont want to make a breaking change

Sources/Tracking/CustomerIO.swift
```diff
@@ -202,7 +202,18 @@ public class CustomerIO: CustomerIOInstance {
     private static func initialize(
         config: SdkConfig
     ) {
-        let newDiGraph = DIGraph(sdkConfig: config)
+        var sdkConfig = config
+
+        // temporary handling for backward compatibility with `autoTrackPushEvents`.
+        // if `autoTrackPushEvents` is false (disabled) and `autoTrackPushMetricEvents` is set to track all events (default),
```
mrehan27, line 208: Shouldn't this be like this? 🤔 
```suggestion
        // if `autoTrackPushEvents` is true (enabled) and `autoTrackPushMetricEvents` is set to track all events (default),
```

Shahroz16: its more like, 

```
if `autoTrackPushEvents` = true
- `autoTrackPushMetricEvents` is enabled for both `delivered` and `opened`

if `autoTrackPushEvents` = false
- `autoTrackPushMetricEvents` is enabled for only `delivered`
```

ami-aman (approved):

Sources/MessagingPush/MessagingPush.swift
```diff
@@ -66,7 +66,7 @@ public class MessagingPush: ModuleTopLevelObject<MessagingPushInstance>, Messagi
         let logger = diGraph.logger
         logger.debug("Setting up MessagingPush module...")
 
-        if diGraph.sdkConfig.autoTrackPushEvents {
+        if diGraph.sdkConfig.autoTrackPushMetricEvents.isOpenedEnabled {
             diGraph.automaticPushClickHandling.start()
         }
```
levibostian, line 71: This is a blocker comment, because it could cause unexpected behaviors for customers. 

The automatic push click handling feature does much more then track opened metrics. It also: 
1. Decides [if a push sent from CIO should be shown while app in foreground](https://github.com/customerio/customerio-ios/blob/873eb43f111a4b451ed271b79e8a101d520ee98d/Sources/MessagingPush/PushHandling/iOSPushEventListener.swift#L55). 
2. Opens a deep link if one is attached to a push and a push is clicked. 

If a customer disables opened metrics tracking, they should be able to expect these other features continue to work.

levibostian (changes requested):

PR #575, opened 2024-02-29
fix: expo users reported app crash on didFailToRegisterForRemoteNotificationsWithError

ami-aman:
part of: https://linear.app/customerio/issue/MBL-141/expo-50-with-version-100-beta15-crash-startup-app-on-ios

### DIFF
--- a/Sources/MessagingPushAPN/MessagingPushAPN+PushConfigs.swift
+++ b/Sources/MessagingPushAPN/MessagingPushAPN+PushConfigs.swift
@@ -45,6 +45,6 @@ extension MessagingPushAPN {
     // Swizzled method for `didFailToRegisterForRemoteNotificationsWithError'
     @objc
     func application(_ application: UIApplication, didFailToRegisterForRemoteNotificationsWithError error: Error) {
-        messagingPush.deleteDeviceToken()
+        Self.shared.deleteDeviceToken()
     }
 }


Sources/MessagingPushAPN/MessagingPushAPN+PushConfigs.swift
```diff
@@ -45,6 +45,6 @@ extension MessagingPushAPN {
     // Swizzled method for `didFailToRegisterForRemoteNotificationsWithError'
     @objc
     func application(_ application: UIApplication, didFailToRegisterForRemoteNotificationsWithError error: Error) {
-        messagingPush.deleteDeviceToken()
+        Self.shared.deleteDeviceToken()
```
Shahroz16, line 48: curious to know why is this only crashing for expo users?

ami-aman: This should be crashing for all users but the issue has only been reported by 1 expo user. The use case is rare and only happens if the device is unable to registerFor remote notifications for some reason.

ami-aman:
> Can you write tests for this change?
> 
> It would be great if we had a test that could reproduce the crash and then show that after this change the test passes.

@levibostian I started writing test cases for this swizzled method but it looks like a larger effort and updates than I anticipated. As this fix is a result of a crash report by a customer, would would mind if I open a tech debt and let go this fix ?

Sources/MessagingPushAPN/MessagingPushAPN+PushConfigs.swift
levibostian, line 48: Great additions to the PR description. Thanks! 

From the updated PR description, this is my understanding. 

This bug only happens when: 
* The messaging push module is never initialized in the app, resulting in `messagingPush` instance never being populated. 
* Device token failing to be generated for the host app. 

This seems like a rare bug to encounter with a workaround of initializing the messagingpush module in the app. 

If my understanding is correct, and there is a workaround available, I would appreciate it if we could prioritize writing automated tests in the PR before merging. This will help ensure that the fix is reliable both now and in the future. 

Besides automated testing, I do have another idea to share. 

Could we instead call one of the public-facing API functions like a customer would call? 

```suggestion
        MessagingPushAPN.shared.deleteDeviceToken()
```

If we call the function like a customer would, we can be extra confident that the app would not crash.

ami-aman: Done. Please re-review the PR on priority as the customer is expecting a solution from past 2 weeks.

levibostian (approved):
LGTM

I checked the CI status checks that are failing in this PR and they are also failing in `main`. Therefore, it's OK with me to merge this PR with the current failed status checks and fix the `main` branch in a separate PR.

PR #582, opened 2024-03-01
test: public api tests for DataPipeline

mrehan27:
closes: [MBL-144](https://linear.app/customerio/issue/MBL-144/write-apitest-cases-to-remove-todos-in-source-code)

### Changes

- Removed completed TODOs
- Added API tests for public APIs exposed in `CustomerIO+Plugins` and `CustomerIO+Segment`
- Added mock plugins to help in testing

### DIFF
--- a/Sources/DataPipeline/CustomerIO+Plugins.swift
+++ b/Sources/DataPipeline/CustomerIO+Plugins.swift
@@ -2,7 +2,6 @@ import CioInternalCommon
 import Foundation
 import Segment
 
-// TODO: Add APITest for it?
 public extension CustomerIO {
     /**
      Applies the supplied closure to the currently loaded set of plugins.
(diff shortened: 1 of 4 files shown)


Tests/Common/APITest.swift
```diff
@@ -37,6 +37,7 @@ class CommonAPITest: UnitTest {
         // Factory method should allow nil values for `SdkConfig` to enable fallback to defaults.
         _ = SdkConfig.Factory.create(logLevel: nil)
         // Wrapper SDKs should be able to create a `SdkConfig` from a dictionary.
-        _ = SdkConfig.Factory.create(from: [:])
+        let configOptions: [String: Any] = [:]
```
mrehan27, line 40: While writing tests for the TODOs, I realized the type was unspecified in these tests. So, I've updated them to ensure the dictionary type is specific to what we expect, rather than generic.

Shahroz16 (approved):
looks good 👍

PR #583, opened 2024-03-01
fix: forward push response events to 3rd party callback functions, for CIO push

levibostian:
Part of: https://linear.app/customerio/issue/MBL-138/[bug]-ios-customers-receive-callbacks-from-3rd-party-sdks-for-pushes

When a push response event (push clicked or swiped away) happens, send the push event to other push event handlers in host app. Giving customers the opportunity to process the push. Previously, only non-CIO pushes would be sent to other push event handlers.

Notes:

* Scenario where CIO pushes are not handled (opened metric not tracked, deep link not opened):

See new test function `test_givenMultiplePushHandlers_givenClickedOnCioPush_givenOtherPushHandlerDoesNotCallCompletionHandler_expectCompletionHandlerDoesNotGetCalled` that reproduces scenario.

When the push event is sent to other push event handlers in the app, this is an async operation. The CIO SDK waits until all of the other push handlers call the async completion handler indicating that they are done processing the push. If 1 of those push handlers does not call the completion handler, the CIO SDK will not process the push (track opened metric, open deep link).

If we do not wait for other push handlers to call the async completion handler, there is a risk that customers push processing logic could get terminated early by iOS.

Waiting for other push handlers aligns with Apple's recommendations. Apple notes in their documentation that you need to call the completion handler after you're done processing a push event. https://developer.apple.com/documentation/usernotifications/unusernotificationcenterdelegate/usernotificationcenter(_:didreceive:withcompletionhandler:)

---

**Stack**:
- #634
- #633
- #601
- #590
- #584
- #583 ⬅

### DIFF
--- a/Sources/MessagingPush/PushHandling/PushEventHandlerProxy.swift
+++ b/Sources/MessagingPush/PushHandling/PushEventHandlerProxy.swift
@@ -26,25 +26,39 @@ class PushEventHandlerProxyImpl: PushEventHandlerProxy {
     public static let shared = PushEventHandlerProxyImpl()
 
     // Use a map so that we only save 1 instance of a given handler.
-    private var nestedDelegates: [String: PushEventHandler] = [:]
+    @Atomic private var nestedDelegates: [String: PushEventHandler] = [:]
 
     func addPushEventHandler(_ newHandler: PushEventHandler) {
         nestedDelegates[String(describing: newHandler)] = newHandler
     }
 
     func onPushAction(_ pushAction: PushNotificationAction, completionHandler: @escaping () -> Void) {
+        let nestedDelegates = self.nestedDelegates // create a scoped copy for this function body.
+
         // If there are no other click handlers, then call the completion handler. Indicating that the CIO SDK handled it.
         guard !nestedDelegates.isEmpty else {
             completionHandler()
             return
         }
 
-        nestedDelegates.forEach { _, delegate in
-            delegate.onPushAction(pushAction, completionHandler: completionHandler)
+        Task {
+            // Wait for all other push event handlers to finish before calling the completion handler.
+            // Each iteration of the loop waits for the push event to be processed by the delegate.
+            for delegate in nestedDelegates.values {
+                await withCheckedContinuation { continuation in
+                    delegate.onPushAction(pushAction) {
+                        continuation.resume()
+                    }
+                }
+            }
+            // After the loop finishes, call the completion handler to indicate the event has been fully processed by all delegates.
+            completionHandler()
         }
     }
 
     func shouldDisplayPushAppInForeground(_ push: PushNotification, completionHandler: @escaping (Bool) -> Void) {
+        let nestedDelegates = self.nestedDelegates // create a scoped copy for this function body.
+
         // If there are no other click handlers, then call the completion handler. Indicating that the CIO SDK handled it.
         guard !nestedDelegates.isEmpty else {
             completionHandler(true)
--- a/Sources/MessagingPush/PushHandling/iOSPushEventListener.swift
+++ b/Sources/MessagingPush/PushHandling/iOSPushEventListener.swift
@@ -44,12 +44,19 @@ class IOSPushEventListener: PushEventHandler {
 
         logger.debug("Push came from CIO. Handle the didReceive event on behalf of the customer.")
 
-        if pushAction.didClickOnPush {
-            pushClickHandler.pushClicked(push)
-        }
+        // Forward event to other push click handlers so they can receive a callback about this push event.
+        pushEventHandlerProxy.onPushAction(pushAction, completionHandler: {
+            // When this block of code executes, the customer is done processing the push event.
 
-        // call the completion handler so the customer does not need to.
-        completionHandler()
+            // We do not open deep link until after customer is done processing the push event in case the deep link would leave the app.
+            // We want to make sure the customer has a chance to process the push event before leaving the app.
+            if pushAction.didClickOnPush {
+                self.pushClickHandler.pushClicked(push)
+            }
+
+            // call the completion handler, indicating to the OS that we are done processing the push.
+            completionHandler()
+        })
     }
 
     func shouldDisplayPushAppInForeground(_ push: PushNotification, completionHandler: @escaping (Bool) -> Void) {
--- a/Tests/MessagingPush/PushHandling/AutomaticPushClickedIntegrationTest.swift
+++ b/Tests/MessagingPush/PushHandling/AutomaticPushClickedIntegrationTest.swift
@@ -25,6 +25,28 @@ class AutomaticPushClickedIntegrationTest: IntegrationTest {
         )
     }
 
+    // MARK: push clicked
+
+    func test_pushClicked_expectRunCodeWhenAPushIsClicked() {
+        let givenPush = PushNotificationStub.getPushSentFromCIO()
+
+        performPushAction(PushNotificationActionStub(push: givenPush, didClickOnPush: true))
+
+        // We expect push click handler to be called when a push is clicked.
+        XCTAssertTrue(pushClickHandler.pushClickedCalled)
+    }
+
+    // MARK: push swiped away
+
+    func test_pushSwipedAway_expectDoNotRunCodeWhenPushIsClicked() {
+        let givenPush = PushNotificationStub.getPushSentFromCIO()
+
+        performPushAction(PushNotificationActionStub(push: givenPush, didClickOnPush: false))
+
+        // We expect push click handler to be called when a push is clicked.
+        XCTAssertFalse(pushClickHandler.mockCalled)
+    }
+
     // MARK: When the CIO SDK is the only click handler in the app.
 
     func test_givenCioSdkOnlyPushHandler_givenClickedOnCioPush_expectEventHandled() {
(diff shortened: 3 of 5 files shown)


Sources/MessagingPush/PushHandling/iOSPushEventListener.swift
```diff
@@ -44,12 +44,19 @@ class IOSPushEventListener: PushEventHandler {
...
+            // When this block of code executes, the customer is done processing the push event.
+
+            // We do not open deep link until after customer is done processing the push event in case the deep link would leave the app.
+            // We want to make sure the customer has a chance to process the push event before leaving the app.
+            if pushAction.didClickOnPush {
+                self.pushClickHandler.pushClicked(push)
+            }
+
+            // call the completion handler, indicating to the OS that we are done processing the push.
+            completionHandler()
```
Ahmed-CIO, line 58: while I can tell how this fixes the existing issue, I see it possibly introducing another issue.
What if the customer didn't call the completionHandler we are passing here `pushEventHandlerProxy.onPushAction` ? neither our code or the customer code will run

Ahmed-CIO (changes requested):

Sources/MessagingPush/PushHandling/iOSPushEventListener.swift
levibostian: I wish that was not the case, but that is correct. 

I called this out and provided more details about this in the PR description. 

The only way that I can think of to avoid this scenario is to call the 3rd party callback functions and not wait for them to call the async completion handler. But, that runs the risk of customer's push processing getting terminated early by the OS which I think is worse. 
 
Terminating customer's processing a push + Apple's docs saying it's required to call completion handler when done processing a push, I do think this is the best implementation we have available.

Ahmed-CIO: What's the problem with doing this
```
 if pushAction.didClickOnPush {
       self.pushClickHandler.pushClicked(push)
}

pushEventHandlerProxy.onPushAction(pushAction, completionHandler: {

            // call the completion handler, indicating to the OS that we are done processing the push.
            completionHandler()
}
 ```
 
 At this point, our click handler guaranteed to run, no?
 If there is a problem on the client side that doesn't call the `completionHandler` then it is a bug in their code and has nothing to do with our implementation, no?

File: Sources/MessagingPush/PushHandling/PushEventHandlerProxy.swift [hunk 2]
```diff
@@ -39,8 +39,18 @@ class PushEventHandlerProxyImpl: PushEventHandlerProxy {
...
         nestedDelegates.forEach { _, delegate in
-            delegate.onPushAction(pushAction, completionHandler: completionHandler)
+            delegate.onPushAction(pushAction, completionHandler: {
+                numberOfAsyncOperationsCompleted += 1
+
+                // Do not call the function's completionHandler until *all* delegates have had their completion handler called.
+                if numberOfAsyncOperationsExpected == numberOfAsyncOperationsCompleted {
+                    completionHandler()
+                }
+            })
```
Shahroz16, line 53: a couple of things, 
- using DispatchGroup over variable tracking could help with synchronization and thread safety
- can we include timeout to ensure, `completionHandler()` is called in case it's not called by the user? something like this? 



```
func onPushAction(_ pushAction: PushNotificationAction, completionHandler: @escaping () -> Void) {
    let dispatchGroup = DispatchGroup()
    let timeoutInterval = 10.0 // Timeout after 10 seconds?

    for delegate in nestedDelegates.values {
        dispatchGroup.enter()
        delegate.onPushAction(pushAction) {
            dispatchGroup.leave()
        }
    }

    // Completion block to execute once all handlers have completed.
    let completionBlock = {
        // All delegates have processed the push action, or a timeout occurred.
        // Proceed with CIO SDK-specific logic.
        completionHandler()
    }

    // Set up the completion notification.
    dispatchGroup.notify(queue: .main, execute: completionBlock)

    // Set up a timeout mechanism.
    DispatchQueue.main.asyncAfter(deadline: .now() + timeoutInterval) {
        completionBlock()
    }
}
```

Sources/MessagingPush/PushHandling/iOSPushEventListener.swift
Ahmed-CIO: > but i believe the callback onPushAction in only called, when all the other delegates have called the completionHandler.

Which is yet another concern. I will respond to your suggestion. At this point my gut feeling is that I need to understand the problem we are solving better. Otherwise, I am feeling like I am contributing to work around a problem that's not fully clear to me

Sources/MessagingPush/PushHandling/PushEventHandlerProxy.swift
Ahmed-CIO: This suggestion is worrisome. If my understand is correct, then if a callback in any delegate isn't called the `dispatchGroup.leave()` won't be called.

Then we pretty much have a queue handing forever, yes it is non-blocking, but it still a resource that's created and for every push and remained there.

Which seems reasonable to say that we are trying to synchronise with async code (the `delegate .onPushAction`) that we don't own. While this can seem like a trivial problem, it is not and we need to be careful how we proceed.

Specially the fact that we needed to introduce timeout, if you think about it, this is what the OS watchdog does, and this is no trivial task either.

Let's talk more about this in the comms window to make sure at least we are on the same page of what's the problem we are solving

Sources/MessagingPush/PushHandling/iOSPushEventListener.swift
levibostian: In a call today, we made the decision to move forward with de-coupling `opened` metrics tracking and opening deep links. That way we can make this feature more reliable, in case someone doesn't call the async completion handler. 

[I created a sub-issue for that change](https://linear.app/customerio/issue/MBL-154/track-opened-push-metrics-3rd-party-push-callback-functions-get-called) and will open another feature for that change onto this stack. 

I believe that this means this conversation can be resolved.

Sources/MessagingPush/PushHandling/PushEventHandlerProxy.swift
levibostian: Conversation happening in [the other PR](https://github.com/customerio/customerio-ios/pull/584#discussion_r1509523944).

Sources/MessagingPush/PushHandling/iOSPushEventListener.swift
levibostian: [New PR](https://github.com/customerio/customerio-ios/pull/590) introduced in this stack to cover this.

Sources/MessagingPush/PushHandling/PushEventHandlerProxy.swift
```diff
@@ -39,8 +39,18 @@ class PushEventHandlerProxyImpl: PushEventHandlerProxy {
             return
         }
 
-        nestedDelegates.forEach { _, delegate in
-            delegate.onPushAction(pushAction, completionHandler: completionHandler)
+        Task {
+            // Wait for all other push event handlers to finish before calling the completion handler.
+            // Each iteration of the loop waits for the push event to be processed by the delegate.
+            for delegate in nestedDelegates.values {
```
Shahroz16, line 45: great start !! but i do believe we are still prone to race condition, because we are not explicitly handling concurrent access to the `nestedDelegates` dictionary, which could lead to race conditions if multiple threads attempted to modify this shared resource simultaneously.

what do you think about making `nestedDelegates` as an actor?

levibostian: I changed `nestedDelegates` to be `@Atomic` which should solve the problem you mentioned. 

I do see a lot of value in `actor`. For this situation, I think it would require a bigger refactor to use it. 

To use `actor` to protect `nestedDelegates`, we would need to make the Proxy class an `actor` instead of a `class`. If we made that change, then we would also need to make functions such as `addPushEventHandler` to be `async`. That change would require a lot of code changes in the project that I do not think is worth it at this time.

Ahmed-CIO: Before guard condition add something like `let nestedDelegatesCopy = nestedDelegates`
 Then use the `nestedDelegatesCopy` throughout the method. This ensures the array copy you checked at the beginning of the method, won't change in the middle of the method as it is a local copy.

levibostian: [Made change](https://github.com/customerio/customerio-ios/compare/97e5e022765adf66f9baf088b160173d9bf284e5..e8c670f6722df673959cd887a98002a1f4f42f81)

Shahroz16: I'll be okay with these changes for now. 

Regarding:
> To use an actor to protect nestedDelegates, we would need to make the Proxy class an actor instead of a class. If we made that change, then we would also need to make functions such as addPushEventHandler to be async

I don't think we need to do all that, we can just do something like this and it should work too. 

```
actor NestedDelegatesActor {
    private var nestedDelegates: [String: PushEventHandler] = [:]

    var hasDelegates: Bool {
        !nestedDelegates.isEmpty
    }

    func addDelegate(_ delegate: PushEventHandler) {
        nestedDelegates[String(describing: delegate)] = delegate
    }

    func removeDelegate(_ delegate: PushEventHandler) {
        nestedDelegates.removeValue(forKey: String(describing: delegate))
    }

    func forEachDelegate(_ action: @escaping (PushEventHandler) async -> Void) async {
        for delegate in nestedDelegates.values {
            await action(delegate)
        }
    }
}
```

```
class PushEventHandlerProxyImpl: PushEventHandlerProxy {
    /*
     # Why is this class not stored in the digraph?

     Similar to why the SDK's `UNUserNotificationCenterDelegate` instance is also not in the digraph. See those comments to learn more.
     */
    public static let shared = PushEventHandlerProxyImpl()

    // Use a map so that we only save 1 instance of a given handler.
    private var nestedDelegates = NestedDelegatesActor()

        Task {
            await nestedDelegates.addDelegate(newHandler)
        }
    }

    func onPushAction(_ pushAction: PushNotificationAction, completionHandler: @escaping () -> Void) {
        Task {
            await nestedDelegates.forEachDelegate { delegate in
                await withCheckedContinuation { continuation in
                    delegate.onPushAction(pushAction) {
                        continuation.resume()
                    }
                }
            }
            completionHandler()
        }
    }
 }
```

but i am okay with revisiting it later.

Shahroz16 (approved):

Ahmed-CIO (approved):

levibostian:

PR #584, opened 2024-03-01
fix: forward push received app in foreground events to 3rd party callbacks

levibostian:
Part of: https://linear.app/customerio/issue/MBL-138/[bug]-ios-customers-receive-callbacks-from-3rd-party-sdks-for-pushes

When a push is received while app in foreground, send the push event to other push event handlers in host app. Giving customers the opportunity to process the push event. Previously, only non-CIO pushes would be sent to other push event handlers.

Notes:

* CIO SDK waits for all push handlers to call completion handler for push processing. Not calling completion handler may result in push not being shown while app is in foreground.

Push event handlers must call async completion handler for CIO SDK to handle push events. Test function `test_givenMultiplePushHandlers_givenCioPushDelivered_givenOtherPushHandlerDoesNotCallCompletionHandler_expectCompletionHandlerDoesNotGetCalled` reproduces scenario.

Waiting for completion handler prevents premature termination of push processing logic by iOS. Apple documentation states completion handler must be called after processing push event. https://developer.apple.com/documentation/usernotifications/unusernotificationcenterdelegate/usernotificationcenter(_:willpresent:withcompletionhandler:)

* The SDK now determines if a push should be shown for non-CIO pushes and fixes bug in `PushEventHandlerProxy`.

This change had to be made as part of waiting for all push handlers to complete async operation.

Behavior before: The push handler that called the async completion handler first would be the one that decides if the push is displayed on the device.

After behavior: The SDK uses Apple's default of not displaying push. The SDK will display the push if at least 1 of the push handlers decides to display push.

This aligns with Apple's default behavior of not displaying a push unless at least one push handler decides to display it.https://developer.apple.com/documentation/usernotifications/unusernotificationcenterdelegate/usernotificationcenter(_:willpresent:withcompletionhandler:)

---

**Stack**:
- #634
- #633
- #601
- #590
- #584 ⬅
- #583

### DIFF
--- a/Sources/MessagingPush/PushHandling/PushEventHandlerProxy.swift
+++ b/Sources/MessagingPush/PushHandling/PushEventHandlerProxy.swift
@@ -65,8 +65,28 @@ class PushEventHandlerProxyImpl: PushEventHandlerProxy {
             return
         }
 
-        nestedDelegates.forEach { _, delegate in
-            delegate.shouldDisplayPushAppInForeground(push, completionHandler: completionHandler)
+        Task {
+            // 2+ other push event handlers may exist in app. We need to decide if a push should be displayed or not, by combining all the results from all other push handlers.
+            // To do that, we start with Apple's default value of: do not display.
+            // If any of the handlers return result indicating push should be displayed, we return true.
+            // Apple docs: https://developer.apple.com/documentation/usernotifications/unusernotificationcenterdelegate/usernotificationcenter(_:willpresent:withcompletionhandler:)
+            var shouldDisplayPush = false
+
+            // Wait for all other push event handlers to finish before calling the completion handler.
+            // Each iteration of the loop waits for the push event to be processed by the delegate.
+            for delegate in nestedDelegates.values {
+                await withCheckedContinuation { continuation in
+                    delegate.shouldDisplayPushAppInForeground(push, completionHandler: { delegateShouldDisplayPushResult in
+                        if delegateShouldDisplayPushResult {
+                            shouldDisplayPush = true
+                        }
+
+                        continuation.resume()
+                    })
+                }
+            }
+            // After the loop finishes, call the completion handler to indicate the event has been fully processed by all delegates.
+            completionHandler(shouldDisplayPush)
         }
     }
 }
--- a/Sources/MessagingPush/PushHandling/iOSPushEventListener.swift
+++ b/Sources/MessagingPush/PushHandling/iOSPushEventListener.swift
@@ -83,9 +83,17 @@ class IOSPushEventListener: PushEventHandler {
 
         logger.debug("Push came from CIO. Handle the willPresent event on behalf of the customer.")
 
-        let shouldShowPush = moduleConfig.showPushAppInForeground
+        // Forward event to other push handlers so they can receive a callback about this push event.
+        pushEventHandlerProxy.shouldDisplayPushAppInForeground(push, completionHandler: { _ in
+            // When this block of code executes, the customer is done processing the push event.
 
-        // Call the completionHandler so customer does not need to. The push came from CIO, so it gets handled by the CIO SDK.
-        completionHandler(shouldShowPush)
+            // Because push came from CIO, ignore the return result of other push handlers.
+            // Determine if CIO push should be shown from SDK config
+            let shouldShowPush = self.moduleConfig.showPushAppInForeground
+
+            // The push came from CIO, so it gets handled by the CIO SDK.
+            // Calling the completion handler indicates to the OS that we are done processing the push.
+            completionHandler(shouldShowPush)
+        })
     }
 }
--- a/Tests/MessagingPush/PushHandling/AutomaticPushDeliveredAppInForegroundTest.swift
+++ b/Tests/MessagingPush/PushHandling/AutomaticPushDeliveredAppInForegroundTest.swift
@@ -4,15 +4,14 @@ import Foundation
 import SharedTests
 import XCTest
 
+// When a push is delivered to an iOS device and the app is in the foreground, iOS gives the app the option to either display that push or to not display that push.
 class AutomaticPushDeliveredAppInForegrondTest: IntegrationTest {
     private var pushEventHandler: PushEventHandler!
 
     private let pushClickHandler = PushClickHandlerMock()
     private let pushEventHandlerProxy = PushEventHandlerProxyImpl()
 
-    // MARK: test handling when push delivered, app in foreground
-
-    // When a push is delivered to an iOS device and the app is in the foreground, iOS gives the app the option to either display that push or to not display that push.
+    // MARK: SDK configuration behavior
 
     func test_givenCioPushDelivered_givenSdkConfigDisplayPush_expectPushDisplayed() {
         configureSdk(shouldDisplayPushAppInForeground: true)
@@ -21,7 +20,7 @@ class AutomaticPushDeliveredAppInForegrondTest: IntegrationTest {
 
         let didPushGetDisplayed = deliverPush(givenPush)
 
-        XCTAssertTrue(didPushGetDisplayed)
+        XCTAssertEqual(didPushGetDisplayed, true)
     }
 
     func test_givenCioPushDelivered_givenSdkConfigDoNotDisplayPush_expectPushNotDisplayed() {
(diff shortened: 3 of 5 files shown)


Sources/MessagingPush/PushHandling/PushEventHandlerProxy.swift
```diff
@@ -61,8 +61,29 @@ class PushEventHandlerProxyImpl: PushEventHandlerProxy {
...
+                if delegateShouldDisplayPushResult {
+                    shouldDisplayPush = true
+                }
+
+                numberOfAsyncOperationsCompleted += 1
+
+                // Do not call the function's completionHandler until *all* delegates have had their completion handler called.
+                if numberOfAsyncOperationsExpected == numberOfAsyncOperationsCompleted {
+                    completionHandler(shouldDisplayPush)
+                }
```
Ahmed-CIO, line 85: There are race conditions in this code that can result in the completionHandler in the if conditions runs too early or never run

Ahmed-CIO (changes requested):

Sources/MessagingPush/PushHandling/PushEventHandlerProxy.swift
levibostian: "never run" - I suggest we cover that use case in [this thread](https://github.com/customerio/customerio-ios/pull/583#discussion_r1509511361). 

"runs too early" - Could you give steps on how to reproduce this?

Ahmed-CIO: > "never run" - I suggest we cover that use case in https://github.com/customerio/customerio-ios/pull/583#discussion_r1509511361.

This has nothing to with the mentioned thread. This could happen if the race condition caused the counter to miss one count. Then it will never reach the `numberOfAsyncOperationsExpected`.

> "runs too early" - Could you give steps on how to reproduce this?

Hard to explain without lengthy explanation but let's try.
in a simple scenario where `numberOfAsyncOperationsExpected` is set to 2.
Now imagine two threads increase the counter; the first one to go past line 80 after the two threads increased the counter, will meet the condition and call the completion handler, in undetermined order. Then the second one, will call it again.

Basically, there is a race condition, all sort of shenanigans can go.

What's even worse with such simple value (int in this case) that it can't be torn apart (there are exceptions but not needed here); meaning you won't access it in a way that ends up accessing half-baked memory or getting seg_fault errors. So you will never see a crash out of that, but you would see random behavior.

Ahmed-CIO (changes requested):

Sources/MessagingPush/PushHandling/PushEventHandlerProxy.swift
Shahroz16: I agree with the race condition possibility here, this is why i suggested using `Dispatch Group` in my comment in the other [PR](https://github.com/customerio/customerio-ios/pull/583/files#r1510386994).

Shahroz16 (commented):
similar comments and concerns as to the other [PR](https://github.com/customerio/customerio-ios/pull/583).

Sources/MessagingPush/PushHandling/PushEventHandlerProxy.swift
levibostian: I am on the same page with what you mean by race condition now. Thanks. 

Going off of Shahroz's suggestion, DispatchGroup could be useful here. I don't currently agree with implementing a timeout, although I do understand the value of it. I think the OS should enforce the timeout. It's important to me that we keep our code as close to OS behavior as possible for customers. 

With the timeout code removed, a DisptchGroup code snippet could look similar to this: 

```swift  
let dispatchGroup = DispatchGroup()

for (_, delegate) in nestedDelegates {
    dispatchGroup.enter()

    delegate.onPushAction(pushAction) {
        dispatchGroup.leave()
    }
}

dispatchGroup.notify(queue: .main) {
    completionHandler()
}
```

Unless I hear differently, I will update both PRs with this pattern. I will do that [after working on this ticket](https://linear.app/customerio/issue/MBL-154/process-push-event-before-we-call-3rd-party-callback-functions) to give the team more time to review this PR and give suggestions.

Shahroz16: Are we okay with the fact, we won't be opening deeplinks for a user if any of the `completionHandler()` is not called from any of the delegates.

Shahroz16: Are we okay with the fact, we won't be opening deeplinks for a user if any of the `completionHandler()` is not called from any of the delegates?

Shahroz16: additionally, i think @Ahmed-CIO mentioned they would not prefer to go with `DispatchGroup`, so I would wait for him to raise any concern here before we proceed with it.

Ahmed-CIO: Thanks both for the farther elaboration here.
The point about deep linking is also really important one. Not opening deep linking if a delegate didn't call the completionHandler can also fire back at us.

I will read through the updates I missed and come back to this one again

Ahmed-CIO: Alright, after quick catch, I also did quick refresher about DispatchGroup and my worry about causing deadlock isn't actually valid. 
The only thing that still valid is using dispatchGroup.enter() must be balanced with dispatchGroup.leave(). While in the above code this is the case and easy to reason about, this is prone to all sort of issues one used to do in the early days of manually calling alloc/release. However, this is not as a hard blocker as the point I initially raised and turns out to be wrong.
My recommendation: if it is not too much work to create a small actor that only contains the parts that needs to synced, then let's do that as it is likely to be much cleaner solution. Otherwise, go with DispatchGroup. If it helps, here is a quick non-fully-validated example of how actor could help
```swift
actor PushDelegateAsyncOperationsTracker {
    private var numberOfAsyncOperationsExpected: Int
    private var shouldShowNotification: Bool = false
    private let completionHandler: (Bool) -> Void
    
    init(_ numberOfAsyncOperationsExpected: Int,
         completionHandler: @escaping (Bool) -> Void) {
        self.numberOfAsyncOperationsExpected = numberOfAsyncOperationsExpected
        self.completionHandler = completionHandler
    }
    
    
    func onAsyncOperationExecuted(_ shouldShowNotification: Bool) {
        self.numberOfAsyncOperationsExpected -= 1
        self.shouldShowNotification = self.shouldShowNotification || shouldShowNotification
        
        if self.numberOfAsyncOperationsExpected == 0 {
            self.completionHandler(self.shouldShowNotification)
        }
    }
}

delegate.shouldDisplayPushAppInForeground(push)  { delegateShouldDisplayPushResult in
        Task {
            await operationsTracker.onAsyncOperationExecuted(delegateShouldDisplayPushResult)
        }
 })
    
```

levibostian: Wrote some code and shared it with team in Slack. After response, I'll update this PR.

Sources/MessagingPush/PushHandling/PushEventHandlerProxy.swift
```diff
@@ -61,8 +61,28 @@ class PushEventHandlerProxyImpl: PushEventHandlerProxy {
...
+        Task {
+            // 2+ other push event handlers may exist in app. We need to decide if a push should be displayed or not, by combining all the results from all other push handlers.
+            // To do that, we start with Apple's default value of: do not display.
+            // If any of the handlers return result indicating push should be displayed, we return true.
+            // Apple docs: https://developer.apple.com/documentation/usernotifications/unusernotificationcenterdelegate/usernotificationcenter(_:willpresent:withcompletionhandler:)
+            var shouldDisplayPush = false
+
+            // Wait for all other push event handlers to finish before calling the completion handler.
+            // Each iteration of the loop waits for the push event to be processed by the delegate.
+            for delegate in nestedDelegates.values {
```
Shahroz16, line 73: same comment about `nestedDelegates` as in the other [PR](https://github.com/customerio/customerio-ios/pull/583#discussion_r1516079941).

Sources/MessagingPush/PushHandling/PushEventHandlerProxy.swift
```diff
@@ -61,8 +61,28 @@ class PushEventHandlerProxyImpl: PushEventHandlerProxy {
...
+                        if delegateShouldDisplayPushResult {
+                            shouldDisplayPush = true
+                        }
+
+                        continuation.resume()
+                    })
+                }
+            }
+            // After the loop finishes, call the completion handler to indicate the event has been fully processed by all delegates.
+            completionHandler(shouldDisplayPush)
```
Shahroz16, line 85: an existing conversation going on it, regarding we may not show the push in forground if the delegates didn't give callback.

Sources/MessagingPush/PushHandling/PushEventHandlerProxy.swift
levibostian: This has been updated and ready for re-review.

Sources/MessagingPush/PushHandling/PushEventHandlerProxy.swift
levibostian: This problem has been resolved with new PR: https://github.com/customerio/customerio-ios/pull/601

Ahmed-CIO (approved):
Check the comments in case you want to address the nit comment

Shahroz16 (approved):

Ahmed-CIO (approved):

levibostian:

PR #587, opened 2024-03-03
chore: linting fixes

Shahroz16:
part 1 of the [linting fixes](https://linear.app/customerio/issue/MBL-151/linting-fixes)

### DIFF
--- a/Sources/Common/CustomerIOInstance.swift
+++ b/Sources/Common/CustomerIOInstance.swift
@@ -22,6 +22,7 @@ public protocol CustomerIOInstance: AutoMockable {
      */
     func identify(userId: String, traits: [String: Any]?)
 
+    // swiftlint:disable orphaned_doc_comment
     /**
      Identify a customer (aka: Add or update a profile).
      [Learn more](https://customer.io/docs/identifying-people/) about identifying a customer in Customer.io
@@ -33,6 +34,7 @@ public protocol CustomerIOInstance: AutoMockable {
      [Learn more](https://customer.io/docs/api/#operation/identify)
      - traits: Request body of identifying profile. Use to define user attributes.
      */
+    // swiftlint:enable orphaned_doc_comment
     // sourcery:Name=identifyEncodable
     // sourcery:DuplicateMethod=identify
     func identify<RequestBody: Codable>(
@@ -95,6 +97,7 @@ public protocol CustomerIOInstance: AutoMockable {
      */
     func track(name: String, properties: [String: Any]?)
 
+    // swiftlint:disable orphaned_doc_comment
     /**
      Track an event
      [Learn more](https://customer.io/docs/events/) about events in Customer.io
@@ -102,6 +105,7 @@ public protocol CustomerIOInstance: AutoMockable {
      - name: Name of the event you want to track.
      - properties: Optional event body of properties about the event.
      */
+    // swiftlint:enable orphaned_doc_comment
     // sourcery:Name=trackEncodable
     // sourcery:DuplicateMethod=track
     func track<RequestBody: Codable>(
@@ -122,6 +126,7 @@ public protocol CustomerIOInstance: AutoMockable {
      */
     func screen(title: String, properties: [String: Any]?)
 
+    // swiftlint:disable orphaned_doc_comment
     /**
      Track a screen view
      [Learn more](https://customer.io/docs/events/) about events in Customer.io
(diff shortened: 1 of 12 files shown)


Sources/Common/Util/KeyValueStorage.swift
```diff
@@ -29,7 +29,7 @@ public protocol SandboxedSiteIdKeyValueStorage {
  Uses UserDefaults to store data in key/value pairs.
  */
 // sourcery: InjectRegisterShared = "SandboxedSiteIdKeyValueStorage"
-public class UserDefaultsSandboxedSiteIdKeyValueStorage: SandboxedSiteIdKeyValueStorage {
+public class UserDefaultsSandboxedSiteIdKVStore: SandboxedSiteIdKeyValueStorage {
```
mrehan27, line 32: Any specific reason to change this class name? I think the name was exceeding the length, correct? 🤔

Tests/DataPipeline/Config/SDKConfigBuilderTest.swift
```diff
@@ -78,7 +78,11 @@ class SDKConfigBuilderTest: UnitTest {
         let analyticsConfig = Configuration(writeKey: givenCdpApiKey).values
 
         var analyticsDefaultExpectedSettings = Settings(writeKey: givenCdpApiKey)
-        analyticsDefaultExpectedSettings.integrations = try! JSON(["Segment.io": true])
+        do {
+            analyticsDefaultExpectedSettings.integrations = try JSON(["Segment.io": true])
+        } catch {
+            XCTFail("Failed to create JSON with error: \(error)")
```
mrehan27, line 84: Might be more descriptive slightly

```suggestion
            XCTFail("Failed to setup test integrations with error: \(error)")
```

Tests/DataPipeline/DataPipelineImplementation+ScreenViewsTest.swift
```diff
@@ -6,25 +6,30 @@ import SwiftUI
 import UIKit
 import XCTest
 
-class DataPipelineImplementationScreenViewsTest: IntegrationTest {
-    private var autoTrackingScreenViews: AutoTrackingScreenViews!
-    private var outputReader: OutputReaderPlugin!
+class DataPipelineScreenViewsTest: IntegrationTest {
+    private var autoTrackingScreenViews: AutoTrackingScreenViews?
+    private var outputReader: OutputReaderPlugin?
```
mrehan27, line 11: Can we consider suppressing the lint rule for these properties instead? It seems to unnecessarily complicate current code and adding more tests to this file. Given that tests run in a controlled environment and we know the cast always succeeds, I would prefer suppressing lint in cases where following it makes code readability more difficult 🤓

Tests/MessagingInApp/MessagingInAppImplementationTest.swift
```diff
@@ -5,9 +5,9 @@ import SharedTests
 import XCTest
 
 class MessagingInAppImplementationTest: IntegrationTest {
-    private var messagingInApp: MessagingInAppImplementation {
+    private var messagingInApp: MessagingInAppImplementation? {
         // get MessagingInAppImplementation instance so we can call its methods directly
-        (MessagingInApp.shared.implementation as! MessagingInAppImplementation)
+        (MessagingInApp.shared.implementation as? MessagingInAppImplementation)
```
mrehan27, line 10: Same as above

mrehan27 (approved):
Non blocking comments, but would prefer to review it before merging.

ami-aman (approved):

Sources/Common/CustomerIOInstance.swift
```diff
@@ -33,6 +34,7 @@ public protocol CustomerIOInstance: AutoMockable {
      [Learn more](https://customer.io/docs/api/#operation/identify)
      - traits: Request body of identifying profile. Use to define user attributes.
      */
+    // swiftlint:enable orphaned_doc_comment
```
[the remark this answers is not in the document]
Shahroz16: none that i know of 😅

Sources/Common/Util/KeyValueStorage.swift
Shahroz16: Yup name of the class, open to suggestions was struggling alot with it 😅

Tests/DataPipeline/DataPipelineImplementation+ScreenViewsTest.swift
Shahroz16: I am open to suppressing these, i do think values being nullable can cause test being flaky, and hence i add to ensure checking null everywhere. 

Since i have two votes on going forward with force casting, i am going to go ahead and update it.

Ahmed-CIO (approved):
Thanks Shahroz for that!
From quick pass, I am approving this on the assumption that the modified files are either auto-generated, a rename or callsite that need to be updated to match the renaming.
If any of these assumptions aren't accurate or if there is any functional change, please request another review pass

PR #590, opened 2024-03-04
refactor: process push click event before we call 3rd party callback functions

levibostian:
Closes: https://linear.app/customerio/issue/MBL-154/process-push-event-before-we-call-3rd-party-callback-functions

If a 3rd party callback function does not call the async `completionHandler`, there is a risk that we will not be able to process the push click event.

Knowing there is always a risk of this happening, we decided to do as much push processing as possible before allowing 3rd party callback functions to process the push. Making this feature more reliable for customers, even if they forget to call the `completionHandler`.

---

**Stack**:
- #634
- #633
- #601
- #590 ⬅
- #584
- #583

### DIFF
--- a/Sources/MessagingPush/PushHandling/ManualPushHandling+UserNotifications.swift
+++ b/Sources/MessagingPush/PushHandling/ManualPushHandling+UserNotifications.swift
@@ -96,7 +96,9 @@ extension MessagingPushImplementation {
         // A hack to get an instance of pushClickHandler without making it a property of the MessagingPushImplementation class. pushClickHandler is not available to app extensions but MessagingPushImplementation is.
         // We get around this by getting a instance in this function, only.
         if let pushClickHandler = sdkInitializedUtil.postInitializedData?.diGraph.pushClickHandler {
-            pushClickHandler.pushClicked(push)
+            pushClickHandler.trackPushMetrics(for: push)
+            pushClickHandler.cleanupAfterPushInteractedWith(for: push)
+            pushClickHandler.handleDeepLink(for: push)
         }
     }
 }
--- a/Sources/MessagingPush/PushHandling/PushClickHandler.swift
+++ b/Sources/MessagingPush/PushHandling/PushClickHandler.swift
@@ -5,7 +5,11 @@ import UserNotifications
 
 @available(iOSApplicationExtension, unavailable)
 protocol PushClickHandler: AutoMockable {
-    func pushClicked(_ push: PushNotification)
+    // Cleanup files on device that were used when the push was displayed. Files are no longer
+    // needed now that the push is no longer shown.
+    func cleanupAfterPushInteractedWith(for push: PushNotification)
+    func trackPushMetrics(for push: PushNotification)
+    func handleDeepLink(for push: PushNotification)
 }
 
 @available(iOSApplicationExtension, unavailable)
@@ -19,19 +23,15 @@ class PushClickHandlerImpl: PushClickHandler {
         self.customerIO = customerIO
     }
 
-    // Note: This function is called from automatic and manual push click handlers.
-    func pushClicked(_ push: PushNotification) {
+    func trackPushMetrics(for push: PushNotification) {
         guard let cioDelivery = push.cioDelivery else {
             return
         }
 
         customerIO.trackMetric(deliveryID: cioDelivery.id, event: .opened, deviceToken: cioDelivery.token)
+    }
 
-        // Cleanup files on device that were used when the push was displayed. Files are no longer
-        // needed now that the push is no longer shown.
-        cleanupAfterPushInteractedWith(push: push)
-
-        // Handle deep link, if there is one attached to push.
+    func handleDeepLink(for push: PushNotification) {
         if let deepLinkUrl = push.cioDeepLink?.url {
             deepLinkUtil.handleDeepLink(deepLinkUrl)
         }
@@ -40,7 +40,7 @@ class PushClickHandlerImpl: PushClickHandler {
     // There are files that are created just for displaying a rich push. After a push is interacted with, those files
     // are no longer needed.
     // This function's job is to cleanup after a push is no longer being displayed.
-    func cleanupAfterPushInteractedWith(push: PushNotification) {
+    func cleanupAfterPushInteractedWith(for push: PushNotification) {
         push.cioAttachments.forEach { attachment in
             let localFilePath = attachment.localFileUrl
 
--- a/Sources/MessagingPush/PushHandling/iOSPushEventListener.swift
+++ b/Sources/MessagingPush/PushHandling/iOSPushEventListener.swift
@@ -44,14 +44,21 @@ class IOSPushEventListener: PushEventHandler {
 
         logger.debug("Push came from CIO. Handle the didReceive event on behalf of the customer.")
 
-        // Forward event to other push click handlers so they can receive a callback about this push event.
+        pushClickHandler.cleanupAfterPushInteractedWith(for: push)
+
+        if pushAction.didClickOnPush {
+            pushClickHandler.trackPushMetrics(for: push)
+        }
+
+        // Forward event to other push click handlers so they can receive a callback about this push event and optionally process the event.
+        // This funcion is an async operation that calls code we do not own. Therefore, there is risk that the completion handler will not be called and the rest of our code will not be executed. That's why it's important that before we perform this call, we do as much push processing as we can to increase SDK reliability.
         pushEventHandlerProxy.onPushAction(pushAction, completionHandler: {
             // When this block of code executes, the customer is done processing the push event.
 
             // We do not open deep link until after customer is done processing the push event in case the deep link would leave the app.
             // We want to make sure the customer has a chance to process the push event before leaving the app.
             if pushAction.didClickOnPush {
-                self.pushClickHandler.pushClicked(push)
+                self.pushClickHandler.handleDeepLink(for: push)
             }
 
             // call the completion handler, indicating to the OS that we are done processing the push.
--- a/Sources/MessagingPush/autogenerated/AutoMockable.generated.swift
+++ b/Sources/MessagingPush/autogenerated/AutoMockable.generated.swift
@@ -449,38 +449,102 @@ class PushClickHandlerMock: PushClickHandler, Mock {
     }
 
     public func resetMock() {
-        pushClickedCallsCount = 0
-        pushClickedReceivedArguments = nil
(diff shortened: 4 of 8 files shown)


Sources/MessagingPush/PushHandling/PushClickHandler.swift
```diff
@@ -20,18 +35,29 @@ class PushClickHandlerImpl: PushClickHandler {
     }
 
     // Note: This function is called from automatic and manual push click handlers.
-    func pushClicked(_ push: PushNotification) {
-        guard let cioDelivery = push.cioDelivery else {
-            return
+    func pushClicked(_ push: PushNotification, skipPushMetricsTracking: Bool, skipDeepLinkHandling: Bool) {
```
Shahroz16, line 38: can't we refactor this method to not have two responsibilities? I understand its being called from both automatic, where we are already utilizing these functionalities separately. 

For manual, can we not call the two methods separately? would make codebase easier to manage and less chance of messing up stuff.

levibostian: Hopefully these [comments](https://github.com/customerio/customerio-ios/pull/590/files#diff-dfdc330b2ab4df1a2983e50045a6a6c29ce35e274532ba3eec5483b01c15056bR9-R10) could help explain the idea? 

I want to avoid this scenario: 
* In the future, we add more capabilities to when a push is clicked. Let's say for example we now track a custom event "iOS is awesome" when a push is clicked. 
* The developer goes into `iOSPushEventListener` and updates `pushAction()` function to make sure and call this new function `pushClickHandler.trackiOSIsAwesome()`
* The developer forgets to update the manual push click handling code: 
```swift
// when a push is manually clicked: 
pushClickHandling.trackOpenedMetric()
pushClickHandling.openDeepLink()
```

* I designed pushClicked to be re-usable by both automatic and manual, to avoid this scenario of forgetting to update 1 or the other. 

I would like to keep encapsulation of push clicked logic for that reason. If you have ideas for how to make the code "easier to manage and less chance of messing stuff up", could you share some ideas?

levibostian: Is this a blocking or non-blocking suggestion?

Ahmed-CIO: I would say it is a blocker. We need to be careful on the design choices that can lead to mix ups and unclear responsibilities.
While this example isn't that massive, but it adds up.

Ahmed-CIO (changes requested):

Sources/MessagingPush/PushHandling/PushClickHandler.swift
Shahroz16: Levi, fair concern, but let's think of it this way, 

building off from your example, we have
```
pushClickHandling.trackOpenedMetric()
pushClickHandling.openDeepLink()
```

and then we have one method where you want to ensure, everything happens. 

```
func pushClicked(_ push: PushNotification) {
    // Now simply orchestrates calls without making decisions
    pushClickHandling.trackOpenedMetric()
    pushClickHandling.openDeepLink()
}
```

now we cover all the cases, if someone wants the whole `push click handling` experience vs when they want to utilize one of the APIs. 

having this doesn't remove the need to have a method that takes optional values and tries to perform multiple actions based on a condition and this API design would align more with the Single Responsibility Principle (SRP)? what do you think?

levibostian: I do agree that the API design in this PR is not ideal. With this change, the most important thing to me was to prevent breaking features by accident in the future. 

Another idea to solve this problem is leveraging automated tests. Tests that check that *all* steps for push handling is done in manual or automatic push handling. If the test breaks, it is an indicator you forgot to update something. 

If we use tests as the solution, we can then use a very simple API of just: 

```swift
pushClickHandling.trackOpenedMetric()
pushClickHandling.openDeepLink()
```

I think all 3 of us would like this simple of an API. Unless I hear differently, I'll update this PR with this solution today.

levibostian: [Made change](https://github.com/customerio/customerio-ios/compare/8133fd953cca2871897b4c7a8ea0e999248fd901..3f3fa43cca68798ae4baf726c27d654dd1fc43dd)

Ahmed-CIO (approved):

Sources/MessagingPush/PushHandling/ManualPushHandling+UserNotifications.swift
```diff
@@ -96,7 +96,9 @@ extension MessagingPushImplementation {
         // A hack to get an instance of pushClickHandler without making it a property of the MessagingPushImplementation class. pushClickHandler is not available to app extensions but MessagingPushImplementation is.
         // We get around this by getting a instance in this function, only.
         if let pushClickHandler = sdkInitializedUtil.postInitializedData?.diGraph.pushClickHandler {
-            pushClickHandler.pushClicked(push)
+            pushClickHandler.trackPushMetrics(for: push)
+            pushClickHandler.cleanupAfterPushInteractedWith(for: push)
+            pushClickHandler.handleDeepLink(for: push)
```
Shahroz16, line 101: does the order matter? i would think so? as `handleDeepLink` might redirect to another app. 

what are we doing to avoid a case where someone might call one method before another? specially when the method is named `cleanup` which would confuse anyone to call at the end?

Shahroz16: if the above supposition is true, then it would incline towards the previous [suggestion](https://github.com/customerio/customerio-ios/pull/590#discussion_r1516694859) 

We should have another method that dictates the ordering of methods, which automated test might not be able to do

Shahroz16 (changes requested):
blocking so we can cater [this](https://github.com/customerio/customerio-ios/pull/590#discussion_r1519441688) concern

Sources/MessagingPush/PushHandling/ManualPushHandling+UserNotifications.swift
levibostian: Great call-out. The order is important. 

From reading code in the link you sent, I may not fully understand the suggestion, but I do not think that the solution will solve the problem we have with multiple parts of the code performing click tracking and the functions being called are unique. 

I think that automated tests are the best place to assert the ordering. I will add this functionality to the tests.

levibostian: [I updated the test suite to assert order](https://github.com/customerio/customerio-ios/compare/3f3fa43cca68798ae4baf726c27d654dd1fc43dd..660a9906d8a8e1a2bc88d1867a4922a43d31c642). 

Notice that I added a `XXX:` suggestion to the test function. To keep the scope small, I did not implement this. If this sounds like a good idea to do, I can create a future PR to do this.

Shahroz16: Great test, I think this test would work for now but my concern regarding the usage of this class remains. This test is great for the methods we already know are utilizing push click handling. 

My concern is, that someone adds another method in `PushEventHandler`, in that method if they are supposed to do push handling, and don't follow the order, we can't stop it. But if we had a method that they can call, that respects the order for them, it would have made it simpler.

Shahroz16 (commented):
additionally, should `shouldDisplayPushAppInForeground` not alse track open metric?

levibostian:
@Shahroz16 

> additionally, shouldn't shouldDisplayPushAppInForeground also track open metric?

No. `shouldDisplayPushAppInForeground` is called when the push is received on the device but the push has not yet been clicked or swiped away. We only track `opened` when the push is clicked on.

Sources/MessagingPush/PushHandling/ManualPushHandling+UserNotifications.swift
levibostian: > But if we had a method that they can call, that respects the order for them, it would have made it simpler.

Previously, we did have 1 function that could be called and all push click handling could be done inside of it: `PushClickHandler.onPushClicked`. 

This PR made this more complex because now the push click handling when we do automatic and manual push click handling is different. The order of operations is identical, but the implementations are still unique. Therefore, I think we would need to implement 2 versions of `onPushClicked`, right? 

Assuming this is what you're trying to say, I think that leveraging automated tests to verify push click handling is what we expect across the whole SDK, is a better solution then using multiple `PushClickHandler` functions that perform all of the push click handling. 

Maybe I don't understand your comment?

Shahroz16: I am okay with moving forward with automated test for now, since its an internal API. I'll create proposal ticket as that might be out of scope for this.

levibostian: That sounds like a plan. 

I believe this is the last unresolved suggestion. Mind approving the PR to unblock merging it?

Shahroz16:
```
    func assertWillHandleDeepLinkLast(for push: PushNotification) {
        // When deep link is handled, we expect all other click handling to have already been done.
        handleDeepLinkClosure = { _ in
            // We expect all other push click handling to have already happened.
            self.assertHandledPushClick(for: push)

            // XXX: This would be a good place to call: mock.verifyNoMoreInteractions(), like in Mockito, to be more confident that we are calling deep link last.
        }
    }
```

are you planing on adding the `XXX` part later? because as i understand this test doesnt currently verify that its called at last?

Shahroz16 (approved):

Sources/MessagingPush/PushHandling/ManualPushHandling+UserNotifications.swift
levibostian: > are you planing on adding the XXX part later?

To keep the scope small, I did not implement it in this PR. I was first going to look for feedback from the team in the PR to see if anyone had objections to it before I spend the effort on it. If no objections, I do plan on implementing this, yes. 

> as i understand this test doesnt currently verify that its called at last?

The test could be improved, but I do think that it verifies that it's called last. The function `assertWillHandleDeepLinkLast`, when deep link is handled, will check if all other push click handling was done. Which, is a form of checking that handling the deep link was done last. 

The test does not verify that no more push click handling operations are done after the deep link is handled. The `XXX:` comment calls this out and offers a potential solution.

levibostian:

PR #591, opened 2024-03-05
refactor: swiftUi sample app with CDP integration

ami-aman:
part of: https://linear.app/customerio/issue/MBL-93/sample-app-fcm-swiftui

### Context
**Integrating CDP iOS SDK into SwiftUI Sample App**
We have two iOS sample apps:

- Based on UIKit: Uses SPM (Swift Package Manager) and APNs (Apple Push Notification service).
- Based on SwiftUI: Uses CocoaPods and FCM (Firebase Cloud Messaging).

Both apps previously used the Journeys SDK. As we've already successfully integrated the CDP iOS SDK into the UIKit sample app, this ticket specifically focuses on integrating CDP iOS SDK SwiftUI sample app.

### PR Summary
PR makes the following fixes/updates:
- Replaces use of Journey SDK with iOS CDP SDK
- Uses CDP methods for tracking, grouping, aliasing, logging etc
- Adds `CioTrackingMigration` podspec and updates relevant files for smooth CI functioning for migration module
- Utilizes CDP configuration
- NSE uses initialization method for extension
- Plugin for screen tracking 
- CDP Api key instead of Journey's apiKey 
- Configurable properties build() for initialization
- Removed usage of CioTracking in app

### Tests
Refer Linear comment for complete all tests.


Apps/CocoaPods-FCM/src/AppDelegate.swift
```diff
@@ -18,25 +18,40 @@ class AppDelegate: NSObject, UIApplicationDelegate {
...
+            config.cdnHost(cdnHost)
+        }
+        CustomerIO.initialize(withConfig: config.build())
 
-            // This line of code is internal to Customer.io for testing purposes. Do not add this code to your app.
-            appSetSettings?.configureCioSdk(config: &config)
+        // Set auto tracking of screen views
+        let autoScreenTrack = appSetSettings?.trackScreens ?? true
+        if autoScreenTrack {
+            CustomerIO.shared.add(plugin: AutoTrackingScreenViews(filterAutoScreenViewEvents: nil, autoScreenViewBody: nil))
```
Shahroz16, line 43: We shouldn't add this plugin in swift UI! and shouldn't recommend users to do so either.

Apps/CocoaPods-FCM/src/View/SettingsView.swift
```diff
@@ -50,57 +52,52 @@ struct SettingsView: View {
...
-
                     let didChangeSiteId = siteIdBeforeEditingSettings != viewModel.settings.siteId
                     if didChangeSiteId { // if siteid changed, we need to re-identify for the Customer.io SDK to get into a working state.
                         userManager.logout()
                     }
-
-                    done()
+                    timer.start(interval: TimeInterval(3)) {
+                        done()
+                    }
```
Shahroz16, line 78: what is it for?

Apps/CocoaPods-FCM/src/View/SettingsView.swift
```diff
@@ -50,57 +52,52 @@ struct SettingsView: View {
...
 
                 Button("Restore default settings") {
                     viewModel.restoreDefaultSettings()
                 }.setAppiumId("Restore Default Settings Button")
             }
+
             .padding([.leading, .trailing], 10)
             .onAppear {
-                siteIdBeforeEditingSettings = CustomerIO.shared.siteId!
+                siteIdBeforeEditingSettings = BuildEnvironment.CustomerIO.siteId
```
Shahroz16, line 88: i don't think it would work for the second, or third time you change siteId ? because it would always check for the first one?

Shahroz16 (changes requested):

Apps/CocoaPods-FCM/src/View/SettingsView.swift
ami-aman: Added a [toast](https://github.com/customerio/customerio-ios/blob/9a6ea4430086c3344dab261dcfcf896c2a3bf347/Apps/CocoaPods-FCM/src/View/SettingsView.swift#L69) on settings screen to maintain consistency across all apps. The toast will disappear after 3 seconds and the user will be taken back to parent controller.

Apps/CocoaPods-FCM/src/View/SettingsView.swift
ami-aman: I am not very clear about the question, but let  me try answering to what I understood. Feel free to correct me. 

This is to restore default value of `siteId` and not to fetch the current value of `siteId`. The default value is the one that we capture in BuildEnvironment.

Shahroz16 (approved):

PR #592, opened 2024-03-06
chore: fcm sample app using autoFetchDeviceToken

ami-aman:
### About the PR

Based on our recent discussion, concerns were raised about the functionality of the `autoFetchDeviceToken` property within SwiftUI applications. While there haven't been any reported user issues, but it was assumed that the feature does not work as expected in SwiftUI.

While being on-call and working on SwiftUI Sample app, I also investigated `autoFetchDeviceToken` propert. The feature functions as intended when implemented correctly. Successful tests were done with both APN and FCM on SwiftUI app. However, this pull request focuses solely on FCM integration, as the SwiftUI iOS Sample app utilizes FCM and CocoaPods.

### About the fix
The fix to make the feature work in SwiftUI apps required some tweaks in AppDelegate file. There have been no major changes but few implementation changes. After the tweaks made, the feature worked smoothly with SwiftUI sample app. 

### Tests
[This user's device ](https://fly.customer.io/workspaces/111525/journeys/people/a5e70600fa01fb01) on fly has been registered after making the updates.

### DIFF
--- a/Apps/CocoaPods-FCM/src/AppDelegate.swift
+++ b/Apps/CocoaPods-FCM/src/AppDelegate.swift
@@ -34,10 +34,6 @@ class AppDelegate: NSObject, UIApplicationDelegate {
         MessagingPushFCM.initialize { config in
             config.autoFetchDeviceToken = true
         }
-
-        // Manually get FCM device token. Swizzling hasn't been working for me.
-        Messaging.messaging().delegate = self
-
         /**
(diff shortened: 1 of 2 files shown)


Ahmed-CIO (approved):

Ahmed-Ali (changes requested):
Moving back to request changes till the merge conflicts are resolved

PR #594, opened 2024-03-06
chore: remove Tracking module from APN UIKit

mrehan27:
helps: [MBL-140](https://linear.app/customerio/issue/MBL-140/remove-tracking-module)

### Changes

- Updated APN UIKit dependencies
- Fixed APN UIKit to use `DIGraphShared` from updated SDK template
- Updated relevant code files to use new dependencies and graph


Apps/APN-UIKit/APN UIKit/DIGraphShared.swift
```diff
@@ -1,7 +1,7 @@
 import Foundation
 
-public class DIGraph {
-    public static let shared: DIGraph = .init()
+public class DIGraphShared {
+    public static let shared: DIGraphShared = .init()
```
mrehan27, line 4: Since our SDK template was updated to support only `DIGraphShared` earlier, and our sample app is reusing the same template, had to rename `DIGraph` to `DIGraphShared` for accurate auto-generated code. `make generate` wasn’t previously executed on sample apps, so I ran it on APN UIKit in this PR to ensure the changes are compilable.

levibostian: I am glad that the sample apps can re-use the template files so I am OK with this change.

levibostian (approved):

Shahroz16 (approved):

PR #600, opened 2024-03-07
refactor: change the OS directory eventbus saves to

levibostian:
Based on what types of files we are creating, I suggest that we change the directory of where files are saved.

To help with this decision, here is a summary of the OS directories of their suggested use cases:

.documentDirectory:
* Data stored here is backed up by iTunes and iCloud.
* Suitable for data that the user would not typically manipulate or interact with directly, such as configuration files or cached data.

.applicationSupportDirectory:
* Use this directory for data that is generated by your application and can be re-created or downloaded again.
* Data stored here helps support the functionality of your app but can be replaced without loss of significant data.

.cachesDirectory:
* Data stored in this directory can be purged by the system at any time, so it should not be relied upon for long-term storage.

.libraryDirectory:
* Use this directory for files that are critical to the functioning of your app but are not user-generated. This could include databases, preferences, or other app-related data.
* Data stored here is not visible to the user through the Files app.

commit-id:90e4b304


Shahroz16 (approved):
Thank you for this, can you verify the behaviour in test too? because we create test directory for event storage there.

PR #601, opened 2024-03-07
refactor: help customers if a push event handler forgot to call completion handler

levibostian:
Part of: https://linear.app/customerio/issue/MBL-154/process-push-event-before-we-call-3rd-party-callback-functions

When we send push events to the host app's other push event handlers to process, we trust that all of the push event handlers will remember to call the async completion handler when they are done processing.

However, there will always be a chance a customer's host app or 3rd party SDK forgets to call it. When this happens, customers may experience frustration because they may have unexpected behavior while using our SDK and not be sure why.

We decided to solve this problem by giving logs to the customer on what push handlers are called and what push handlers remember to call the completion handler. These logs can then be used to identify 1 class that forgot to call the async completion handler.

---

**Stack**:
- #634
- #633
- #601 ⬅
- #590
- #584
- #583

### DIFF
--- a/Sources/MessagingPush/PushHandling/PushEventHandlerProxy.swift
+++ b/Sources/MessagingPush/PushHandling/PushEventHandlerProxy.swift
@@ -1,4 +1,5 @@
 import CioInternalCommon
+import CioTracking
 import Foundation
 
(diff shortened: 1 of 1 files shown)


Ahmed-CIO (approved):

Shahroz16 (approved):

levibostian:

PR #603, opened 2024-03-07
chore: remove Tracking module from FCM SwiftUI

mrehan27:
helps: [MBL-140](https://linear.app/customerio/issue/MBL-140/remove-tracking-module)

### Changes

- Updated pod files for FCM SwiftUI
- Fixed imports

### Notes

The app still does not compile because of `CioTrackingMigration` error. We'll cover this later in ticket for updating FCM SwiftUI app to make it compatible with new changes.


Shahroz16 (approved):
Looks good, continuation of it would be part of https://github.com/customerio/customerio-ios/pull/591

PR #604, opened 2024-03-08
docs: [vision-os-sample-app][2/n] App State Storage

Ahmed-CIO:
* #617
* #613
* #612
* #611
* #610
* #609
* #608
* #606
* #605
* __->__ #604

## Context

For more context:

## PR Summary
In this PR I added storage utilities to persiste/restore the app satate as well as few payloads that will be used for SDK initialization, and identify

## Test plan
- Open `VisionOs.xcproject`

### DIFF
--- a/Apps/VisionOS/VisionOS/Storage/AppState.swift
+++ b/Apps/VisionOS/VisionOS/Storage/AppState.swift
@@ -1,0 +1,41 @@
+import Foundation
+
+struct ScreenTitleConfig {
+    let screenTitle: String
+    let menuTitle: String
+    let showVisionProLogo: Bool
+
+    init(_ screenTitle: String, menuTitle: String? = nil, showVisionProLogo: Bool = false) {
+        self.screenTitle = screenTitle
+        self.menuTitle = menuTitle ?? screenTitle
+        self.showVisionProLogo = showVisionProLogo
+    }
+}
+
+class AppState: ObservableObject {
+    static let shared = AppState()
+
+    @Published var profile: Profile = .loadFromStorage() {
+        didSet {
+            UserDefaults.standard.setValue(
+                profile.toJson(),
+                forKey: Profile.storageKey()
+            )
+        }
+    }
+
+    @Published var workspaceSettings: WorkspaceSettings = .loadFromStorage() {
+        didSet {
+            UserDefaults.standard.setValue(
+                workspaceSettings.toJson(),
+                forKey: WorkspaceSettings.storageKey()
+            )
+        }
+    }
+
+    // MARK: non persistent state
+
+    @Published var titleConfig: ScreenTitleConfig = .init("")
+    @Published var errorMessage: String = ""
+    @Published var successMessage: String = ""
+}
--- a/Apps/VisionOS/VisionOS/Storage/AppState.swift
+++ b/Apps/VisionOS/VisionOS/Storage/AppState.swift
@@ -1,17 +1,5 @@
 import Foundation
 
-struct ScreenTitleConfig {
-    let screenTitle: String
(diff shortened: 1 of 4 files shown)


Apps/VisionOS/VisionOS/Storage/Payloads.swift
```diff
@@ -0,0 +1,71 @@
...
+
+    static func storageKey() -> String {
+        "Profile"
+    }
+}
+
+struct Event {
+    var name: String = ""
+    var properties: [Property]
+}
```
levibostian, line 43: I can understand why Profile, Workspace settings are needed, but not sure what this is used for? 

I would be OK if the data type was removed from this PR and introduced in a PR it's being used in. Or, explain what it's used for if you believe it will be required in the future.

Apps/VisionOS/VisionOS/Storage/Payloads.swift
```diff
@@ -0,0 +1,71 @@
...
+
+struct Property: Codable, Identifiable, Comparable {
```
levibostian, line 45: Same comment as `Event`. I don't understand what this data type is used for? 

I would be OK if the data type was removed from this PR and introduced in a PR it's being used in. Or, explain what it's used for if you believe it will be required in the future.

Apps/VisionOS/VisionOS/Storage/AppState.swift
```diff
@@ -0,0 +1,41 @@
+import Foundation
+
+struct ScreenTitleConfig {
```
levibostian, line 3: "ScreenTitleConfig" seems like it's a data type that is scoped to 1 specific screen in the app. Assuming that's correct, I think we should move the data type out of this `AppState` file.

Apps/VisionOS/VisionOS/Storage/AppState.swift
```diff
@@ -0,0 +1,41 @@
...
+                forKey: WorkspaceSettings.storageKey()
+            )
+        }
+    }
+
+    // MARK: non persistent state
+
+    @Published var titleConfig: ScreenTitleConfig = .init("")
+    @Published var errorMessage: String = ""
+    @Published var successMessage: String = ""
```
levibostian, line 40: These properties look like properties useful to the UI of the app. Assuming that's true, I prefer not coupling business logic data structures with UI. 

I would be OK if UI properties are in a ViewModel, but this data type does not appear to be a ViewModel.

levibostian (commented):
Great changes. 

Any interest in storing these inside of the [Common Sample apps SDK](https://github.com/customerio/customerio-ios/tree/main/Apps/Common)? So our other sample apps can also use these util classes?

Apps/VisionOS/VisionOS/Storage/AppState.swift
Ahmed-CIO: Agreed. Will remove these non-persistent UI states from this PR and include them in separate PR as a view model

Apps/VisionOS/VisionOS/Storage/Payloads.swift
Ahmed-CIO: Unlike the Event, the Property is used in Profile struct too. Hence it is appropriate to remain in this PR

Ahmed-CIO:
I am on board for including these in the shared sample apps code base. But I would hold on this for now given that this is a brand new one and we have intention of revising the existing sample apps; so once this one is landed and we plan how we want to revise the rest of the sample apps we will have better idea of what to reuse from this one, existing ones and what to add as a shared code base. Then we can do that as one piece with all the information needed for it already in place

PR #605, opened 2024-03-08
docs: [vision-os-sample-app][3/n] Markdown utils

Ahmed-CIO:
* #617
* #613
* #612
* #611
* #610
* #609
* #608
* #606
* __->__ #605
* #604

## Context

For more context:

## PR Summary
In this PR I added few utils for code syntax highlighting and the Markdown theme that will use it

## Test plan
- Open `VisionOs.xcproject`

### DIFF
--- a/Apps/VisionOS/VisionOS/ViewUtils/MarkdownTheme.swift
+++ b/Apps/VisionOS/VisionOS/ViewUtils/MarkdownTheme.swift
@@ -1,0 +1,63 @@
+import MarkdownUI
+import SwiftUI
+
+private extension Color {
+    static let codeBackground = Color(rgba: 0x333336FF)
+    static let blockquoteIcon = Color(rgba: 0xFECD00FF)
+    static let blockquoteBackground = Color(rgba: 0xAF65FFFF)
+    static let linkColor = Color(rgba: 0x04ECBBFF)
+}
+
+extension Theme {
+    static let tutorial = Theme.docC
+        .code {
+            FontFamilyVariant(.monospaced)
+            FontSize(16)
+            FontStyle(.italic)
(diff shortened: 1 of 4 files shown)


Apps/VisionOS/VisionOS/ViewUtils/MarkdownTheme.swift
```diff
@@ -0,0 +1,63 @@
+import MarkdownUI
```
levibostian, line 1: Is this a 3rd party dependency? If so, how is it installed? The xcode project diff doesn't look like any dependency was installed? 

What doe the design of this app supposed to look like? Do we have a design created that needs to be replicated?

Ahmed-CIO: > Is this a 3rd party dependency? If so, how is it installed? The xcode project diff doesn't look like any dependency was installed?

In the very first PR I added it alongside CIO SDK and another dependency called Splash.

> What doe the design of this app supposed to look like? Do we have a design created that needs to be replicated?

We don't have any design for the app, I am pretty much putting it together using out of the box components with minimum styling.

The only non-standard piece is the markdown and code syntax highlighting. The markdown used with the font type/size is likely to need a better designer eyes if we are keen on perfecting the UI.
The call for using markdown is due to the fact that the app intends to contain textual content/docs like that mandates it to be easy to read, update, mix with code blocks, call outs, etc and not having to deal with laying out punch of different types of views. This is on purpose so we can use help from @glosier later to update that content

Ahmed-CIO:
Apologies for that, gh-stack pushes seems to cause too much noise due to the fact that it pushes the entire stack for every push. These changes will carry on #619

PR #606, opened 2024-03-08
docs: [vision-on-sample-app][4/n] InlineNavigationLinks

Ahmed-CIO:
* #617
* #613
* #612
* #611
* #610
* #609
* #608
* __->__ #606
* #605
* #604

## Context

For more context:

## PR Summary
In this PR I added the InlineNavigationLinks enum and its handling. These enum values represent the screens that will be in the app. It will be used within the markdown content to act as links to navigate to different screens and include some associated values for these screens.

For example:
```
To try this out, see the [identify](InlineNavigationLink.identify) tutorial
```

In this content, the "identify" link when clicked:
- Will go through a specialised URL handler (will be introduced later) that will update the navigation path to navigate to the Identify screen.
- Will be stored in the persistent app state so the app starts at that screen next time
- Will be marked as visited, so the user can tell which screens they saw before. This might be handy if we end up with many content. Which would happen when we add push and in-app

Moreover, the enum cases will be used to build the kind of "table of contents" alike menu for easier navigation throughout the app.

## Test plan
- Open `VisionOs.xcproject`

ghstack-source-id: 12dd5f7f2d3b47f5b92501150c956fbcf886a8cd
Pull Request resolved: https://github.com/customerio/customerio-ios/pull/606

### DIFF
--- a/Apps/VisionOS/VisionOS/MainScreen.swift
+++ b/Apps/VisionOS/VisionOS/MainScreen.swift
@@ -1,11 +1,22 @@
 import SwiftUI
 
 struct MainScreen: View {
+    @ObservedObject var state: AppState = .shared
     var body: some View {
         VStack {
             Text("Hello, world!")
         }
         .padding()
+        .environment(
+            \.openURL,
+            OpenURLAction { url in
+                guard let link = InlineNavigationLink(fromUrl: url) else {
+                    return .systemAction
+                }
+                state.navigationPath.append(link)
+                return .handled
+            }
+        )
     }
 }
 
--- a/Apps/VisionOS/VisionOS/Storage/AppState.swift
+++ b/Apps/VisionOS/VisionOS/Storage/AppState.swift
@@ -1,5 +1,25 @@
 import Foundation
 
+extension [InlineNavigationLink]: UserDefaultsCodable {
+    static func storageKey() -> String {
+        "navigationPath"
+    }
+
+    static func empty() -> [Element] {
+        []
+    }
+}
+
+extension Set<InlineNavigationLink>: UserDefaultsCodable {
+    static func storageKey() -> String {
+        "visitedLinks"
+    }
+
+    static func empty() -> Set<Element> {
+        .init()
+    }
+}
+
 struct ScreenTitleConfig {
     let screenTitle: String
     let menuTitle: String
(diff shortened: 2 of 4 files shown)


Apps/VisionOS/VisionOS/MainScreen.swift
```diff
@@ -1,11 +1,23 @@
...
             Text("Hello, world!")
         }
         .padding()
+        .environment(
+            \.openURL,
+            OpenURLAction { url in
+                guard let link = InlineNavigationLink(fromUrl: url) else {
+                    return .systemAction
+                }
+                state.navigationPath = [link]
```
levibostian, line 16: Why do we need to save this navigation link into the AppState? 

I was expecting the app to navigate to a screen and that's it.

Apps/VisionOS/VisionOS/Storage/AppState.swift
```diff
@@ -1,5 +1,25 @@
...
+
+extension Set<InlineNavigationLink>: UserDefaultsCodable {
+    static func storageKey() -> String {
+        "visitedLinks"
+    }
+
+    static func empty() -> Set<Element> {
+        .init()
+    }
+}
```
levibostian, line 21: Why are we saving this information to persistent storage?

Apps/VisionOS/VisionOS/ViewUtils/InlineNavigationLink.swift
```diff
@@ -0,0 +1,25 @@
+import Foundation
+
+enum InlineNavigationLink: String, Codable, CaseIterable {
```
levibostian, line 3: What does "Inline" in the data type name mean?

Ahmed-CIO: I updated the PR description. The name is coming from the fact that these links will appear as a markdown inline links. Something like [this](https://github.com/customerio/customerio-ios/pull/606) link which is just a clickable word(s) within a continuous line of text

Apps/VisionOS/VisionOS/Storage/AppState.swift
Ahmed-CIO: This is a persistent UI state. Helps the user identify screens they already visited before.
We have few things to store
Persistent
- Data used for the SDK like the workspace settings and initialize (profile) details
- Navigation state so the user don't always start on the landing screen (a bit of a UI convenience)

Non-Persistent
These are moved to a ViewModel that won't be persistent
- Current screen title - updated on each screen onAppear
- error and success message

Now we can argue that navigation is part of the view model. Which isn't wrong. It would just mean:
- Persistent storage used for SDK related payloads
- Persistent storage for parts of the view model
- Non-persistent storage for other parts of the view model

Again, nothing wrong with that, just didn't encounter a reason to see the full value of adding this strict separation yet in the app

Apps/VisionOS/VisionOS/MainScreen.swift
Ahmed-CIO: this will cause the next app run to start on the screen it was terminated at instead of the very initial screen as the first app run

PR #608, opened 2024-03-08
docs: [vision-on-sample-app][6/n] AppDelegate

Ahmed-CIO:
* (to be filled)

## Context

For more context:

## PR Summary
This PR introduces fthe app delegate which will be used to initialize `CustomerIO`

## Test plan
- Open `VisionOs.xcproject`

### DIFF
--- a/Apps/VisionOS/VisionOS/AppDelegate.swift
+++ b/Apps/VisionOS/VisionOS/AppDelegate.swift
@@ -1,0 +1,39 @@
+import CioTracking
+import UIKit
+
+class AppDelegate: NSObject, UIApplicationDelegate {
+    func application(
+        _ application: UIApplication,
+        didFinishLaunchingWithOptions launchOptions: [UIApplication
+            .LaunchOptionsKey: Any]? = nil
+    ) -> Bool {
+        /*
+
+         How do I get the values for the siteId and apiKy?
+         1- Login or Signup in CustomerIO platform: https://fly.customer.io/
+         2- On the top right section click the
+         Settings Icon >> Workspace Settings >> API and webhook credentials
+         You will see list of one or more workspaces and each has
+         its own pair of siteId and apiKey, these are the ones
+         to use here.
+         */
+
+        let workspaceSettings = AppState.shared.workspaceSettings
+        if workspaceSettings.isSet() {
+            CustomerIO.initialize(
+                siteId: workspaceSettings.siteId,
+                apiKey: workspaceSettings.apiKey,
+                region: workspaceSettings.region
+            ) { config in
+                // Debug config just to make the demo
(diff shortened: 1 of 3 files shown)


Apps/VisionOS/VisionOS/AppDelegate.swift
```diff
@@ -0,0 +1,39 @@
...
+    ) -> Bool {
+        /*
+
+         How do I get the values for the siteId and apiKy?
+         1- Login or Signup in CustomerIO platform: https://fly.customer.io/
+         2- On the top right section click the
+         Settings Icon >> Workspace Settings >> API and webhook credentials
+         You will see list of one or more workspaces and each has
+         its own pair of siteId and apiKey, these are the ones
+         to use here.
```
levibostian, line 18: I do greatly appreciate your efforts to help customers find the information that they need in this code! 

I am concerned of this documentation getting out-of-date. Instead, do we have a webpage in our CIO docs site that we could point to here?

Apps/VisionOS/VisionOS/AppDelegate.swift
```diff
@@ -0,0 +1,39 @@
...
+         */
+
+        let workspaceSettings = AppState.shared.workspaceSettings
+        if workspaceSettings.isSet() {
```
levibostian, line 22: What happens if workspace settings are *not* set? 

At a minimum, should we create a log warning or something?

Ahmed-CIO: Not really. The goal here is not to ensure full setup ready for every app run; the gaol is to ensure the user don't have to go to the initialize screen and call initialize manually. More of a UX piece instead of production-ready setup. At the end of the day, it is a sample app meant to make it easy to see the SDK usage in action instead of being in-itself a production ready app

Apps/VisionOS/VisionOS/AppDelegate.swift
Ahmed-CIO: I had similar comment earlier about the fact that the proximity to the code should give it better chance of not being outdated; with that said, a link here to a full document webpage won't hurt. On it

Apps/VisionOS/VisionOS/AppDelegate.swift
Ahmed-CIO: But one thing I found missing while working on this part, is that the callsite have noway to tell if they enter wrong credentials. The SDK won't report anything, as far as I know at least. Something we might want to revise at some point later.

Ahmed-CIO:
Apologies for that, gh-stack pushes seems to cause too much noise due to the fact that it pushes the entire stack for every push. These changes will carry on #621 

Meanwhile, I addressed the comments in this PR

PR #609, opened 2024-03-08
docs: [vision-on-sample-app][6/n] Common Views

Ahmed-CIO:
* #617
* #613
* #612
* #611
* #610
* __->__ #609
* #608
* #606
* #605
* #604

## Context

For more context:

## PR Summary
This PR adds punch of generic SwiftUI view
- `CommonLayoutView` this view will be used to wrap all content and handle bunch of the navigation logic. Including displaying the `Logo`, the `TutroialsMenuView`, and error messages
- `LineView` custom view for flexible divider
- `FloatingTitleTextView` custom text field to ensure the field placeholder becomes labels when the field is filled.

## Test plan
- Open `VisionOs.xcproject`


Apps/VisionOS/VisionOS/CommonViews/CommonLayoutView.swift
```diff
@@ -0,0 +1,65 @@
...
+        .buttonStyle(CustomButtonStyle())
+    }
+}
+
+struct CustomButtonStyle: PrimitiveButtonStyle {
+    typealias Body = Button
+    func makeBody(configuration: Configuration) -> some View {
+        Button(configuration)
+            .buttonBorderShape(.roundedRectangle(radius: 10))
+            .background(Color(rgba: 0x333333FF))
```
levibostian, line 53: Will this color be used in multiple parts of the app? If so, I think it's best to not have it hard-coded in a View. 

Perhaps use extensions so we could do: `Color.foo` to reference it?

Apps/VisionOS/VisionOS/CommonViews/TutorialsMenuView.swift
```diff
@@ -0,0 +1,45 @@
...
+    if state.workspaceSettings.isSet(), state.profile.loggedIn {
+        return InlineNavigationLink.allCases
+    }
+
+    if state.workspaceSettings.isSet() {
+        return InlineNavigationLink.preIdentifyLinks
+    }
+
+    return InlineNavigationLink.preInitializationLinks
+}
```
levibostian, line 27: This code seems un-related to `TutorialsMenuView`. If that's true, could we move this code into another file?

Apps/VisionOS/VisionOS/ViewUtils/InlineNavigationLink.swift
```diff
@@ -22,4 +22,49 @@ enum InlineNavigationLink: String, Codable, CaseIterable {
...
+        case .deviceAttributes:
+            ""
+        case .howToTestDeviceAttributes:
+            ""
+        case .track:
+            ""
+        case .howToTestTrack:
+            ""
+        }
+    }
```
levibostian, line 54: Non-blocking question: Is it possible for each screen's View to decide what the menu title is? I am OK with this if that's preferred.

levibostian (approved):
This may not be in scope for this app, but I still want to ask. 

Could some of these generic SwiftUI views be re-used with the SwiftUI iOS sample app?

Apps/VisionOS/VisionOS/ViewUtils/InlineNavigationLink.swift
Ahmed-CIO: You are spot on. Each of these cases will get the title from the relevant screen. Hence it is currently blank "to indicate the missing title". The only reason this is added here as an associated property, is to avoid the need for each callsite to do the logic of this switch case.

Apps/VisionOS/VisionOS/CommonViews/TutorialsMenuView.swift
Ahmed-CIO: It belongs here. The button view builder decides whether to show a checkmark in the menu entry based on whether the said link is visited. The `getEnabledNavigationLinks` we can argue whether it belongs here or in the shared view model. Since it is only used here, it seems appropriate to keep it here.

Apps/VisionOS/VisionOS/CommonViews/CommonLayoutView.swift
Ahmed-CIO: So far this is only used here. Agreed if at any point I needed to repeat it, it will be extracted out

Ahmed-CIO:
Apologies for that, gh-stack pushes seems to cause too much noise due to the fact that it pushes the entire stack for every push. These changes will carry on #622 

Meanwhile, I addressed the comments in this PR

PR #610, opened 2024-03-08
docs: [vision-on-sample-app][7/n] App Intro View

Ahmed-CIO:
* #617
* #613
* #612
* #611
* __->__ #610
* #609
* #608
* #606
* #605
* #604

## Context

For more context:

## PR Summary
This PR introduces the first screen in the app when the app runs for the first time.
Intended to give the user an idea what to expect to see while using the app and quick call out on where they will find the actual `CustomerIO.[method]` calls if they all they care about is the code samples.

Beside the regular code review, I would really appreciate any suggestion to improve the visual/texutal content.

## Test plan
- Open `VisionOs.xcproject`
- See the intro page similar to the one in the screenshot

AppIntro


Apps/VisionOS/VisionOS/AppIntroView.swift
```diff
@@ -0,0 +1,41 @@
...
+            This sample app is more than just a demonstration—it's a hands-on guide
+            designed to familiarize you with **Customer.io Swift SDK**. You'll learn not only how to integrate the SDK into your VisionPro app but also how
+            it contributes to leveraging **Customer.io**'s marketing prowess to enhance the user engagement.
+
+            **P.S.:** For those eager to jump straight into the code,
+            the `MainScreen.swift` file in the sample app provides a quick overview of the `CustomerIO` class in action.
+
+            Ready to embark on this immersive journey? Let's start by
+            [installing CustomerIO SDK](\(InlineNavigationLink.install))
+            """
```
levibostian, line 28: Non-blocking suggestion: Instead of hard-coding English into the code, could we localize this string easily in the future?

This isn't a requirement for any of our sample apps today so I suppose it's OK to do this for now. By the use of markdown here, I was not sure how easy this would be to localize later?

levibostian (commented):
The screenshot in the description is very helpful, thanks! 

I would prefer for others to make the design decisions. This is only my opinion and feedback. 

The screenshot of the View has a lot of copy and the width of the copy is quite wide. Making me think that most people will not read it all and decide to skip around.  

Besides that, I will refrain from sharing my other opinions about the app design since there might be some decisions already made about the design somewhere. If you're looking for feedback on it, we can discuss it somewhere else.

Ahmed-CIO:
> I would prefer for others to make the design decisions. This is only my opinion and feedback.

I would love that too, hence the screenshot. We had no design or designer and while I believe I put some effort in ensuring good UX, I am not the best when it comes to UI design. The only note is that it is also essential to keep it as simple as possible. At the end of the day, it is a sample app and not an app that will go to the app store. So the investment is justified as long as it is core to deliver the value of a sample app.

> The screenshot of the View has a lot of copy and the width of the copy is quite wide. Making me think that most people will not read it all and decide to skip around.

I had the same concern, but luckily it exists plainly in the changed files where reviewers will take a look and hopefully read.

> Besides that, I will refrain from sharing my other opinions about the app design since there might be some decisions already made about the design somewhere. If you're looking for feedback on it, we can discuss it somewhere else.

No specific decisions set in stones. We don't have a design and I invested a bit to deliver a minimum accepted UX for a sample app based on my own expectations only. Any feedback is more than welcome

Ahmed-CIO:
Apologies for that, gh-stack pushes seems to cause too much noise due to the fact that it pushes the entire stack for every push. These changes will carry on #623

PR #612, opened 2024-03-08
docs: [vision-on-sample-app][9/n] SDK Setup view

Ahmed-CIO:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #617
* #613
* __->__ #612
* #611
* #610
* #609
* #608
* #606
* #605
* #604

## Context

This is a PR in a series of PRs to build sample app that runs Identify and Track in VisionPro
The app is built to be an interactive walk through guidance for developers who are integrating Customer.io Swift SDK in their VisioPro apps.
For more context:

- See the project [one pager](https://www.notion.so/custio/Spacial-Identify-and-Tracking-6587d848bbde494095c0585237326ed3?pvs=4)
- See [Linear Issue](https://linear.app/customerio/issue/MBL-115/sample-app-for-visionpro)

## PR Summary
This PR introduces a the Setup view where we instruct the user step by step how to get the SITE_ID and API_KEY from Fly and use them to initialize the SDK.

Beside the regular code review, I would really appreciate any suggestion to improve the visual/textual content.

## Test plan

**Initialization**
- App shows an error if the user hits Initialize button without entering SITE_ID and API_KEY
- The app persist the SITE_ID and API_KEY when the app restarts
- Hitting initialize button with values for the site id and api key fields, calls `CustomerIO.initilize`

https://github.com/customerio/customerio-ios/assets/142905621/b5de40c2-a2fd-46cf-9414-c1f7fbd34a55

**Navigation**
- Ensure the link in the "SDK Installation" view that links to the setup view does indeed navigate to the setup view
 - Pre-initialization: Ensure the app menu shows only "Intro, Install, and Setup" and the setup entry does navigate to the setup view
- Post-initilzation: Ensure the app menu shows other entries (which will show as blank lines for now)

https://github.com/customerio/customerio-ios/assets/142905621/67a1a6bf-5352-47c8-b8f9-c1f41deff0eb

PR #611, opened 2024-03-08
docs: [vision-on-sample-app][8/n] SDK Installation view

Ahmed-CIO:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #617
* #613
* #612
* __->__ #611
* #610
* #609
* #608
* #606
* #605
* #604

## Context

This is a PR in a series of PRs to build sample app that runs Identify and Track in VisionPro
The app is built to be an interactive walk through guidance for developers who are integrating Customer.io Swift SDK in their VisioPro apps.
For more context:

- See the project [one pager](https://www.notion.so/custio/Spacial-Identify-and-Tracking-6587d848bbde494095c0585237326ed3?pvs=4)
- See [Linear Issue](https://linear.app/customerio/issue/MBL-115/sample-app-for-visionpro)

## PR Summary
This PR introduces a very brief SDK installation tutorial screen. It is intentially very brief as assume the developer knows how to use Swift Package Manager or Cocoapods if their app uses any of them already.

Beside the regular code review, I would really appreciate any suggestion to improve the visual content.

## Test plan

Beside testing the newly added screenand it associated behavior, since this is the first change in the app where we have multiple screen, I will also add a test to ensure the app remembers the last opened screen.

### New user experience
This scenario validates that:
1- Top left menu "opened" status is updated when we visit the SDK Installation screen correctly.
2- Taping/chosing the "installing CustomerIO SDK" direct to the correct screen
3- Tapping/chosing the "SDK Installtion" menu entry navigates to the correct screen

**Pre condition:**
Make sure the app is not installed on the device or simulator to run through the new user experience.

**Test steps**
- Open `VisionOs.xcproject`
- Hit CMD+R to run the project and make sure you have VisionOS simulator installed and selected
- Expected behavior:
   - The app opens on the App Intro View.
   - Top left menu shows no checkmark beside the "SDK Installation" entry because we never opened it before.
   - Tapping/chosing the "installing CustomerIO SDK" should open the SDK Installation View
   - Now the top left button should show the checkmark beside the "SDK Installation"

https://github.com/customerio/customerio-ios/assets/142905621/c4faca1a-7c4a-4550-8bb6-0dc478c948f9

### Existing user experience
This scenario validates that the app opens on the last visited screen and the menu state is persisted cross app runs.

https://github.com/customerio/customerio-ios/assets/142905621/ca9670d4-ee28-4277-8611-6ef31aff666b

PR #613, opened 2024-03-08
docs: [vision-on-sample-app][11/n] PostInitialization - CIO Intro

Ahmed-CIO:
* #617
* __->__ #613
* #612
* #611
* #610
* #609
* #608
* #606
* #605
* #604

## Context

For more context:

## PR Summary
This PR introduces a readonly screen to introduce the developer to CIO as a paltform, what different termenologies used in CIO and the SDK means and how they can be used in a very high level. the usage of the SDK

## Test plan

Make sure the different links in the page navigate to the correct (not yet built) screen or open safari for external URLs


Apps/VisionOS/VisionOS/Assets.xcassets/PostInitialize.imageset/Image@2x.png
levibostian, line 1: My opinion: If this graphic was someone's first exposure to CIO and what it does, this could be overwhelming and a bit complex. 

Is there a way we could "explain it like I'm 5" to the user?

levibostian (changes requested):
I do like the idea of our sample apps sharing the value of CIO. That's exciting. However, it's also confusing to me because up until now, our sample apps were designed to be QA apps. This is getting into the territory of a showcase app. 

If that is true, I would prefer that this be a team decision to make this sample app have a different purpose then our other sample apps we have built.

Apps/VisionOS/VisionOS/Assets.xcassets/PostInitialize.imageset/Image@2x.png
Ahmed-CIO: This is app is on purpose a customer-first app not a QA app. 
So partially showcase, partially explanation and tutorial-alike.

I am not the best document writer, what you see is best attempt. If you have suggestions please suggest and we can also iterate over it as we go.

Also @glosier will take a look into this content later in this week or the week after and, again, we will iterate as needed.

Ahmed-CIO:
Apologies for that, gh-stack pushes seems to cause too much noise due to the fact that it pushes the entire stack for every push. These changes will carry on #626

PR #615, opened 2024-03-08
refactor: reduce tearDown flakiness in test suite

levibostian:
Part of: https://linear.app/customerio/issue/MBL-95/improve-test-suite-compatibility-with-async-eventbus-operations

As part of improving test suite compatibility with the EventBus, this refactor reduces flakiness by replacing the async operations in tearDown() with synchronous operations. The async code (`Task {}`) in the synchronous function (`deleteAllPersistentData`) is not guaranteed to complete by the time tearDown() is complete. This can create flaky test results where, for example, the EventBus does not reset before the next test function runs. 

This synchronous solution was chosen because it has another benefit for us: by using this more generic solution, we do not need to reset the data store for just the EventBus, we reset the data stores for all objects. Making the test suite easier to maintain. 

As an alternative solution, we could convert all test classes in the test suite to use `tearDown() async` instead of `tearDown()`. Then, we would guarantee that all async operations would complete before `tearDown()` is complete. 

commit-id:42b484a0

---

**Stack**:
- #615 ⬅
- #614

### DIFF
--- a/Tests/Shared/Core/UnitTestBase.swift
+++ b/Tests/Shared/Core/UnitTestBase.swift
@@ -108,29 +108,42 @@ open class UnitTestBase<Component>: XCTestCase {
         diGraphShared.reset()
     }
 
+    // All data that the SDK writes, delete it here so each test function has a clean environment to run and does not depend on the result of the previous test.
     open func deleteAllPersistentData() {
-        var expectations: [XCTestExpectation] = []
+        deleteAllFiles()
+        deleteKeyValueStoredData()
+    }
 
-        let resetEventBusExpectation = XCTestExpectation(description: "reset EventBus to initial state")
-        expectations.append(resetEventBusExpectation)
-        Task {
-            await diGraphShared.eventBusHandler.reset()
-            resetEventBusExpectation.fulfill()
+    // Deletes all key/value storage pairs that the SDK could save
+    private func deleteKeyValueStoredData() {
+        diGraphShared.sharedKeyValueStorage.deleteAll()
+    }
+
+    private func deleteAllFiles() {
+        let fileManager = FileManager.default
+
+        let deleteFromSearchPath: (FileManager.SearchPathDirectory) -> Void = { path in
+            // OK to use try! here as we want tests to crash if for some reason we are not able to delete files from the
+            // device.
+            // if files do not get deleted between tests, we could have false positive tests.
+            // swiftlint:disable:next force_try
+            let pathUrl = try! fileManager.url(for: path, in: .userDomainMask, appropriateFor: nil, create: false)
+            // swiftlint:disable:next force_try
+            let fileURLs = try! fileManager.contentsOfDirectory(
+                at: pathUrl,
+                includingPropertiesForKeys: nil,
+                options: .skipsHiddenFiles
+            )
+            for fileURL in fileURLs {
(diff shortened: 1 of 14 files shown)


levibostian:
Building of cocoapods sample app on CI failing on parent branch. Will not fix in this PR. All other status checks should pass.

levibostian:
The PR title and description might be a little miss-leading. Looks like the speedup of the test suite is minimal to the parent branch. 

The change is still valuable, but I would not say it's because of a speed improvement.

Tests/Shared/Core/UnitTestBase.swift
```diff
@@ -108,29 +108,42 @@ open class UnitTestBase<Component>: XCTestCase {
...
+
+    private func deleteAllFiles() {
+        let fileManager = FileManager.default
+
+        let deleteFromSearchPath: (FileManager.SearchPathDirectory) -> Void = { path in
+            // OK to use try! here as we want tests to crash if for some reason we are not able to delete files from the
+            // device.
+            // if files do not get deleted between tests, we could have false positive tests.
+            // swiftlint:disable:next force_try
+            let pathUrl = try! fileManager.url(for: path, in: .userDomainMask, appropriateFor: nil, create: false)
```
Shahroz16, line 130: even though we run tests in a controlled environment, I would have hoped we had a dedicated directory for testing and we just cleaned it. 

currently the files were being deleted using 

```
 /// Resets the EventBus to initial state by removing all observers and stored events.
    public func reset() async {
        await eventStorage.removeAll()
        await eventBus.removeAllObservers()
    }
    
    // and removeAll looks like this
    func removeAll() async {
        try? fileManager.removeItem(at: baseDirectory)
    }
```

Shahroz16 (commented):
Has this change sped up? if so do we have any metrics for it? curious because i see cleanup happening before too.

Tests/Shared/Core/UnitTestBase.swift
```diff
@@ -108,29 +108,42 @@ open class UnitTestBase<Component>: XCTestCase {
...
     }
 
+    // All data that the SDK writes, delete it here so each test function has a clean environment to run and does not depend on the result of the previous test.
     open func deleteAllPersistentData() {
-        var expectations: [XCTestExpectation] = []
-
-        let resetEventBusExpectation = XCTestExpectation(description: "reset EventBus to initial state")
-        expectations.append(resetEventBusExpectation)
-        Task {
-            await diGraphShared.eventBusHandler.reset()
```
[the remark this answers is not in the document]
levibostian: I do not believe we need to because all objects in the digraph are re-created before each test function. This includes the eventbus so all observers would be removed when we re-create the eventbus.

Tests/Shared/Core/UnitTestBase.swift
levibostian: I don't understand the question, sorry.

levibostian:
@mrehan27 @Shahroz16 I updated the PR title and description to be more accurate to the problem that's being solved. 

Sorry for the confusion. I realized that when I edited the ticket's description, I did not edit this PR's description to match it.

Shahroz16 (approved):
good enough solution for now

mrehan27 (approved):

PR #621, opened 2024-03-11
docs: [vision-os-sample-app][5/n] AppDelegate

Ahmed-CIO:
For more context:

This PR introduces fthe app delegate which will be used to initialize `CustomerIO`

- Open `VisionOs.xcproject`

### DIFF
--- a/Apps/VisionOS/VisionOS/AppDelegate.swift
+++ b/Apps/VisionOS/VisionOS/AppDelegate.swift
@@ -1,0 +1,29 @@
+import UIKit
+
+import CioDataPipelines
+
+class AppDelegate: NSObject, UIApplicationDelegate {
+    func application(
+        _ application: UIApplication,
+        didFinishLaunchingWithOptions launchOptions: [UIApplication
+            .LaunchOptionsKey: Any]? = nil
+    ) -> Bool {
+        /*
+
+         You will need cdpApiKey to initialize CustomerIO
+         Learn more about it here: https://customer.io/docs/sdk/ios/quick-start-guide/#prerequisites
+         */
+
+        let workspaceSettings = AppState.shared.workspaceSettings
+        if workspaceSettings.isSet() {
+            CustomerIO.initialize(
+                withConfig:
+                SDKConfigBuilder(cdpApiKey: workspaceSettings.cdpApiKy)
+                    .region(workspaceSettings.region)
+                    .logLevel(.debug)
+                    .build())
+        }
+
+        return true
+    }
+}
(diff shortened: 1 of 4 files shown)


Ahmed-CIO:
Closing pull request: commit has gone away

Ahmed-CIO:
Closing pull request: commit has gone away

Apps/VisionOS/VisionOS/AppDelegate.swift
```diff
@@ -0,0 +1,36 @@
...
+            CustomerIO.initialize(
+                siteId: workspaceSettings.siteId,
+                apiKey: workspaceSettings.apiKey,
+                region: workspaceSettings.region
+            ) { config in
+                // Debug config just to make the demo
+                // easier. You can learn more about these configs
+                // here: https://customer.io/docs/sdk/ios/getting-started/#configuration-options
+                config.backgroundQueueMinNumberOfTasks = 1
+                config.backgroundQueueSecondsDelay = 0
```
ami-aman, line 29: Despite the fact that we provide these configurations (`backgroundQueueMinNumberOfTasks`, `backgroundQueueMinNumberOfTasks` to be specific), [we do not recommend our users to update these values](https://customer.io/docs/sdk/ios/getting-started/#configuration-options) as it might affect the performance of the app and battery life of the device.

Just thinking out loud if we would want to update these values to minimum for the demo app ? 🤔

Apps/VisionOS/VisionOS/AppDelegate.swift
```diff
@@ -0,0 +1,36 @@
...
+
+class AppDelegate: NSObject, UIApplicationDelegate {
+    func application(
+        _ application: UIApplication,
+        didFinishLaunchingWithOptions launchOptions: [UIApplication
+            .LaunchOptionsKey: Any]? = nil
+    ) -> Bool {
+        /*
+
+         How do I get the values for the siteId and apiKy?
```
ami-aman, line 12:
```suggestion
         How do I get the values for the siteId and apiKey?
```

Apps/VisionOS/VisionOS/AppDelegate.swift
ami-aman: Also wondering that we might have to do additional efforts on this after we release CDP. There is a huge change when migrating from Journeys to CDP !

Ahmed-Ali: Not that much huge change, IMO. Just updating the callistes to use DataPipeline APIs instead of Tracking

PR #629, opened 2024-03-11
chore: fix data pipeline tests

mrehan27:
### Changes

- Fixed tests breaking due to reinitialization of `analytics` with same `writeKey`
- Fixed `logLevel` for `DataPipeline` module

### DIFF
--- a/Tests/DataPipeline/Core/UnitTest.swift
+++ b/Tests/DataPipeline/Core/UnitTest.swift
@@ -76,5 +76,6 @@ open class UnitTest: SharedTests.UnitTestBase<CustomerIO> {
     override open func cleanupTestEnvironment() {
         super.cleanupTestEnvironment()
         CustomerIO.resetTestEnvironment()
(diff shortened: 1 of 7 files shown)


Tests/DataPipeline/Core/UnitTest.swift
```diff
@@ -75,6 +75,7 @@ open class UnitTest: SharedTests.UnitTestBase<CustomerIO> {
 
     override open func cleanupTestEnvironment() {
         super.cleanupTestEnvironment()
+        Analytics.removeActiveWriteKey(dataPipelineConfigOptions.cdpApiKey)
```
Shahroz16, line 78: curious if we are removing `active write` key, why do we still need to set a random on each test case?

Shahroz16 (approved):

Tests/DataPipeline/Core/UnitTest.swift
mrehan27: Mainly for api tests as we are not actually removing the provided write keys for them and they may have multiple initializations as they are for only validating public apis.

PR #631, opened 2024-03-11
test: remove flakiness in HttpTest

levibostian:

An attempt at reducing flakiness of HttpTest tests.

The changes done thus far to try and fix issue:
* Chose a different remote image. Maybe the web server we were using blocks github actions?
* Moved the test out of /Shared/ directory and into /Common/. The test suite assumes that test functions are not written in /Shared/ so maybe that was causing issues?

How did I test this change? 
* Ran test 100 times on local dev machine with success. 

commit-id:0467ebc9

### DIFF
--- a/Tests/Common/Service/HttpRequestRunnerTest.swift
+++ b/Tests/Common/Service/HttpRequestRunnerTest.swift
@@ -39,4 +39,44 @@ class HttpRequestRunnerTest: HttpTest {
 
         waitForExpectations()
     }
+
+    func testParallelDownloadFileCreatesUniquePaths() {
+        let expectation1 = expectation(description: "Parallel download file 1")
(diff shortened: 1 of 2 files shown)


levibostian:
Unfortunately with test flakiness, the best we can do is to try and determine what the issue is, push a fix, and over time see if flakiness is better. 

I ran the test on my local dev machine 100 times and never got it to fail after these changes. Also, the tests are all passing on the CI on this PR. Making me feel good about the changes I made, but only time will tell if the flakiness is actually removed. 

If you have suggestions for the code, feel free to suggest it.

Shahroz16 (approved):
I am also unsure this is the solution, since it would pass on the local CI before as well. But lets see

ami-aman (approved):

PR #632, opened 2024-03-12
fix: open deep link from main thread as required by UIKit

levibostian:
While QA testing https://linear.app/customerio/issue/MBL-138/[bug]-ios-customers-receive-callbacks-from-3rd-party-sdks-for-pushes, I noticed Xcode give a warning that we were opening deep links from a background thread instead of main thread. UIKit operations must be executed on a main thread.

Because our proxy is creating a new async Task that by default runs on a background thread, the proxy will forward all push handling events on a background thread which is what causes this issue. We can fix this by telling Swift to run the async block on the main thread.

---

**Stack**:
- #632 ⬅
- #601
- #590
- #584
- #583


levibostian:
I think this was created by spr by accident? Closing.

PR #634, opened 2024-03-12
fix: call completionHandler if push handler does not implement option…

levibostian:
…al function

While QA testing https://linear.app/customerio/issue/MBL-138/[bug]-ios-customers-receive-callbacks-from-3rd-party-sdks-for-pushes, I noticed the 3rd party push handler was not calling the completionHandler, which resulted in our SDK not able to finish processing the push event.

The reason this issue can occur is because the UserNotificationCenterDelegate functions are optional. So, customers do not need to implement functions in their app, if they do not want to. Check the FCM sample app AppDelegate for an example where we only implement 1 of the 2 functions for push event handling.

This commit fixes the issue by checking if the push handler implements the optional function or not. If not, we call the completion handler for you.

Testing performed:
* manual QA testing iOS sample apps. Checked the value of `delegateGotCalled` when the optional function is and is not implemented in host iOS app. 

Note: Because UserNotification center functions are not able to execute in test suite, the code in this PR can only be QA tested and automated tests cannot be written for it. 

commit-id:c6506c76

---

**Stack**:
- #634 ⬅
- #633
- #601
- #590
- #584
- #583

### DIFF
--- a/Apps/CocoaPods-FCM/src/AppDelegate.swift
+++ b/Apps/CocoaPods-FCM/src/AppDelegate.swift
@@ -73,9 +73,9 @@ extension AppDelegate: UNUserNotificationCenterDelegate {
         completionHandler()
     }
 
-    // For testing purposes, it's suggested to not include the willPresent function in the AppDelegate.
-    // The automatic push click handling feature uses swizzling. A good edge case to test with swizzling is when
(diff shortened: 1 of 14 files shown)


Shahroz16 (approved):
couldn't find any documentation that mentioned about `optional` methods returning `nil` but looks a safe change and as long manual testing passes I am good. 

it would be good to write an automated test too, that returns nil

PR #635, opened 2024-03-12
test: fix flaky tests when running on CI

levibostian:
Part of: https://linear.app/customerio/issue/MBL-164/cdp-branch-flaky-tests-on-ci

We have been experiencing some flaky tests, only when running on our CI. After reviewing the test functions and the error messages thrown on CI, I think that the issue is because of slow performance when running the test suite on the CI. After doing some research online, others were sharing similar issues, specifically when running tests on Xcode 15.0 and 15.1. Xcode release notes also share known issues when running tests on CI and slow performance.

By updating our CI to macos 14 github action runners, we are installing Xcode 15.2 which has better test performance.

---

**Stack**:
- #636
- #635 ⬅

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #637, opened 2024-03-13
test: fix flaky tests that use convenient function `runTest(numberOfTimes)`

levibostian:

I have noticed a pattern where test functions that use the convenient function, runTest(numberOfTimes:), are flaky on CI. `runTest` was introduced as a way to try and catch race conditions by running a test many times.

While doing deeper research to running test functions multiple times, the solutions I found were either (1) working for older versions of Xcode but not the latest versions or (2) identical solution as `runTest(numberOfTimes:)` but with an additional `sleep` function after setup and teardown.

I could not find a solution that was reliable and up-to-date. Therefore, I think the best solution is remove `runTest`.

If you want to run a test multiple times, you can inside of Xcode: https://stackoverflow.com/a/67963997

---

**Stack**:
- #638
- #637 ⬅
- #636
- #635


levibostian:
🤦🏻 seems as though this PR's changes still result in flaky tests. 

I'll try and figure out why this is.

levibostian:
Solution didn't remove flakiness. Dropping PR change.

PR #639, opened 2024-03-14
chore: auto screen tracking plugin using config

mrehan27:
closes: https://linear.app/customerio/issue/MBL-166/update-autotrack-screen-view-config

### Changes

- Added `autoTrackScreenViews` to `SDKConfigBuilder` so customers can attach screen plugin using config
- Added `autoConfiguredPlugins` to `DataPipelineConfigOptions` so builder can provide finals plugins using convenient config options for customers

### DIFF
--- a/Apps/APN-UIKit/APN UIKit/AppDelegate.swift
+++ b/Apps/APN-UIKit/APN UIKit/AppDelegate.swift
@@ -57,11 +57,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         if let cdnHost = storage.cdnHost, !cdnHost.isEmpty {
             config.cdnHost(cdnHost)
         }
-        CustomerIO.initialize(withConfig: config.build())
-        let autoScreenTrack = storage.isTrackScreenEnabled ?? true
-        if autoScreenTrack {
-            CustomerIO.shared.add(plugin: AutoTrackingScreenViews(filterAutoScreenViewEvents: nil, autoScreenViewBody: nil))
+        if storage.isTrackScreenEnabled == true {
+            config.autoTrackScreenViews()
         }
+        CustomerIO.initialize(withConfig: config.build())
 
         // Initialize messaging features after initializing Customer.io SDK
         MessagingInApp
--- a/Apps/APN-UIKit/APN UIKit/AppDelegate.swift
+++ b/Apps/APN-UIKit/APN UIKit/AppDelegate.swift
@@ -58,7 +58,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
             config.cdnHost(cdnHost)
         }
         if storage.isTrackScreenEnabled == true {
-            config.autoTrackScreenViews()
+            config.autoTrackUIKitScreenViews()
         }
         CustomerIO.initialize(withConfig: config.build())
 
--- a/Sources/DataPipeline/Config/DataPipelineConfigOptions.swift
+++ b/Sources/DataPipeline/Config/DataPipelineConfigOptions.swift
@@ -30,4 +30,7 @@ public struct DataPipelineConfigOptions {
 
     /// Configuration options required for migration from earlier versions
     public let migrationSiteId: String?
+
+    /// Plugins identified based on configuration provided by the user
+    let autoConfiguredPlugins: [Plugin]
 }
--- a/Sources/DataPipeline/Config/SDKConfigBuilder.swift
+++ b/Sources/DataPipeline/Config/SDKConfigBuilder.swift
@@ -1,6 +1,9 @@
 import CioInternalCommon
 import Foundation
 import Segment
+#if canImport(UIKit)
+import UIKit
+#endif
 
 /// Builder class designed to facilitate the creation of `SdkConfig` and `DataPipelineConfigOptions`.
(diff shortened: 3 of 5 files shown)


Sources/DataPipeline/Config/DataPipelineConfigOptions.swift
```diff
@@ -30,4 +30,7 @@ public struct DataPipelineConfigOptions {
 
     /// Configuration options required for migration from earlier versions
     public let migrationSiteId: String?
+
+    /// Plugins identified based on configuration provided by the user
+    let autoConfiguredPlugins: [Plugin]
```
mrehan27, line 35: I introduced this to simplify adding plugins based on configurations, see my comment below on `SDKConfigBuilder` for more details

Sources/DataPipeline/Config/SDKConfigBuilder.swift
```diff
@@ -137,12 +168,22 @@ public class SDKConfigBuilder {
...
         )
 
+        // create plugins based on given configurations
+        var configuredPlugins: [Plugin] = []
+        if autoTrackUIKitScreenViews {
+            configuredPlugins.append(AutoTrackingScreenViews(
+                filterAutoScreenViewEvents: filterAutoScreenViewEvents,
+                autoScreenViewBody: autoScreenViewBody
+            ))
+        }
```
mrehan27, line 185: `autoConfiguredPlugins` allows us providing more config options in builder to help customers setup the SDK. We don't need to pass and store all these configs in `DataPipelineConfigOptions` again. We can instead translate the config into a plugin, and `DataPipelineImplementation` will automatically attach all these plugins whenever initialized.

Open to suggestions though. And if this sounds good, I can do the changes for Logger plugin also to keep it consistent.

Sources/DataPipeline/Config/SDKConfigBuilder.swift
```diff
@@ -137,12 +168,22 @@ public class SDKConfigBuilder {
         return self
     }
 
+    @available(iOSApplicationExtension, unavailable)
```
mrehan27, line 171: We need this because we are accessing `AutoTrackingScreenViews` in it. I don't think this is an issue as [initialize method is already annotated similar](https://github.com/customerio/customerio-ios/blob/082d21b70e829381d7f4bc2c76b99f3c070c1185/Sources/DataPipeline/CustomerIO.swift#L9). So customers may never need to build config outside the normal initialization of SDK.

Apps/APN-UIKit/APN UIKit/AppDelegate.swift
```diff
@@ -57,11 +57,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
             config.cdnHost(cdnHost)
         }
-        CustomerIO.initialize(withConfig: config.build())
-        let autoScreenTrack = storage.isTrackScreenEnabled ?? true
-        if autoScreenTrack {
-            CustomerIO.shared.add(plugin: AutoTrackingScreenViews(filterAutoScreenViewEvents: nil, autoScreenViewBody: nil))
+        if storage.isTrackScreenEnabled == true {
+            config.autoTrackUIKitScreenViews()
         }
+        CustomerIO.initialize(withConfig: config.build())
```
Shahroz16, line 63: were we not building config before? 😅

Sources/DataPipeline/Config/SDKConfigBuilder.swift
Shahroz16: have you verified this change with Cocoapods FCM app? i remember last time either not having it was causing issue or having it. Not 100% sure.

Sources/DataPipeline/Config/SDKConfigBuilder.swift
```diff
@@ -137,12 +168,22 @@ public class SDKConfigBuilder {
...
     public func build() -> SDKConfigBuilderResult {
-        // create `SdkConfig`` from given configurations
+        // create `SdkConfig` from given configurations
         let sdkConfig = SdkConfig.Factory.create(
             logLevel: logLevel
         )
 
+        // create plugins based on given configurations
+        var configuredPlugins: [Plugin] = []
```
Shahroz16, line 179: maybe on `de init` we should empty the array?

Shahroz16 (approved):
Looks good, added minor suggestions.

Apps/APN-UIKit/APN UIKit/AppDelegate.swift
mrehan27: We were, I think its just code movement that Github is showing it as deleted and added again 😅

Sources/DataPipeline/Config/SDKConfigBuilder.swift
mrehan27: Might not be needed here as it is in function. And can't add to `DataPipelineConfigOptions` as it is struct.

Sources/DataPipeline/Config/SDKConfigBuilder.swift
mrehan27: All CI checks are working, that too might help us giving confidence that it's working for Cocoapods FCM app 🤔

PR #640, opened 2024-03-14
chore: device attributes update

Shahroz16:
closes: https://linear.app/customerio/issue/MBL-168/device-attributes-in-journeys

It does the following:

- Remove CIO device attributes being sent into `context.device` (it was unnecessary)
- if autoTrackDeviceAttributes is false, remove all device attributes (which was happening before) from context except `token` and `type` which other destinations depend on
- Selected context attributes are now part of device attributes by adding an event plugin
- Remove `library` from `context` as part of https://linear.app/customerio/issue/MBL-173/remove-library-value

### DIFF
--- a/Sources/DataPipeline/DataPipelineImplementation.swift
+++ b/Sources/DataPipeline/DataPipelineImplementation.swift
@@ -39,6 +39,9 @@ class DataPipelineImplementation: DataPipelineInstance {
             analytics.add(plugin: customerIODestination)
         }
 
+        // plugin to add contextual inforamation to device attributes
+        analytics.add(plugin: DeviceCreateOrUpdate())
+
         // plugin to update context properties for each request
         analytics.add(plugin: contextPlugin)
 
--- a/Sources/DataPipeline/Plugins/Context.swift
+++ b/Sources/DataPipeline/Plugins/Context.swift
@@ -40,12 +40,19 @@ class Context: Plugin {
             // set the user agent
             context["userAgent"] = userAgentUtil.getUserAgentHeaderValue()
 
-            // set the device attributes
-            if let device = context[keyPath: "device"] as? [String: Any], autoTrackDeviceAttributes {
-                context["device"] = device.mergeWith(attributes)
-            } else {
-                context["device"] = attributes
+            // remove library from context
+            context.removeValue(forKey: "library")
+
+            // if autoTrackDeviceAttributes is false, remove all device attributes except token and type which other destination might depend on
+            if let device = context[keyPath: "device"] as? [String: Any] {
+                if !autoTrackDeviceAttributes {
+                    // Keep only device.token and device.type, remove everything else
+                    let token = device["token"]
+                    let type = device["type"]
+                    context["device"] = ["token": token, "type": type]
+                }
             }
+
             workingEvent.context = try JSON(context)
         } catch {
             analytics?.reportInternalError(error)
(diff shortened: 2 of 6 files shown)


Sources/DataPipeline/Plugins/Context.swift
```diff
@@ -40,12 +37,9 @@ class Context: Plugin {
             // set the user agent
             context["userAgent"] = userAgentUtil.getUserAgentHeaderValue()
 
-            // set the device attributes
```
Shahroz16, line 43: we are not sending CIO device attributes in context, it was providing no value

ami-aman:
> Remove CIO device attributes being sent into context.device (it was unnecessary)

@Shahroz16 In tracking migration, we are [doing the same](https://github.com/customerio/customerio-ios/blob/082d21b70e829381d7f4bc2c76b99f3c070c1185/Sources/Migration/DataPipelineMigrationAssistant.swift#L146) i.e. sending the attributes in `context.device`. I believe this change will affect it too ?

Shahroz16:
> > Remove CIO device attributes being sent into context.device (it was unnecessary)
> 
> @Shahroz16 In tracking migration, we are [doing the same](https://github.com/customerio/customerio-ios/blob/082d21b70e829381d7f4bc2c76b99f3c070c1185/Sources/Migration/DataPipelineMigrationAssistant.swift#L146) i.e. sending the attributes in `context.device`. I believe this change will affect it too ?

@ami-aman nope, `token` etc is not a device attribute, its model, OS version etc. `Context.device` already collects them, we were just adding ours alongside Segments, which was causing duplications.

ami-aman:
> nope, `token` etc is not a device attribute, its model, OS version etc. `Context.device` already collects them, we were just adding ours alongside Segments, which was causing duplications.

Hmm..Device attributes does not only contain `token` it also contain model, OS version etc as collected by our current Journey module. When we do migration, we are sending the data as is to CDP which is kinda duplicate ! Just thinking out loud.

Sources/DataPipeline/DataPipelineImplementation.swift
```diff
@@ -39,6 +44,11 @@ class DataPipelineImplementation: DataPipelineInstance {
             analytics.add(plugin: customerIODestination)
         }
 
+        // plugin to add contextual information to device attributes
+        if let plugin = deviceUpdatePlugin {
+            analytics.add(plugin: plugin)
+        }
```
mrehan27, line 50: Do we need to store it at class level? Or could this be simply follows? 🤔 

```suggestion
        if moduleConfig.autoTrackDeviceAttributes {
            analytics.add(plugin: DeviceContexualAttributes())
        }
```

Sources/DataPipeline/Plugins/Context.swift
```diff
@@ -40,12 +37,9 @@ class Context: Plugin {
...
-            if let device = context[keyPath: "device"] as? [String: Any], autoTrackDeviceAttributes {
-                context["device"] = device.mergeWith(attributes)
-            } else {
-                context["device"] = attributes
-            }
+            // remove library from context
```
mrehan27, line 40: Can we please add some context here in comment for future reference 🤓

mrehan27 (approved):
LGTM 👍🏻

Sources/DataPipeline/DataPipelineImplementation.swift
Shahroz16: ahhh forgot we have module config here

PR #641, opened 2024-03-15
chore: logger fixes and improvements

mrehan27:
### Changes

- Fixed singleton logger instance
- Based on changes done in #639, added logger plugin to builder to utilize `autoConfiguredPlugins`
- Removed `setDebugLogsEnabled` extension as I think it was initially added to help `Tracking` module and no longer needed
- Fixed logLevel for NSE
- Updated tests and sample apps

### DIFF
--- a/Sources/DataPipeline/CustomerIO+Plugins.swift
+++ b/Sources/DataPipeline/CustomerIO+Plugins.swift
@@ -57,21 +57,4 @@ public extension CustomerIO {
     func find(key: String) -> DestinationPlugin? {
         DataPipeline.shared.analytics.find(key: key)
     }
-
-    func setDebugLogsEnabled(_ enabled: Bool = true) {
-        // enable debug logs in analytics
-        Analytics.debugLogsEnabled = enabled
-
-        let analytics = DataPipeline.shared.analytics
-        let loggerPlugin = analytics.find(pluginType: ConsoleLogger.self)
-        if enabled {
-            // if logs are enabled but plugin is not added, add it to enable logs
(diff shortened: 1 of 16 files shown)


Sources/DataPipeline/CustomerIO+Plugins.swift
```diff
@@ -57,21 +57,4 @@ public extension CustomerIO {
     func find(key: String) -> DestinationPlugin? {
         DataPipeline.shared.analytics.find(key: key)
     }
-
-    func setDebugLogsEnabled(_ enabled: Bool = true) {
```
mrehan27, line 61: Removed this as I think it was initially added to help Tracking module and no longer needed with this change

Sources/MessagingPush/MessagingPush.swift
```diff
@@ -68,7 +68,9 @@ public class MessagingPush: ModuleTopLevelObject<MessagingPushInstance>, Messagi
     @discardableResult
     public static func initializeForExtension(withConfig config: MessagingPushConfigOptions) -> MessagingPushInstance {
         shared.initializeModuleIfNotAlready {
-            shared.getImplementation(config: config)
+            // Set logLevel of shared logger only when module is initialized from NotificationServiceExtension.
+            DIGraphShared.shared.logger.setLogLevel(config.logLevel)
```
mrehan27, line 72: While a discussion, Shahroz pointed out we are not setting log level for NSE, so I updated this to fix for NSE.

Shahroz16 (approved):
Looks good, thanks for adding

PR #642, opened 2024-03-15
fix: use git commit instead of git branch for segment dependency

levibostian:
Part of: https://linear.app/customerio/issue/MBL-167/segment-dependency-currently-pointing-to-main-branch

In the iOS SDK today, with the Segment CDP SDK, we are currently specifying the version of the segment SDK as "main branch". Using a git branch as the version of a dependency poses risks, potentially leading to difficulties for customers during SDK updates and app compilation, increasing the chance of crashes or compilation issues.

This change uses a hard-coded git commit. The git commit used is the latest commit on the "main" branch.

There are pros and cons to using a git commit as the version, as specified in the ticket description. We can change to a different version such as a git tag in the future.

commit-id:45c2d17e

### DIFF
--- a/CustomerIODataPipelines.podspec
+++ b/CustomerIODataPipelines.podspec
@@ -24,5 +24,5 @@ Pod::Spec.new do |spec|
   spec.dependency "CustomerIOTrackingMigration", "= #{spec.version.to_s}"
 
   # Add Segment SDK as a dependency, as this module is designed to be compatible with it.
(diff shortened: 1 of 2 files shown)


Shahroz16 (approved):

Shahroz16 (commented):
Lets wait till we decide on..

https://linear.app/customerio/issue/MBL-167/segment-dependency-currently-pointing-to-main-branch

Shahroz16 (approved):
Looks good!

PR #643, opened 2024-03-15
test: fix flaky tests that run code on another thread

levibostian:

There is a set of flaky tests that all have 1 thing in common: They are the only tests that call the functions `runOnBackground` or `runOnMain`. 

From my own personal experience and research online, tests that use GCD can have unexpected/flaky results, especially on simulators. Recent versions of Xcode or Swift may also have changed the behavior to become flaky. 

This commit changes to using swift concurrency instead of GCD to run test functions on background threads. We hope that all of these tests pass on the CI with no more flaky results. 

commit-id:cc25358b

### DIFF
--- a/Tests/Common/Module/ModuleTopLevelObjectTest.swift
+++ b/Tests/Common/Module/ModuleTopLevelObjectTest.swift
@@ -37,7 +37,7 @@ class ModuleTopLevelObjectTest: UnitTest {
                 expectAllThreadsToComplete.fulfill()
             }
 
-            waitForExpectations(1) // test may take up to 1 second to finish because it is running so many times. CI server is a less powerful machine and this test is flaky when we set wait() for < 1 second.
+            waitForExpectations()
 
             // Even though we call initialize twice, the initialize count should only be 1.
             XCTAssertEqual(ModuleTopLevelObjectStub.shared.initializeCount, 1)
--- a/Tests/MessagingPush/PushHandling/PushEventHandlerProxyTest.swift
+++ b/Tests/MessagingPush/PushHandling/PushEventHandlerProxyTest.swift
@@ -28,7 +28,7 @@ class PushEventHandlerProxyTest: UnitTest {
             delegate1.onPushActionClosure = { _, completion in
                 expectDelegatesReceiveEvent.fulfill()
 
-                self.runOnMain {
+                self.runOnBackground {
                     completion()
                 }
             }
@@ -69,7 +69,7 @@ class PushEventHandlerProxyTest: UnitTest {
             delegate1.shouldDisplayPushAppInForegroundClosure = { _, completion in
                 expectDelegatesReceiveEvent.fulfill()
 
-                self.runOnMain {
+                self.runOnBackground {
                     completion(false)
                 }
             }
--- a/Tests/Shared/Core/UnitTestBase.swift
+++ b/Tests/Shared/Core/UnitTestBase.swift
@@ -165,13 +165,7 @@ open class UnitTestBase<Component>: XCTestCase {
     }
 
     public func runOnBackground(_ block: @escaping () -> Void) {
-        CioThreadUtil().runBackground {
-            block()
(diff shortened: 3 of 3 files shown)


levibostian:
Yay! The test suite has passed successfully 3 out of 3 times I have run the suite on this PR! 

The tests could still be flaky however, it's unlikely as I would have at least 1 flaky test fail over 50% of the time while working on this ticket.

Tests/Common/Module/ModuleTopLevelObjectTest.swift
```diff
@@ -37,7 +37,7 @@ class ModuleTopLevelObjectTest: UnitTest {
                 expectAllThreadsToComplete.fulfill()
             }
 
-            waitForExpectations(1) // test may take up to 1 second to finish because it is running so many times. CI server is a less powerful machine and this test is flaky when we set wait() for < 1 second.
+            waitForExpectations()
```
levibostian, line 40: Increasing the timeout originally was to make the test less flaky on CI. Since then, I have realized that this timeout does not reduce flakiness so, removing it.

Tests/Shared/Core/UnitTestBase.swift
```diff
@@ -165,13 +165,7 @@ open class UnitTestBase<Component>: XCTestCase {
     }
 
     public func runOnBackground(_ block: @escaping () -> Void) {
-        CioThreadUtil().runBackground {
-            block()
-        }
-    }
-
-    public func runOnMain(_ block: @escaping () -> Void) {
```
levibostian, line 173: Removed `runOnMain` since there is no more purpose behind it. The reason it was added was to make sure that a test function could create 2+ different threads to run code on. However, the existing function `runOnBackground` already creates a new thread every time that the function is called. Therefore, no reason for `runOnMain`.

Tests/Shared/Core/UnitTestBase.swift
```diff
@@ -165,13 +165,7 @@ open class UnitTestBase<Component>: XCTestCase {
...
-        CioThreadUtil().runMain {
+        Task {
```
levibostian, line 168: This is the bug fix. Switching from GCD to swift concurrency to see if that increases the reliability of the test suite.

Tests/MessagingPush/PushHandling/PushEventHandlerProxyTest.swift
```diff
@@ -28,7 +28,7 @@ class PushEventHandlerProxyTest: UnitTest {
             delegate1.onPushActionClosure = { _, completion in
                 expectDelegatesReceiveEvent.fulfill()
 
-                self.runOnMain {
+                self.runOnBackground {
```
Shahroz16, line 31: shouldn't notification callback be running on main?

levibostian: The `delegate1`  calling the completion handler represents a 3rd party push handler. They can call the completion handler on whatever thread they want. Our SDK can call `delegate1`  on the main thread, but the delegate gets to decide what thread they complete on.

Tests/Shared/Core/UnitTestBase.swift
Shahroz16: How are we calculating the reliability? because the CI run before this PR also passed?

Shahroz16 (approved):
> This is the bug fix. Switching from GCD to swift concurrency to see if that increases the reliability of the test suite.

Tests/Shared/Core/UnitTestBase.swift
levibostian: While working on this ticket, I have run the test suite on the CI a lot.

Before this PR, I was noticing the test suite that would call `runOnMain` or `runOnBackground` would fail 75% of the time. The fact that the test suite has executed 4 times thus far on this PR and none of the tests have failed thus far, that shows increased reliability to me.

PR #648, opened 2024-03-18
test: added integration test for event bus with datapipeline implementation

Shahroz16:
closes: https://linear.app/customerio/issue/MBL-179/increase-integration-test-coverage-for-eventbus-events-in-the-cdp

- Added `DataPipelineEventBustTests` which checks once EventBus receives an event in `DataPipelineImplementation` its sending it to CDP
- Added test so that given AutoTrackDeviceAttributes is Disabled, we verify attributes are not being added in properties


Sources/Common/Communication/EventBusHandler.swift
```diff
@@ -144,7 +144,6 @@ public class CioEventBusHandler: EventBusHandler {
 
     /// Resets the EventBus to initial state by removing all observers and stored events.
     public func reset() async {
-        await eventStorage.removeAll()
```
mrehan27, line 147: Since this is public api, I suggest we keep the name clear and not misleading. So may be rename this to `removeAllObservers` instead? 🤔

Tests/DataPipeline/DataPipelineEventBustTests.swift
```diff
@@ -0,0 +1,81 @@
...
+
+    override func setUp() {
+        super.setUp(modifySdkConfig: nil)
+        // OutputReaderPlugin helps validating interactions with analytics
+        outputReader = (customerIO.add(plugin: OutputReaderPlugin()) as? OutputReaderPlugin)
+        userAgentUtil = UserAgentUtilImpl(deviceInfo: deviceInfoStub)
+    }
+
+    func testSubscribeToJourneyEvents_DataPipelineHandlesTrackMetricEvent() async {
+        diGraphShared.override(value: eventBusHandler, forType: EventBusHandler.self)
```
mrehan27, line 34: Why we didn't do this in `setUpDependencies`?

Tests/Shared/Core/UnitTestBase.swift
```diff
@@ -112,6 +112,9 @@ open class UnitTestBase<Component>: XCTestCase {
 
     // All data that the SDK writes, delete it here so each test function has a clean environment to run and does not depend on the result of the previous test.
     open func deleteAllPersistentData() {
+        // need to remove the observers for integration tests that utilizes actual NotificationCenter
+        // otherwise, results are flaky
+        Task { await diGraphShared.eventBusHandler.reset() }
```
mrehan27, line 117: 👍🏻

Tests/DataPipeline/Support/Segment.swift
```diff
@@ -67,7 +67,7 @@ extension RawEvent {
         return nil
     }
 
-    var deviceAttributes: [String: Any]? {
+    var contextDeviceAttributes: [String: Any]? {
```
mrehan27, line 70: Why did we rename this? 🤔

mrehan27 (approved):
None blocker comments. LGTM 👍🏻

Tests/DataPipeline/DataPipelineEventBustTests.swift
Shahroz16: good catch, it was added by mistake

Tests/DataPipeline/Support/Segment.swift
Shahroz16: to be able to differentiate between device attributes added as property vs device object in context

Tests/Shared/Core/UnitTestBase.swift
```diff
@@ -112,6 +112,9 @@ open class UnitTestBase<Component>: XCTestCase {
 
     // All data that the SDK writes, delete it here so each test function has a clean environment to run and does not depend on the result of the previous test.
     open func deleteAllPersistentData() {
+        // need to remove the observers for integration tests that utilizes actual NotificationCenter
+        // otherwise, results are flaky
+        Task { await diGraphShared.eventBusHandler.removeAllObservers() }
```
levibostian, line 117: This change will probably result in flaky test results. 

When you create a `Task{}` we do not know when the task will execute. There is a chance that `diGraphShared.eventBusHandler.removeAllObservers()` will happen after tearDown is done running and so it can't be reliable. We need a guarantee that `removeAllObservers` will be done executing before tearDown() is done running. 

Ideas to fix this: 
* Make `eventBusHandler.removeAllObservers()` not an `async` function. Make it sync.
* Modify EventBusHandler and EventBus `deinit` function to call `removeAllObservers`. Then, when our test suite runs DIGraph.reset(), the `removeAllObservers` should be called for us, no longer needing this line of code in our tearDown method at all. 
* There are other options such as making `deleteAllPersistentData` an `async` function but that would be a much larger change. I don't suggest doing that at this time.

PR #650, opened 2024-03-19
chore: revert push module config to use static

mrehan27:
### Background

The PR fixes an issue where `RichPushHttpClient` was trying to access push module config before it was set as it is initialized with `MessagingPushImplementation` due to changes made in #560 

### Changes

- Reverted push module to be static because it matches with `DataPipeline` module implementation
- Event though we could update `RichPushHttpClient` to use config from implementation, but this appears to be error-prone and has a chance of breaking things in future. So we have reverted back to use static module config directly.

### DIFF
--- a/Sources/MessagingPush/MessagingPush.swift
+++ b/Sources/MessagingPush/MessagingPush.swift
@@ -7,11 +7,11 @@ import Foundation
   */
 public class MessagingPush: ModuleTopLevelObject<MessagingPushInstance>, MessagingPushInstance {
     @Atomic public private(set) static var shared = MessagingPush()
+    @Atomic public private(set) static var moduleConfig: MessagingPushConfigOptions = MessagingPushConfigBuilder().build()
 
     private static let moduleName = "MessagingPush"
 
     private var globalDataStore: GlobalDataStore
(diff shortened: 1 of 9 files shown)


Sources/MessagingPush/MessagingPush.swift
```diff
@@ -50,6 +53,8 @@ public class MessagingPush: ModuleTopLevelObject<MessagingPushInstance>, Messagi
     @available(iOSApplicationExtension, unavailable)
     public static func initialize(withConfig config: MessagingPushConfigOptions = MessagingPushConfigBuilder().build()) -> MessagingPushInstance {
         shared.initializeModuleIfNotAlready {
+            // set moduleConfig before creating implementation instance as dependencies inside instance may directly use moduleConfig from MessagingPush.
+            Self.moduleConfig = config
```
levibostian, line 57: Thanks for adding these comments! If the order matters, a comment is good to have!

levibostian (approved):

Sources/MessagingPush/MessagingPush.swift
Shahroz16: Since all modules rely on the same order, that is Config gets built first and then implemented, maybe we should have a structure in place so that whenever a new module is built it follows a strict protocol of config and implementation.

Shahroz16 (approved):

PR #651, opened 2024-03-19
chore: remove unwanted public apis

mrehan27:
closes: https://linear.app/customerio/issue/MBL-181/remove-config-settings-from-public-api

### Changes

- Removed following public apis as discussed in the ticket
  - `operatingMode`
  - `flushInterval`
  - `flushAt`
  - `flushPolicies`
  - `traits`
  - `version`
  - `settings`
  - `manuallyEnableDestination`
  - `openURL`
- Fixed API tests

### DIFF
--- a/Sources/DataPipeline/Config/SDKConfigBuilder.swift
+++ b/Sources/DataPipeline/Config/SDKConfigBuilder.swift
@@ -60,6 +60,7 @@ public class SDKConfigBuilder {
         return self
     }
 
+    #if canImport(UIKit)
     /// Enable this property if you want SDK to automatically track screen views for UIKit based apps.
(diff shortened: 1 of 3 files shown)


Sources/DataPipeline/Config/SDKConfigBuilder.swift
```diff
@@ -60,6 +60,7 @@ public class SDKConfigBuilder {
         return self
     }
 
+    #if canImport(UIKit)
```
mrehan27, line 63: I realized it was missed earlier, so updated to expose only with UIKit

Shahroz16 (approved):

PR #655, opened 2024-03-20
chore: hide object APIs for initial release

Shahroz16:
Until the group traits format is sorted out, we have decided to hide the group APIs.


ami-aman (approved):
lgtm

PR #660, opened 2024-03-21
fix: install 3.0 with Swift Package Manager from Xcode

levibostian:
Resolve error, "Failed to resolve dependencies Dependencies could not be resolved because root depends on 'customerio-ios' 3.0.0..<4.0.0. 'customerio-ios' >= 3.0.0 cannot be used because no versions of 'customerio-ios' match the requirement and package is required using a stable-version but depends on an unstable-version package"

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -45,7 +45,7 @@ let package = Package(
         .package(name: "Firebase", url: "https://github.com/firebase/firebase-ios-sdk.git", "8.7.0"..<"11.0.0"),
 
         // The version is a git commit hash. Make sure the commit is the same as what the DataPipelines CocoaPods is using.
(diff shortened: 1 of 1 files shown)


Package.swift
```diff
@@ -45,7 +45,7 @@ let package = Package(
         .package(name: "Firebase", url: "https://github.com/firebase/firebase-ios-sdk.git", "8.7.0"..<"11.0.0"),
 
         // The version is a git commit hash. Make sure the commit is the same as what the DataPipelines CocoaPods is using.
-        .package(name: "Segment", url: "https://github.com/customerio/cdp-analytics-swift.git", .revision("b2f8f5b68dd0d8796b4bacdec39aacb5b6387a41"))
+        .package(name: "Segment", url: "https://github.com/customerio/cdp-analytics-swift.git", "1.5.5-cio.1")
```
Shahroz16, line 48: do we not need `.exact("1.5.5-cio.1")`?

Shahroz16 (approved):

mrehan27 (approved):

PR #663, opened 2024-03-23
docs: [vision-os-sample-app][5/n] AppDelegate

Ahmed-CIO:
This is a PR in a series of PRs to build sample app that runs Identify and Track in VisionPro The app is built to be an interactive walk through guidance for developers who are integrating Customer.io Swift SDK in their VisioPro apps. For more context:

- See the project [one pager](https://www.notion.so/custio/Spacial-Identify-and-Tracking-6587d848bbde494095c0585237326ed3?pvs=4)
- See [Linear Issue](https://linear.app/customerio/issue/MBL-115/sample-app-for-visionpro)

This PR introduces fthe app delegate which will be used to initialize `CustomerIO`

- Open `VisionOs.xcproject`
- Hit CMD+R to run the project and make sure you have VisionOS simulator installed and selected

PR #666, opened 2024-03-23
docs: [vision-os-sample-app][8/n] Identify view

Ahmed-CIO:
This is a PR in a series of PRs to build sample app that runs Identify and Track in VisionPro The app is built to be an interactive walk through guidance for developers who are integrating Customer.io Swift SDK in their VisioPro apps. For more context:

## In this PR
This PR introduces a the "Identify" example screen. It demonstrates the requirements of the API, allows the user to enter profile attributes to be send alongside the identify call.

## Test Plan
Ensure identify works as expected end to end

https://github.com/customerio/customerio-ios/assets/142905621/b2e0c805-2a2a-4a2a-9503-992d4baf3704

--

## PR Stack

- #668 
- #667 
- #666 👈
- #665 
- #664   
- #663

### DIFF
--- a/Apps/VisionOS/VisionOS/APIViews/IdentifyView.swift
+++ b/Apps/VisionOS/VisionOS/APIViews/IdentifyView.swift
@@ -1,0 +1,95 @@
+import MarkdownUI
+import Splash
+import SwiftUI
+
+private func identifyCodeSnippet(_ profile: Profile) -> String {
+    let propertiesStr = propertiesToTutorialString(
+        profile.traits,
+        linePrefix: "\t\t"
+    )
+
+    if !profile.userId.isEmpty, !profile.traits.isEmpty {
+        return
+            """
+            CustomerIO.shared.identify(
+                userId: "\(profile.userId)",
+                traits: \(propertiesStr)
+            )
+            """
+    } else if !profile.userId.isEmpty {
+        return
+            """
+            CustomerIO.shared.identify(
+                userId: "\(profile.userId)"
+            )
+            """
+    } else if !profile.traits.isEmpty {
+        return
+            """
+            CustomerIO.shared.identify(",
+                traits: \(propertiesStr)
+            )
+            """
+    }
+    return "// enter user id or add a trait"
+}
+
+struct IdentifyView: View {
+    @ObservedObject var state = AppState.shared
+    @State private var profile = AppState.shared.profile
+    @EnvironmentObject private var viewModel: ViewModel
+
+    let onSuccess: (Profile) -> Void
+
+    var body: some View {
+        ScrollView(.vertical) {
+            VStack(alignment: .leading) {
+                Markdown {
+                    """
+                    Before you can track events, set devices attributes or update user traits, you must identify the user first.
+                    To identify a user, you must set a userId or traits or both.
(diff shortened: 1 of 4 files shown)


Apps/VisionOS/VisionOS/APIViews/IdentifyView.swift
```diff
@@ -0,0 +1,95 @@
...
+        profile.traits,
+        linePrefix: "\t\t"
+    )
+
+    if !profile.userId.isEmpty, !profile.traits.isEmpty {
+        return
+            """
+            CustomerIO.shared.identify(
+                userId: "\(profile.userId)",
+                traits: \(propertiesStr)
```
ami-aman, line 16: I believe this would throw error if I copy this snippet and use in some app 🤔 Should this be "\(propertiesStr)" to make it work ?

Apps/VisionOS/VisionOS/APIViews/IdentifyView.swift
```diff
@@ -0,0 +1,95 @@
...
+        return
+            """
+            CustomerIO.shared.identify(
+                userId: "\(profile.userId)"
+            )
+            """
+    } else if !profile.traits.isEmpty {
+        return
+            """
+            CustomerIO.shared.identify(",
```
ami-aman, line 29:
```suggestion
            CustomerIO.shared.identify("",
```

ami-aman: Hmm.. I do not think calling `CustomerIO.shared.identify("", someTraits)` will call the identify method as we have a [check in place](https://github.com/customerio/customerio-ios/blob/e9ad4acb9a72348380180848ead8d9ade8fe4f0e/Sources/DataPipeline/DataPipelineImplementation.swift#L145) for empty or blank userID.

Apps/VisionOS/VisionOS/APIViews/IdentifyView.swift
```diff
@@ -0,0 +1,95 @@
...
+    @EnvironmentObject private var viewModel: ViewModel
+
+    let onSuccess: (Profile) -> Void
+
+    var body: some View {
+        ScrollView(.vertical) {
+            VStack(alignment: .leading) {
+                Markdown {
+                    """
+                    Before you can track events, set devices attributes or update user traits, you must identify the user first.
```
ami-aman, line 49: Bit [outdated](https://customer.io/docs/sdk/ios/tracking/track-events/#anonymous-activity) after 3.x

Apps/VisionOS/VisionOS/APIViews/IdentifyView.swift
Ahmed-CIO: You wouldn't copy this code. This is for rendering on the screen and it is valid compilable string. If you see the test scenario in the video, you will see this code rendering in the code snippet

Apps/VisionOS/VisionOS/APIViews/IdentifyView.swift
Ahmed-CIO: it was meant to use the alternative, thank you for calling it out!

PR #671, opened 2024-03-25
test: replace async events in synchronous functions to reduce test flake

levibostian:
Closes: https://linear.app/customerio/issue/MBL-198/fix-flaky-tests-on-main

The test suite at the moment is trying to perform an async operation inside of a synchronous tearDown() function. I believe that this is the reason that some tests have been flaky on the CI lately. There is no guarantee the async operation will be completed by the time tearDown() is done executing. This may cause tearDown() to not properly reset the test suite state, but also may cause crashes when running tests if the async operation tries to access memory that no longer exists.

The goal of this change was to get tests passing again. I modified the async tearDown() operation to be a sync operation. To do that, I had to change the EventBus from an actor to a class. The EventBus remains to be thread safe after this change.

I went with this solution because it is small in scope and I am not yet sure that the EventBus should be an actor. With the little support our SDK has for swift concurrency, I think using an actor is causing more harm then good. As our codebase evolves to include more swift concurrency functionality, an actor may make sense again.

### DIFF
--- a/Sources/Common/Communication/EventBus.swift
+++ b/Sources/Common/Communication/EventBus.swift
@@ -26,7 +26,7 @@ public protocol EventBus: AutoMockable {
     /// Removes all observers from the EventBus.
     ///
     /// The function can be used for cleaning up or resetting the event handling system.
-    func removeAllObservers() async
+    func removeAllObservers()
 }
 
 /// EventBusObserversHolder is a private helper class used within SharedEventBus.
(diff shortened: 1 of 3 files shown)


Shahroz16 (commented):
If the goal is to just make the method sync, why can't we just make it `nonisolated`? we can mark this method under `debug` if needed. 

Changing EventBus from being an actor to back to utilizing `@Atomic` which then would use `DispatchQueue` internally seems like out of scope for this ticket and might require a longer conversation. 

I have created a PR and ran it multiple times and no test failed so far,
https://github.com/customerio/customerio-ios/pull/674

levibostian:
I made the decision to move EventBus back to a `class` because: 
1. Our code has been using `class` with `@Atomic` properties for a few years now and it's been stable for us. The change I made in this PR, I trust it like I do other classes in production today. 
2. `actor` is a tool. Like all tools, there are use cases where the tool makes sense to use and where it doesn't. I am not trying to say that the EventBus should never be an `actor`, but I am saying that an `actor` should be used correctly in the codebase. The EventBus being an `actor` in the code today is causing more harm then good, which is why I am OK with using a `class` for it again.

levibostian:
Closing in favor of https://github.com/customerio/customerio-ios/pull/674

PR #672, opened 2024-03-25
chore: remove unused script

levibostian:
This script is not being used anywhere in the code. This caused confusion for the team recently because we were not sure if we should update this script or update the `deploy-sdk` github actions workflow (which is where the deployment script actually lives).

This change deletes this file to remove confusion.


Shahroz16 (approved):
I am not sure if we would ever need to push cocoapods locally, but deleting this would mean pushing all modules ourselves via command rather than just running this 1 script.

levibostian:
> deleting this would mean pushing all modules ourselves via command rather than just running this 1 script.

At this time, this script is only causing confusion because of it not being used and it's out-dated. A script like this could create convenience over pushing each pod 1 by 1, yes, but I don't think keeping this script in the codebase for that reason is worth it.

PR #673, opened 2024-03-25
ci: print the binary size of the SDK

levibostian:

Modify the CI to print the binary size of the SDK files in a compiled sample app.

To view the SDK size report, you must view the CI logs. This was a decision the team made as the initial scope of the feature.

commit-id:7c521be9


Shahroz16 (commented):
> To view the SDK size report, you must view the CI logs. This was a decision the team made as the initial scope of the feature.

Thanks, Levi, in the Linear ticket, it doesn't say that but rather have mulitple options presented from which one can be opted for. I don't recall knowing this decision either. 

Because I feel opening up CI logs every time to check SDK size is going to be so tedious, would rather just appreciate a comment on PR.

Shahroz16 (approved):

PR #675, opened 2024-03-26
docs: VisionOS Sample App Readme

Ahmed-CIO:
Readme file for the VisionOS sample app

### DIFF
--- a/Apps/VisionOS/README.md
+++ b/Apps/VisionOS/README.md
@@ -32,7 +32,7 @@ To run the app, follow these steps:
 `MainScreen.swift` contains the executable code samples that the app's UI will use during interaction. Place breakpoints in the `switch` case within this file to inspect any action performed by the app.
 
 ### AppDelegate.swift
-The SDK is initialized when you enter the `CDP API key` and tap the `Initialize` button on the **Initialize** screen. In a production app, you would typically initialize the SDK in your `AppDelegate`. Therefore, you will find SDK initialization code in both `MainScreen.swift` and `AppDelegate.swift`.
+The SDK is initialized when you enter the `CDP API key` and tap the `Initialize` button on the **Initialize** screen. In a production app, you would typically initialize the SDK in your `AppDelegate`. Therefore, you will find SDK initialization code in both `MainScreen.swift` and `AppDelegate.swift`. Moreover, if you don't want to set the CDP API Key in the UI, you can hardcode it in the app delegate while you are exploring other APIs in the app.
 
 ### The UI
 The app features a navigation menu that allows you to move between different examples:
--- a/Apps/VisionOS/README.md
+++ b/Apps/VisionOS/README.md
@@ -1,0 +1,59 @@
+# VisionOS Sample App
+
+The VisionOS Sample App is a practical demonstration of how to integrate and utilize the Customer.io Data Pipelines iOS/Swift SDK.
+
+## Who Is This App For?
+
+This sample app is designed for VisionOS and iOS developers who are looking to get hands-on experience with Customer.io's code samples. It serves as a tool to validate, debug, or enhance their understanding of the different Customer.io APIs.
+
+## Run the App
+
+Before running the app, ensure you have one of the following prerequisites:
+* Apple Vision Pro simulator, which can be downloaded from Xcode.
+* An Apple Vision Pro device with developer mode enabled.
+
+To run the app, follow these steps:
+
+1. Clone the [customerio-ios](https://github.com/customerio/customerio-ios) repository by executing:
+   ```
+   git clone https://github.com/customerio/customerio-ios
+   ```
+2. Navigate to `<REPO_ROOT_DIR>/Apps/VisionOs` and open the `VisionOs.xcodeproj` file with Xcode.
+3. Ensure the `VisionOS` scheme is selected and set to run on either the simulator or your device.
+4. Click the run button in Xcode or press `CMD+R`.
+
+## How to Use the App
+
+### Prerequisites
+1. You must have a Customer.io account. Sign up for a free one if you haven't already.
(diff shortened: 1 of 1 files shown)


Apps/VisionOS/README.md
```diff
@@ -0,0 +1,59 @@
...
+* Apple Vision Pro simulator, which can be downloaded from Xcode.
+* An Apple Vision Pro device with developer mode enabled.
+
+To run the app, follow these steps:
+
+1. Clone the [customerio-ios](https://github.com/customerio/customerio-ios) repository by executing:
+   ```
+   git clone https://github.com/customerio/customerio-ios
+   ```
+2. Navigate to `<REPO_ROOT_DIR>/Apps/VisionOs` and open the `VisionOs.xcodeproj` file with Xcode.
```
ami-aman, line 21: Suggestion - Folder name is [VisionOS](https://github.com/customerio/customerio-ios/tree/main/Apps/VisionOS).

```suggestion
2. Navigate to `<REPO_ROOT_DIR>/Apps/VisionOS` and open the `VisionOS.xcodeproj` file with Xcode.
```

Apps/VisionOS/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+### Prerequisites
+1. You must have a Customer.io account. Sign up for a free one if you haven't already.
+2. While a full workspace setup is not required for this sample app, you do need to set up an iOS Data Pipeline source and use its CDP API key to initialize the SDK. [Learn more](https://customer.io/docs/sdk/ios/getting-started/auth/#set-up-a-new-source).
+
+### MainScreen.swift
+`MainScreen.swift` contains the executable code samples that the app's UI will use during interaction. Place breakpoints in the `switch` case within this file to inspect any action performed by the app.
+
+### AppDelegate.swift
+The SDK is initialized when you enter the `CDP API key` and tap the `Initialize` button on the **Initialize** screen. In a production app, you would typically initialize the SDK in your `AppDelegate`. Therefore, you will find SDK initialization code in both `MainScreen.swift` and `AppDelegate.swift`.
```
ami-aman, line 35: > Therefore, you will find SDK initialization code in both `MainScreen.swift` and `AppDelegate.swift`.

Does this mean that once you enter the CDP API Key, it is stored on the device and next time when someone re-launches the app, mainScreen automatically picks up the last set CDP API Key and initializes the SDK ? I am wondering, what is the use of having initialization at two places. Is it only for the demo or there's some difference in the way the initialization happens ?

Apps/VisionOS/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+- Detailed logs offering immediate feedback during development and debugging. For instance, try using the initialize functionality again after the first setup to observe the corresponding log message.
+- The `CustomerIO.shared.flush()` call circumvents network queue optimizations, allowing immediate visibility of your data in Customer.io data pipelines.
+
+### End-to-End Examples
+To fully grasp the integration, observe how data sent from the SDK appears in the data pipelines:
+
+1. Visit your [data pipeline dashboard](https://fly.customer.io/workspaces/sources).
+2. Select 'Sources' from the left menu and choose the source associated with the CDP API key mentioned in the prerequisites.
+3. Click on the `Data In` tab.
```
ami-aman, line 54: Wondering if we should also talk about `Data Out` since this is also a part of the workflow. In case the data in'd is not processed and passed to data out, I believe that is considered a problem either in data parsing or data received ! Thoughts ?

Apps/VisionOS/README.md
Ahmed-CIO: > Does this mean that once you enter the CDP API Key, it is stored on the device and next time when someone re-launches the app

Yes. Matter of convenience to avoid the need to re-enter it when you are trying to explore other functionalities

Apps/VisionOS/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+### Prerequisites
+1. You must have a Customer.io account. Sign up for a free one if you haven't already.
+2. While a full workspace setup is not required for this sample app, you do need to set up an iOS Data Pipeline source and use its CDP API key to initialize the SDK. [Learn more](https://customer.io/docs/sdk/ios/getting-started/auth/#set-up-a-new-source).
+
+### MainScreen.swift
+`MainScreen.swift` contains the executable code samples that the app's UI will use during interaction. Place breakpoints in the `switch` case within this file to inspect any action performed by the app.
+
+### AppDelegate.swift
+The SDK is initialized when you enter the `CDP API key` and tap the `Initialize` button on the **Initialize** screen. In a production app, you would typically initialize the SDK in your `AppDelegate`. Therefore, you will find SDK initialization code in both `MainScreen.swift` and `AppDelegate.swift`.
```
scotttwittrockcio, line 35: Since the most likely audience of this is a developer, should we allow them to set their CDP API Key via code? I think this might actually be easier than typing it via the "keyboard". Especially since the keyboard seems to be a bit of a pain with vision pro.

Apps/VisionOS/README.md
scotttwittrockcio: I thought something similar, if we think of this as a CDP SDK just talking about data in is fine. If we think of this as a Journeys SDK then we should talk about setting up the connection to Journeys. 

In this case I think we are ok to lean towards it being a CDP SDK. We have other info and docs about setting up that Data Pipelines to Journeys connection, and by not mentioning that here, it means one less place to keep in sync as things change.

Apps/VisionOS/README.md
Ahmed-CIO: They definitely can. If they want to, they will set it up in the app delegate as a hardcoded value instead of the value loaded from storage and based on the UI.

PR #677, opened 2024-03-28
feat: privacy manifest files

levibostian:
Closes: https://linear.app/customerio/issue/MBL-187/update-values-based-on-linear-doc
Part of: https://linear.app/customerio/issue/MBL-189/release-ios

Create privacy manifest files for all target modules that customers could decide to install in an app. Update package manager config files to include these new resource files.

Testing criteria CocoaPods:
* `pod update`, opened app in xcode
* Verified that in 'Pods' project settings, there are new targets added for each module (example: `CustomerIODataPipelines-CustomerIODataPipelines_Privacy`). I then verified in each of the `_Privacy targets > Build phases > compile sources` list, the privacy manifest files were listed and they resource to a valid path.
* App compiled without any new warnings.

Testing criteria SPM:
* Opened sample app in xcode.
* Fixed all warnings xcode gave me while resolving the CIO package.
* Compiled app without any new warnings.


levibostian:
Notes for reviewers:
* The values in the manifest files came from internal docs we created. See [ticket](https://linear.app/customerio/issue/MBL-122/apple-privacy-requirements-for-sdks-by-spring-2024) for link to the docs. 
* See description for testing criteria on how I validated this.

Sources/DataPipeline/Resources/PrivacyInfo.xcprivacy
```diff
@@ -0,0 +1,79 @@
...
+	<array>
+	
+	  <!-- 
+	  Section for SDKs that capture "Such as screen name, handle, account ID, assigned user ID, customer number, or other user- or account-level ID that can be used to identify a particular user or account."
+	  -->
+		<dict>
+			<key>NSPrivacyCollectedDataType</key>
+			<string>NSPrivacyCollectedDataTypeUserID</string>
+			<key>NSPrivacyCollectedDataTypeLinked</key>
+			<false/> <!-- because of anonymous profiles, linking to a person is optional -->
```
Shahroz16, line 22: added a comment in notion about it

Sources/DataPipeline/Resources/PrivacyInfo.xcprivacy
```diff
@@ -0,0 +1,79 @@
...
+	This section is describing what information our SDK collects about the app user. 
+	
+	It's my understanding that this section is for what our SDK requires, not what a customer can do with our SDK. Therefore, if something is optional, use `false`.
+	Customers must create their own privacy file for what their app does and provide that to Apple. That's where customers report what they might be using our SDK for. 
+   -->   
+	<key>NSPrivacyCollectedDataTypes</key>
+	<array>
+	
+	  <!-- 
+	  Section for SDKs that capture "Such as screen name, handle, account ID, assigned user ID, customer number, or other user- or account-level ID that can be used to identify a particular user or account."
```
Shahroz16, line 16: do we have to specify the things we capture, like app name etc?

levibostian: No, you do not need to specify what you capture. Instead specify what categories that you capture.

levibostian: I may have misunderstood your question. 

If you're asking if app name needs to be called out, [if you see a category in this apple doc](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_data_use_in_privacy_manifests) that is missing in this file, please mention the missing category in our notion doc.

Sources/DataPipeline/Resources/PrivacyInfo.xcprivacy
```diff
@@ -0,0 +1,79 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<!--
+	Docs for this section: https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_data_use_in_privacy_manifests
+	This section is describing what information our SDK collects about the app user. 
+	
+	It's my understanding that this section is for what our SDK requires, not what a customer can do with our SDK. Therefore, if something is optional, use `false`.
+	Customers must create their own privacy file for what their app does and provide that to Apple. That's where customers report what they might be using our SDK for.
```
scotttwittrockcio, line 10: Customers do not need to declare their dependancies privacy requirements based on the second line in this call out at the top of [this page](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_data_use_in_privacy_manifests).
image

Sources/DataPipeline/Resources/PrivacyInfo.xcprivacy
levibostian: I am keeping this value as-is until we give others on team time to read documentation and give suggestions for this value.

Sources/DataPipeline/Resources/PrivacyInfo.xcprivacy
levibostian: Re-reviewing the comments in this file and the documentation you linked to, I think the comments in this file are up-to-date and accurate. 

Do you have a suggestion of a change to make to these comments?

Sources/DataPipeline/Resources/PrivacyInfo.xcprivacy
scotttwittrockcio: Since to utilize basically any functionality of Customer.io, we need to link usage data to a user typically email or a device. This is the rational for marking this as true. If we mark this as false, it means that Customer.io does not have any data linked to a user anywhere in our systems.

Sources/DataPipeline/Resources/PrivacyInfo.xcprivacy
```diff
@@ -0,0 +1,79 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<!--
+	Docs for this section: https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_data_use_in_privacy_manifests
+	This section is describing what information our SDK collects about the app user. 
+	
+	It's my understanding that this section is for what our SDK requires, not what a customer can do with our SDK. Therefore, if something is optional, use `false`.
```
scotttwittrockcio, line 9:
```suggestion
	It's my understanding that this section is for what our SDK collects automatically, not the data a customer passes to our SDK. Therefore, if something is optional, use `false`.
```

Sources/DataPipeline/Resources/PrivacyInfo.xcprivacy
scotttwittrockcio: I think the issue is around the interpretation of "required". Since their app is collecting an email address and passing it to us, they would have to declare that. If we ever created a UI to collect the email and link to a user then we'd have to declare the email address. Therefore in the first case email isn't a "required" part of our SDK. 

So this does work, but the wording made that tough to interpret. I provided a suggestion to the line that might make that more clear.

scotttwittrockcio, line 10:
```suggestion
	Customers must create their own privacy file for what their app collects and provide that to Apple. That's where customers report what data they are passing to our SDK. 
```

Sources/DataPipeline/Resources/PrivacyInfo.xcprivacy
mrehan27: Added comment on notion too. We only need a profile to function, so even if they use aliases to create profiles, sending push notifications and in-app messages should still be possible based on the data linked to the user, which I believe aligns with [Apple's privacy details](https://developer.apple.com/app-store/app-privacy-details/#linked-data). For sending email, it's true that linking an actual email is necessary, but we don’t force it and it likely falls under the customer app's privacy? 🤔

Sources/DataPipeline/Resources/PrivacyInfo.xcprivacy
scotttwittrockcio: I'm open to skipping adding any custom data types for now. Apples documents aren't always the most clear, so it will be tough to determine what qualifies as a custom data type that we might need to declare.

Sources/DataPipeline/Resources/PrivacyInfo.xcprivacy
levibostian: This PR thread has been discussed in a Notion comment. So, I'll resolve it here. 

I suggest reviewing [this updated section in Notion](https://www.notion.so/custio/Apple-SDK-Privacy-Settings-8a36323e410b44028ebe6ffb1fd6fe30?pvs=4#801bf0270c994aa1bcd7903b9571dcdb) and comment there if you have more feedback.

kvdesa:
Any updates on when this will be released?

levibostian:
@kvdesa Unfortunately, we do not have a target date when to expect this feature. Know that it's a high priority of our team and we're working to release it as soon as we are able.

PR #680, opened 2024-03-28
chore: Use actual os name for the user agent

Ahmed-CIO:
The user agent is currently hard-coded to use iOS as the os name. I changed it to pick the value from the `deviceInfo` object and default to `iOS`

### DIFF
--- a/Sources/Common/Service/UserAgentUtil.swift
+++ b/Sources/Common/Service/UserAgentUtil.swift
@@ -23,7 +23,7 @@ public class UserAgentUtilImpl: UserAgentUtil {
      * `Customer.io iOS Client/1.0.0-alpha.16`
      */
     public func getUserAgentHeaderValue() -> String {
-        var userAgent = "Customer.io iOS Client/\(deviceInfo.sdkVersion)"
+        var userAgent = "Customer.io \(deviceInfo.osName ?? "iOS" ) Client/\(deviceInfo.sdkVersion)"
 
         if let deviceModel = deviceInfo.deviceModel,
            let deviceOsVersion = deviceInfo.osVersion,
--- a/Sources/Common/Service/UserAgentUtil.swift
+++ b/Sources/Common/Service/UserAgentUtil.swift
@@ -23,7 +23,7 @@ public class UserAgentUtilImpl: UserAgentUtil {
      * `Customer.io iOS Client/1.0.0-alpha.16`
      */
     public func getUserAgentHeaderValue() -> String {
-        var userAgent = "Customer.io \(deviceInfo.osName ?? "iOS" ) Client/\(deviceInfo.sdkVersion)"
+        var userAgent = "Customer.io \(deviceInfo.osName ?? "iOS") Client/\(deviceInfo.sdkVersion)"
 
         if let deviceModel = deviceInfo.deviceModel,
            let deviceOsVersion = deviceInfo.osVersion,
--- a/Tests/Common/Service/UserAgentUtilTest.swift
+++ b/Tests/Common/Service/UserAgentUtilTest.swift
@@ -39,4 +39,19 @@ class UserAgentUtilTest: UnitTest {
 
         XCTAssertEqual(expected, actual)
     }
+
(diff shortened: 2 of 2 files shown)


Sources/Common/Service/UserAgentUtil.swift
```diff
@@ -23,7 +23,7 @@ public class UserAgentUtilImpl: UserAgentUtil {
      * `Customer.io iOS Client/1.0.0-alpha.16`
      */
     public func getUserAgentHeaderValue() -> String {
-        var userAgent = "Customer.io iOS Client/\(deviceInfo.sdkVersion)"
+        var userAgent = "Customer.io \(deviceInfo.osName ?? "iOS" ) Client/\(deviceInfo.sdkVersion)"
```
Shahroz16, line 26: Lets please add test case for this, because our mix-panel tracking is checking for keyword "Customer.io iOS" and so we should make sure it doesn't break

Shahroz16 (changes requested):

Sources/Common/Service/UserAgentUtil.swift
Ahmed-CIO: Good call out. I will add additional test case for `visionos`. However, this won't make the backend work for `visionos`. I have separate PRs for tha backend.

Sources/Common/Service/UserAgentUtil.swift
```diff
@@ -23,7 +23,7 @@ public class UserAgentUtilImpl: UserAgentUtil {
      * `Customer.io iOS Client/1.0.0-alpha.16`
      */
     public func getUserAgentHeaderValue() -> String {
-        var userAgent = "Customer.io iOS Client/\(deviceInfo.sdkVersion)"
+        var userAgent = "Customer.io \(deviceInfo.osName ?? "iOS") Client/\(deviceInfo.sdkVersion)"
```
Shahroz16, line 26: I haven't checked the documentation yet, but we need to be certain the only possible values from `osName` are `iOS` and `visionOS` otherwise our metrics can be skewed.

Shahroz16 (approved):

Tests/Common/Service/UserAgentUtilTest.swift
```diff
@@ -39,4 +39,19 @@ class UserAgentUtilTest: UnitTest {
 
         XCTAssertEqual(expected, actual)
     }
+
+    func test_getUserAgent_givenDeviceInfoAvailable_expectLongVisonOsUserAgent() {
```
levibostian, line 43: I think this test is a duplicate of the `test_getUserAgent_givenDeviceInfoAvailable_expectLongUserAgent` test. Just with different mock values, which I don't think adds much value to the suite. 

Both test functions could be renamed to be more accurate of what logic is under test: *given device info values, expect the user agent is generated in the format that we expect*. 

If others think this is a valuable test, I am OK with it.

Sources/Common/Service/UserAgentUtil.swift
levibostian: [Docs are pretty generic](https://developer.apple.com/documentation/uikit/uidevice/1620054-systemname) unfortunately. 

I think the only way we can verify `deviceInfo.osName` is with QA testing. Print `deviceInfo.osName` to the console in iOS sample app and then in visionOS app and see what value gets printed.

levibostian (approved):
Assuming QA tests pass where the user agent string is what we expect, I approve.

Sources/Common/Service/UserAgentUtil.swift
Ahmed-CIO: We can also use some constraints that ensures the `.osName` always returns a valid value. Realistically speaking though, if we are restricting our packages to iOS and VisionOS only, I don't see scenario where a third value comes in.

Tests/Common/Service/UserAgentUtilTest.swift
Ahmed-CIO: I agree. My only justification for not changing it is that we only expect two variations. Keeping them explicit like this is easier to understand. If we introduce more, then it is much more justifiable to improve the test structure

PR #681, opened 2024-03-29
chore: update to latest analytics fork

Shahroz16:
closes: https://linear.app/customerio/issue/MBL-223/update-segment-fork

The latest fork contains the Apple privacy changes along with other improvements added that can be found here
https://github.com/customerio/cdp-analytics-swift/compare/1.5.5-cio.1...1.5.9+cio.1

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -45,7 +45,7 @@ let package = Package(
         .package(name: "Firebase", url: "https://github.com/firebase/firebase-ios-sdk.git", "8.7.0"..<"11.0.0"),
 
         // The version is a git commit hash. Make sure the commit is the same as what the DataPipelines CocoaPods is using.
-        .package(name: "Segment", url: "https://github.com/customerio/cdp-analytics-swift.git", .exact("1.5.5-cio.1"))
+        .package(name: "Segment", url: "https://github.com/customerio/cdp-analytics-swift.git", .exact("1.5.9+cio.1"))
     ],
     targets: [ 
         // Common - Code used by multiple modules in the SDK project.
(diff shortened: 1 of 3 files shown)


Package.swift
```diff
@@ -45,7 +45,7 @@ let package = Package(
         .package(name: "Firebase", url: "https://github.com/firebase/firebase-ios-sdk.git", "8.7.0"..<"11.0.0"),
 
         // The version is a git commit hash. Make sure the commit is the same as what the DataPipelines CocoaPods is using.
-        .package(name: "Segment", url: "https://github.com/customerio/cdp-analytics-swift.git", .exact("1.5.5-cio.1"))
+        .package(name: "Segment", url: "https://github.com/customerio/cdp-analytics-swift.git", .exact("1.5.9+cio.1"))
```
mrehan27, line 48: Shouldn't it be the same as in podspec?

```suggestion
        .package(name: "Segment", url: "https://github.com/customerio/cdp-analytics-swift.git", .exact("1.5.9+cio.2"))
```

levibostian (changes requested):
I think this PR goes beyond the scope of the ticket. 

The scope of the ticket is to update the segment fork to include the privacy manifest files. Seeing the updates to the .podspec and Package.swift files, those look good to me. 

The rest of the changes go beyond this scope and I don't think are required for the ticket?

Shahroz16:
> I think this PR goes beyond the scope of the ticket.
> 
> The scope of the ticket is to update the segment fork to include the privacy manifest files. Seeing the updates to the .podspec and Package.swift files, those look good to me.
> 
> The rest of the changes go beyond this scope and I don't think are required for the ticket?

Correct !!! they were added by mistake, reverting those.

Tests/DataPipeline/DataPipelineCompatibilityTests.swift
levibostian, line 1: Could you share the reason for this change?

Shahroz16: Storage access methods were updated in the segment, which we utilize in our testing and hence had to make those changes accordingly.

levibostian (approved):

PR #684, opened 2024-04-01
chore: Support visionOS 1.0

Ahmed-CIO:
We have APIs, mainly for push, that's restricted to iOS 13.0 and above. They can safely be enabled for VisionOS 1.0 as well.

## Test Plan
Add NSE to a VisionOS app and make sure it builds and works as expected when using rich push


levibostian (commented):
The change looks good. Before I approve, could you share testing criteria for this PR?

Ahmed-CIO:
Fair point, added.

Shahroz16 (approved):
Looks good, I haven't tested it myself but since its not taking away any of the available platform but rather adding another, so  I am good with it.

PR #690, opened 2024-04-04
chore: Overriding device.type for VisionOS

Ahmed-CIO:
The push notification logic won't work if the device.type isn't ios. This PR is a mitigation for now to enable push notifications on VisionOS till we have a better handling on the backend


Shahroz16 (approved):
Awesome, one funny thing, we had a test to validate `type` being sent is `ios` hence when you run the tests suite on the Vision Pro simulator, it would fail. But with this change it doesn't break verifying that it indeed is now converting the `device.type` to `ios`. 😄 

Screenshot 2024-04-05 at 4 47 15 AM

PR #691, opened 2024-04-05
ci: cache sample app builds on the main branch

levibostian:

In order for us to perform SDK binary size changes in each PR, each PR needs to know the SDK binary size of the main branch.

This commit sets up the CI to cache sample app builds for the main branch. In future commits, PR builds can then download this cache and generate SDK binary size comparison reports.

### DIFF
--- a/.github/workflows/build-sample-apps.yml
+++ b/.github/workflows/build-sample-apps.yml
@@ -72,6 +72,7 @@ jobs:
       run: |
         brew install sd # used in CI script as an easier to use sed CLI. Replaces text in files. 
         brew install xcbeautify # used by fastlane for output 
+        brew install bloaty # used to generate SDK binary size reports 
     
     - name: Install tools from Gemfile (ruby language) used for building our apps with 
       uses: ruby/setup-ruby@v1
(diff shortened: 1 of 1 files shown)


levibostian:
Unfortunately, we will not be able to 100% validate this works until we merge into `main` since some of the CI config files have `if` statements checking for `main` branch. 

I did test this code on the CI before opening this PR so the CI config syntax should all be correct.

.github/workflows/build-sample-apps.yml
```diff
@@ -124,11 +125,10 @@ jobs:
...
 
-    - name: Setup to determine binary size of SDK
-      working-directory: Apps/${{ matrix.sample-app }}/build/    
-      run: |
-        brew install bloaty
-        mv *.xcarchive App.xcarchive # rename the file to a static value. Bloaty requires a hard-coded path to files to work. 
+    # xcodebuild creates builds that include a timestamp in the name. In order for bloaty to read the build, we need to rename it to a static name.
+    - name: Rename the same app build to a static name that bloaty can understand  
+      working-directory: Apps/${{ matrix.sample-app }}/build/
+      run: mv *.xcarchive App.xcarchive
```
Shahroz16, line 131: can there be multiple `.xcarchive` generated? or is it only one with timestamp in the name?

Shahroz16 (approved):

.github/workflows/build-sample-apps.yml
levibostian: A new `.xcarchive` directory gets generated each time we call `fastlane ios build`. Which, we only run once in this job. 

If you're concerned about what would happen if there are multiple `.xcarchive` directories in the `build/` directory, from my testing, the `mv` command fails if multiple `.xcarchive` directories are found.

PR #693, opened 2024-04-09
fix: added alias for Region

Shahroz16:
closes: [MBL-201 : \[iOS-CDP\] Internal Region enum](https://linear.app/customerio/issue/MBL-201/[ios-cdp]-internal-region-enum)

- Added class in `Alias`
- Updated public test to decouple data pipeline public APIs with Common


Tests/DataPipeline/APITest.swift
```diff
@@ -1,5 +1,4 @@
 import CioDataPipelines // do not use `@testable` so we can test functions are made public and not `internal`.
-import CioInternalCommon
```
Shahroz16, line 2: now we can find out missing aliases this way as well

Tests/DataPipeline/APITest.swift
```diff
@@ -34,52 +30,37 @@ class DataPipelineAPITest: UnitTest {
...
         CustomerIO.shared.trackMetric(deliveryID: "", event: metric, deviceToken: "")
-        mock.trackMetric(deliveryID: "", event: metric, deviceToken: "")
 
         // profile attributes
         CustomerIO.shared.profileAttributes = dictionaryData
-        mock.profileAttributes = dictionaryData
 
         // device attributes
         CustomerIO.shared.deviceAttributes = dictionaryData
-        mock.deviceAttributes = dictionaryData
```
ami-aman, line 82: why all mocks have been removed ?

Tests/DataPipeline/APITest.swift
ami-aman: > this way as well
Thanks for adding the comment but wondering what "this way" is ? 🤔

Tests/DataPipeline/APITest.swift
ami-aman: I see. It is because you moved all mocks to another method above

Tests/DataPipeline/APITest.swift
Shahroz16: By removing the `CioCommon` from the test class. It would give an error if we utilize something from common but isn't part of alias

Tests/DataPipeline/APITest.swift
Shahroz16: Yup they required Common import

ami-aman (approved):

Tests/DataPipeline/APITest.swift
```diff
@@ -110,6 +91,7 @@ class DataPipelineAPITest: UnitTest {
         try skipRunningTest()
 
         _ = SDKConfigBuilder(cdpApiKey: .random)
+            .region(.US)
```
mrehan27, line 94: I think it would be better to use `Region.US` instead as `.US` doesn't require import while `Region.US` will ask for import if not already there 🤔

mrehan27 (approved):

Tests/DataPipeline/APITest.swift
Shahroz16: > instead as .US doesn't require import while Region.US will ask for import

are you sure about this? i would have thought both are the same thing? and would require import either way? 🤔

mrehan27: Discussed on slack, it does require import for `Region.US`

levibostian: I fixed this same bug a long time ago where Region was not being exposed in the Track module. 

If I remember correctly, this is the behavior that I experienced in APITest class when I fixed that bug. 

```swift
public class APITest {
  // Before I added the `Region` typelias...
  _ = SDKConfigBuilder(cdpApiKey: .random)
    .region(.US) // This would compile just fine. 
  let _: Region = .US // But this would not compile. 

  // After I added the `Region` typealias, the let statement would compile 🤷‍♀️
}
```

So, Rehan is on the right track here. But I do think the only way we can verify if `Region` is truly accessible in the public API is to try and directly reference it [like this APITest is doing](https://github.com/customerio/customerio-ios/blob/33861dbf6d2b465567472ba38e8b78aff377baca/Tests/Common/APITest.swift#L34-L36), not use the Builder function parameter.

PR #694, opened 2024-04-09
ci: generate SDK binary size diff between PRs and main branch

levibostian:

When on a PR, run bloaty to generate a SDK binary size diff report between main branch and PR. This diff report will tell us if the SDK binary size will increases, decreases, or stays the same if the PR gets merged into main.

In order to view the size report, you must view the CI logs. We plan to make the reports more convenient to view by creating PR comments with these reports. I plan to add the PR comments feature in another PR.

---

**Stack**:
- #699
- #697
- #696
- #695
- #694 ⬅

### DIFF
--- a/.github/workflows/build-sample-apps.yml
+++ b/.github/workflows/build-sample-apps.yml
@@ -124,14 +124,19 @@ jobs:
       env: 
         GOOGLE_CLOUD_MATCH_READONLY_SERVICE_ACCOUNT_B64: ${{ secrets.GOOGLE_CLOUD_MATCH_READONLY_SERVICE_ACCOUNT_B64 }}
         FIREBASE_APP_DISTRIBUTION_SERVICE_ACCOUNT_CREDS_B64: ${{ secrets.FIREBASE_APP_DISTRIBUTION_SERVICE_ACCOUNT_CREDS_B64 }}
-
+    
     # xcodebuild creates builds that include a timestamp in the name. In order for bloaty to read the build, we need to rename it to a static name.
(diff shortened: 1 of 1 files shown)


levibostian:
# Testing criteria 

Looking at the CI logs for this PR, we can see that the CI successfully downloaded the latest `main` branch build and ran bloaty to generate a size diff: 

CleanShot 2024-04-09 at 15 04 06@2x

The generated report looks accurate to me so testing is successful. 

### The report shows 3 modified files though? 

The report shows that 3 files have been modified in this PR compared to the `main` branch. However, you can see that this PR has not modified any of the source code of the SDK! 

When the CI builds the sample apps, [it updates the SDK's `Version.swift` file](https://github.com/customerio/customerio-ios/blob/5503f738da00a9d320bb01a43512f35085ecee31/Apps/fastlane/Fastfile#L46-L49). The `Version.swift` file in the `main` and PR sample app builds are therefore unique. 

Because of the `Version.swift` file being different, the diff report says that the SDK binary size is 32 bytes different between the PR and `main`. 32 bytes is a diff that would not stop us from merging a PR. 

I do not consider this report inaccuracy to be a blocker for this ticket. This 3 file inaccuracy is due to us trying to reduce the number of sample app builds we need to run on the CI. I think CI performance is more important then a SDK binary size diff that is perfectly accurate.

ami-aman (approved):

ami-aman:
Thanks for explaining the reason behind 32 bytes difference.

levibostian:
✓ Commit merged in pull request [#697](https://github.com/customerio/customerio-ios/pull/697)

PR #695, opened 2024-04-10
ci: show SDK binary size reports in PRs as comments for convenient viewing

levibostian:

This commit saves all of the bloaty SDK size reports and then shows them in a PR comment. This makes the reports much more convenient for viewing.

Because the reports can be very long in size, I used markdown collapse/expand sections to hide the SDK reports in the PR unless you expand the report.

Testing criteria:
* You should see a PR comment created inside of this PR which shows 3 SDK size reports.

---

**Stack**:
- #699
- #697
- #696
- #695 ⬅
- #694

### DIFF
--- a/.github/workflows/build-sample-apps.yml
+++ b/.github/workflows/build-sample-apps.yml
@@ -136,23 +136,35 @@ jobs:
 
     # We only need to run bloaty on 1 app. Therefore, we are hard-coding APN-UIKit into these steps below.
 
-    - name: Print the binary size of our SDK in an app, minus dependencies
+    - name: Get the binary size of our SDK in an app, minus dependencies
+      id: sdk-size-report-minus-dependencies
       if: ${{ matrix.sample-app == 'APN-UIKit' }}
       working-directory: Apps/${{ matrix.sample-app }}/build/
       run: |
-        bloaty --source-filter ".*(customerio-ios\/Sources).*" \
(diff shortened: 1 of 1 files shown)


.github/workflows/build-sample-apps.yml
```diff
@@ -223,11 +235,62 @@ jobs:
...
+      if: github.event_name == 'pull_request' && matrix.sample-app == 'APN-UIKit'
+      uses: peter-evans/find-comment@v3
+      id: find-sdk-size-report-comment
+      with:
+        issue-number: ${{ github.event.pull_request.number }}
+        comment-author: 'github-actions[bot]'
+        body-includes: <!-- sdk size reports -->
+
+    - name: Send SDK size reports to the PR for convenient viewing 
+      if: github.event_name == 'pull_request' && matrix.sample-app == 'APN-UIKit'
```
ami-aman, line 261: I am wondering that we have hard-coded APN-UIKit as the reference sample app but do we expect the same SDK size reports in FCM sample app ? My question is more around did we do at least one time size check for SDK with FCM implementation? Am I making sense ? 😅

levibostian: Great question. I have generated size reports for the FCM app. Because of the way cocoapods works compared to SPM apps, I found generating the reports to be more work with the FCM app. 

I'll make a comment inside of the Linear ticket talking about this. It would be good for the rest of the team to understand this, too.

levibostian:
✓ Commit merged in pull request [#697](https://github.com/customerio/customerio-ios/pull/697)

PR #702, opened 2024-04-15
fix: track bq migration for identify profile events with no attributes

levibostian:
Part of: https://linear.app/customerio/issue/MBL-251/track-bq-cdp-migration-causes-crash-in-ios-301-for-identifying-a

If there is an event in the Track BQ for identifying a profile with no profile attributes, the CDP migration could cause a crash. This commit fixes this crash by performing a safe optional unwrap.

Testing consideration:
* Added new unit test reproducing the issue.

---

**Stack**:
- #704
- #702 ⬅

### DIFF
--- a/Sources/Migration/DataPipelineMigrationAssistant.swift
+++ b/Sources/Migration/DataPipelineMigrationAssistant.swift
@@ -108,15 +108,15 @@ public class DataPipelineMigrationAssistant {
         guard let trackTaskData: IdentifyProfileQueueTaskData = jsonAdapter.fromJson(taskData) else {
             return false
         }
-        if let attributedString = trackTaskData.attributesJsonString, attributedString.contains("null") {
+
+        // If there are no profile attributes or profile attributes not in a valid format, JSON adapter will return nil and we will perform a migration without the profile attributes.
+        guard let profileAttributesString: String = trackTaskData.attributesJsonString, let profileAttributes: [String: Any] = jsonAdapter.fromJsonString(profileAttributesString) else {
             migrationHandler.processIdentifyFromBGQ(identifier: trackTaskData.identifier, timestamp: timestamp, body: nil)
             return true
         }
-        guard let profileAttributes: [String: Any] = jsonAdapter.fromJsonString(trackTaskData.attributesJsonString!) else {
-            migrationHandler.processIdentifyFromBGQ(identifier: trackTaskData.identifier, timestamp: timestamp, body: nil)
-            return true
-        }
+
         migrationHandler.processIdentifyFromBGQ(identifier: trackTaskData.identifier, timestamp: timestamp, body: profileAttributes)
+
         return true
     }
 
--- a/Tests/Migration/DataPipelineMigrationAssistantTests.swift
+++ b/Tests/Migration/DataPipelineMigrationAssistantTests.swift
@@ -139,12 +139,30 @@ class DataPipelineMigrationAssistantTests: UnitTest {
 
     // MARK: getAndProcessTask
 
-    func test_givenIdentifyProfileWithoutBody_expectTaskRunAndProcessedDeleted() {
+    func test_givenIdentifyProfileWithEmptyBody_expectTaskRunAndProcessedDeleted() {
         let givenType = QueueTaskType.identifyProfile
 
         let givenCreatedTask = QueueTaskMetadata.random
 
         let trackDeliveryMetricData = IdentifyProfileQueueTaskData(identifier: String.random, attributesJsonString: "")
+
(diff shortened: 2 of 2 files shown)


Sources/Migration/DataPipelineMigrationAssistant.swift
Shahroz16, line 1: could be considered scope creep but i think it would make sense, to cater this as well as its the same issue and can cause the same crash? also would save us from doing another release. 
```
jsonAdapter.fromJsonString(registerPushTaskData.attributesJsonString!)
```

Shahroz16 (commented):
Looks good, added 1 ask to also cater

Sources/Migration/DataPipelineMigrationAssistant.swift
mrehan27: Good call, I think we should handle all these cases in this file instead of waiting for issues to be reported. I'm okay if we don't add tests for them in this PR, but I would like to fix possible issues with similar cases in this file before releasing.

Sources/Migration/DataPipelineMigrationAssistant.swift
```diff
@@ -108,15 +108,15 @@ public class DataPipelineMigrationAssistant {
...
             return false
         }
-        if let attributedString = trackTaskData.attributesJsonString, attributedString.contains("null") {
-            migrationHandler.processIdentifyFromBGQ(identifier: trackTaskData.identifier, timestamp: timestamp, body: nil)
-            return true
-        }
-        guard let profileAttributes: [String: Any] = jsonAdapter.fromJsonString(trackTaskData.attributesJsonString!) else {
+
+        // If there are no profile attributes or profile attributes not in a valid format, JSON adapter will return nil and we will perform a migration without the profile attributes.
+        guard let profileAttributes: [String: Any] = jsonAdapter.fromJsonString(trackTaskData.attributesJsonString ?? "") else {
```
mrehan27, line 113: Wouldn't it be better if we don't even try to parse if String value is `nil`?

```suggestion
        guard let profileAttributesString: String = trackTaskData.attributesJsonString, let profileAttributes: [String: Any] = jsonAdapter.fromJsonString(profileAttributesString) else {
```

mrehan27 (approved):

ami-aman (approved):

Sources/Migration/DataPipelineMigrationAssistant.swift
```diff
@@ -108,15 +108,15 @@ public class DataPipelineMigrationAssistant {
         guard let trackTaskData: IdentifyProfileQueueTaskData = jsonAdapter.fromJson(taskData) else {
             return false
         }
-        if let attributedString = trackTaskData.attributesJsonString, attributedString.contains("null") {
```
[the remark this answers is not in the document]
levibostian: Thanks for sharing your expertise here. And thanks for sharing the steps. 

I *think* the reason you're seeing the value "null" is a behavior of our sample app's implementation. 

In the APN sample app, [we are only calling](https://github.com/customerio/customerio-ios/blob/7aa128c1e8031ad9388937c109ea71d432096245/Apps/APN-UIKit/APN%20UIKit/View/LoginViewController.swift#L95) `CustomerIO.identify(id: "", body: "")`. If the sample app was modified to call `CustomerIO.identify(id: "")`, I would no longer expect "null" to be the value of `attributes_json_string`. 

Therefore, I think this bug is happening for customers who are calling the `identify` function without the `body` param.

Sources/Migration/DataPipelineMigrationAssistant.swift
levibostian: I originally had that code but wasn't sure if it was necessary. It felt redundant when the json adapter returns an optional result if parsing is not successful. 

It could increase performance if we skip json parsing. I would be OK with this suggestion.

Sources/Migration/DataPipelineMigrationAssistant.swift
levibostian: I considered that, too. I don't anticipate that the register device BQ event will ever have a null attributes string because a device token should always be included in the attributes. 

However, as we chatted about yesterday, we want to avoid force optional unwraps in the codebase. So, this extra bit of safety could be a good thing. 

I do agree, I would prefer to ship this as far of this fix, too. I'll add a new PR to the stack with this change and we can ship them both together.

Sources/Migration/DataPipelineMigrationAssistant.swift
levibostian: I committed the suggestion.

levibostian:
✓ Commit merged in pull request [#704](https://github.com/customerio/customerio-ios/pull/704)

PR #706, opened 2024-04-22
fix: apple privacy mission permission for extension

Shahroz16:
closes: https://linear.app/customerio/issue/MBL-263/missing-privacy-manifest-for-notificationservice-extension

adds `1C8F.1`

> Declare this reason to access user defaults to read and write information that is only accessible to the apps, app extensions, and App Clips that are members of the same App Group as the app itself.


Shahroz16:
closing until we hear more about it from.

PR #708, opened 2024-04-22
fix: update segment internal SPM dependencies with fixed version

Shahroz16:
in the new Segment release, the internal dependencies (sovran, JSON encoder) versions are fixed.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -44,7 +44,7 @@ let package = Package(
         // Update to exact version until wrapper SDKs become part of testing pipeline.
         .package(name: "Firebase", url: "https://github.com/firebase/firebase-ios-sdk.git", "8.7.0"..<"11.0.0"),
 
-        // The version is a git commit hash. Make sure the commit is the same as what the DataPipelines CocoaPods is using.
+        // Make sure the version number is same for DataPipelines cocoapods.
         .package(name: "Segment", url: "https://github.com/customerio/cdp-analytics-swift.git", .exact("1.5.11+cio.1"))
     ],
     targets: [ 
--- a/Package.swift
+++ b/Package.swift
@@ -45,7 +45,7 @@ let package = Package(
         .package(name: "Firebase", url: "https://github.com/firebase/firebase-ios-sdk.git", "8.7.0"..<"11.0.0"),
 
         // The version is a git commit hash. Make sure the commit is the same as what the DataPipelines CocoaPods is using.
-        .package(name: "Segment", url: "https://github.com/customerio/cdp-analytics-swift.git", .exact("1.5.9+cio.2"))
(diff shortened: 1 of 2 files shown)


Shahroz16:
*Note* This needs to go in before https://github.com/customerio/customerio-ios/pull/707. 

In-case we think https://github.com/customerio/customerio-ios/pull/707 isn't going, we need to update the title to `fix`.

Package.swift
```diff
@@ -45,7 +45,7 @@ let package = Package(
         .package(name: "Firebase", url: "https://github.com/firebase/firebase-ios-sdk.git", "8.7.0"..<"11.0.0"),
 
         // The version is a git commit hash. Make sure the commit is the same as what the DataPipelines CocoaPods is using.
```
mrehan27, line 47: I just noticed this comment needs to be updated 😅

Package.swift
```diff
@@ -45,7 +45,7 @@ let package = Package(
...
-        .package(name: "Segment", url: "https://github.com/customerio/cdp-analytics-swift.git", .exact("1.5.9+cio.1"))
+        .package(name: "Segment", url: "https://github.com/customerio/cdp-analytics-swift.git", .exact("1.5.9+cio.2"))
```
mrehan27, line 48: We don't need the same change for pods, correct?

mrehan27 (approved):

Package.swift
levibostian: I think we do need to update the podspec files in this PR, too. 

I say that because in version 1.5.9+cio.2 of our fork, the Package.swift file was modified to hard-code dependencies. But .podspec file in 1.5.9+cio.2 [does not hard-code the dependencies versions](https://github.com/customerio/cdp-analytics-swift/blob/3b96fe7d92f498d113da97042a066fa9871d9edf/AnalyticsSwiftCIO.podspec#L24-L25). 

I think what needs to happen is:
* Our iOS SDK uses a hard-coded version of our fork in Package.swift and .podspec files. 
* Our fork uses hard-coded versions of dependencies in both Package.swift and .podspec

levibostian (changes requested):
[1 blocker conversation](https://github.com/customerio/customerio-ios/pull/708/files#r1574645675) to resolve

Package.swift
Shahroz16: Both of the versions that you pointed to are not dependent on `Segment`, they are also CIO forks (Segment don't support their cocoapods) so we are responsible for updating them, and hence even if they are not fixed its okay, they are not in hands of Segment.

Shahroz16: I update the forks and cocoapods releases to match the same versions.

PR #716, opened 2024-05-17
chore: create inline UIView that checks if there are inline messages in local queue

levibostian:

Create a new public UIView that customers can add to their app's UI to display inline Views. This commit does not display messages yet. It only checks if there is an in-app message available to display inline with the equal elementId.

Testing:
Automated tests added in the commit. Some boilerplate code was added to allow mocking some dependencies and testing functions that were previously not very testable.

Reviewer notes:
This commit only handles the use case of inline Views being able to display inline messages after a fetch has already completed in the SDK. Future changes will handle when an inline View is visible and a fetch completes.

---

**Stack**:
- #720
- #718
- #717
- #716 ⬅


Sources/MessagingInApp/Gist/Gist.swift
```diff
@@ -146,4 +155,21 @@ public class Gist: GistDelegate {
     func removeMessageManager(instanceId: String) {
         messageManagers.removeAll(where: { $0.currentMessage.instanceId == instanceId })
     }
+
+    // Inline messages
+
+    func getInlineMessages(forElementId elementId: String) -> [Message] {
+        messageQueueManager.getInlineMessages(forElementId: elementId)
+    }
```
levibostian, line 163: Only logic added for inline. Rest of code modified is to make the code testable.

Sources/MessagingInApp/Gist/Managers/MessageQueueManager.swift
```diff
@@ -59,7 +80,11 @@ class MessageQueueManager {
         localMessageStore.removeValue(forKey: queueId)
     }
 
-    private func addMessageToLocalStore(message: Message) {
+    func getInlineMessages(forElementId elementId: String) -> [Message] {
+        localMessageStore.filter { $0.value.elementId == elementId }.map(\.value)
+    }
```
levibostian, line 85: Returns all inline messages with a matching element id.

Sources/MessagingInApp/Gist/Managers/MessageQueueManager.swift
```diff
@@ -123,12 +152,22 @@ class MessageQueueManager {
...
+            // So, add the message to the local store and when inline Views are constructed, they will check the store.
+
+            addMessageToLocalStore(message: message)
+
+            // In a future PR, we will want to notify all currently visible inline Views that new messages are available in local store.
+            //
+            // At that time, we may decide we do not need these lines anymore. Keeping them in until we implement this notify piece.
+            //            Logger.instance.info(message: "Found a message meant to be shown inline. Element Id \(elementId)")
+            //            Gist.shared.embedMessage(message: message, elementId: elementId)
+
```
levibostian, line 167: Modified logic here to add all inline messages to local queue.

levibostian: // TODO: I need to verify that when page rule triggers and SDK checks if we need to display any modal messages it filters out inline messages.

Sources/MessagingInApp/Gist/Managers/Models/Message.swift
```diff
@@ -18,7 +18,7 @@ public class GistProperties {
 
 public class Message {
     public private(set) var instanceId = UUID().uuidString.lowercased()
-    public let queueId: String?
+    public let queueId: String? // uniquely identifies an in-app message in the backend system
```
levibostian, line 21: Added this comment because[ I asked in-app squad about it](https://customerio.slack.com/archives/C036FTE6UJ0/p1715967637970009). I wasn't sure if this impacts the messages we return in queuemanager.getInlineMessages and it doesn't.

Tests/MessagingInApp/Extensions/GistExtensions.swift
levibostian, line 1: Boilerplate for testing. Allows us to create test instances of Message for inline tests

Tests/Shared/NSCoder.swift
levibostian, line 1: Required to add this for `InAppMessageViewTest` to create instances of `InAppMessageView`.

Sources/MessagingInApp/Gist/Managers/MessageQueueManager.swift
```diff
@@ -72,21 +97,16 @@ class MessageQueueManager {
...
                             guard let responses else {
                                 return
                             }
-                            // To prevent us from showing expired / revoked messages, clear user messages from local queue.
-                            self.clearUserMessagesFromLocalStore()
-                            Logger.instance.info(message: "Gist queue service found \(responses.count) new messages")
-                            for queueMessage in responses {
-                                let message = queueMessage.toMessage()
-                                self.handleMessage(message: message)
-                            }
```
levibostian, line 89: I moved this block into a class function so I could write automated tests against the logic.

Sources/MessagingInApp/Gist/Managers/MessageQueueManager.swift
Ahmed-Ali: what is the use case where we can have multiple messages matching the same element id?

Sources/MessagingInApp/Gist/Managers/Models/Message.swift
Ahmed-Ali, line 21: based on the comment beside it, shouldn't this be a message id? the comment says it identifies the message but the property name is queue id

Sources/MessagingInApp/Views/InAppMessageView.swift
```diff
@@ -0,0 +1,69 @@
...
+    public init(elementId: String) {
+        super.init(frame: .zero)
+        self.elementId = elementId
+
+        // Setup the View and display a message, if one available. Since an elementId has been set.
+        setupView()
+        checkIfMessageAvailableToDisplay()
+    }
+
+    // This is called when the View is created from a Storyboard.
```
Ahmed-Ali, line 42: 👍

Sources/MessagingInApp/Views/InAppMessageView.swift
```diff
@@ -0,0 +1,69 @@
...
+            return
+        }
+
+        let queueOfMessagesForGivenElementId = localMessageQueue.getInlineMessages(forElementId: elementId)
+        guard let messageToDisplay = queueOfMessagesForGivenElementId.first else {
+            return // no messages to display, exit early. In the future we will dismiss the View.
+        }
+
+        displayInAppMessage(messageToDisplay)
+        //      }
```
Ahmed-Ali, line 65: delete

Sources/MessagingInApp/Gist/Managers/MessageQueueManager.swift
```diff
@@ -1,13 +1,34 @@
...
+}
+
+// sourcery: InjectRegisterShared = "MessageQueueManager"
+class MessageQueueManagerImpl: MessageQueueManager {
     var interval: Double = 600
     private var queueTimer: Timer?
-    // The local message store is used to keep messages that can't be displayed because the route rule doesnt match.
-    private var localMessageStore: [String: Message] = [:]
+    // The local message store is used to keep messages that can't be displayed because the route rule doesnt match and inline messages.
+    var localMessageStore: [String: Message] = [:]
```
Ahmed-Ali, line 22: public mutable property is almost always not good choice.
let alone the no-threading-protection

Ahmed-Ali (changes requested):
requesting changes mainly for the exposed mutable var in the message queue maanger

Sources/MessagingInApp/Gist/Managers/Models/Message.swift
Shahroz16: I think the comment is a bit confusing, it probably uniquely identifies an in-app message in the queue for this user. 

Message-id is the unique identifier for the in-app message.

Sources/MessagingInApp/Gist/Managers/MessageQueueManager.swift
levibostian: [Gist's backend service can return multiple in-app messages with the same element id. You have to use the priority property on the message](https://customerio.slack.com/archives/C036FTE6UJ0/p1716306131792019?thread_ts=1716304555.465419&cid=C036FTE6UJ0). 

I will modify this function in this PR to sort the array by priority before returning.

Sources/MessagingInApp/Gist/Managers/Models/Message.swift
levibostian: [I was told that the `queueId` is a unique identifier for a message](https://customerio.slack.com/archives/C036FTE6UJ0/p1715967782604259?thread_ts=1715967637.970009&cid=C036FTE6UJ0). 

[The `messageId` refers to templates and so it may not be unique](https://customerio.slack.com/archives/C036FTE6UJ0/p1716291837428599?thread_ts=1715967637.970009&cid=C036FTE6UJ0).

Sources/MessagingInApp/Gist/Managers/MessageQueueManager.swift
```diff
@@ -1,13 +1,34 @@
...
+    private var gist: GistInstance {
+        DIGraphShared.shared.gist
+    }
 
-    func setup(skipQueueCheck: Bool = false) {
+    func setup() {
+        setup(skipQueueCheck: false)
+    }
```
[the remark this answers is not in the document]
levibostian: I needed to generate a mock for testing. We generate mocks by writing protocols. You can't generate a protocol with default parameters. Instead, you must write all versions of the functions you plan to provide. So, we need 2 `setup()` functions to replace the 1 `setup()` function that used default parameter values.

Sources/MessagingInApp/Gist/Gist.swift
```diff
@@ -113,9 +122,13 @@ public class Gist: GistDelegate {
     }
 
     func logMessageView(message: Message) {
+        guard message.isModalMessage else { // we do not expect this function to be called for inline Views. Adding this to be safe.
```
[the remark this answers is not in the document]
levibostian: The comment might be adding confusion. It's not meant to say that we will never have inline messages run the logic in this function, but instead do not call it at this time. 

To make sure that `shownModalMessageQueueIds` does not get modified for inline messages, I added the `guard`. 

In milestone 1, the only way for a message to no longer be shown is have it expire. So, removing from the local store and modifying `shownModalMessageQueueIds` at this time is not wanted.

Sources/MessagingInApp/Gist/Managers/MessageQueueManager.swift
levibostian: The property is `internal` as that's the swift default. It's made internal so we can reference it in our automated tests. 

As far as no-threading, that's a good call-out. I can make it have safe reading and writing access as we do in the codebase today.

Sources/MessagingInApp/Gist/Managers/MessageQueueManager.swift
levibostian: [List is sorted by priority before being returned](https://github.com/customerio/customerio-ios/pull/716/files#diff-ec01291f32e74af14510929ca3d297f876bb1882bd4fdd5008cadca4dff331bcR73).

Sources/MessagingInApp/Gist/Managers/MessageQueueManager.swift
levibostian: I [updated](https://github.com/customerio/customerio-ios/pull/716/files#diff-ec01291f32e74af14510929ca3d297f876bb1882bd4fdd5008cadca4dff331bcR19-R23) the `localMessageStore` and `interval` to have atomic read and write access.

Sources/MessagingInApp/Gist/Gist.swift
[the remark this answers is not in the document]
levibostian: [I updated the comment to try and be more clear](https://github.com/customerio/customerio-ios/pull/716/files#diff-05212de0e1f067b635cf22ad2778b1d51110ddeb8d5615b893ea42b4324e9ca0R125-R127). 

The biggest thing I am trying to do is protecting the `shownModalMessageQueueIds` from being inserted with inline messages.

Sources/MessagingInApp/Gist/Managers/Models/Message.swift
Shahroz16: I agree `messageId` refers to template, so we can have multiple messages with the same message id in the queue. But all of them will have a unique `queueId` in that queue and the same message when expired and added again will have a new `queueId` so instead of `backend system` I would rather put it as `queue`

Sources/MessagingInApp/Gist/Managers/MessageQueueManager.swift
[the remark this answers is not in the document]
Ahmed-Ali: I would agree that it is better to keep one version with no default value. Default value can make it harder to understand the default behavior, or even worse, change the default behavior without introducing API change

Sources/MessagingInApp/Gist/Managers/MessageQueueManager.swift
```diff
@@ -99,9 +108,35 @@ class MessageQueueManager {
...
+            // Inline Views show inline messages by getting messages stored in the local queue on device.
+            // So, add the message to the local store and when inline Views are constructed, they will check the store.
+
+            addMessageToLocalStore(message: message)
+
+            // In a future PR, we will want to notify all currently visible inline Views that new messages are available in local store.
+            //
+            // At that time, we may decide we do not need these lines anymore. Keeping them in until we implement this notify piece.
+            //            Logger.instance.info(message: "Found a message meant to be shown inline. Element Id \(elementId)")
+            //            Gist.shared.embedMessage(message: message, elementId: elementId)
```
Ahmed-Ali, line 131: Let's make sure we don't merge changes with commented out code please

Sources/MessagingInApp/Gist/Managers/Models/Message.swift
Ahmed-Ali: I still find the name `queueId` confusing. It makes me think that there are multiple queues, and here we are referencing one specific queue by its id.

The `messageId` referencing a template is also confusing but I would be okay that it is out of scope for this PR.

For the property referenced above we can come up with something better, maybe something like `messageInstanceId` or anything else that associates the uniqueness and id directly to the message not the queue

Sources/MessagingInApp/Gist/Managers/Models/Message.swift
```diff
@@ -18,9 +18,9 @@ public class GistProperties {
...
     public let priority: Int?
-    public let messageId: String
+    public let messageId: String // the messageId refers to the template used to render. For non-HTML messages, that messageId is something like "welcome-demo"
```
Ahmed-Ali, line 23: if it is not big change, let's rename this to `templateId` as a messageId referencing a template isn't very conveying to what it actually is.

Sources/MessagingInApp/Views/InAppMessageView.swift
```diff
@@ -0,0 +1,69 @@
...
+
+    // This is called when the View is created from a Storyboard.
+    required init?(coder: NSCoder) {
+        super.init(coder: coder)
+
+        setupView()
+        // An element id will not be set yet. No need to check for messages to display.
+    }
+
+    private func setupView() {}
```
Ahmed-Ali, line 50: why an empty method?
if it will be implemented later, then let's add it later.

Sources/MessagingInApp/Views/InAppMessageView.swift
Ahmed-Ali, line 65: left over bracket, please clean it up

Sources/MessagingInApp/Views/InAppMessageView.swift
```diff
@@ -0,0 +1,69 @@
...
+        let queueOfMessagesForGivenElementId = localMessageQueue.getInlineMessages(forElementId: elementId)
+        guard let messageToDisplay = queueOfMessagesForGivenElementId.first else {
+            return // no messages to display, exit early. In the future we will dismiss the View.
+        }
+
+        displayInAppMessage(messageToDisplay)
+        //      }
+    }
+
+    private func displayInAppMessage(_ message: Message) {}
```
Ahmed-Ali, line 68: another empty private method, can be added when it is implemented

Ahmed-Ali (changes requested):

Sources/MessagingInApp/Gist/Managers/Models/Message.swift
Ahmed-Ali: I am fine with not doing this change in this PR for sure. Let's just not lose track of it as these hard to reason about names can easily fire back later.

Ahmed-Ali (approved):

Sources/MessagingInApp/Gist/Managers/Models/Message.swift
```diff
@@ -18,9 +18,9 @@ public class GistProperties {
 
 public class Message {
     public private(set) var instanceId = UUID().uuidString.lowercased()
-    public let queueId: String?
+    public let queueId: String? // used to uniquely identify an in-app message.
```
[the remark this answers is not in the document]
levibostian: Opened a new PR to do the renaming instead of a TODO. https://github.com/customerio/customerio-ios/pull/730

PR #717, opened 2024-05-17
chore: get WebView for inline message and add to Inline View to display

levibostian:

Create a WebView to display inline in-app message and display in the Inline UIView. To do this, we can re-use a lot of the same logic that Modal messages use to create WebViews and display them.

Testing:
There are some automated tests added in this commit.

Reviewer notes:
This commit will not show an in-app message if you were to add it to a sample app. Autolayout constraints need to be added to have Views resize and that will come in a future commit.

---

**Stack**:
- #720
- #718
- #717 ⬅
- #716

### DIFF
--- a/Sources/MessagingInApp/Gist/Managers/MessageManager.swift
+++ b/Sources/MessagingInApp/Gist/Managers/MessageManager.swift
@@ -5,13 +5,18 @@ public enum GistMessageActions: String {
     case close = "gist://close"
 }
 
-class MessageManager: EngineWebDelegate {
+/**
+ Handles business logic for in-app message events such as loading messages and handling when action buttons are clicked.
+
+ This class is meant to be extended and not constructed directly. It holds the common logic between all in-app message types.
+
+ Usage:
+ * Extend class.
+ * Override any of the abstract functions in class to implement custom logic for when certain events happen. Depending on the type of message you are displaying, you may want to handle events differently.
+ */
+class MessageManager {
     private var engine: EngineWeb?
     private let siteId: String
-    private var messagePosition: MessagePosition = .top
-    private var messageLoaded = false
-    private var modalViewManager: ModalViewManager?
-    var isMessageEmbed = false
     let currentMessage: Message
     var gistView: GistView!
     private var currentRoute: String
@@ -32,51 +37,49 @@ class MessageManager: EngineWebDelegate {
             properties: message.toEngineRoute().properties
         )
 
+        // When EngineWeb instance is constructed, it will begin the rendering process for the in-app message.
+        // This means that the message begins the process of loading.
+        // Start a timer that helps us determine how long a message took to load/render.
+        elapsedTimer.start(title: "Loading message with id: \(currentMessage.messageId)")
         self.engine = EngineWeb(configuration: engineWebConfiguration)
+
         if let engine = engine {
             engine.delegate = self
             self.gistView = GistView(message: currentMessage, engineView: engine.view)
         }
     }
 
-    func showMessage(position: MessagePosition) {
-        elapsedTimer.start(title: "Displaying modal for message: \(currentMessage.messageId)")
-        messagePosition = position
+    deinit {
+        engine?.cleanEngineWeb()
+        engine = nil
     }
 
-    func getMessageView() -> GistView {
-        isMessageEmbed = true
-        return gistView
+    // MARK: event listeners that subclasses override to handle events.
+
+    // Called when close action button pressed.
+    func onCloseAction() {
+        // Expect subclass implements this.
     }
 
-    private func loadModalMessage() {
-        if messageLoaded {
-            modalViewManager = ModalViewManager(gistView: gistView, position: messagePosition)
-            modalViewManager?.showModalView { [weak self] in
-                guard let self = self else { return }
-                self.delegate?.messageShown(message: self.currentMessage)
-                self.elapsedTimer.end()
-            }
-        }
+    // Called when a deep link action button was clicked in a message and the SDK opened the deep link.
+    func onDeepLinkOpened() {
+        // expect subclass implements this.
     }
 
-    func dismissMessage(completionHandler: (() -> Void)? = nil) {
-        if let modalViewManager = modalViewManager {
-            modalViewManager.dismissModalView { [weak self] in
-                guard let self = self else { return }
-                self.delegate?.messageDismissed(message: self.currentMessage)
-                completionHandler?()
-            }
-        }
+    // Called when the message has finished loading and the WebView is ready to display the message.
+    func onDoneLoadingMessage(routeLoaded: String, onComplete: @escaping () -> Void) {
+        // expect subclass implements this.
     }
 
-    func removePersistentMessage() {
-        if currentMessage.gistProperties.persistent == true {
-            Logger.instance.debug(message: "Persistent message dismissed, logging view")
-            Gist.shared.logMessageView(message: currentMessage)
-        }
+    // Called when an action button is clicked and the action is to show a different in-app message.
+    func onReplaceMessage(newMessageToShow: Message) {
+        // subclass should implement
     }
+}
 
+// The main logic of this class is being the delegate for the EngineWeb instance.
+// This class's delegate responsibilities are to run the logic that's common to all types of in-app messages and call the event listeners that subclasses override.
+extension MessageManager: EngineWebDelegate {
     func bootstrapped() {
         Logger.instance.debug(message: "Bourbon Engine bootstrapped")
 
@@ -86,7 +89,6 @@ class MessageManager: EngineWebDelegate {
         }
     }
 
-    // swiftlint:disable cyclomatic_complexity
     func tap(name: String, action: String, system: Bool) {
         Logger.instance.info(message: "Action triggered: \(action) with name: \(name)")
         delegate?.action(message: currentMessage, currentRoute: currentRoute, action: action, name: name)
@@ -96,8 +98,7 @@ class MessageManager: EngineWebDelegate {
             switch url.host {
             case "close":
                 Logger.instance.info(message: "Dismissing from action: \(action)")
-                removePersistentMessage()
-                dismissMessage()
+                onCloseAction()
             case "loadPage":
                 if let page = url.queryParameters?["url"],
                    let pageUrl = URL(string: page),
@@ -105,13 +106,7 @@ class MessageManager: EngineWebDelegate {
                     UIApplication.shared.open(pageUrl)
                 }
             case "showMessage":
-                if currentMessage.isEmbedded {
-                    showNewMessage(url: url)
-                } else {
-                    dismissMessage {
-                        self.showNewMessage(url: url)
-                    }
-                }
+                showNewMessage(url: url)
             default: break
             }
         } else {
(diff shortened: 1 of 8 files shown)


Sources/MessagingInApp/Gist/Managers/MessageManager.swift
```diff
@@ -39,35 +108,21 @@ class MessageManager: EngineWebDelegate {
...
-    func showMessage(position: MessagePosition) {
-        elapsedTimer.start(title: "Displaying modal for message: \(currentMessage.messageId)")
-        messagePosition = position
+    func loadingMessage() {
+        elapsedTimer.start(title: "Loading message with id: \(currentMessage.messageId)")
+    }
+
+    func doneLoadingMessage() {
+        elapsedTimer.end()
     }
```
levibostian, line 118: // TODO: call these from inline manager.

Sources/MessagingInApp/Gist/Managers/MessageManager.swift
levibostian, line 1: This is the main logic of this PR. `MessageManager` contains a lot of code that's common between inline and modal messages. `ModalMessageManager` contains code I pulled out of `MessageManager` that's only used for modal messages.

Sources/MessagingInApp/Gist/Managers/MessageManager.swift
```diff
@@ -208,18 +267,6 @@ class MessageManager: EngineWebDelegate {
...
-            if isMessageEmbed {
-                delegate?.messageShown(message: currentMessage)
-            } else {
-                if UIApplication.shared.applicationState == .active {
-                    loadModalMessage()
-                } else {
-                    Gist.shared.removeMessageManager(instanceId: currentMessage.instanceId)
-                }
-            }
-        }
```
levibostian, line 222: This is code that should only be called for modal messages. So I moved this block into the Modal manager subclass.

Sources/MessagingInApp/Gist/Managers/MessageManager.swift
```diff
@@ -39,35 +108,25 @@ class MessageManager: EngineWebDelegate {
...
-    }
-
     func dismissMessage(completionHandler: (() -> Void)? = nil) {
-        if let modalViewManager = modalViewManager {
-            modalViewManager.dismissModalView { [weak self] in
-                guard let self = self else { return }
-                self.delegate?.messageDismissed(message: self.currentMessage)
-                completionHandler?()
-            }
-        }
```
levibostian, line 70: This is code that should only be called for modal messages. So I moved this block into the Modal manager subclass.

Sources/MessagingInApp/Views/InAppMessageView.swift
```diff
@@ -65,5 +71,36 @@ public class InAppMessageView: UIView {
         //      }
     }
 
-    private func displayInAppMessage(_ message: Message) {}
+    private func displayInAppMessage(_ message: Message) {
```
levibostian, line 74: This function's body is the logic that creates a WebView and renders the in-app message in the WebView. Most of this logic comes from `InlineMessageManager`. So all we need to do in the inline View is create a `InlineMessageManager` instance and setup the WebView for displaying in this View.

Sources/MessagingInApp/Views/InAppMessageView.swift
```diff
@@ -65,5 +71,36 @@ public class InAppMessageView: UIView {
...
+        // There might already be a message displayed. If so, remove it and cleanup the resources.
+        // First, remove the subview that the inline manager has a reference to.
+        // Lastly, cleanup the inline manager.
+        subviews.first?.removeFromSuperview()
+        inlineMessageManager = nil
+
+        // Create a new manager for this new message to display and then display the manager's WebView.
+        let newInlineMessageManager = InlineMessageManager(siteId: gist.siteId, message: message)
+        newInlineMessageManager.startLoadingMessage()
+
```
Shahroz16, line 84: shouldn't we preload the new message before removing the first one to avoid visual delays?

Sources/MessagingInApp/Gist/Managers/MessageManager.swift
```diff
@@ -5,12 +5,81 @@ public enum GistMessageActions: String {
...
+    private var messageLoaded = false
+    private var modalViewManager: ModalViewManager?
+    var messagePosition: MessagePosition = .top
+
+    override func routeLoaded(route: String) {
+        super.routeLoaded(route: route)
+
+        if route == currentMessage.messageId, !messageLoaded {
+            messageLoaded = true
+            if isMessageEmbed {
```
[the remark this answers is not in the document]
levibostian: `isMessageEmbed` is existing code in the codebase today, not code I wrote. So it may or may not be in reference to inline messaging. 

My intention with this PR was to create `InlineMessageManager` without modifying the logic used for Modals. 

We may be able to make more modifications to `ModalMessageManager` and `MessageManager`. Because of scope, I wanted to leave the Modal code alone in this change.

Sources/MessagingInApp/Views/InAppMessageView.swift
levibostian: 🤔 Not a bad idea. 

If a message is being shown on the device, the message expires, and then we want to show a new in-app message, I think this idea would be possible since we can wait a few seconds before we remove the expired message. 

But for the scenario where a close button is clicked and we want to show another message, I am not sure we want to pre-render the next message. If we never show the next message, that pre-render could be wasteful. I think this would be considered to be a future enhancement, agree?

Sources/MessagingInApp/Views/InAppMessageView.swift
```diff
@@ -65,5 +71,36 @@ public class InAppMessageView: UIView {
         //      }
     }
 
-    private func displayInAppMessage(_ message: Message) {}
+    private func displayInAppMessage(_ message: Message) {
+        // There might already be a message displayed. If so, remove it and cleanup the resources.
+        // First, remove the subview that the inline manager has a reference to.
+        // Lastly, cleanup the inline manager.
+        subviews.first?.removeFromSuperview()
+        inlineMessageManager = nil
```
levibostian, line 79: // TODO as we talked about in the call today, I'll comment this out and we can implement this in a future PR. 

I was trying to be resourceful by cleaning up old resources but I can see how this would change the functionality of the View. We decided to implement this later.

levibostian: Done. Added a `guard` statement to exit early if we are already showing a message.

Shahroz16 (approved):

Sources/MessagingInApp/Gist/Managers/MessageManager.swift
```diff
@@ -5,12 +5,81 @@ public enum GistMessageActions: String {
...
 
+/**
+ Class that implements the business logic for a inline message being displayed.
+ */
+class InlineMessageManager: MessageManager {}
+
+/**
+ Class that implements the business logic for a modal message being displayed.
+ */
+class ModalMessageManager: MessageManager {
```
Ahmed-Ali, line 16: Why are we using inheritance here?
I understand there is common logic, but we can extract the common logic in its own utility type, and use it in two different implementations.
Inheritance obscures the side effect of doing a single change and can be hard to trace at times

Sources/MessagingInApp/Gist/Managers/MessageManager.swift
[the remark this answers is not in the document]
Ahmed-Ali: embed messages in the existing code base meant to reference inline messages. So I would indeed work on different approach to reduce the complexity of the code branches

Sources/MessagingInApp/Gist/Managers/MessageManager.swift
```diff
@@ -5,12 +5,81 @@ public enum GistMessageActions: String {
...
+        startLoadingMessage()
+        messagePosition = position
+    }
+
+    override func dismissMessage(completionHandler: (() -> Void)? = nil) {
+        if let modalViewManager = modalViewManager {
+            modalViewManager.dismissModalView { [weak self] in
+                guard let self = self else { return }
+                self.delegate?.messageDismissed(message: self.currentMessage)
+                completionHandler?()
```
Ahmed-Ali, line 48: shouldn't this get called even if self is nil?

Sources/MessagingInApp/Gist/Managers/MessageManager.swift
```diff
@@ -39,35 +108,25 @@ class MessageManager: EngineWebDelegate {
...
-
     func dismissMessage(completionHandler: (() -> Void)? = nil) {
-        if let modalViewManager = modalViewManager {
-            modalViewManager.dismissModalView { [weak self] in
-                guard let self = self else { return }
-                self.delegate?.messageDismissed(message: self.currentMessage)
-                completionHandler?()
-            }
-        }
+        // expect subclass implements this.
```
Ahmed-Ali, line 129: I am honestly not big fan of subclassing when there are different approaches that makes the code easier to follow.
MessageManager can simply be a protocol with two different implementations if needed. The actual implementations can use similar utility functions for the common logic.
Inheritance is a very strong relation and unless absolutely a must, we can almost always find simpler approaches

Sources/MessagingInApp/Views/InAppMessageView.swift
```diff
@@ -65,5 +71,43 @@ public class InAppMessageView: UIView {
...
+            return
+        }
+
+        /** Code commented out to be used for reference when we implement replacing of 1 message with another */
+        // There might already be a message displayed. If so, remove it and cleanup the resources.
+        // First, remove the subview that the inline manager has a reference to.
+        // Lastly, cleanup the inline manager.
+//        subviews.first?.removeFromSuperview()
+//        inlineMessageManager = nil
+
```
Ahmed-Ali, line 87: let's remove dead code instead of commenting it out please

Ahmed-Ali (changes requested):

Sources/MessagingInApp/Gist/Managers/MessageManager.swift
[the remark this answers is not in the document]
levibostian: I just pushed a commit that removed `isMessageEmbed` from the `ModalMessageManager` class as it was indeed related to inline messaging. This removal did increase the amount of code changes in this PR, but the amount of code change was small and it only removed unused code so I thought it was worth it.

levibostian:
I just pushed a change based on review suggestions. 

I kept `InlineMessageManager` and `ModalMessageManager` being subclasses of a `MessageManager` superclass. I agree with the feedback about abstract classes not always being the best tool. I did attempt to create a protocol that inline and modal classes inherit but I did not like the result. 

I modified the `MessageManager` abstract class to make it more clear what code is common between inline and modal messages and what code is unique. All of the common logic for modal and inline messages is [the `EngineWebDelegate` implementation](https://github.com/customerio/customerio-ios/pull/717/files#diff-70b8fb68d815f2ebfa219cf5e9332ad86a7b2081a7a87e6dbcc43c0ddd850e4aR82-R223). All of the code that is unique is [implemented by the abstract functions](https://github.com/customerio/customerio-ios/pull/717/files#diff-70b8fb68d815f2ebfa219cf5e9332ad86a7b2081a7a87e6dbcc43c0ddd850e4aR57-R78). 

There could be a better design pattern over this approach, but I think it will require more refactoring to the existing Modal in-app message code in the SDK.

Sources/MessagingInApp/Gist/Managers/MessageManager.swift
levibostian: See [my latest update](https://github.com/customerio/customerio-ios/pull/717#issuecomment-2135841512).

Sources/MessagingInApp/Gist/Managers/MessageManager.swift
levibostian: See [my latest update](https://github.com/customerio/customerio-ios/pull/717#issuecomment-2135841512).

PR #718, opened 2024-05-17
chore: add AutoLayout constraints to make inline messages visible in UIKit app

levibostian:

This commit implements UIKit AutoLayout constraints to the Inline View and the WebView. Adding these constraints make inline messages visible inside of UIKit apps! The Inline View will start at height 0 and then after the webview content is loaded, it will instantly change the height to display the full in-app message.

Testing:
All testing is done manually in sample apps. AutoLayout constraints change the visuals of Views so viewing how the View looks in an app is required to test.

If you want to test this commit in a sample app build, follow these instructions:
* Open the UIKit app on device.
* Immediately after the app opens, send yourself a test in-app message with element ID "test".
* Wait 10 seconds. In this time, the in-app SDK will fetch the test in-app message and then will display it inline on the screen.

#### Demo video showing this testing:

When I began recording the video, I had already sent 2 messages from Fly to my test device. That way when the app opens and fetches messages, it would find these 2 messages. Then, the app displays those 2 messages in 2 separate inline Views on the screen. The 2 inline Views are expected to display all of the content with the correct height and width to match the web content's aspect ratio. 

https://github.com/customerio/customerio-ios/assets/2041082/a00ed038-e882-425b-afe9-542683436645

---

**Stack**:
- #720
- #718 ⬅
- #717
- #716

### DIFF
--- a/Apps/APN-UIKit/APN UIKit/Base.lproj/Main.storyboard
+++ b/Apps/APN-UIKit/APN UIKit/Base.lproj/Main.storyboard
@@ -1,9 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.CocoaTouch.Storyboard.XIB" version="3.0" toolsVersion="22154" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" useSafeAreas="YES" colorMatched="YES" initialViewController="37F-6d-U3o">
+<document type="com.apple.InterfaceBuilder3.CocoaTouch.Storyboard.XIB" version="3.0" toolsVersion="32700.99.1234" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" useSafeAreas="YES" colorMatched="YES" initialViewController="37F-6d-U3o">
     <device id="retina6_5" orientation="portrait" appearance="light"/>
     <dependencies>
         <deployment identifier="iOS"/>
-        <plugIn identifier="com.apple.InterfaceBuilder.IBCocoaTouchPlugin" version="22130"/>
+        <plugIn identifier="com.apple.InterfaceBuilder.IBCocoaTouchPlugin" version="22685"/>
         <capability name="Safe area layout guides" minToolsVersion="9.0"/>
         <capability name="System colors in document resources" minToolsVersion="11.0"/>
         <capability name="documents saved in the Xcode 8 format" minToolsVersion="8.0"/>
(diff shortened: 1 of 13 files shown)


Apps/APN-UIKit/APN UIKit/Base.lproj/Main.storyboard
levibostian, line 1: storyboard modified because I added inline Views to the screen. 

I have been talking with Jason on how to QA test this feature. He didn't have specific requirements as far as what screen these Views go, so I added to Dashboard for convenience.

Apps/APN-UIKit/APN UIKit/View/DashboardViewController.swift
levibostian, line 1: I added 2 inline Views to the Dashboard screen. So we can test customers can use storyboards or use code to add inline Views to their app.

Sources/MessagingInApp/Gist/Managers/MessageQueueManager.swift
levibostian, line 1: Because queue manager stores data in-memory, I had to make it a singleton. All changes in this commit are to make it a singleton.

Sources/MessagingInApp/Views/InAppMessageView.swift
levibostian, line 1: The main logic of this PR - AutoLayout constraints so messages show up on screen.

Sources/MessagingInApp/Gist/Gist.swift
```diff
@@ -156,12 +159,6 @@ public class Gist: GistInstance, GistDelegate {
     func removeMessageManager(instanceId: String) {
         messageManagers.removeAll(where: { $0.currentMessage.instanceId == instanceId })
     }
-
-    // Inline messages
-
-    func getInlineMessages(forElementId elementId: String) -> [Message] {
-        messageQueueManager.getInlineMessages(forElementId: elementId)
-    }
 }
```
levibostian, line 162: I decided to call `getInlineMessages` directly from `MessageQueueManager` instead of `Gist` to reduce amount of boilerplate code needed.

Sources/MessagingInApp/Gist/Gist.swift
```diff
@@ -8,8 +8,11 @@ protocol GistInstance: AutoMockable {
...
 
 public class Gist: GistInstance, GistDelegate {
-    var messageQueueManager: MessageQueueManager = DIGraphShared.shared.messageQueueManager
```
levibostian, line 11: Changed because `MessageQueueManager` is a singleton now. Use a property getter to get the shared singleton instance.

Sources/MessagingInApp/Gist/Managers/MessageQueueManager.swift
```diff
@@ -15,8 +16,9 @@ protocol MessageQueueManager: AutoMockable {
 }
 
 // sourcery: InjectRegisterShared = "MessageQueueManager"
+// sourcery: InjectSingleton
```
levibostian, line 19: This class has always been used as a singleton. The whole SDK would use `Gist.messageQueue`. To make the code testable, I switched to instead make the class a singleton in the digraph as we do for all other code in the SDK.

Sources/MessagingInApp/Gist/Managers/MessageQueueManager.swift
```diff
@@ -3,7 +3,8 @@ import Foundation
 import UIKit
 
 protocol MessageQueueManager: AutoMockable {
-    var interval: Double { get set }
```
levibostian, line 6: Although I do prefer to use property getter/setters for an API, I found this API to be safer with having the class be a singleton. 

Here is some sample code with what I mean: 

```swift
// If you use a property getter/setter API:
let queueManager: MessageQueueManager = ... get singleton instance 
queueManager.interval = 500
// 🤔 Did we update the shared singleton `inventory` property? Or did we update the `queueManager` local variable, only? 

// if we change to getter/setter functions:
MessageQueueManager.singletonInstance.setInventory(500)
// This is much more clear that we are editing the singleton, not a locally scoped instance. 
```

Sources/MessagingInApp/Views/InAppMessageView.swift
levibostian, line 1: Added autolayout code to the View so we can now display the web content in the View.

Apps/APN-UIKit/APN UIKit/View/DashboardViewController.swift
```diff
@@ -28,6 +31,26 @@ class DashboardViewController: BaseViewController {
...
+            // Add the View to the screen.
+            view.addSubview(newInlineViewUsingUIAsCode)
+
+            // We are adding the new View between 2 Buttons in the StackView. No specific reason why, not sure where else to put it.
+            newInlineViewUsingUIAsCode.widthAnchor.constraint(equalToConstant: UIScreen.main.bounds.width).isActive = true
+            newInlineViewUsingUIAsCode.topAnchor.constraint(equalTo: randomEventButton.bottomAnchor).isActive = true
+            newInlineViewUsingUIAsCode.bottomAnchor.constraint(equalTo: customEventButton.topAnchor).isActive = true
+            newInlineViewUsingUIAsCode.leadingAnchor.constraint(equalTo: view.leadingAnchor).isActive = true
+            newInlineViewUsingUIAsCode.trailingAnchor.constraint(equalTo: view.trailingAnchor).isActive = true
+            newInlineViewUsingUIAsCode.centerXAnchor.constraint(equalTo: randomEventButton.centerXAnchor).isActive = true
```
levibostian, line 51: During QA testing, I noticed what might be a bug with this code. 

steps: 
* Send yourself 2 inline test messages so you have both inline Views showing messages in the app. 
* Scroll down the screen. When you put your finger in the middle of 1 of the inline Views and scroll your finger, the screen moves with your finger. But for the other inline View (this one I added via code), the screen does not move with your finger. 

I think that this is a bug with the AutoLayout code here which means it's a bug in the sample app and not with the SDK.

Ahmed-Ali (changes requested):
This is a very UI heavy code. A small video showing how it looks like and what have been tested would make the code review much easier.

levibostian:
Video added to description of PR.

Ahmed-Ali (approved):

PR #720, opened 2024-05-20
chore: display inline messages fetched after View constructed

levibostian:

Fetching of in-app message is an async operation running in the background. There is a chance that in-app messages are not available when the inline View is constructed but become available after. This commit gets notified when a fetch is complete so it can check if there are in-app messages to display after the View has already been constructed.

Testing:
* Automated tests are added.
* Verified that messages display in sample app.
If you want to test this commit in a sample app build, follow these instructions:
1. Kill the app on device.
2. Send yourself a test inline message from Fly.
3. Open app. After you wait a few seconds, the inline message you sent yourself will display.

### DIFF
--- a/Apps/APN-UIKit/APN UIKit/View/DashboardViewController.swift
+++ b/Apps/APN-UIKit/APN UIKit/View/DashboardViewController.swift
@@ -32,24 +32,22 @@ class DashboardViewController: BaseViewController {
     override func viewDidLoad() {
         super.viewDidLoad()
 
-        // In a future PR, we will remove the asyncAfter(). this is only for testing in sample apps because when app opens, the local queue is empty. so wait to check messages until first fetch is done.
-        DispatchQueue.main.asyncAfter(deadline: .now() + 5) { [self] in
-            inlineInAppView.elementId = "dashboard-announcement"
+        // For inline Views added with Storyboard, set the elementId to finish setup of the View and begin showing messages.
+        inlineInAppView.elementId = "dashboard-announcement"
 
-            // We want to test that Inline Views can be used by customers who prefer to use code to make the UI.
-            // Construct a new instance of the View, add it to the ViewController, then set constraints to make it visible.
-            let newInlineViewUsingUIAsCode = InAppMessageView(elementId: "dashboard-announcement-code")
-            // Add the View to the screen.
-            view.addSubview(newInlineViewUsingUIAsCode)
+        // We want to test that Inline Views can be used by customers who prefer to use code to make the UI.
+        // Construct a new instance of the View, add it to the ViewController, then set constraints to make it visible.
+        let newInlineViewUsingUIAsCode = InAppMessageView(elementId: "dashboard-announcement-code")
+        // Add the View to the screen.
+        view.addSubview(newInlineViewUsingUIAsCode)
 
-            // We are adding the new View between 2 Buttons in the StackView. No specific reason why, not sure where else to put it.
-            newInlineViewUsingUIAsCode.widthAnchor.constraint(equalToConstant: UIScreen.main.bounds.width).isActive = true
-            newInlineViewUsingUIAsCode.topAnchor.constraint(equalTo: randomEventButton.bottomAnchor).isActive = true
-            newInlineViewUsingUIAsCode.bottomAnchor.constraint(equalTo: customEventButton.topAnchor).isActive = true
-            newInlineViewUsingUIAsCode.leadingAnchor.constraint(equalTo: view.leadingAnchor).isActive = true
-            newInlineViewUsingUIAsCode.trailingAnchor.constraint(equalTo: view.trailingAnchor).isActive = true
-            newInlineViewUsingUIAsCode.centerXAnchor.constraint(equalTo: randomEventButton.centerXAnchor).isActive = true
-        }
+        // We are adding the new View between 2 Buttons in the StackView. No specific reason why, not sure where else to put it.
+        newInlineViewUsingUIAsCode.widthAnchor.constraint(equalToConstant: UIScreen.main.bounds.width).isActive = true
+        newInlineViewUsingUIAsCode.topAnchor.constraint(equalTo: randomEventButton.bottomAnchor).isActive = true
+        newInlineViewUsingUIAsCode.bottomAnchor.constraint(equalTo: customEventButton.topAnchor).isActive = true
+        newInlineViewUsingUIAsCode.leadingAnchor.constraint(equalTo: view.leadingAnchor).isActive = true
+        newInlineViewUsingUIAsCode.trailingAnchor.constraint(equalTo: view.trailingAnchor).isActive = true
+        newInlineViewUsingUIAsCode.centerXAnchor.constraint(equalTo: randomEventButton.centerXAnchor).isActive = true
 
         configureDashboardRouter()
         addNotifierObserver()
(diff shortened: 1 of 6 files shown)


Sources/MessagingInApp/Gist/Gist.swift
```diff
@@ -121,10 +121,6 @@ public class Gist: GistInstance, GistDelegate {
         delegate?.action(message: message, currentRoute: currentRoute, action: action, name: name)
     }
 
-    public func embedMessage(message: Message, elementId: String) {
```
levibostian, line 124: `embedMessage` function call [removed in this commit](Sources/MessagingInApp/Gist/Managers/MessageQueueManager.swift). Because it's no longer needed, I removed it here. 

This is a public function of Gist, but it's not part of our advertised public API. So removing it should not be considered a breaking change to our SDK.

Ahmed-Ali (changes requested):
I would really appreciate a demo video showing the scenarios where this has been tested and how it looks like. Downloading the sample app and recreating a scenario to see how the change looks like is an overkill for the code review

Apps/APN-UIKit/APN UIKit/View/DashboardViewController.swift
```diff
@@ -33,21 +33,19 @@ class DashboardViewController: BaseViewController {
...
-            let newInlineViewUsingUIAsCode = InAppMessageView(elementId: "dashboard-announcement-code")
-            // Because the Dashboard screen contains a lot of Views and it's designed using Storyboard, we are
-            // adding this inline View into the UI by adding to a StackView. This allows us to dyanamically add to the Dashboard screen without complexity or breaking any of the constraints set in Storyboard.
-            buttonStackView.addArrangedSubview(newInlineViewUsingUIAsCode)
-
-            // Customers are responsible for setting the width of the View.
-            newInlineViewUsingUIAsCode.translatesAutoresizingMaskIntoConstraints = false
-            newInlineViewUsingUIAsCode.widthAnchor.constraint(equalTo: buttonStackView.widthAnchor).isActive = true
-        }
+        // For inline Views added with Storyboard, set the elementId to finish setup of the View and begin showing messages.
```
levibostian, line 36: Just removed the `DispatchQueue.main.asyncAfter(deadline: .now() + 5) { [self]` and formatted the code. 

This delay is no longer needed after this commit since View will display messages that are fetched *after* the UI is in the foreground.

levibostian:
Video added to PR description.

Sources/MessagingInApp/Gist/Gist.swift
Shahroz16: I think we still need to provide customers with a way to add messages locally, we might need to check with in-app squad about this. We could probably give a method on top of `InAppMessageView`.  Doesn't need to happen in this PR but maybe a ticker or todo so we don't forget.

Sources/MessagingInApp/Views/InAppMessageView.swift
```diff
@@ -94,6 +106,11 @@ public class InAppMessageView: UIView {
             return // no messages to display, exit early. In the future we will dismiss the View.
         }
 
+        // Do not re-show the existing message if already shown to prevent the UI from flickering as it loads the same message again.
+        if let currentlyShownMessage = inlineMessageManager?.currentMessage, currentlyShownMessage.messageId == messageToDisplay.messageId {
```
Shahroz16, line 110: `currentlyShownMessage.messageId == messageToDisplay.messageId` isn't this template id? so we can have two messages with same template ids and it won;t show the second message because its template id is same as previous one?

Shahroz16 (changes requested):
blocking because of : `currentlyShownMessage.messageId == messageToDisplay.messageId`

Sources/MessagingInApp/Gist/Gist.swift
levibostian: Since this feature isn't on the project roadmap, I'll hold off on creating a ticket or todo. If you think it's important, please pitch it to the team, perhaps in the project Slack channel.

Sources/MessagingInApp/Views/InAppMessageView.swift
levibostian: You're correct, it should be `queueId` and not `messageId`. 

I am pushing up a commit with this change. Unfortunately, we don't have a good way to write a test for this line in this PR. 

[I am adding this case to another ticket](https://linear.app/customerio/issue/MBL-313/after-close-action-pressed-replace-in-app-message-with-next-in-queue) where I'll include an automated test there.

Sources/Common/Communication/Event.swift
```diff
@@ -195,3 +195,16 @@ public struct NewSubscriptionEvent: EventRepresentable {
         self.params = params
     }
 }
+
+/// When in-app SDK has fetched in-app messages from the server.
+public struct InAppMessagesFetchedEvent: EventRepresentable {
```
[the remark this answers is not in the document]
levibostian: I am OK with having the event in common or in-app. 

I put it here because right now, *all* of the eventbus events are in this file. So for organization purposes, I put it in here. 

I am happy to move it if you think that's best.

Apps/APN-UIKit/APN UIKit/View/DashboardViewController.swift
```diff
@@ -33,21 +33,19 @@ class DashboardViewController: BaseViewController {
...
+        inlineInAppViewCreatedInStoryboard.elementId = "dashboard-announcement"
+
+        // We want to test that Inline Views can be used by customers who prefer to use code to make the UI.
+        // Construct a new instance of the View, add it to the ViewController, then set constraints to make it visible.
+        let newInlineViewUsingUIAsCode = InAppMessageView(elementId: "dashboard-announcement-code")
+        // Because the Dashboard screen contains a lot of Views and it's designed using Storyboard, we are
+        // adding this inline View into the UI by adding to a StackView. This allows us to dyanamically add to the Dashboard screen without complexity or breaking any of the constraints set in Storyboard.
+        buttonStackView.addArrangedSubview(newInlineViewUsingUIAsCode)
+
+        // Customers are responsible for setting the width of the View.
```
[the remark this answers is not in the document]
levibostian: The default behavior is that the view does not add any width constraint. We assume that the developer set one for us. As far as what will happen if you don't set a width, that would depend on your app's UI code. AutoLayout can be unpredictable as every layout is different in how it behaves. 

I think this requirement is reasonable when working with AutoLayout in UIKit. Especially when working with Storyboard, you're used to setting all of the constraints of all of the Views in your app. 

Think of it like in Android, you *can* set a default width to let's say match parent. But really, it's reasonable to assume that you're going to set a width on your View when you add it to your app, especially when using XML layouts and getting lint warnings for not setting a width on a View. Compose might be different as we may have default parameters set for width. SwiftUI could probably behavior similarly to Compose in that way. 

Happy to talk about this in more depth if you have more questions.

Sources/Common/Communication/Event.swift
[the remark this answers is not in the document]
Shahroz16: Right now all the events in there are being utilized cross-module, as in track metrics, reset, screen etc. So this would be the only one relevant to `in-app` internal only, so my vote would be to move it there.

Shahroz16 (approved):
Thanks for the changes and additional tickets. 👍

PR #724, opened 2024-05-24
chore: when in-app web content rendered, animate visibility of inline View

levibostian:
Part of: https://linear.app/customerio/issue/MBL-311/provide-positive-ux-when-inline-messages-displayed-by-animating-height

Fetching of in-app message is an async operation running in the background. There is a chance that an inline in-app message is rendered and ready to display on the screen currently in the foreground. We want to display the inline message when it's ready to show, and we also want to display these messages without causing a negative UX.

This commit does the following:
1. Hides the inline View when the View is constructed by setting the height to 0.
2. When the in-app message web content is rendered, the inline View is given a height that matches the web content aspect ratio. The inline View animates it's height from 0 to the new height to make the inline View visible.

Testing:
This change is heavy on UI so testing is done mostly in UIKit sample app QA testing. If you want to test this commit in a sample app build, follow these instructions:
1. Open the app on device.
2. Send yourself a test inline message from Fly to one of the inline Views on Dashboard screen.
3. After you wait a few seconds, you will see the inline message animate into visibility.

Notes for reviewer:
* During development, I encountered difficulties with getting the animation to work when the inline View is embedded in a StackView. Because of this, I modified the Dashboard screen's UI to nest the inline Views inside of the StackView to make sure we QA test this scenario.

Demo: 

https://github.com/customerio/customerio-ios/assets/2041082/a67f9f25-eca6-461a-8726-4381ac3a9fe2

---

Sorry, the tool I use for making stacked PRs, `spr`, was not able to create this PR for me. To help navigate the PR stack, here are some of the previous PRs in the stack:

**Stack**:
- #724 ⬅
- #720

### DIFF
--- a/Sources/MessagingInApp/Views/InAppMessageView.swift
+++ b/Sources/MessagingInApp/Views/InAppMessageView.swift
@@ -67,7 +67,7 @@ public class InAppMessageView: UIView {
     private func setupView() {
         // Customer did not set a height constraint. Create one so the View has one.
         // It's important to have only 1 active constraint for height or UIKit will ignore some constraints.
-        // Try to re-use a constraint is one is already added instead of replacing it. Some scenarios such as
+        // Try to re-use a constraint if one is already added instead of replacing it. Some scenarios such as
         // when UIView is nested in a UIStackView and distribution is .fillProportionally, the height constraint StackView adds is important to keep.
         if viewHeightConstraint == nil {
             heightAnchor.constraint(equalToConstant: 0).isActive = true


Ahmed-Ali (approved):
Thank you Levi for all the work and the impressive outcome.

**Non-blockers:**
* In the sample app, it might be worth it to create separate sub-screen where you can put variety of examples that don't have to match the dashboard layout. This would be great for testing different use cases and also testing.

**Blockers:**
We need to test that model views are still working. A test video showing the scenarios used to verify inline and model views working would also a neat way to show the changes.

I am approving the PR so once you address the above comments, you don't have to wait long before merging.

levibostian:
0 failed tests on CI, even though the run failed. Going to merge this.

PR #728, opened 2024-05-28
fix: action tracking for in-app

Shahroz16:
closes: https://linear.app/customerio/issue/INAPP-12554/in-app-tracked-responses-does-not-match-clicked-metrics-in-env-154794

updates keys of `actionValue` and `actionName` in properties, for `Report Delivery Event` event.


mrehan27 (approved):
Looks good 👍🏻

PR #729, opened 2024-05-28
chore: Log the correct push delegate name

Ahmed-Ali:
Our current push handling logging, like `self.logger.info("Received async completion handler from \(nameOfDelegateClass).")` will always log
"Received async completion handler from UNUserNotificationCenterDelegateWrapper"

This change makes it log the actual user provided delegate.


Shahroz16 (changes requested):
Looks good but lint is failing, once that's fixed, we can merge it

Shahroz16 (approved):
👍

PR #732, opened 2024-06-03
chore: improve reliability dismissing modal messages on page rule change

levibostian:

There is a scenario where a modal message can be displayed on the wrong screen due to a timing issue. Although the time window where this could happen is small, it was encountered during QA testing.

Test case to reproduce:
* In-app message with page rule "Dashboard" sent to device. User on "Dashboard" screen. In-app message begins to render.
* Similar to what is explained [here](https://customerio.slack.com/archives/C06H7MMLH6C/p1716391939537599), if the user of the app clicks a button to begin navigating to another screen in the app and the modal message finishes rendering during the screen transition animation, the modal will be displayed and will remain displayed after the animation is complete and the next screen is shown.

Solution:
Dismiss the modal when the screen transition animation ends and a new page rule is triggered. Ensure that when a Modal message is shown, it does not trigger an automatic screenview event. This allows the SDK to recognize user navigation to a new screen accurately.

Testing:
* Automated tests added to verify auto screenview tracking.
* QA test verified that when a Gist Modal message shown, an automatic screenview event not tracked.

# Based on PR #731

## Full chain of PRs as of 2024-06-14

### DIFF
--- a/Sources/Common/Type/DoNotTrackAutomaticScreenViewEvent.swift
+++ b/Sources/Common/Type/DoNotTrackAutomaticScreenViewEvent.swift
@@ -1,0 +1,5 @@
+import Foundation
+
+/// Inherit in a UIViewController subclass to avoid having an automatic screenview event tracked for it.
+/// Currently only being used internally. `public` only because multiple SDK modules use it.
+public protocol DoNotTrackScreenViewEvent {}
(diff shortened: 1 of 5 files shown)


Shahroz16 (approved):
👍

PR #733, opened 2024-06-04
chore: dismiss message if message expires

levibostian:
https://linear.app/customerio/issue/MBL-315/dismiss-message-if-message-expires

As a marketer I want to stop displaying an in-app message inline by setting a expiration on the in-app message. If a message is being displayed in the app > a fetch gets message from backend for the user > fetch no longer contains the message being displayed, it can be considered expired. If this scenario happens, we want the inline View to gracefully stop showing the in-app message. This is done by animating the height from N to 0 which makes the inline View no longer visible. 

Here is a demo of this behavior: 

https://github.com/customerio/customerio-ios/assets/2041082/b2120702-618e-4787-9897-48999b8aa4b5
> The most important part of the video is when the in-app message animates it's height back to 0 so the in-app message is no longer visible. This is the new behavior implemented in this PR. 

Testing: 
* First, I QA tested this feature in our sample app. 

I tested... 
  1. the animation looks good
  2. if the inline View is currently animating the height from 0 to N, we cancel that animation and begin a new animation back down to 0. 

Testing in-app message expiration is difficult because [messages have a minimum of 1 hour of expiration](https://customerio.slack.com/archives/C036FTE6UJ0/p1717051120718159?thread_ts=1717003069.505409&cid=C036FTE6UJ0). 

Instead of waiting for 1 hour, I simulated this behavior by adding this line of code to the inline View's `setupView()`: 
```swift
// Run block of code in 10 seconds giving time to the SDK to perform fetch, find 1 in-app message I sent myself and render it. 
DispatchQueue.main.asyncAfter(deadline: .now() + 10) { 
   print("InAppMessageView: Simulating message expired.")
   Gist.shared.messageQueueManager.processFetchedMessages([]) // simulate message expired
   // This line of code above executes in the SDK after a fetch to the backend server is completed. Provide an empty array to indicate that the in-app message previously fetched has expired on the backend and so it's no longer being returned. 
}
```

The demo video was recorded with this block of code used. 

* After QA testing was done, I wrote automated tests. All of the test cases I was QA testing have been added to the automated test suite. 

# Based on PR #724

## Full chain of PRs as of 2024-06-04

* PR #733: `levi/inline-dismiss-when-expire` ➔ `levi/inline-animate-open`
* PR #724: `levi/inline-animate-open` ➔ `levi/display-inline-uiview-afterfetch`
* PR #720: `levi/display-inline-uiview-afterfetch` ➔ `spr/main/5ab2c403`
* PR #718: `spr/main/5ab2c403` ➔ `spr/main/025c7bf5`
* PR #717: `spr/main/025c7bf5` ➔ `feature-inline-inapp`


Makefile
levibostian, line 1: After adding EngineWeb and EngineWebProvider to auto-generated files, auto-generated files need `UIKit`.

levibostian: I copied this change from [this PR](https://github.com/customerio/customerio-ios/pull/731) which also adds testing support to the in-app SDK.

Sources/MessagingInApp/Gist/EngineWeb/EngineWeb.swift
levibostian, line 1: I copied this change from [this PR](https://github.com/customerio/customerio-ios/pull/731) which also adds testing support to the in-app SDK. 

By mocking the EngineWeb, we are able to do the same thing as if we were mocking HTTP requests in our SDK. We can mock the process of rendering in-app messages from the web server.

Sources/MessagingInApp/Gist/EngineWeb/EngineWebProvider.swift
levibostian, line 1: See `MessageManager` for use of the provider.

Sources/MessagingInApp/Views/InAppMessageView.swift
```diff
@@ -132,32 +134,48 @@ public class InAppMessageView: UIView {
 
         inlineMessageManager = newInlineMessageManager
     }
+
+    private func dismissInAppMessage() {
+        // If this function gets called a lot in a short amount of time (eventbus triggers multiple events), the dismiss animation does not look as expected.
+        // To fix this, exit early if dimiss has already been triggered.
+        if inlineMessageManager?.inlineMessageDelegate == nil {
+            return // dismiss already triggered, exit early
```
levibostian, line 142:
```suggestion
            return
```

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
```diff
@@ -46,42 +56,49 @@ class InAppMessageViewTest: UnitTest {
...
 
-    func test_displayInAppMessage_givenNoMessageAvailable_expectDoNotDisplayAMessage() {
+    @MainActor
+    func test_displayInAppMessage_givenNoMessageAvailable_expectDoNotDisplayAMessage() async {
         queueMock.getInlineMessagesReturnValue = []
 
         let inlineView = InAppMessageView(elementId: .random)
 
-        XCTAssertNil(getInAppMessageWebView(fromInlineView: inlineView))
+        XCTAssertFalse(isDisplayingInAppMessage(inlineView))
```
levibostian, line 65: I changed some of the `getInAppMessageWebView` calls to the new `isDisplayingInAppMessage` to make the test function more accurate checking if the inline View is displaying a in-app message. By testing the height of the View is not 0, we are more certain that the inline View is visible/displayed to the user.

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
levibostian, line 1: Besides adding new test functions, this file needed other changes. 

* I added a `onDoneRenderingInAppMessage` function to run the code that animates the height from 0 to N to display in-app messages. I designed `onDoneRenderingInAppMessage` to not return until the `sizeChanged()` function is called on the inline View. 

* Because `onDoneRenderingInAppMessage` is now `async`, I needed to make all test functions `async` to use it. When you do that, you can no longer create `InAppMessageView` instances in the test function unless you assign the test function to the main thread (`@MainActor`).

Ahmed-Ali:
How will the changes in [that pr](https://github.com/customerio/customerio-ios/pull/720) impact this one?
Also is there a chance to break down this PR farther? I wanted to review today but there were many things that I thought I would get back to it later when I have more capacity to go through it all properly. There are other changes that I could have easily approved. hence the ask for a break down.

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
Shahroz16, line 1: Great collection of tests! few areas that I feel could use some tests
- Multiple event bus events, 

something like

```
queueMock.getInlineMessagesReturnValue = []
    DIGraphShared.shared.eventBusHandler.postEvent(InAppMessagesFetchedEvent())
    await waitForMainThreadToFinishPendingTasks()

    queueMock.getInlineMessagesReturnValue = [secondMessage]
    DIGraphShared.shared.eventBusHandler.postEvent(InAppMessagesFetchedEvent())
    await onDoneRenderingInAppMessage(secondMessage)
```

- Multiple messages returned, should we be getting the one with higher priority to be displayed? 
```
    queueMock.getInlineMessagesReturnValue = [firstMessage, secondMessage]
```

- Delayed Setting of element ID? 

- Multiple instances of InAppMessageView operate independently without interference

- Changes to dynamic constraints? we are already changing height constraints, if user changes height/width what end height width constraint are we expecting?

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
```diff
@@ -104,42 +121,80 @@ class InAppMessageViewTest: UnitTest {
...
-        let givenNewInlineMessage = Message(messageId: .random, campaignId: .random, elementId: givenOldInlineMessage.elementId)
+        let webViewBeforeFetch = getInAppMessageWebView(fromInlineView: inlineView)
 
-        simulateSdkFetchedMessages([givenNewInlineMessage])
+        await simulateSdkFetchedMessages([givenInlineMessage])
 
         let webViewAfterFetch = getInAppMessageWebView(fromInlineView: inlineView)
 
-        // If the WebViews are different, it means the message was reloaded.
+        // If the WebViews are the same instance, it means the message was not reloaded.
```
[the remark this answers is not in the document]
levibostian: Since the single-source-of-truth is the UI, I like how this test does an assertion on the View layer. I don't know if the test would increase in accuracy by doing this check. Can you think of a reason to?

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
```diff
@@ -6,15 +6,24 @@ import XCTest
...
+        engineWebMock.view = UIView()
+
         DIGraphShared.shared.override(value: queueMock, forType: MessageQueueManager.self)
+        DIGraphShared.shared.override(value: engineProvider, forType: EngineWebProvider.self)
     }
 
     // MARK: View constructed
 
+    @MainActor
     func test_whenViewConstructedUsingStoryboards_expectCheckForMessagesToDisplay() {
```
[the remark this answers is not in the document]
levibostian: Adding.

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
levibostian: Thanks for the test suggestions. You put a lot of thought into that, thanks! 

Let me try and put these in my own words to see if we align. 

* Multiple event bus events

I can think of these tests: 
1. Given fetch event returning 0 events, expect View to not display any message. Then another fetch happens, event gets returned, we expect the View to display that message. 
2. Given fetch event returning N events, but none with element Id matching the View. We expect that the View does not display any messages. Then a fetch happens returning a message with element id, we expect the View to display that message. 

* Multiple messages returned, should we be getting the one with higher priority to be displayed?

Priority logic is in the business layer and tests exist for that so I don't see a need to add a test for that in the View. 

* Delayed Setting of element ID?

A test like this? Given no element id set of View yet, fetch occurs returns messages with a matching element id, then an element id is set on View, we expect that the inline View displays a message. 

* Multiple instances of InAppMessageView operate independently without interference

Good idea. Tests where 2 Views are created with matching IDs and non-matching IDs would be good tests. 

* Changes to dynamic constraints? we are already changing height constraints, if user changes height/width what end height width constraint are we expecting?

I will need some help with this one. Could you give an example of tests cases?

Shahroz16: You are spot on all the additions of tests you mentioned. We are on the same page.

> Changes to dynamic constraints? we are already changing height constraints, if user changes height/width what end height width constraint are we expecting?

I was referring to, if user set the height and width of the view at beginning, and when we inflate the in-app view, the final height and width of the controller should be what we expect it not what user set its.

Similarly, if we inflate the view and set the height and width of the controller, if user updates the width/height, what size are we expecting at the end?

levibostian: I've looked into each of the suggestions. Either the test case is now covered, or I suggest a follow-up PR to introduce it. See my responses below: 

### Multiple event bus events

* Test case: Given fetch event returning 0 events, expect View to not display any message. Then another fetch happens, event gets returned, we expect the View to display that message. 

This is covered in test `test_givenInAppMessageFetchedAfterViewConstructed_expectShowInAppMessageFetched`

* Test case: Given fetch event returning N events, but none with element Id matching the View. We expect that the View does not display any messages. Then a fetch happens returning a message with element id, we expect the View to display that message. 

This is going to require some refactors to the test class to be more of an integration test then a unit test as it is now. Because this will require some refactors to the test suite, I suggest we don't block this PR for it and we introduce this change in a follow-up PR. 

### Delayed Setting of element ID

* Test case: Given no element id set of View yet, fetch occurs returns messages with a matching element id, then an element id is set on View, we expect that the inline View displays a message. 

This is covered in test `test_whenViewConstructedUsingStoryboards_expectCheckForMessagesToDisplay`, since a delay in setting element ID is only with constructing a View used in a Storyboard. 

### Multiple instances of InAppMessageView operate independently without interference

Same response as above when referring to integration tests. Same suggestion applies - unblock this PR and introduce follow-up PR. 

### Changes to dynamic constraints

If user set the height and width of the view at beginning, and when we inflate the in-app view, the final height and width of the controller should be what we expect it not what user set its.

Similarly, if we inflate the view and set the height and width of the controller, if user updates the width/height, what size are we expecting at the end?

I'm looking into these. Haven't written tests like this before so I'll come back with results.

Tests/MessagingInApp/MessagingInAppIntegrationTests.swift
```diff
@@ -264,6 +264,7 @@ extension MessagingInAppIntegrationTest {
     }
 
     func onDoneFetching(messages: [Message]) {
+        // swiftlint:disable:next force_cast
         (Gist.shared.messageQueueManager as! MessageQueueManagerImpl).processFetchedMessages(messages)
```
levibostian, line 268: I prefer to not have private functions such as `processFetchedMessages` as part of the `MessageQueueManager` protocol as the protocol is meant to be the `public` API that the SDK calls and only the tests uses `processFetchedMessages`. 

Instead of directly calling `processFetchedMessages`, we could mock the `GistQueueNetwork` HTTP request which then calls `processFetchedMessages`. But I believe that change would go beyond the scope of this PR and may not be worth the effort when this call does the same thing.

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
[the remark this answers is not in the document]
levibostian: Added 2 new tests to check that after Views are constructed they start with a height of 0 (aka: dismissed state).

Shahroz16 (approved):
Looking forward to the later PRs with the tests.

PR #738, opened 2024-06-13
test: convert async operations to sync to increase reliability of modal page rule integration tests

levibostian:
Follow-up PR to [dismissing modal messages when changing screens](https://github.com/customerio/customerio-ios/pull/731). 

We have automated tests for testing if a modal message is dismissed when you change the page rule. However, the tests could be more reliable. There are some test cases with false positives in them because of async operations (view animations). 

This PR is one step towards making the tests more reliable by disabling modal view animations in automated tests. By having the code run synchronously, checking the state of the modal messages is more reliable.


levibostian:
Note: There are 2 test functions skipped [in this PR](https://github.com/customerio/customerio-ios/pull/742). Assuming that PR gets merged and then this one, make sure to remove that skip in this PR.

Sources/MessagingInApp/Gist/Managers/ModalViewManager.swift
```diff
@@ -10,8 +11,10 @@ class ModalViewManager {
     var window: UIWindow?
     var viewController: GistModalViewController!
     var position: MessagePosition
-    var isShowingMessage: Bool {
-        window != nil
+    var isShowingMessage: Bool = false // true if the modal is visible to the user. Meaning, all operations are complete including animations.
```
levibostian, line 14: One part of making the tests more reliable is changing `isShowingMessage` to only be `true` if *all* of the animations have been completed. 

In an app, we do not consider a message as shown until all View animations are completed. Our tests should match that and now they do.

Sources/MessagingInApp/Gist/Managers/ModalViewManager.swift
```diff
@@ -40,13 +43,15 @@ class ModalViewManager {
...
             self.viewController.view.center.y = finalPosition
         }, completion: { _ in
-            UIView.animate(withDuration: 0.1, delay: 0, options: [.curveEaseIn], animations: {
+            self.animationRunner.animate(withDuration: 0.1, delay: 0, options: [.curveEaseIn], animations: {
                 self.viewController.view.backgroundColor = UIColor.black.withAlphaComponent(0.2)
-            }, completion: nil)
-            completionHandler()
+            }, completion: { _ in
+                self.isShowingMessage = true
+                completionHandler()
```
levibostian, line 53: `self.isShowingMessage = true` is only set until after all animations are done running.

Tests/MessagingInApp/MessagingInAppIntegrationTests.swift
```diff
@@ -47,6 +47,48 @@ class MessagingInAppIntegrationTest: IntegrationTest {
         XCTAssertFalse(didCallGlobalEventListener)
     }
 
+    func test_givenNavigateToDifferentScreenWhileMessageAnimatingIntoView_expectDoNotShowModalMessage() {
```
levibostian, line 50: `ViewAnimationRunnerStub` allows us to have control over when an animation finishes running. 

Because of this, I wrote new tests for scenarios that could happen in an app. We expect the SDK dismisses a modal if the rendering is done but the modal has not animated into view yet.

Tests/MessagingInApp/Util/ViewAnimationRunnerStub.swift
```diff
@@ -0,0 +1,19 @@
+@testable import CioMessagingInApp
+import Foundation
+import UIKit
+
+class ViewAnimationRunnerStub: ViewAnimationRunner {
```
levibostian, line 5: This is how we can mock view animations to make them synchronous which makes our tests more reliable and easier to write.

Sources/MessagingInApp/Gist/Gist.swift
```diff
@@ -32,7 +32,11 @@ public class Gist: GistDelegate {
     func reset() {
         clearUserToken()
         messageQueueManager = MessageQueueManager()
-        messageManagers = []
+
+        for messageManager in messageManagers {
+            removeMessageManager(instanceId: messageManager.currentMessage.instanceId)
+        }
```
levibostian, line 38: After mocking animations, I was getting crashes in the test suite when `messageManagers = []` was running. The error message was a memory access error because `messageManagers` was being accessed while clearing it. 

After debugging the error, I decided to only modify the test suite code instead of making a refactor to the SDK code and possibly break a feature in the SDK.

Tests/MessagingInApp/MessagingInAppIntegrationTests.swift
```diff
@@ -269,15 +311,23 @@ extension MessagingInAppIntegrationTest {
...
+     The goal of this function is to determine the message that is visible on the screen to the user.
+     To try and make the tests the most accurate, access the currently displayed message from the UI layer of the code instead of business logic.
+     */
     var currentlyShownModalMessage: Message? {
-        guard let messageManager = Gist.shared.getModalMessageManager() else {
-            return nil // no modal message shown or loading
+        // Unfortunately, this method of determining visible modals is not 100% reliable.
+        // We are accessing the visible ViewController by getting manager instances from the Gist class and the Gist class destroys and creates
+        // instances of managers.
+        // It would be better if we could talk directly to the UI layer to get the visible ViewController or Gist has only 1 modal manager like the Android SDK days.
```
levibostian, line 322: As stated in the PR description, this PR is one step to making our modal integration tests more reliable. This call-out is suggesting another step to make the tests more reliable. 

There are some tests that check if a modal has been dismissed. While the test pass, it is not truly testing is a modal is shown in the UI or not. 

I think a great follow-up PR would be a refactor to the Gist class which makes modal message manager a singleton, just like the Android SDK does today. This would allow us to check the state of the 1 modal message in the app, which would make our tests more reliable.

Sources/MessagingInApp/Gist/Managers/ModalViewManager.swift
Shahroz16, line 1: should we be checking 
`guard !isShowingMessage else { return }` in methods like `showModalView`?

Sources/MessagingInApp/Gist/Managers/ModalViewManager.swift
```diff
@@ -40,13 +43,15 @@ class ModalViewManager {
             finalPosition = viewController.view.center.y - viewController.view.bounds.height
         }
 
-        UIView.animate(withDuration: 0.3, delay: 0, options: [.curveEaseIn], animations: {
+        animationRunner.animate(withDuration: 0.3, delay: 0, options: [.curveEaseIn], animations: {
```
Shahroz16, line 46: more of an SDK change so happy to not include but would it make more sense to use `weak self`  over here?  incase If the ModalViewManager is deallocated during an ongoing animation, the animation will simply stop without causing a crash.

Tests/MessagingInApp/Util/ViewAnimationRunnerStub.swift
Shahroz16, line 1: should we add something like 

```
func testSlowMotionAnimation() {
        viewAnimationRunnerStub.animationSpeed = 0.1 // 10x slower
        modalViewManager.showModalView {}
        XCTAssertFalse(modalViewManager.isShowingMessage)
        
        // Wait for animation to complete
        let expectation = XCTestExpectation(description: "Animation completed")
        DispatchQueue.main.asyncAfter(deadline: .now() + 3) { // 0.3 seconds * 10
            expectation.fulfill()
        }
        wait(for: [expectation], timeout: 5)
        
        XCTAssertTrue(modalViewManager.isShowingMessage)
    }
 ```
 
- By using DispatchQueue.main.asyncAfter, we're simulating the actual timing of animations more accurately.
- This allows us to test scenarios where timing is crucial, such as interrupting an ongoing animation or checking the state of the UI at different points during an animation.

Shahroz16 (approved):

PR #739, opened 2024-06-13
chore: instantly dismiss modal instead of async animation dismissal

levibostian:

When the page route changes in the SDK, we need to have the current modal dismissed and ready to display a new message. Currently, the code  dismisses the modal with an animation, so it's an async operation. To reduce edge cases in the code, modify the dismissing of messages to be a synchronous/instant operation.

Testing:
* In sample app, I manually triggered the cancel/dismiss code to confirm that the modal dismisses instantly and when I change screens, I expect that a new message is able to show.

Note to reviewers
* This is the behavior implemented in the Android SDK.

# Based on PR #732

## Full chain of PRs as of 2024-06-13

* PR #739: `levi/instantly-dismiss-modals` ➔ `levi/ignore-cio-autoscreenviews`


Ahmed-Ali (approved):
Overall nothing stood out except the logic to dismiss the in-app on page rule change.
Are there scenarios where a page rule can match more than one page?

If the answer is yes, then please don't merge till we discuss it

PR #740, opened 2024-06-14
chore: do not dismiss modal when page route changed and regex pattern matches

levibostian:
# Based on PR #739

## Full chain of PRs as of 2024-06-14

* PR #740: `levi/compare-regex-page-rules` ➔ `levi/instantly-dismiss-modals`
* PR #739: `levi/instantly-dismiss-modals` ➔ `levi/ignore-cio-autoscreenviews`

Because page rules can contain regex, you could change screens in the app and the currently shown message can still match that screen. Do not cancel if the message matches the new route.

Testing
* Added automated tests.

### DIFF
--- a/Sources/MessagingInApp/Gist/Managers/MessageQueueManager.swift
+++ b/Sources/MessagingInApp/Gist/Managers/MessageQueueManager.swift
@@ -122,17 +122,10 @@ class MessageQueueManager {
 
         let position = message.gistProperties.position
 
-        if let routeRule = message.gistProperties.routeRule {
-            let cleanRouteRule = routeRule.replacingOccurrences(of: "\\", with: "/")
-            if let regex = try? NSRegularExpression(pattern: cleanRouteRule) {
-                let range = NSRange(location: 0, length: Gist.shared.getCurrentRoute().utf16.count)
-                if regex.firstMatch(in: Gist.shared.getCurrentRoute(), options: [], range: range) == nil {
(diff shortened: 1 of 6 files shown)


Sources/MessagingInApp/Gist/Managers/MessageQueueManager.swift
```diff
@@ -122,17 +122,10 @@ class MessageQueueManager {
...
-            let cleanRouteRule = routeRule.replacingOccurrences(of: "\\", with: "/")
-            if let regex = try? NSRegularExpression(pattern: cleanRouteRule) {
-                let range = NSRange(location: 0, length: Gist.shared.getCurrentRoute().utf16.count)
-                if regex.firstMatch(in: Gist.shared.getCurrentRoute(), options: [], range: range) == nil {
-                    Logger.instance.debug(message: "Current route is \(Gist.shared.getCurrentRoute()), needed \(cleanRouteRule)")
-                    return // exit early to not show the message since page rule doesnt match
-                }
-            } else {
-                Logger.instance.info(message: "Problem processing route rule message regex: \(cleanRouteRule)")
-                return // exit early to not show the message since we cannot parse the page rule for message.
```
levibostian, line 135: This PR does not introduce new logic for page rule regex matching. Here is the code that exists in production today. This PR takes this logic, puts it into a re-usable extension in `Message.swift`, and then calls that logic in multiple places.

Ahmed-Ali (approved):
Thanks Levi for following up the earlier [comments with these changes](https://github.com/customerio/customerio-ios/pull/739#pullrequestreview-2118311775)

PR #743, opened 2024-06-17
fix: differentiate every push event handler installed in app

levibostian:
https://linear.app/customerio/issue/MBL-383/[bug]-customer-unable-to-handle-push-notification-event-on-ios-with

At runtime when `PushEventHandlerProxy.addPushEventHandler` is called, the given push event handler will overrwrite any previously added push event handlers. This results in the SDK only being aware of 1 push event handler even if the iOS app has N number installed.

Testing
* In Flutter sample app, install CIO SDK, FlutterFire, and set `UNNotificationCenterDelegate.current().delegate = self` in `AppDelegate` for the host app to also handle pushes. This reproduces the bug because we now have at least 2 other push click handlers besides the CIO SDK installed. Set a Xcode breakpoint in `PushEventHandlerProxy.addPushEventHandler` and check what dictionary key is generated. 
* I added new automated tests around `PushEventHandlerProxy.addPushEventHandler` function.

### DIFF
--- a/Sources/MessagingPush/PushHandling/PushEventHandlerProxy.swift
+++ b/Sources/MessagingPush/PushHandling/PushEventHandlerProxy.swift
@@ -65,11 +65,11 @@ class PushEventHandlerProxyImpl: PushEventHandlerProxy {
 
                     // Using logs to give feedback to customer if 1 or more delegates do not call the async completion handler.
                     // These logs could help in debuggging to determine what delegate did not call the completion handler.
-                    self.logger?.info("Sending push notification, \(pushAction.push.title), event to: \(nameOfDelegateClass)). Customer.io SDK will wait for async completion handler to be called...")
+                    self.logger?.info("Sending push notification, \(pushAction.push.title), action event to: \(nameOfDelegateClass)). Customer.io SDK will wait for async completion handler to be called...")
 
                     delegate.onPushAction(pushAction) {
                         Task { @MainActor in // in case the delegate calls the completion handler on a background thread, we need to switch back to the main thread.
-                            self.logger?.info("Received async completion handler from \(nameOfDelegateClass).")
+                            self.logger?.info("Received async completion handler from \(nameOfDelegateClass) for action push event.")
 
                             if !hasResumed {
                                 hasResumed = true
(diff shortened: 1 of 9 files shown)


Tests/MessagingPush/PushHandling/PushEventHandlerProxyTest.swift
```diff
@@ -3,7 +3,7 @@ import Foundation
 import SharedTests
 import XCTest
 
-class PushEventHandlerProxyTest: UnitTest {
+class PushEventHandlerProxyTest: IntegrationTest {
```
levibostian, line 6: Only changed so the class had access to the new `getNewPushEventHandler()` function. Should have no effect on the tests besides that.

Tests/MessagingPush/PushHandling/PushEventHandlerProxyTest.swift
```diff
@@ -12,13 +12,47 @@ class PushEventHandlerProxyTest: UnitTest {
         proxy = PushEventHandlerProxyImpl()
     }
 
+    // MARK: addPushEventHandler
```
levibostian, line 15: These new tests catch the bug that this PR fixes.

Sources/MessagingPush/UserNotificationsFramework/Wrappers.swift
```diff
@@ -120,6 +120,10 @@ public struct UNNotificationWrapper: PushNotification {
 class UNUserNotificationCenterDelegateWrapper: PushEventHandler {
     private let delegate: UNUserNotificationCenterDelegate
 
+    var identifier: String {
+        String(describing: delegate)
```
levibostian, line 124: By returning the instance description of the nested delegate, this is what will uniquely identify `UNUserNotificationCenterDelegateWrapper` instances from 1 another. Since `UNUserNotificationCenterDelegateWrapper` is just a wrapper, it's the nested delegate that we really care about.

Sources/MessagingPush/PushHandling/PushEventHandlerProxy.swift
```diff
@@ -27,7 +27,7 @@ class PushEventHandlerProxyImpl: PushEventHandlerProxy {
     public static let shared = PushEventHandlerProxyImpl()
 
     // Use a map so that we only save 1 instance of a given handler.
-    @Atomic private var nestedDelegates: [String: PushEventHandler] = [:]
+    @Atomic var nestedDelegates: [String: PushEventHandler] = [:]
```
levibostian, line 30: made `internal` for tests to access

Sources/MessagingPush/PushHandling/PushEventHandlerProxy.swift
```diff
@@ -37,7 +37,7 @@ class PushEventHandlerProxyImpl: PushEventHandlerProxy {
     }
 
     func addPushEventHandler(_ newHandler: PushEventHandler) {
-        nestedDelegates[String(describing: newHandler)] = newHandler
+        nestedDelegates[newHandler.identifier] = newHandler
```
levibostian, line 40: The bug fix. 

At runtime, multiple `UNUserNotificationCenterDelegateWrapper` instances will call this function. Using `identifier` makes sure that each handler is uniquely identified.

Shahroz16 (approved):
LGTM

PR #746, opened 2024-06-21
fix: display in-app messages with page rules set and kill/re-open app

levibostian:

Persist the queue of in-app messages to disk on iOS, similar to the existing implementation on Android and web platforms.

Testing:
* Automated tests have been written and will be introduced in a future PR to keep the amount of changes low in this PR. The SDK is not setup for testing this code easily so some test setup code needs added first.

Reviewer notes:
* I implemented a network-level HTTP response body cache, similar to the Android and web SDKs. I considered caching the `MessageQueueManager` local message queue, but it required more code changes then this. I like the network-level approach since this is a requirement of the SDK based on the HTTP communication with the Gist backend server.


Sources/MessagingInApp/Gist/Managers/QueueManager.swift
```diff
@@ -18,24 +29,28 @@ class QueueManager {
                         self.updatePollingInterval(headers: response.allHeaderFields)
                         switch response.statusCode {
                         case 204:
+                            self.cachedFetchUserQueueResponse = nil
```
Shahroz16, line 32: hmm thats bit is different from android, we are not clearing cache on `204`.  Logic behind it being, `304` is content not changed and `204` is when there is no response needed from the server. 

So i don't expect this server to return `204`, this was probably wrong code being sent and then they updated it to `304`. We should probably confirm it.

Shahroz16 (approved):

Sources/MessagingInApp/Gist/Managers/QueueManager.swift
levibostian: The fact that the SDK is returning an empty array on the next line of code is what makes me believe that the cache should be cleared. 

If the SDK returns an empty string, it would mean that the user has no messages in their queue and therefore, we should not have a cache. Or, we could still have a cache, but it would be an empty array instead of nil. 

[I am seeing that Android is not handling 204](https://github.com/customerio/customerio-android/blob/bfd8e28158c56e97e74f6ab8634200756b6545c3/messaginginapp/src/main/java/io/customer/messaginginapp/gist/data/listeners/Queue.kt#L74), which seems odd to me. Before merging this, I am going to check with the in-app squad.

levibostian: [Looks like 204 does not need handled](https://customerio.slack.com/archives/C036FTE6UJ0/p1719331133922019?thread_ts=1719323682.517399&cid=C036FTE6UJ0). I'll remove that code.

PR #748, opened 2024-06-21
test: write integration tests for caching in-app messages on 304

levibostian:
# Based on PR #747

## Full chain of PRs as of 2024-06-21

* PR #748: `levi/cache-inapp-responses-integration-tests` ➔ `levi/cache-inapp-response-test-setup`
* PR #747: `levi/cache-inapp-response-test-setup` ➔ `levi/cache-inapp-responses-at-network-level`
* PR #746: `levi/cache-inapp-responses-at-network-level` ➔ `main`

Write integration tests for the new network-layer in-app message queue cache. In particular, these tests need to test the managing of the network cache as well as how the cache is used when the SDK receives a 304 response.

Reviewer notes:
* The cache logic exists in `QueueManager` and so I could have created unit tests around functions in that class. Originally I did this but found that the tests were really hard to write because I was dealing with multiple different data types and converting between them all. Also, the network cache exists ultimately to populate the `MessageQueueManager` local message queue so an integration test made the most sense for me.
* For re-usability and simplicity, I opted into creating a .json file with a sample HTTP response body in it for the test sample data. There are cons to this approach, in particular it's hard to modify the sample data if you want. Although I prefer using Swift structs for mock HTTP responses, it requires significant code refactoring. Specifically, the use of `[String: Any]` in `UserQueueResponse` and `Message` complicates making these structures Codable, which is necessary for JSON conversion.
* For the contents of the JSON file, I copy/pasted it from a HTTP response I received while running a sample app and I sent in-app messages from Fly to my device.


mrehan27 (approved):
Looks good 👍🏻

PR #749, opened 2024-06-21
chore: clear in-app message cache on sdk logout

levibostian:
# Based on PR #748

## Full chain of PRs as of 2024-06-21

* PR #749: `levi/cache-inapp-response-clear-on-logout` ➔ `levi/cache-inapp-responses-integration-tests`
* PR #748: `levi/cache-inapp-responses-integration-tests` ➔ `levi/cache-inapp-response-test-setup`
* PR #747: `levi/cache-inapp-response-test-setup` ➔ `levi/cache-inapp-responses-at-network-level`
* PR #746: `levi/cache-inapp-responses-at-network-level` ➔ `main`

When SDK is logged out, clear the in-app network cache to prevent displaying messages to the wrong Journeys profile.

Testing:
* Automated tests added.
* QA test case in the linear ticket passed in sample apps.

Reviewer notes:
* This implementation was chosen to match the behavior of Android 3.11.0

### DIFF
--- a/Tests/MessagingInApp/MessagingInAppIntegrationTests.swift
+++ b/Tests/MessagingInApp/MessagingInAppIntegrationTests.swift
@@ -176,6 +176,27 @@ class MessagingInAppIntegrationTest: IntegrationTest {
         XCTAssertNotNil(currentlyShownModalMessage)
     }
 
+    // The in-app SDK maintains a cache of messages that are returned from the backend. When a profile is logged out of the SDK, we expect the message cache is cleared otherwise we run the risk of displaying messages meant for profile A to profile B.
+    func test_clearUserToken_givenProfileLoggedOutAndNewProfileLoggedIn_expectLocalMessageCacheCleared() {
+        Gist.shared.setUserToken("profile-A")
+
+        XCTAssertTrue(Gist.shared.messageQueueManager.localMessageStore.isEmpty)
+        setupHttpResponse(code: 200, body: readSampleDataFile(subdirectory: "InAppUserQueue", fileName: "fetch_response.json").data)
+        Gist.shared.messageQueueManager.fetchUserMessages()
+        XCTAssertFalse(Gist.shared.messageQueueManager.localMessageStore.isEmpty)
+
+        // Expect no messages immediately after logging into another profile.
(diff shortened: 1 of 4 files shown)


Tests/MessagingInApp/MessagingInAppIntegrationTests.swift
```diff
@@ -176,6 +176,27 @@ class MessagingInAppIntegrationTest: IntegrationTest {
...
+        Gist.shared.clearUserToken()
+        XCTAssertTrue(Gist.shared.messageQueueManager.localMessageStore.isEmpty)
+        Gist.shared.setUserToken("profile-B")
+        XCTAssertTrue(Gist.shared.messageQueueManager.localMessageStore.isEmpty)
+
+        // Expect that after first fetch with new profile logged in, the message cache remains empty.
+        setupHttpResponse(code: 304, body: "".data)
+        Gist.shared.messageQueueManager.fetchUserMessages()
+        XCTAssertTrue(Gist.shared.messageQueueManager.localMessageStore.isEmpty)
+    }
```
mrehan27, line 198: IIRC, this creates an issue if `"profile-A"` logs in back and keep getting 304, as we can't load messages we cleared before. But that's something we might need to discuss and fix for Android too. Since this implementation matches with current one on Android, I'm good with it.

mrehan27 (approved):

Tests/MessagingInApp/MessagingInAppIntegrationTests.swift
levibostian: You are correct, this is an issue on Android as well. 

I talked with Shahroz about this scenario, too. This could require a backend change to handle this because this is possible: 

* Profile-A logs in. Gets a 200. 
* User uninstalls the app. 
* User re-installs the app, logs in as Profile-A. 
* Gets a 304 and SDK has no cache to read from because of app re-install. 

This might be a rare scenario that only a small number of users will encounter so I am not sure how high of a priority it is. But for a scenario like this, the SDK is limited on what it can do.

mrehan27: Agreed, thanks for the details 👍🏻

PR #751, opened 2024-06-25
fix: differentiate every push event handler installed in app

levibostian:
https://linear.app/customerio/issue/MBL-383/[bug]-customer-unable-to-handle-push-notification-event-on-ios-with

(This PR is identical to the one merged into v2: https://github.com/customerio/customerio-ios/pull/743)

Summary of bug: The iOS SDK had a bug where it was only able to store 1 3rd party push click handing SDK and ignore others that were installed. After this fix, all 3rd party SDKs installed in the app are called for push events in the app.

Testing: 
* Automated tests in this PR reproduce the bug and confirm the bug has been fixed. 
* I tested this in FCM sample app. See [this follow-up PR](https://github.com/customerio/customerio-ios/pull/752) to learn how I tested this.


Shahroz16 (commented):
Changes look similar to v2 ! 👍

PR #756, opened 2024-06-28
chore: after close button pressed show next message in local queue 

levibostian:
https://linear.app/customerio/issue/MBL-313/after-close-action-pressed-replace-in-app-message-with-next-in-queue 

Problem to solve: 
As a marketer, I want to display another in-app message inline when close button pressed so I can rotate between multiple in-app messages.

Solution: 
When close button is clicked on an inline View, check the local queue to see if there is another message to display for the element id. 

If there is another message...
1. Remove the current WebView from the inline View and cleanup resources. 
2. Begin rendering the next message. 
3. When next message ready to display, display it by animating the height to match the new message to display. 

If there is not another message, dismiss the inline View. 

Not in scope: 
* When you click the close button and there is another message to display, the currently shown WebView is instantly removed from the UI. It doesn't look the best. Follow-up PRs will add [animations to make this transition look nicer](https://linear.app/customerio/issue/MBL-313/after-close-action-pressed-replace-in-app-message-with-next-in-queue#comment-17c168a1). 

Testing: 
* I wrote automated tests for all use cases I could think of. 
* I tested in UIKit sample app. QA test case: 
1. Send 2 in-app messages to a device with same element id. 
2. You should see 1 message visible on the screen. Click the close button on the message. 
3. You should see the message instantly disappear. ~1-2 seconds later, after the next message is rendered, you should see the next message animate into visiblity on the screen. 
4. Click the close button 1 more time. You should see the inline View dismiss by animating height to 0. 

Here is a demo video showing this off: 

https://github.com/customerio/customerio-ios/assets/2041082/a167dbac-d077-4814-bc01-37f89a9d2c09

### DIFF
--- a/Sources/MessagingInApp/Gist/Managers/InlineMessageManager.swift
+++ b/Sources/MessagingInApp/Gist/Managers/InlineMessageManager.swift
@@ -27,6 +27,12 @@ class InlineMessageManager: MessageManager {
 
     weak var inlineMessageDelegate: InlineMessageManagerDelegate?
 
+    override func stopAndCleanup() {
+        inlineMessageDelegate = nil
+
+        super.stopAndCleanup()
+    }
+
     override func onReplaceMessage(newMessageToShow: Message) {
         // Not yet implemented. Planned in future update.
     }
--- a/Sources/MessagingInApp/Gist/Managers/InlineMessageManager.swift
+++ b/Sources/MessagingInApp/Gist/Managers/InlineMessageManager.swift
@@ -18,9 +18,11 @@ protocol InlineMessageManagerDelegate: AnyObject {
  */
 class InlineMessageManager: MessageManager {
     var inlineMessageView: GistView? {
-        let view = super.gistView
-        view?.delegate = self
-        return view
+        if super.gistView.delegate == nil {
+            super.gistView.delegate = self
+        }
+
+        return super.gistView
     }
 
     weak var inlineMessageDelegate: InlineMessageManagerDelegate?
--- a/Sources/MessagingInApp/Gist/Managers/InlineMessageManager.swift
+++ b/Sources/MessagingInApp/Gist/Managers/InlineMessageManager.swift
@@ -17,7 +17,7 @@ protocol InlineMessageManagerDelegate: AnyObject {
  ```
  */
 class InlineMessageManager: MessageManager {
-    var inlineMessageView: GistView? {
+    var inlineMessageView: GistView {
         if super.gistView.delegate == nil {
             super.gistView.delegate = self
         }
--- a/Sources/MessagingInApp/Gist/Managers/MessageManager.swift
+++ b/Sources/MessagingInApp/Gist/Managers/MessageManager.swift
@@ -19,7 +19,7 @@ class MessageManager {
     var engine: EngineWebInstance
     private let siteId: String
     let currentMessage: Message
-    var gistView: GistView!
+    let gistView: GistView
     private var currentRoute: String
     private var elapsedTimer = ElapsedTimer()
     weak var delegate: GistDelegate?
@@ -45,11 +45,22 @@ class MessageManager {
         elapsedTimer.start(title: "Loading message with id: \(currentMessage.templateId)")
 
         self.engine = engineWebProvider.getEngineWebInstance(configuration: engineWebConfiguration)
-        engine.delegate = self
         self.gistView = GistView(message: currentMessage, engineView: engine.view)
+        engine.delegate = self
     }
 
     deinit {
+        self.stopAndCleanup()
+    }
+
+    // The manager instance is no longer needed. Create a new instance when you want to display another message.
+    // Note: This function does not remove the WebView from the view hierarchy. Do that from the UI layer.
+    func stopAndCleanup() {
+        // First, stop sending events to the delegates.
+        delegate = nil
+        gistView.delegate = nil
+
+        // Then, cleanup resources.
         engine.cleanEngineWeb()
     }
 
--- a/Sources/MessagingInApp/Gist/Managers/Models/Message.swift
+++ b/Sources/MessagingInApp/Gist/Managers/Models/Message.swift
@@ -164,7 +164,16 @@ extension Array where Element == Message {
             switch ($0.priority, $1.priority) {
             case (let priority0?, let priority1?):
                 // Both messages have a priority, so we compare them.
-                return priority0 < priority1
+
+                // if the messages do not have the same priority, sort by the priority.
+                if priority0 != priority1 {
+                    return priority0 < priority1
+                }
+
+                // If messages have same priority, sort by something else to assert that the function return value is always the same no matter if the order of the input array is different.
+
+                // Because the priorities are the same, it doesn't matter what message is next. Use a unique value to perform the comparison.
+                return $0.instanceId < $1.instanceId
             case (nil, _):
                 // The first message has no priority, it should be considered greater so that it ends up at the end of the sorted array.
                 return false
--- a/Sources/MessagingInApp/Gist/Managers/Models/Message.swift
+++ b/Sources/MessagingInApp/Gist/Managers/Models/Message.swift
@@ -164,16 +164,7 @@ extension Array where Element == Message {
             switch ($0.priority, $1.priority) {
             case (let priority0?, let priority1?):
                 // Both messages have a priority, so we compare them.
-
-                // if the messages do not have the same priority, sort by the priority.
-                if priority0 != priority1 {
-                    return priority0 < priority1
-                }
-
-                // If messages have same priority, sort by something else to assert that the function return value is always the same no matter if the order of the input array is different.
-
-                // Because the priorities are the same, it doesn't matter what message is next. Use a unique value to perform the comparison.
-                return $0.instanceId < $1.instanceId
+                return priority0 < priority1
             case (nil, _):
                 // The first message has no priority, it should be considered greater so that it ends up at the end of the sorted array.
                 return false
--- a/Sources/MessagingInApp/Gist/Views/GistView.swift
+++ b/Sources/MessagingInApp/Gist/Views/GistView.swift
@@ -8,7 +8,7 @@ public protocol GistViewDelegate: AnyObject {
 
 public class GistView: UIView {
     public weak var delegate: GistViewDelegate?
-    private var message: Message?
+    var message: Message?
 
     convenience init(message: Message, engineView: UIView) {
         self.init()
--- a/Sources/MessagingInApp/Views/InAppMessageView.swift
+++ b/Sources/MessagingInApp/Views/InAppMessageView.swift
@@ -38,6 +38,13 @@ public class InAppMessageView: UIView {
         }
     }
 
+    // Inline messages that have already been shown by this View instance.
+    // This is used to prevent showing the same message multiple times when the close button is pressed.
+    //
+    // When persistent vs non-persistent messages and metrics features are implemented in the SDK, this array may be
+    // replaced with a global list of shown messages.
+    var previouslyShownMessages: [Message] = []
+
     var runningHeightChangeAnimation: UIViewPropertyAnimator?
 
     private var inlineMessageManager: InlineMessageManager?
@@ -90,7 +97,11 @@ public class InAppMessageView: UIView {
         }
 
         let queueOfMessagesForGivenElementId = localMessageQueue.getInlineMessages(forElementId: elementId)
-        let messageToDisplay = queueOfMessagesForGivenElementId.first
+        let messageToDisplay = queueOfMessagesForGivenElementId.first { potentialMessageToDisplay in
+            let didPreviouslyShowMessage = previouslyShownMessages.contains(where: { $0.id == potentialMessageToDisplay.id })
+
+            return !didPreviouslyShowMessage
+        }
 
         if let messageToDisplay {
             displayInAppMessage(messageToDisplay)
@@ -178,7 +189,11 @@ extension InAppMessageView: InlineMessageManagerDelegate {
 
     func onCloseAction() {
         Task { @MainActor in
-            self.dismissInAppMessage()
+            if let currentlyShownMessage = inlineMessageManager?.currentMessage {
+                previouslyShownMessages.append(currentlyShownMessage)
+            }
+
+            self.checkIfMessageAvailableToDisplay()
         }
     }
 }
--- a/Sources/MessagingInApp/Views/InAppMessageView.swift
+++ b/Sources/MessagingInApp/Views/InAppMessageView.swift
@@ -97,7 +97,7 @@ public class InAppMessageView: UIView {
     }
 
     // Updates the state of the View, if needed. Call as often as you need if an event happens that may cause the View to need to update.
-    private func refreshView(showNextMessage: Bool = false) {
+    private func refreshView(forceShowNextMessage: Bool = false) {
         guard let elementId = elementId else {
             return // we cannot check if a message is available until element id set on View.
         }
@@ -109,8 +109,7 @@ public class InAppMessageView: UIView {
             return !didPreviouslyShowMessage
         }
 
-        let shouldCheckIfAlreadyShowingAMessage = !showNextMessage
-        if shouldCheckIfAlreadyShowingAMessage, isRenderingOrDisplayingAMessage {
+        if !forceShowNextMessage, isRenderingOrDisplayingAMessage {
             // We are already displaying or rendering a messsage. Do not show another message until the current message is closed.
             // The main reason for this is when a message is tracked as "opened", the Gist backend will not return this message on the next fetch call.
             // We want to coninue showing a message even if the fetch no longer returns the message and the message is currently visible.
@@ -210,7 +209,7 @@ extension InAppMessageView: InlineMessageManagerDelegate {
                 previouslyShownMessages.append(currentlyShownMessage)
             }
 
-            self.refreshView(showNextMessage: true)
+            self.refreshView(forceShowNextMessage: true)
         }
     }
 }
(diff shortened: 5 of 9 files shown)


Sources/MessagingInApp/Gist/Managers/MessageManager.swift
```diff
@@ -19,7 +19,7 @@ class MessageManager {
     var engine: EngineWebInstance
     private let siteId: String
     let currentMessage: Message
-    var gistView: GistView!
+    let gistView: GistView
```
levibostian, line 22: This PR adds code that removes the WebView from the UI and performs resource cleanup. Using `!` adds risk to a runtime crash and I thought there would be a possibility in the future where someone tries to use `gistView = nil` and cause the SDK to crash. To prevent this, I refactored `gistView` to be immutable.

Sources/MessagingInApp/Gist/Managers/InlineMessageManager.swift
```diff
@@ -18,9 +18,11 @@ protocol InlineMessageManagerDelegate: AnyObject {
...
 class InlineMessageManager: MessageManager {
     var inlineMessageView: GistView? {
-        let view = super.gistView
-        view?.delegate = self
-        return view
+        if super.gistView.delegate == nil {
+            super.gistView.delegate = self
+        }
+
+        return super.gistView
```
levibostian, line 25: See [these notes](https://github.com/customerio/customerio-ios/pull/756/files#diff-70b8fb68d815f2ebfa219cf5e9332ad86a7b2081a7a87e6dbcc43c0ddd850e4aR22) to explain this change.

Sources/MessagingInApp/Gist/Managers/Models/Message.swift
levibostian, line 1: This change was to fix a bug I found while doing sample app testing. The bug is that when you have multiple inline messages with same element id and same priority, the order of the messages in the local queue are in an in-consistent order. 

See the [commit message and test function](https://github.com/customerio/customerio-ios/pull/756/commits/f519aebf8124d432a047ef00afcc2059a200e959) to learn more about the details.

Sources/MessagingInApp/Views/InAppMessageView.swift
```diff
@@ -38,6 +38,13 @@ public class InAppMessageView: UIView {
         }
     }
 
+    // Inline messages that have already been shown by this View instance.
+    // This is used to prevent showing the same message multiple times when the close button is pressed.
+    //
+    // When persistent vs non-persistent messages and metrics features are implemented in the SDK, this array may be
+    // replaced with a global list of shown messages.
+    var previouslyShownMessages: [Message] = []
```
levibostian, line 46: Call-out to the comments for this property. 

When persistent vs non-persistent messages and metrics features are implemented for inline in-app messages, I am expecting that this property will be replaced. 

For modal messages today, once a message is shown, the SDK tracks a metric and doesn't show the same messages again. I imagine that inline will behave similarly.

Sources/MessagingInApp/Gist/Views/GistView.swift
```diff
@@ -8,7 +8,7 @@ public protocol GistViewDelegate: AnyObject {
 
 public class GistView: UIView {
     public weak var delegate: GistViewDelegate?
-    private var message: Message?
+    var message: Message?
```
levibostian, line 11: `internal` scope for tests to access

Sources/MessagingInApp/Views/InAppMessageView.swift
```diff
@@ -38,9 +38,16 @@ public class InAppMessageView: UIView {
...
+
     var runningHeightChangeAnimation: UIViewPropertyAnimator?
 
-    private var inlineMessageManager: InlineMessageManager?
+    var inlineMessageManager: InlineMessageManager?
```
levibostian, line 50: `internal` scope for tests to access

Tests/MessagingInApp/EngineWeb/EngineWebProviderStub.swift
```diff
@@ -12,3 +13,17 @@ class EngineWebProviderStub: EngineWebProvider {
         engineWebMock
     }
 }
+
+// In order to test multiple instances of inline Views, you need to create multiple separate EngineWebInstance instances for each View instance.
+// This is a new stub that allows this.
+// A follow-up refactor that makes this stub the default would be a good chance to the test suite.
```
levibostian, line 19: I am planning on adding another PR to this stack with this refactor. 

Feel free to not approve this PR until that PR is added to the stack.

Tests/MessagingInApp/Extensions/GistExtensions.swift
```diff
@@ -20,10 +20,10 @@ extension Message {
...
 
     static var random: Message {
-        Message(messageId: .random, campaignId: .random)
+        Message(elementId: nil)
     }
 
     static var randomInline: Message {
-        Message(messageId: .random, campaignId: .random, elementId: .random)
+        Message(elementId: .random)
     }
```
levibostian, line 28: This was only changed out of convenience. The `Message` constructor used in tests has default values for all parameters so some parameters are no longer needed.

Sources/MessagingInApp/Gist/Managers/Models/Message.swift
```diff
@@ -164,7 +164,16 @@ extension Array where Element == Message {
...
+
+                // if the messages do not have the same priority, sort by the priority.
+                if priority0 != priority1 {
+                    return priority0 < priority1
+                }
+
+                // If messages have same priority, sort by something else to assert that the function return value is always the same no matter if the order of the input array is different.
+
+                // Because the priorities are the same, it doesn't matter what message is next. Use a unique value to perform the comparison.
+                return $0.instanceId < $1.instanceId
```
levibostian, line 176: I was encountering a bug during sample app testing that this code fixes. 

Reproduce steps:
* Send 2+ messages with same element id and priority to device
* Open app and wait until 1 inline View is in foreground and View is showing a message. 
* Wait 10-60 seconds. You should notice that the inline View will randomly change to displaying the other message that you sent to the device, without you touching the screen. 

This bug fix makes it so that when `sortByMessagePriority` is called, as output will always be the same for any given input. [See added tests to learn more](https://github.com/customerio/customerio-ios/pull/756/files#diff-d167fc3626f2b47618e1db865ee4d5631fd993ba64fd17a303fe77c1078f9d0fR68)

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
```diff
@@ -47,11 +43,11 @@ class InAppMessageViewTest: UnitTest {
 
         let view = InAppMessageView(coder: EmptyNSCoder())!
 
-        XCTAssertFalse(isDisplayingInAppMessage(view)) // Assert View is in dismissed state
+        XCTAssertFalse(isInlineViewVisible(view)) // Assert View is in dismissed state
```
levibostian, line 46: I decided to rename this test function to be more accurate for what the function checks. 

When I was reading `isDisplayingInAppMessage`, I read that name and thought that it was also checking a specific in-app message was being shown, which is not true. You must also call `getInAppMessage` to check that.

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
```diff
@@ -160,25 +174,6 @@ class InAppMessageViewTest: UnitTest {
         XCTAssertTrue(webViewBeforeFetch === webViewAfterFetch)
     }
 
-    @MainActor
-    func test_givenAlreadyShowingInAppMessage_whenNewMessageFetched_expectDoNotReplaceContents() async {
```
levibostian, line 164: Removed this test because `test_expiration_givenMessageExpired_givenNewMessageFetched_expectDisplayNextMessage` is testing this logic for us already, making this test no longer valuable.

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
```diff
@@ -188,31 +183,35 @@ class InAppMessageViewTest: UnitTest {
...
         // Simulate message expiration.
         await simulateSdkFetchedMessages([])
 
-        XCTAssertFalse(isDisplayingInAppMessage(inlineView))
+        XCTAssertFalse(isInlineViewVisible(inlineView))
+        XCTAssertNil(getInAppMessage(forView: inlineView))
     }
 
-    // Once an in-app message has been displayed it will not be replaced with another message.
-    // We plan to change this behavior in the future. Test function can be modified to match the new behavior at that time.
```
levibostian, line 202: This PR is the first PR that deals with replacing 1 in-app message with another one. Previously, once a View was showing a message we would not replace it's contents at all. 

There are some test functions, including this one, that were created matching the original behavior. Now that we are replacing in-app messages, these tests needed modified to show new expected behavior of replacing in-app messages.

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
```diff
@@ -223,29 +222,73 @@ class InAppMessageViewTest: UnitTest {
...
+        XCTAssertEqual(getInAppMessage(forView: view), givenMessages[1])
 
-        XCTAssertFalse(isDisplayingInAppMessage(inlineView))
+        await onCloseActionButtonPressed(onInlineView: view)
+        XCTAssertFalse(isInlineViewVisible(view))
+        XCTAssertNil(getInAppMessage(forView: view))
     }
 
-    // Once an in-app message has been closed it will not be replaced with another message.
-    // We plan to change this behavior in the future. Test function can be modified to match the new behavior at that time.
```
levibostian, line 235: See https://github.com/customerio/customerio-ios/pull/756/files#r1661305328

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
```diff
@@ -223,29 +222,73 @@ class InAppMessageViewTest: UnitTest {
...
     @MainActor
-    func test_onCloseAction_givenMessageClosed_givenNewMessageFetched_expectIgnoreMessage() async {
+    func test_onCloseAction_givenMessageClosed_givenNewMessageFetched_expectDisplayNewMessage() async {
```
levibostian, line 257: Test function modified to match new expected behavior. 

If we close a message and then a new message is fetched in the future, that new fetched message will be shown.

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
```diff
@@ -312,33 +355,44 @@ class InAppMessageViewTest: UnitTest {
...
+        getWebEngineForInlineView(inlineView)?.delegate?.routeLoaded(route: message.templateId)
+        getWebEngineForInlineView(inlineView)?.delegate?.sizeChanged(width: widthOfRenderedMessage, height: heightOfRenderedMessage)
 
-        return viewHeightConstraint.constant > 0
+        // When sizeChanged() is called on the inline View, it adds a task to the main thread queue. Our test wants to wait until this task is done running.
+        await waitForMainThreadToFinishPendingTasks()
+    }
+
+    func getWebEngineForInlineView(_ view: InAppMessageView) -> EngineWebInstance? {
+        view.inlineMessageManager?.engine
```
levibostian, line 395: I wanted to write test functions that involved testing multiple separate instances of inline View. In order to do that, I needed to make a refactor in the test class to have 1 `engineWebMock` instance per inline View instead of 1 mock used for all View instances. 

These utility functions were modified to take in the inline View instance, get the engine mock for that inline View, and then perform the actions on the engine instance. 

This allows us to, for example, click the close button on 1 View instance but not the other.

Tests/MessagingInApp/EngineWeb/EngineWebProviderStub.swift
levibostian: See https://github.com/customerio/customerio-ios/pull/756/files#r1661310325 to understand why this new stub was created.

Sources/MessagingInApp/Gist/Managers/Models/Message.swift
levibostian: FYI, I have a follow-up PR that modifies this logic. I found a bug in it. https://github.com/customerio/customerio-ios/pull/760

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
```diff
@@ -84,7 +80,8 @@ class InAppMessageViewTest: UnitTest {
 
         let inlineView = InAppMessageView(elementId: .random)
 
-        XCTAssertFalse(isDisplayingInAppMessage(inlineView))
+        XCTAssertFalse(isInlineViewVisible(inlineView))
+        XCTAssertNil(getInAppMessage(forView: inlineView)) // expect not in process of rendering a message
```
levibostian, line 84: It's important to test not only if a View is visible but also check if the View is currently in the process of rendering a message. 

I improved this test function to check both of these states.

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
```diff
@@ -93,9 +90,23 @@ class InAppMessageViewTest: UnitTest {
...
         let inlineView = InAppMessageView(elementId: givenInlineMessage.elementId!)
-        await onDoneRenderingInAppMessage(givenInlineMessage)
+        await onDoneRenderingInAppMessage(givenInlineMessage, insideOfInlineView: inlineView)
+
+        XCTAssertTrue(isInlineViewVisible(inlineView))
+        XCTAssertEqual(getInAppMessage(forView: inlineView), givenInlineMessage)
+    }
+
+    @MainActor
+    func test_displayInAppMessage_givenMultipleMessagesInQueue_expectDisplayFirstMessage() async {
```
levibostian, line 100: Added missing test to check what message is displayed when the local queue has multiple messages in it.

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
```diff
@@ -168,51 +182,43 @@ class InAppMessageViewTest: UnitTest {
...
         // If the WebViews are different, it means the message was reloaded.
         XCTAssertTrue(webViewBeforeFetch === webViewAfterFetch)
+        XCTAssertEqual(getInAppMessage(forView: inlineView), givenOldInlineMessage)
     }
 
     // MARK: expiration of in-app messages
 
     @MainActor
-    func test_expiration_givenDisplayedMessageExpires_expectDismissView() async {
+    func test_expiration_givenDisplayedMessageExpires_expectContinueShowingMessageUntilClose() async {
```
levibostian, line 200: While this PR focuses on replacing an in-app message with another, we expect that a message is replaced only after the close button has been clicked. 

That means that this test function needed modified because the expected behavior has changed for messages that expire.

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
```diff
@@ -223,29 +229,73 @@ class InAppMessageViewTest: UnitTest {
...
+        XCTAssertTrue(isInlineViewVisible(inlineView))
+
+        await onCloseActionButtonPressed(onInlineView: inlineView)
+
+        XCTAssertFalse(isInlineViewVisible(inlineView))
+        XCTAssertNil(getInAppMessage(forView: inlineView))
+    }
+
+    @MainActor
+    func test_onCloseAction_givenMultipleMessagesInQueue_expectDisplayNextMessageInQueueAfterClose() async {
```
levibostian, line 242: This test function tests the main use case that this PR is based on.

Sources/MessagingInApp/Views/InAppMessageView.swift
```diff
@@ -34,21 +34,33 @@ public class InAppMessageView: UIView {
...
+            refreshView()
         }
     }
 
+    // Inline messages that have already been shown by this View instance.
+    // This is used to prevent showing the same message multiple times when the close button is pressed.
+    //
+    // When persistent vs non-persistent messages and metrics features are implemented in the SDK, this array may be
+    // replaced with a global list of shown messages.
+    var previouslyShownMessages: [Message] = []
```
Shahroz16, line 46: how about we use [OrderedSet](https://github.com/apple/swift-collections/blob/main/Documentation/OrderedSet.md) to manage duplicates?

Sources/MessagingInApp/Views/InAppMessageView.swift
```diff
@@ -79,45 +91,52 @@ public class InAppMessageView: UIView {
...
+            return // we cannot check if a message is available until element id set on View.
         }
 
         let queueOfMessagesForGivenElementId = localMessageQueue.getInlineMessages(forElementId: elementId)
-        let messageToDisplay = queueOfMessagesForGivenElementId.first
+        let messageAvailableToDisplay = queueOfMessagesForGivenElementId.first { potentialMessageToDisplay in
+            let didPreviouslyShowMessage = previouslyShownMessages.contains(where: { $0.id == potentialMessageToDisplay.id })
+
+            return !didPreviouslyShowMessage
+        }
```
Shahroz16, line 110: minor but can be simplified to read, i had to double check if what i am understanding is whats happenings

```
// Function to check if a message has been previously shown
func hasNotBeenPreviouslyShown(_ message: Message) -> Bool {
    return !previouslyShownMessages.contains { $0.id == message.id }
}

// Find the first message that has not been previously shown
let messageAvailableToDisplay = queueOfMessagesForGivenElementId.first(where: hasNotBeenPreviouslyShown)
```

Shahroz16 (approved):

Sources/MessagingInApp/Views/InAppMessageView.swift
levibostian: Done https://github.com/customerio/customerio-ios/pull/756/commits/c70e9691db346ba4dc16a5c0fff2bb209a724d18

PR #758, opened 2024-07-01
chore: add animations and activity indicator while rendering next message

levibostian:
https://linear.app/customerio/issue/MBL-313/after-close-action-pressed-replace-in-app-message-with-next-in-queue

### Problem: 
As a marketer, I want to display another in-app message inline when close button pressed so I can rotate between multiple in-app messages.

When you click close button and a new inline message needs to be displayed, it could take a couple seconds for renderer response. The transition between messages should be seamless and not disrupt the user experience.

### Solution: 
When the close button is pressed...
1. Fade out the currently shown in-app message while also fading in an iOS [Activity Indicator](https://developer.apple.com/design/human-interface-guidelines/progress-indicators). The height of the View remains the same as what the old in-app message's height was. 
2. Show the activity indicator while the next message is rendering. 
3. After rendering is done, fade out the activity indicator, fade in the next in-app message, and animate the height to match the new in-app message. 

Here is a demo of what this looks like in the UIKit sample app: 

https://github.com/customerio/customerio-ios/assets/2041082/7483321a-51a9-4520-bab4-f9c1564a347c

### Testing:
* Automated test suite should pass and not need any modifications since this is only adding animations between state changes. 
* Testing for this PR is entirely in sample app QA testing. [Test case added to Qase](https://app.qase.io/project/SM?previewMode=side&suite=112&case=1187).

### DIFF
--- a/Sources/MessagingInApp/Views/InAppMessageView.swift
+++ b/Sources/MessagingInApp/Views/InAppMessageView.swift
@@ -158,7 +158,7 @@ public class InAppMessageView: UIView {
                 activityIndicator.heightAnchor.constraint(equalTo: heightAnchor)
             ])
 
-            animateCrossFade(fromView: currentlyDisplayedInAppWebView, toView: activityIndicator) {
+            animateFadeInOutInlineView(fromView: currentlyDisplayedInAppWebView, toView: activityIndicator) {
                 // After animation is over, cleanup resources and begin rendering of the next message.
                 self.stopShowingMessageAndCleanup()
                 self.beginShowing(message: message)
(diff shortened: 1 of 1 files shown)


Sources/MessagingInApp/Views/InAppMessageView.swift
```diff
@@ -191,6 +231,28 @@ public class InAppMessageView: UIView {
 
         runningHeightChangeAnimation?.startAnimation()
     }
+
+    // Takes in 2 Views. In 1 single animation, fades in 1 View while fading out the other.
+    private func animateCrossFade(fromView: UIView, toView: UIView, onComplete: (() -> Void)?) {
```
ami-aman, line 236: Reading the name of the method, for a moment I thought this method animates when the "cross" button is tapped.
but the actual functioning is to animate in and out the inline views so how about `animateFadeInOutInlineView` for more clarity ?

ami-aman (approved):
Non-blocking suggestion.

Sources/MessagingInApp/Views/InAppMessageView.swift
levibostian: Done https://github.com/customerio/customerio-ios/pull/758/commits/76399b6fe2e8e6408968a0632fcbf9317979eedb

PR #759, opened 2024-07-02
feat: add page rule for inline inapp messages

ami-aman:
[Implement page rules with inline messages](https://linear.app/customerio/issue/MBL-314/implement-page-rules-with-inline-messages)

### Problem:
As a marketer I want some inline in-app messages to only be displayed on certain screens of my app. The in-app message does not display inline on the screen that was set in page rules.

### Testing:
For QA, there are three inline views in UIKit Sample app. 
2 on the dashboard screen - `dashboard-announcement` and `dashboard-announcement-code`
1 on custom attributes screen -  `custom-screen`

To avoid any confusion, since same controller `CustomDataViewController` is used for Custom Attributes, Device Attributes,  Profile Attributes so if an in-app message is targeted for element Id `custom-screen` it will show every time a user visits any of the above screens. 

**Steps to test:**
- Using a campaign/broadcast send an inline inapp message with _no page rule_ targeted for one of the element ids given above example dashboard-announcement
- You'll see this inline message on the targeted element id. 
- Now send another inline inapp message _with page rule_ (assume Custom) targeted for another element id example custom-screen
- Note that the inline inapp message targeted with page rule (Custom) and element Id custom-screen will only be shown on custom view controllers like Custom Attributes, Device Attributes, Profile Attributes since they all share common view controller. 

https://github.com/customerio/customerio-ios/assets/91549653/f5c03705-4e1b-4bcd-85ab-c05b419cb37b

### DIFF
--- a/Apps/APN-UIKit/APN UIKit/Base.lproj/Main.storyboard
+++ b/Apps/APN-UIKit/APN UIKit/Base.lproj/Main.storyboard
@@ -375,113 +375,132 @@
                         <rect key="frame" x="0.0" y="0.0" width="414" height="896"/>
                         <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
                         <subviews>
-                            <view contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="2e8-CT-ayB">
-                                <rect key="frame" x="15" y="273" width="384" height="350"/>
+                            <view contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="Mh2-zI-sF9">
+                                <rect key="frame" x="20" y="64" width="374" height="400"/>
                                 <subviews>
-                                    <label opaque="NO" userInteractionEnabled="NO" contentMode="left" horizontalHuggingPriority="251" verticalHuggingPriority="251" text="Set Custom Device Attribute" textAlignment="natural" lineBreakMode="tailTruncation" baselineAdjustment="alignBaselines" minimumScaleFactor="0.5" translatesAutoresizingMaskIntoConstraints="NO" id="6Pd-I5-ES2">
-                                        <rect key="frame" x="20" y="20.000000000000004" width="344" height="34.333333333333343"/>
-                                        <fontDescription key="fontDescription" name="Avenir-Medium" family="Avenir" pointSize="25"/>
-                                        <color key="textColor" red="0.24313725489999999" green="0.24313725489999999" blue="0.24313725489999999" alpha="1" colorSpace="custom" customColorSpace="displayP3"/>
-                                        <nil key="highlightedColor"/>
-                                    </label>
-                                    <button opaque="NO" contentMode="scaleToFill" contentHorizontalAlignment="center" contentVerticalAlignment="center" buttonType="system" lineBreakMode="middleTruncation" translatesAutoresizingMaskIntoConstraints="NO" id="frl-I2-mJf" customClass="ThemeButton" customModule="APN_UIKit" customModuleProvider="target">
-                                        <rect key="frame" x="40" y="250.33333333333337" width="304" height="50"/>
-                                        <color key="backgroundColor" red="0.23529411759999999" green="0.26274509800000001" blue="0.49019607840000001" alpha="1" colorSpace="custom" customColorSpace="displayP3"/>
+                                    <view contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="kvM-jQ-73n" customClass="InAppMessageView" customModule="CioMessagingInApp">
+                                        <rect key="frame" x="0.0" y="0.0" width="374" height="20"/>
                                         <constraints>
-                                            <constraint firstAttribute="height" constant="50" id="JkB-FX-Tyf"/>
+                                            <constraint firstAttribute="height" constant="20" id="g15-GG-CD3"/>
                                         </constraints>
-                                        <color key="tintColor" white="1" alpha="1" colorSpace="custom" customColorSpace="genericGamma22GrayColorSpace"/>
-                                        <state key="normal" title="Button"/>
-                                        <buttonConfiguration key="configuration" style="plain" title="Send Event"/>
-                                        <connections>
-                                            <action selector="sendCustomData:" destination="L6G-xX-ThQ" eventType="touchUpInside" id="aJ3-Cr-NYE"/>
-                                        </connections>
-                                    </button>
-                                    <stackView opaque="NO" contentMode="scaleToFill" alignment="top" spacing="16" translatesAutoresizingMaskIntoConstraints="NO" id="go1-R4-suI">
-                                        <rect key="frame" x="20" y="74.333333333333314" width="344" height="136"/>
+                                    </view>
+                                    <view contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="2e8-CT-ayB">
+                                        <rect key="frame" x="0.0" y="20" width="374" height="350"/>
                                         <subviews>
-                                            <stackView opaque="NO" contentMode="scaleToFill" axis="vertical" distribution="fillEqually" translatesAutoresizingMaskIntoConstraints="NO" id="g0C-Az-c2s">
-                                                <rect key="frame" x="0.0" y="0.0" width="109.66666666666667" height="136"/>
+                                            <label opaque="NO" userInteractionEnabled="NO" contentMode="left" horizontalHuggingPriority="251" verticalHuggingPriority="251" text="Set Custom Device Attribute" textAlignment="natural" lineBreakMode="tailTruncation" baselineAdjustment="alignBaselines" minimumScaleFactor="0.5" translatesAutoresizingMaskIntoConstraints="NO" id="6Pd-I5-ES2">
+                                                <rect key="frame" x="20" y="20.000000000000004" width="334" height="34.333333333333343"/>
+                                                <fontDescription key="fontDescription" name="Avenir-Medium" family="Avenir" pointSize="25"/>
+                                                <color key="textColor" red="0.24313725489999999" green="0.24313725489999999" blue="0.24313725489999999" alpha="1" colorSpace="custom" customColorSpace="displayP3"/>
+                                                <nil key="highlightedColor"/>
+                                            </label>
+                                            <button opaque="NO" contentMode="scaleToFill" contentHorizontalAlignment="center" contentVerticalAlignment="center" buttonType="system" lineBreakMode="middleTruncation" translatesAutoresizingMaskIntoConstraints="NO" id="frl-I2-mJf" customClass="ThemeButton" customModule="APN_UIKit" customModuleProvider="target">
+                                                <rect key="frame" x="40" y="250.33333333333331" width="294" height="50"/>
+                                                <color key="backgroundColor" red="0.23529411759999999" green="0.26274509800000001" blue="0.49019607840000001" alpha="1" colorSpace="custom" customColorSpace="displayP3"/>
+                                                <constraints>
+                                                    <constraint firstAttribute="height" constant="50" id="JkB-FX-Tyf"/>
+                                                </constraints>
+                                                <color key="tintColor" white="1" alpha="1" colorSpace="custom" customColorSpace="genericGamma22GrayColorSpace"/>
+                                                <state key="normal" title="Button"/>
+                                                <buttonConfiguration key="configuration" style="plain" title="Send Event"/>
+                                                <connections>
+                                                    <action selector="sendCustomData:" destination="L6G-xX-ThQ" eventType="touchUpInside" id="aJ3-Cr-NYE"/>
+                                                </connections>
+                                            </button>
+                                            <stackView opaque="NO" contentMode="scaleToFill" alignment="top" spacing="16" translatesAutoresizingMaskIntoConstraints="NO" id="go1-R4-suI">
+                                                <rect key="frame" x="20" y="74.333333333333343" width="334" height="136"/>
                                                 <subviews>
-                                                    <label opaque="NO" userInteractionEnabled="NO" contentMode="left" horizontalHuggingPriority="251" verticalHuggingPriority="251" text="Event Name" textAlignment="justified" lineBreakMode="tailTruncation" baselineAdjustment="alignBaselines" adjustsFontSizeToFit="NO" translatesAutoresizingMaskIntoConstraints="NO" id="BBq-Fn-h9P">
-                                                        <rect key="frame" x="0.0" y="0.0" width="109.66666666666667" height="45.333333333333336"/>
-                                                        <fontDescription key="fontDescription" type="system" pointSize="16"/>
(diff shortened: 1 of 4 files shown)


Sources/MessagingInApp/Gist/Managers/MessageQueueManager.swift
```diff
@@ -87,7 +87,16 @@ class MessageQueueManagerImpl: MessageQueueManager {
...
+            if $0.value.elementId == elementId {
+                if $0.value.doesHavePageRule() {
+                    return $0.value.doesPageRuleMatch(route: Gist.shared.getCurrentRoute())
+                }
+                return true
+            }
+            return false
+        }.map(\.value).sortByMessagePriority()
+        return messages
     }
```
ami-aman, line 100: There are two ways inline messages are checked, fetched and shown to the user if found in the local storage: 
- Through polling ([EventBus event posted](https://github.com/customerio/customerio-ios/blob/edcd7765e3388ba4d16424a2c798149d09b4b15c/Sources/MessagingInApp/Gist/Managers/MessageQueueManager.swift#L143))
- When an [element id is set on a view](https://github.com/customerio/customerio-ios/blob/edcd7765e3388ba4d16424a2c798149d09b4b15c/Sources/MessagingInApp/Views/InAppMessageView.swift#L92) 

Either ways, the messages are fetched using  `getInlineMessages` method and then shown.

Tests/MessagingInApp/Gist/Managers/MessageQueueManagerTest.swift
```diff
@@ -64,6 +64,45 @@ class MessageQueueManagerTest: UnitTest {
...
+        let givenMessage3 = Message(elementId: givenElementId3, priority: 0)
+
+        manager.localMessageStore = [
+            "1": givenMessage1,
+            "2": givenMessage2,
+            "3": givenMessage3
+        ]
+        Gist.shared.setCurrentRoute(currentRoute)
+        let actualMessages = manager.getInlineMessages(forElementId: givenElementId1)
+        XCTAssertEqual(actualMessages, [givenMessage1])
```
ami-aman, line 84: 1 message returned, that matches page rule and element id condition.

Tests/MessagingInApp/Gist/Managers/MessageQueueManagerTest.swift
```diff
@@ -64,6 +64,45 @@ class MessageQueueManagerTest: UnitTest {
...
+        let givenMessage2 = Message(pageRule: currentRoute, elementId: givenElementId2, priority: 0)
+        let givenMessage3 = Message(elementId: givenElementId3, priority: 0)
+
+        manager.localMessageStore = [
+            "1": givenMessage1,
+            "2": givenMessage2,
+            "3": givenMessage3
+        ]
+        let actualMessages = manager.getInlineMessages(forElementId: givenElementId1)
+        XCTAssertEqual(actualMessages, [])
```
ami-aman, line 103: 0 messages returned since none of the messages matched page rule and element id.

Apps/APN-UIKit/APN UIKit/View/Customisation/CustomDataViewController.swift
levibostian, line 1: To help with QA testing, I added a label to the Dashboard screen which tells you what inline Views are present on that screen. I suggest adding a label to this screen, too. What do you think? 

CleanShot 2024-07-02 at 09 40 05@2x

Apps/APN-UIKit/APN UIKit/Base.lproj/Main.storyboard
levibostian, line 1: Could you record a video or take screenshots of what the UI looks like now that the inline Views are added? This would help with reviewing the code since it's heavily reliant on UI. 

[This PR is an example of recording a demo showing the UI changes done in a PR](https://github.com/customerio/customerio-ios/pull/758).

Sources/MessagingInApp/Gist/Managers/MessageQueueManager.swift
```diff
@@ -87,7 +87,16 @@ class MessageQueueManagerImpl: MessageQueueManager {
...
-        localMessageStore.filter { $0.value.elementId == elementId }.map(\.value).sortByMessagePriority()
+        let messages = localMessageStore.filter {
+            if $0.value.elementId == elementId {
+                if $0.value.doesHavePageRule() {
+                    return $0.value.doesPageRuleMatch(route: Gist.shared.getCurrentRoute())
+                }
+                return true
+            }
+            return false
+        }.map(\.value).sortByMessagePriority()
```
levibostian, line 98:
```suggestion
        let messages = localMessageStore
            .map(\.value) // give us the Message from dictionary 
            .filter { $0.value.elementId == elementId } // Only get messages for a specific elementid. 
            .filter { $0.value.doesHavePageRule() && $0.value.doesPageRuleMatch(route: Gist.shared.getCurrentRoute()) } // if page rules enabled, filter what match             
            .sortByMessagePriority()
```

The logic looks good. I do find the code hard to read at the moment. I suggest using multiple filters, comments, or extensions to make this more usable.

Tests/MessagingInApp/Gist/Managers/MessageQueueManagerTest.swift
```diff
@@ -64,6 +64,45 @@ class MessageQueueManagerTest: UnitTest {
...
 
+    func test_getInlineMessages_givenQueueMessageHasPageRule_pageRuleMatch_expectInlineMessage() {
+        let givenElementId1 = String.random
+        let givenElementId2 = String.random
+        let givenElementId3 = String.random
+        let currentRoute = String.random
+
+        let givenMessage1 = Message(pageRule: currentRoute, elementId: givenElementId1, priority: 1)
+        let givenMessage2 = Message(pageRule: currentRoute, elementId: givenElementId2, priority: 0)
+        let givenMessage3 = Message(elementId: givenElementId3, priority: 0)
```
levibostian, line 75: For the `Message pageRule`, I don't recommend using a `String.random` value. A `Message` page rule is [a regex string](https://github.com/customerio/customerio-ios/blob/edcd7765e3388ba4d16424a2c798149d09b4b15c/Sources/MessagingInApp/Gist/Managers/Models/Message.swift#L118-L127). To make the test more accurate, I suggest using a regex pattern for `Message(pageRule)` parameter.

levibostian: From reading the test function name, it looks like the test is scoped to only test the page rules. I agree with that scope. 

To keep this test more focused on just testing page rules, I suggest making the element id and priorities of all of the given messages to be the same. So the test is only testing 1 Message property instead of 3 separate ones.

Tests/MessagingInApp/Gist/Managers/MessageQueueManagerTest.swift
```diff
@@ -64,6 +64,45 @@ class MessageQueueManagerTest: UnitTest {
...
+            "1": givenMessage1,
+            "2": givenMessage2,
+            "3": givenMessage3
+        ]
+        Gist.shared.setCurrentRoute(currentRoute)
+        let actualMessages = manager.getInlineMessages(forElementId: givenElementId1)
+        XCTAssertEqual(actualMessages, [givenMessage1])
+    }
+
+    func test_getInlineMessages_givenQueueMessageHasPageRule_pageRuleMisMatch_expectEmptyArray() {
```
levibostian, line 87:
```suggestion
    func test_getInlineMessages_givenQueueMessageHasPageRule_givenNoSetCurrentRoute_expectEmptyArray() {
```

I think this is a good test function, testing if `Gist.shared.setCurrentRoute(currentRoute)` is not called. 

The test function name made me think that we were testing if a page rule is set, but the Messages have a page rule *not* matching the currently set route in the SDK.

Tests/MessagingInApp/Gist/Managers/MessageQueueManagerTest.swift
```diff
@@ -64,6 +64,45 @@ class MessageQueueManagerTest: UnitTest {
...
+
+        manager.localMessageStore = [
+            "1": givenMessage1,
+            "2": givenMessage2,
+            "3": givenMessage3
+        ]
+        let actualMessages = manager.getInlineMessages(forElementId: givenElementId1)
+        XCTAssertEqual(actualMessages, [])
+    }
+
```
levibostian, line 105: Great tests so far. I think there is 1 test missing: given the SDK has a route set (`Gist.shared.setCurrentRoute(currentRoute)`) but the route not does match any of the page rules set on messages.

levibostian: Another use case to test: given the SDK has a route set (Gist.shared.setCurrentRoute(currentRoute)) but a `Message` does not have a page rule added to it (`Message(pageRule: nil)`), you expect that Message to be returned.

levibostian (commented):
Great work on your first PR for the inline feature! 

Gave some small suggestions. Overall, it looks great.

levibostian:
I have some questions about how to test this feature. 

* Could you add to the PR description QA test cases for how to test this code? [Example PR with test cases in description](https://github.com/customerio/customerio-ios/pull/756). 
* Also, could you add the QA test cases [to Qase](https://app.qase.io/project/SM?previewMode=side&suite=112) for regression testing in the future?

Apps/APN-UIKit/APN UIKit/View/Customisation/CustomDataViewController.swift
ami-aman: This sounds great. ✅

Tests/MessagingInApp/Gist/Managers/MessageQueueManagerTest.swift
ami-aman: Correct, it has been scoped to test only page rules. To your suggestion to make all element Ids and priorities same, the test would look something like:

```
let givenMessage1 = Message(pageRule: currentRoute, elementId: givenElementId, priority: 0)
let givenMessage2 = Message(pageRule: currentRoute, elementId: givenElementId, priority: 0)
let givenMessage3 = Message(elementId: givenElementId, priority: 0)
```

Since all of of same priority then ordering can be random, that might end up test cases to fail at times. Instead I have updated the test case to have same element ids but different priorities.

Sources/MessagingInApp/Gist/Managers/MessageQueueManager.swift
ami-aman: Thanks for the suggestion.
Small tweak was required in this suggestion. With the suggested logic, the message that matches the element id _and_ page rule were only filtered whereas the requirement is:

- Element id does not match then don't filter
- Element Id matches and page rule is inapplicable then filter
- Element Id matches and page rule is applicable and matches current route then filter
- Element Id matches and page rule is applicable but does not match current route then don't filter
 
 The suggestion was not satisfying the 2nd requirement and hence tweaks.

Tests/MessagingInApp/Gist/Managers/MessageQueueManagerTest.swift
levibostian: Good point about the randomness. 

I also encountered that and implemented a fix [in this PR](https://github.com/customerio/customerio-ios/pull/756/files#r1661299354). 

To not block this PR, I am OK with have same element ids and different priorities. If you would rather change the base branch of your PR to `levi/is-there-a-message-next-in-queue-after-close-button` so you can re-use that bug fix, I would be OK with that, too.

ami-aman: Updated the base branch to `levi/is-there-a-message-next-in-queue-after-close-button`

levibostian: Very sorry to complicate this, Aman. 

I am making some modifications to the `levi/is-there-a-message-next-in-queue-after-close-button` branch to [prevent some bugs I discovered](https://linear.app/customerio/issue/MBL-313/after-close-action-pressed-replace-in-app-message-with-next-in-queue#comment-68684ba6). 

Part of these modifications is reverting the fix I suggested in this thread. 

Therefore, I suggest moving forward with your original plan: "updated the test case to have same element ids but different priorities."

ami-aman: No worries, Levi ! Thanks for letting me know.

Tests/MessagingInApp/Gist/Managers/MessageQueueManagerTest.swift
ami-aman: Addressed. ✅

Tests/MessagingInApp/Gist/Managers/MessageQueueManagerTest.swift
```diff
@@ -64,6 +64,79 @@ class MessageQueueManagerTest: UnitTest {
...
+            "1": givenMessage1,
+            "2": givenMessage2,
+            "3": givenMessage3
+        ]
+        Gist.shared.setCurrentRoute("Home")
+        let actualMessages = manager.getInlineMessages(forElementId: givenElementId)
+        XCTAssertEqual(actualMessages, [])
+    }
+
+    func test_getInlineMessages_givenQueueMessageHasPageRule_givenNoSetCurrentRoute_expectEmptyArray() {
```
levibostian, line 121: I do think that it's important to test `givenNoSetCurrentRoute` as the function already does. I see the function is also testing the result when the elementIds are unique between all of the messages. 

To avoid testing multiple cases in 1 test function, I suggest having all of the messages use the same elementid and only test the result when no current route is set. 

I believe the function `test_getInlineMessages_givenQueueNotEmpty_expectGetInlineMessagesWithElementId` is testing the case when element ids do not match.

levibostian (approved):
1 non-blocking suggestion for tests. Otherwise, looks great! Thanks for all the automated tests and QA testing you added to this PR.

Apps/APN-UIKit/APN UIKit/Base.lproj/Main.storyboard
levibostian, line 1: for QA testing page rules, we should have multiple screens in the app have inline Views installed with the same elementid. 

For example, if the Dashboard screen has an inline View with elementId "announcement" and the CustomData screen has an inline View with elementId "announcement", then we can test different scenarios such as: 
* Given send a message with elementId "announcement" with no page rule set, expect to see the message on any screen. 
* Given send a message with elementId "announcement" with page rule set to "Custom", expect to see the message only on CustomData screen and not Dashboard.

levibostian (approved):
I thought of a suggestion for how we could improve the QA testing for the feature.

Tests/MessagingInApp/Gist/Managers/MessageQueueManagerTest.swift
ami-aman: Catered. ✅

Apps/APN-UIKit/APN UIKit/Base.lproj/Main.storyboard
ami-aman: There are multiple ways to test this feature and this is a great one as I believe our customers could use this scenario. For the sake of this PR, I will let the changes as is and test this scenario and if it works as expected then cool and if any weird behaviour is found, we can then talk about it.

PR #760, opened 2024-07-02
fix: getting messages by elementId has a consistent Array order

levibostian:
[linear.app/customerio/issue/MBL-313/after-close-action-pressed-replace-in-app-message-with-next-in-queue](https://linear.app/customerio/issue/MBL-313/after-close-action-pressed-replace-in-app-message-with-next-in-queue)

I was encountering a bug during sample app testing that resulted in the inline View randomly changing what message it was showing.

Reproduce steps:
* Send 2+ messages with same element id and priority to device
* Open app and wait until 1 inline View is in foreground and View is showing a message.
* Wait 10-60 seconds. You should notice that the inline View will randomly change to displaying the other message that you sent to the device, without you touching the screen.

Testing
* All logic tested from the added automated tests.


levibostian:
This logic is no longer needed. Closing.

PR #762, opened 2024-07-03
test: given expired message not yet shown expect to not show message

levibostian:
https://linear.app/customerio/issue/MBL-381/after-message-expires-do-not-display-it-in-app-if-it-has-not-already

As a marketer, I don't want messages that are no longer relevant to be displayed in an app. If a message expires, I expect it to not be displayed in an app if it has not been shown yet.

The logic of this ticket is already implemented in the SDK. This PR is only adding test cases that are missing for the acceptance criteria of the ticket.

From the acceptance criteria of the ticket, there are 2 test cases that need tested:
* Given a message expires and is already being shown, expect continue showing it until message is closed. (test function `test_expiration_givenDisplayedMessageExpires_expectContinueShowingMessageUntilClose` already covers this)
* Given a message expires and not yet shown, expect to not show that message. This PR adds this test case.

### DIFF
--- a/Tests/MessagingInApp/Views/InAppMessageViewTest.swift
+++ b/Tests/MessagingInApp/Views/InAppMessageViewTest.swift
@@ -177,7 +177,7 @@ class InAppMessageViewTest: UnitTest {
     // MARK: expiration of in-app messages
 
     @MainActor
-    func test_expiration_givenDisplayedMessageExpires_expectDismissView() async {
+    func test_expiration_givenDisplayedMessageExpires_expectContinueShowingMessageUntilClose() async {
         let givenInlineMessage = Message.randomInline
         queueMock.getInlineMessagesReturnValue = [givenInlineMessage]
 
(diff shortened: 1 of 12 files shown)


Tests/MessagingInApp/Views/InAppMessageViewTest.swift
```diff
@@ -221,6 +221,32 @@ class InAppMessageViewTest: UnitTest {
...
+        await simulateSdkFetchedMessages([givenMessageDisplayed])
+
+        // Expect still showing the same message as before the fetch call.
+        XCTAssertEqual(getInAppMessage(forView: inlineView), givenMessageDisplayed)
+
+        await onCloseActionButtonPressed(onInlineView: inlineView)
+
+        // Expect we do not show the expired message but instead close the View.
+        XCTAssertFalse(isInlineViewVisible(inlineView))
+        XCTAssertNil(getInAppMessage(forView: inlineView))
```
ami-aman, line 247: If expired message is not displayed then is it removed from the queue ? This test focuses on what is displayed to the user but wondering if what isn't displayed stays in the queue or is removed and is covered in another test case?

levibostian: > If expired message is not displayed then is it removed from the queue ?

The local SDK queue will not contain a message if it is expired. So, yes, if a message is not displayed, it should also not be in the local queue. 

The backend server is what expires messages. When the mobile SDK performs a fetch for messages from the backend, if the backend does not return a message, it could be because the message has expired. 

the function 

```swift
await simulateSdkFetchedMessages([givenMessageDisplayed])
```

in the test function is what modifies the SDK's queue. Because `givenMessageThatExpires` is not included in the `simulateSdkFetchedMessages` call, then that message will not be in the SDK's local queue.

ami-aman (approved):

PR #763, opened 2024-07-09
chore: handle custom button action on user tap

ami-aman:
https://linear.app/customerio/issue/MBL-328/implement-custom-action-buttons

### Problem
As a marketer, I want to add buttons to inline in-app messages that perform custom logic when they are clicked in my app. 

As an iOS developer, I want to write code in my app that executes when a specific button is pressed in an inline in-app message.

### Solution
When a custom button is tapped by the user
- If the customer _has_ set the delegate for inline messages custom button actions, the SDK will notify the customer's host app where they can act upon this button being pressed
- If the customer _has not_ set the delegate for inline messages custom button actions, the SDK will notify global in-app action handler method `messageActionTaken`
- Just like a modal in-app message then do not close the inline message after button has been pressed.

### Demo

https://github.com/customerio/customerio-ios/assets/91549653/f3287cfb-04b3-4a9f-8c01-29bc53b17d34

### Testing
- Automated test suite should pass (another PR for test cases)
- For testing, user tap actions are tracked:
- If delegate is set then an event is tracked with name `inline custom button action`
- If delegate is not set then an event is tracked with name `inapp action` (same as other in-app action)


Sources/MessagingInApp/Views/InAppMessageView.swift
```diff
@@ -38,6 +44,9 @@ public class InAppMessageView: UIView {
         }
     }
 
+    // Delegate to handle custom action button tap.
+    public var inlineMessageDelegate: InlineMessageDelegate?
```
levibostian, line 48:
```suggestion
    public weak var inlineMessageDelegate: InlineMessageDelegate?
```

Delegates are commonly `weak` in iOS development. I consider this a requirement for this feature as well.

Apps/APN-UIKit/APN UIKit/View/DashboardViewController.swift
```diff
@@ -36,6 +36,11 @@ class DashboardViewController: BaseViewController {
         // For inline Views added with Storyboard, set the elementId to finish setup of the View and begin showing messages.
         inlineInAppViewCreatedInStoryboard.elementId = "dashboard-announcement"
 
+        // This is optional. If set, the delegate method `onInlineCustomButtonAction`
+        // will receive callbacks.
+        // If not set, the global method `messageActionTaken` will handle the callbacks.
+        inlineInAppViewCreatedInStoryboard.inlineMessageDelegate = self
```
levibostian, line 42: The [tech proposal's public API](https://www.notion.so/custio/Mobile-inline-in-app-messages-technical-proposal-7efb22e2af1c40449c72f9cf73770584?pvs=4#d722b4ab9b854d84b86d744cd840d307) specifies the delegate name to be `onActionDelegate`. Could you rename the property name to match this?

Apps/APN-UIKit/APN UIKit/View/DashboardViewController.swift
```diff
@@ -172,3 +177,14 @@ class DashboardViewController: BaseViewController {
         UNUserNotificationCenter.current().add(request, withCompletionHandler: nil)
     }
 }
+
+extension DashboardViewController: InlineMessageDelegate {
+    func onInlineCustomButtonAction(message: Message, action: String, name: String) {
```
levibostian, line 182: The [tech proposal's public API](https://www.notion.so/custio/Mobile-inline-in-app-messages-technical-proposal-7efb22e2af1c40449c72f9cf73770584?pvs=4#d722b4ab9b854d84b86d744cd840d307) specifies the function signature to be `onActionClick(message: InAppMessage, actionValue: String, actionName: String)`. Could you rename the signature to match this?

Apps/APN-UIKit/APN UIKit/View/DashboardViewController.swift
```diff
@@ -172,3 +177,14 @@ class DashboardViewController: BaseViewController {
         UNUserNotificationCenter.current().add(request, withCompletionHandler: nil)
     }
 }
+
+extension DashboardViewController: InlineMessageDelegate {
```
levibostian, line 181: The [tech proposal's public API](https://www.notion.so/custio/Mobile-inline-in-app-messages-technical-proposal-7efb22e2af1c40449c72f9cf73770584?pvs=4#d722b4ab9b854d84b86d744cd840d307) specifies the delegate name to be `InAppMessageViewActionDelegate`. Could you rename this?

Sources/MessagingInApp/Gist/Managers/MessageManager.swift
```diff
@@ -71,6 +71,11 @@ class MessageManager {
         // Expect subclass implements this.
     }
 
+    // Called when custom action button is tapped.
+    func onInlineButtonAction(message: Message, currentRoute: String, action: String, name: String) {
+        // Expect subclass implements this.
+    }
```
levibostian, line 77: Do we need to add any code to the `MessageManager`? [The inline message manager already gets a callback when the action button is clicked](https://github.com/customerio/customerio-ios/pull/763/files#diff-f93ca71bad51e58601b7bc5e6ae8db007cbdbe9ca700730d611472245b322e4bR64) so I don't think we need to add a function to this class? 

Also, this class is designed to be generic without knowledge of modal or inline messages. From the name of the function, this function looks to me like it should belong only in the inline message manager.

levibostian (changes requested):
Great start! There are some important changes to be made but the logic LGTM.

Apps/APN-UIKit/APN UIKit/View/DashboardViewController.swift
ami-aman: I didn't notice it. Thanks for pointing out.

Apps/APN-UIKit/APN UIKit/View/DashboardViewController.swift
ami-aman: ✅

Sources/MessagingInApp/Gist/Managers/MessageManager.swift
ami-aman: Ah ! We don't need it. Thank you.

levibostian (approved):

PR #764, opened 2024-07-09
fix: memory access error when constructing StringAnyEncodable instance

levibostian:
Ticket: https://linear.app/customerio/issue/MBL-404/[bug]-rn-customer-having-ios-app-crash-randomly-when-setting-profile

Based on the stacktrace provided, there is a memory access error when multiple objects are trying to use logger instance. This commit is an attempt at fixing the issue by not passing the logger instance between multiple objects.


levibostian:
Closing in favor of https://github.com/customerio/customerio-ios/pull/767

PR #765, opened 2024-07-09
test: handle custom button action on user tap

ami-aman:
https://linear.app/customerio/issue/MBL-328/implement-custom-action-buttons

As a marketer, I want to add buttons to inline in-app messages that perform custom logic when they are clicked in my app. 
This ticket only caters test cases for the main [PR](https://github.com/customerio/customerio-ios/pull/763).

From the acceptance criteria of the ticket, there are 2 test cases that need tested:

- Given inline delegate is set expect a callback to the custom button action handler (test method `test_onInlineButtonAction_givenDelegateSet_expectCustomCallback`)
- Given inline delegate is not set expect a callback to the in-app global button action handler (test method `test_onInlineButtonAction_givenDelegateNotSet_expectGlobalCallback`)


Tests/MessagingInApp/Views/InAppMessageViewTest.swift
levibostian, line 1: Valuable test cases added thus far. 

One test case I think is missing is: 
* Given 2 inline View instances where both of the inline Views have their own separate inline message delegate set, when a custom action button is clicked on 1 of the messages, expect that only 1 of the inline View instances has it's delegate function called and not the other. 

If it's helpful, [here is a test function example that tests multiple View instances](https://github.com/customerio/customerio-ios/blob/3ad08b7b0c7b5d823b99d59fd2f6dd029d495d71/Tests/MessagingInApp/Views/InAppMessageViewTest.swift#L311-L328).

levibostian (changes requested):

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
```diff
@@ -17,7 +22,7 @@ class InAppMessageViewTest: UnitTest {
 
         DIGraphShared.shared.override(value: queueMock, forType: MessageQueueManager.self)
         DIGraphShared.shared.override(value: engineProvider, forType: EngineWebProvider.self)
-        DIGraphShared.shared.override(value: inlineMessageDelegateMock, forType: InlineMessageDelegate.self)
+        DIGraphShared.shared.override(value: inlineMessageDelegateMock, forType: InAppMessageViewActionDelegate.self)
```
levibostian, line 25:
```suggestion
```

I don't think this is needed. The delegate is set on the View instances in the test functions, not injected through DI.

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
```diff
@@ -7,8 +7,13 @@ import XCTest
...
     private let queueMock = MessageQueueManagerMock()
     private let engineWebMock = EngineWebInstanceMock()
-    private let inlineMessageDelegateMock = InlineMessageDelegateMock()
+    private let inlineMessageDelegateMock = InAppMessageViewActionDelegateMock()
     private var engineProvider: EngineWebProviderStub2!
+    private let eventListenerMock = InAppEventListenerMock()
+    private var messagingInApp: MessagingInAppImplementation {
+        // get MessagingInAppImplementation instance so we can call its methods directly
+        (MessagingInApp.shared.implementation as! MessagingInAppImplementation) // swiftlint:disable:this force_cast
+    }
```
levibostian, line 16: I think it would be helpful if we moved this into the parent `UnitTest` class so it can be re-used by other tests in the suite.

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
```diff
@@ -421,7 +435,48 @@ class InAppMessageViewTest: UnitTest {
...
+        let gistInAppInlineMessage1 = InAppMessage(gistMessage: givenInlineMessage1)
+
+        let givenInlineMessage2 = Message.randomInline
+        queueMock.getInlineMessagesReturnValue = [givenInlineMessage1, givenInlineMessage2]
+
+        // Create two inline in app views
+        let inlineView1 = InAppMessageView(elementId: givenInlineMessage1.elementId!)
+        inlineView1.onActionDelegate = inlineMessageDelegateMock
+        let inlineView2 = InAppMessageView(elementId: givenInlineMessage2.elementId!)
+        inlineView2.onActionDelegate = inlineMessageDelegateMock
```
levibostian, line 459:
```suggestion
        let inlineOnActionDelegate1 = InAppMessageViewActionDelegateMock()
        let inlineOnActionDelegate2 = InAppMessageViewActionDelegateMock()

        // Create two inline in app views
        let inlineView1 = InAppMessageView(elementId: givenInlineMessage1.elementId!)
        inlineView1.onActionDelegate = inlineOnActionDelegate1
        let inlineView2 = InAppMessageView(elementId: givenInlineMessage2.elementId!)
        inlineView2.onActionDelegate = inlineOnActionDelegate2
```

The test function looks good so far. I think the test function could test our View better if each View had it's own delegate instance. 

Then in the assertions, we verify that only 1 of the delegates was called and not the other delegate: 

```swift
XCTAssertFalse(inlineOnActionDelegate2.mockCalled)

XCTAssertTrue(inlineOnActionDelegate1.onActionClickCalled)
...
```

levibostian (approved):
Assuming that you agree with updating the test function `test_onInlineButtonAction_multipleInlineMessages_givenDelegateSetForAll_expectSingleCustomCallback` with separate delegate instances, I'll approve the PR so you can merge after making updates.

PR #767, opened 2024-07-10
fix: memory access error when constructing StringAnyEncodable instance

levibostian:
Ticket: https://linear.app/customerio/issue/MBL-404/[bug]-rn-customer-having-ios-app-crash-randomly-when-setting-profile

The PR is an attempt to fix a bug reported to us. I recommending viewing the ticket for details of the issue, including the stacktrace of the crash. 

Based on the stacktrace provided to us, there is a memory access error when multiple objects are trying to use `logger` instance. Following the stacktrace, I believe that the logger instance is being garbage collected when the `StringAnyInstance` is no longer needed in `CustomerIOImplementation`. By restricting the `logger` instance to only the `CustomerIOImplementation` class, we are removing this scenario from being possible.

### DIFF
--- a/Sources/Common/Util/StringAnyEncodable.swift
+++ b/Sources/Common/Util/StringAnyEncodable.swift
@@ -1,10 +1,9 @@
 import Foundation
 
 public struct StringAnyEncodable: Encodable {
-    private let logger: Logger
     private let data: [String: AnyEncodable]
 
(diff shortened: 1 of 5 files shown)


levibostian:
### Do not merge 

Customer has been testing this fix for us and has reported success thus far. However, we are not yet ready to make a public release with this fix in it. This is why the PR is a draft at the moment until testing has been completed.

PR #768, opened 2024-07-11
feat: support "show another message" action buttons

levibostian:
Part of: https://linear.app/customerio/issue/MBL-360/show-another-message-action-buttons

# Problem 

In the in-app editor, you can create action buttons that have the action "Show another message" that when clicked, they show a different in-app message. This PR is to add support for this. 

# Solution 

The only thing that we need to do in the SDK to support this feature is to show a loading view when you click the "Show another message" button and have that loading view go away when the new message is done loading. The HTML webview handles everything else for us. 

# Testing
* I tested this in UIKit sample app. Compose an in-app message with a "Show another message" action button in it and send to the app. 

Here is a demo: 

https://github.com/customerio/customerio-ios/assets/2041082/a712e89b-4583-4ba5-aad9-dffbaba5603e

> The first message has a "Show another message" action button. When clicked, it animates in the loading view. The 2nd message is the message that I setup in the in-app editor. 

* Automated tests have been added to this PR, testing all QA test cases.


Sources/MessagingInApp/Gist/Managers/MessageManager.swift
```diff
@@ -190,6 +196,8 @@ extension MessageManager: EngineWebDelegate {
 
     func routeChanged(newRoute: String) {
```
levibostian, line 197: [I asked the in-app squad how to get a callback when the "Show another message" button is clicked. They suggested using this function.](https://customerio.slack.com/archives/C036FTE6UJ0/p1720463076324169)

Sources/MessagingInApp/Gist/Managers/MessageManager.swift
```diff
@@ -190,6 +196,8 @@ extension MessageManager: EngineWebDelegate {
...
         Logger.instance.info(message: "Message route changed to: \(newRoute)")
+
+        willChangeMessage(newTemplateId: newRoute)
```
levibostian, line 200: I thought that the name "will change message" and "template id" was a better name for this inline feature. The Gist SDK tends to use "route" when internally on the mobile team, we call that a "template". Also, I thought `willChangeMessage` is more descriptive because when a route is changed, it's because we will be showing a new message.

Sources/MessagingInApp/Views/InAppMessageView.swift
```diff
@@ -150,26 +150,9 @@ public class InAppMessageView: UIView {
...
-            // Set autolayout constraints to position the activity indicator.
-            activityIndicator.translatesAutoresizingMaskIntoConstraints = false
-            NSLayoutConstraint.activate([
-                activityIndicator.centerXAnchor.constraint(equalTo: centerXAnchor),
-                activityIndicator.centerYAnchor.constraint(equalTo: centerYAnchor),
-                activityIndicator.widthAnchor.constraint(equalTo: widthAnchor),
-                activityIndicator.heightAnchor.constraint(equalTo: heightAnchor)
-            ])
-
-            animateFadeInOutInlineView(fromView: currentlyDisplayedInAppWebView, toView: activityIndicator) {
```
levibostian, line 171: All of this code is to animate an activity indicator when a close button is pressed. We can re-use all of this code for this feature when a "show another message" action button is clicked. So I refactored the View to make this code re-usable.

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
Shahroz16, line 1: great tests, can we also a test, when the next message being displayed has a button which also leads to another message?

Shahroz16 (approved):

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
levibostian: added

PR #769, opened 2024-07-11
test: open deep links when an action button is clicked

ami-aman:
Linear ticket : https://linear.app/customerio/issue/MBL-326/open-deep-links-when-an-action-button-is-clicked

As a marketer, I want to be able to open deep links when a specific button in an in-app message is clicked so that I can seamlessly navigate to relevant content within my app or external websites. When the message is displayed inline and the action button is clicked, the SDK should open the deep link, just like it does for push notifications and Modal in-app messages today.

This PR adds test cases to check if the deeplinks are opened when the button on inline inapp message is tapped. Two checks have been added to test the behaviour with a valid link and an invalid link. In both the cases, the inline message does not dismiss and stays on the screen just like in-app messages today.


Sources/MessagingInApp/Gist/Managers/MessageManager.swift
```diff
@@ -24,12 +24,13 @@ class MessageManager {
...
     weak var delegate: GistDelegate?
     private let engineWebProvider: EngineWebProvider = DIGraphShared.shared.engineWebProvider
+    private var deeplinkUtil: DeepLinkUtil = DIGraphShared.shared.deepLinkUtil
 
     init(siteId: String, message: Message) {
         self.siteId = siteId
         self.currentMessage = message
         self.currentRoute = message.templateId
-
+//        self.deeplinkUtil = DIGraphShared.shared.deepLinkUtil
```
levibostian, line 33:
```suggestion
```

Delete unused code?

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
```diff
@@ -476,14 +477,59 @@ class InAppMessageViewTest: IntegrationTest {
...
 
 @MainActor
 extension InAppMessageViewTest {
-    func onCustomActionButtonPressed(onInlineView inlineView: InAppMessageView) {
+    func onCustomActionButtonPressed(onInlineView inlineView: InAppMessageView, isSystem: Bool = false, action: String = "Test") {
         // Triggering the custom action button on inline message from the web engine
         // mocks the user tap on custom action button
-        getWebEngineForInlineView(inlineView)?.delegate?.tap(name: "", action: "Test", system: false)
+        getWebEngineForInlineView(inlineView)?.delegate?.tap(name: "", action: action, system: isSystem)
     }
```
levibostian, line 533: I find this function when it's being used for clicking custom action buttons and deep links hard to read. 

I think it would be easier if there were 2 separate functions: 
1. `onCustomActionButtonPressed(onInlineView inlineView: InAppMessageView, action: String)`
2. `onDeepLinkActionButtonPressed(onInlineView inlineView: InAppMessageView, deepLink: URL)`

levibostian (approved):

Sources/MessagingInApp/Gist/Managers/MessageManager.swift
ami-aman: My bad. I thought I removed it.

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
ami-aman: I do not mind having two methods but the working is same, just the parameter values are changing.

PR #770, opened 2024-07-12
chore: track opened and clicked metrics

ami-aman:
https://linear.app/customerio/issue/MBL-331/track-opened-and-clicked-metrics

### Problem
As a marketer, I want to see how effective my inline messages so that I can make better messaging campaigns in the future. I care about how many messages were shown and how many were acted upon (clicked).

### Solution
- When a customer receives an inline message and is displayed on the screen, the SDK is supposed to track `opened` metric for this message.
- When a customer taps on any custom button and takes an action on the button, the SDK is supposed to track `clicked` metric for this message. 

### Demo

https://github.com/user-attachments/assets/694483af-53a5-433d-b443-9a11625af092

### Testing
- Automated test cases in [another PR](https://github.com/customerio/customerio-ios/pull/774)
- For manual testing, send an inline in-app message to the device
- Once it displays, check activity log for `opened` metric
- Tap on action button on inline message 
- Check activity log for `clicked` metric
- Tap multiple times on the same button
- Check activity log, note that there is just once instance of metric tracking. Activity name `inline custom button action` is related to button action but not tracking the inline metric.

### DIFF
--- a/Sources/MessagingInApp/Gist/Managers/MessageManager.swift
+++ b/Sources/MessagingInApp/Gist/Managers/MessageManager.swift
@@ -21,7 +21,6 @@ class MessageManager {
     let currentMessage: Message
     let gistView: GistView
     private var currentRoute: String
-    private var previouslyProcessedMessages: [Message] = []
     private var elapsedTimer = ElapsedTimer()
     weak var delegate: GistDelegate?
     private let engineWebProvider: EngineWebProvider = DIGraphShared.shared.engineWebProvider
@@ -101,18 +100,7 @@ extension MessageManager: EngineWebDelegate {
         }
     }
 
-    // Function to check if a message has been previously clicked tracked
-    // If not, then track once & store in previouslyProcessedMessages
-    func trackInlineMessageIfNotPreviouslyTracked(action: String, name: String) {
-        let hasBeenTracked = previouslyProcessedMessages.contains { $0.id == currentMessage.id }
-        if !hasBeenTracked {
-            // Add the message to the list of previously processed messages
-            previouslyProcessedMessages.append(currentMessage)
-            trackTapAction(action: action, name: name)
-        }
-    }
-
-    func trackTapAction(action: String, name: String) {
+    func trackClickedMetric(action: String, name: String) {
         // a close action does not count as a clicked action.
         if action != "gist://close" {
             if let deliveryId = getDeliveryId(from: currentMessage) {
@@ -126,12 +114,8 @@ extension MessageManager: EngineWebDelegate {
         // This condition executes only for modal messages and not inline messages.
         if currentMessage.isModalMessage {
             delegate?.action(message: currentMessage, currentRoute: currentRoute, action: action, name: name)
-            trackTapAction(action: action, name: name)
-        } else {
-            // For inline messages, it prevents duplicate tracking and
-            // avoids making multiple event listener calls to delegate methods.
-            trackInlineMessageIfNotPreviouslyTracked(action: action, name: name)
         }
+        trackClickedMetric(action: action, name: name)
         gistView.delegate?.action(message: currentMessage, currentRoute: currentRoute, action: action, name: name)
 
         if let url = URL(string: action), url.scheme == "gist" {
--- a/Sources/MessagingInApp/Gist/Managers/MessageManager.swift
+++ b/Sources/MessagingInApp/Gist/Managers/MessageManager.swift
@@ -86,6 +86,10 @@ class MessageManager {
     func onReplaceMessage(newMessageToShow: Message) {
         // subclass should implement
     }
+
+    func onTapAction(message: Message, currentRoute: String, action: String, name: String) {
+        // subclass should implement
+    }
 }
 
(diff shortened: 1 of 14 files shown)


levibostian (changes requested):
Thank you for continuously improving the PR. 


I think there continues to be a simpler solution that involves re-using the Modal logic. I'll send a DM to discuss this idea further.

ami-aman:
> Thank you for continuously improving the PR.
> 
> I think there continues to be a simpler solution that involves re-using the Modal logic. I'll send a DM to discuss this idea further.
 I’m open to discussing this idea further. Looking forward to your DM so we can explore it in more detail.

Sources/MessagingInApp/Views/InAppMessageView.swift
```diff
@@ -183,7 +180,9 @@ public class InAppMessageView: UIView {
         // Create a new manager for this new message to display and then display the manager's WebView.
         let newInlineMessageManager = InlineMessageManager(siteId: gist.siteId, message: message)
         newInlineMessageManager.inlineMessageDelegate = self
-
+        // Set delegate to track `opened` metric for inline messages
+        // just like modal inapp messages.
```
levibostian, line 184: Great idea to add a comment!

Sources/MessagingInApp/Views/InAppMessageView.swift
```diff
@@ -49,9 +49,6 @@ public class InAppMessageView: UIView {
 
     // Inline messages that have already been shown by this View instance.
     // This is used to prevent showing the same message multiple times when the close button is pressed.
-    //
-    // When persistent vs non-persistent messages and metrics features are implemented in the SDK, this array may be
-    // replaced with a global list of shown messages.
```
levibostian, line 54: Agreed, this comment is no longer relevant.

levibostian (changes requested):
Looking great! I think it's really close to being complete.

Sources/MessagingInApp/Gist/Managers/MessageManager.swift
```diff
@@ -99,9 +101,37 @@ extension MessageManager: EngineWebDelegate {
...
-        delegate?.action(message: currentMessage, currentRoute: currentRoute, action: action, name: name)
+        // This condition executes only for modal messages and not inline messages.
+        if currentMessage.isModalMessage {
+            delegate?.action(message: currentMessage, currentRoute: currentRoute, action: action, name: name)
+            trackTapAction(action: action, name: name)
+        } else {
+            // For inline messages, it prevents duplicate tracking and
+            // avoids making multiple event listener calls to delegate methods.
+            trackInlineMessageIfNotPreviouslyTracked(action: action, name: name)
+        }
```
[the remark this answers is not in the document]
ami-aman: > I don't understand why this PR is only calling delegate.action() for modal messages?

It is because for inline messages we want to [call global event listener only when the delegate is not set](https://github.com/customerio/customerio-ios/blob/dd997c1a0da03fde4b267fc1284460fe651df609/Sources/MessagingInApp/Views/InAppMessageView.swift#L303) and if the delegate is set then we are required to call the custom callback.

To prevent multiple calls to the global listener, this check is required since we are catering to it in [InAppMessageView](https://github.com/customerio/customerio-ios/blob/dd997c1a0da03fde4b267fc1284460fe651df609/Sources/MessagingInApp/Views/InAppMessageView.swift#L301). I hope this make sense.


> With this updated description, I consider the prevent duplicate code in this class to be out-of-scope and it can be removed.

Right, I am updating the code based on the latest acceptance criteria as we talk.

Sources/MessagingInApp/Gist/Managers/MessageManager.swift
```diff
@@ -99,9 +100,22 @@ extension MessageManager: EngineWebDelegate {
...
+    }
+
     func tap(name: String, action: String, system: Bool) {
         Logger.instance.info(message: "Action triggered: \(action) with name: \(name)")
-        delegate?.action(message: currentMessage, currentRoute: currentRoute, action: action, name: name)
+        // This condition executes only for modal messages and not inline messages.
+        if currentMessage.isModalMessage {
+            delegate?.action(message: currentMessage, currentRoute: currentRoute, action: action, name: name)
+        }
```
levibostian, line 117: I do understand [why the delegate is only being called for modal messages](https://github.com/customerio/customerio-ios/pull/770#discussion_r1681111606). I do see value in conditionally deciding if the `delegate.action` should be called or not. 

I think it would be better if the modal and inline `MessageManager` subclasses decided when the `delegate.action` function is called instead of this logic being inside of the generic `MessageManager` instance. 

As a suggestion, [here is a commit explaining what I mean](https://github.com/customerio/customerio-ios/pull/770#discussion_r1681111606). I especially like how the commit keeps the `delegate.action` call contained in the `MessageManager` and removes the `Gist.shared.action` call from the inline View.

levibostian (approved):

ami-aman:
Lint is failing because of FIXME https://github.com/customerio/customerio-ios/pull/774#discussion_r1686982361 that is handled in another https://github.com/customerio/customerio-ios/pull/776.

PR #771, opened 2024-07-12
feat: persistent and non-persistent inline in-app messages

levibostian:
https://linear.app/customerio/issue/MBL-329/after-inline-view-shown-determine-if-it-should-show-again-or-not

# Problem 

As a marketer, I want to customize an inline in-app message to be shown only once or shown multiple times. 

According [to the in-app docs](https://customer.io/docs/journeys/send-in-app-message/#persist-message), persistent messages are defined as: 

> By default, in-app messages only show once—the message disappears after the customer clicks, taps, or refreshes the page. You can check the Persist Message option to show your message across multiple sessions and page views. Persisted messages continue being displayed until the expiration date or until the customer clicks Close or the icon. Make sure your messages include a Close button or an icon so that customers can permanently dismiss your message.

# Solution 

Re-use the existing logic that Modal messages use for persistent and non-persistent messaging. 

# Testing 

* Automated tests will be added in a follow-up PR as there are required changes to the test suite to write these tests. 
* In UIKit sample app, I tested these QA use cases:

1. Given non-persistent message, given message shown, kill-app and re-open it. Expect that we do not show the message again. 
2. Given persistent message, given message shown, kill-app and re-open it. Expect that we do show the message again. Click the close button and kill app, re-open app, expect to not see the message again.

### DIFF
--- a/Sources/MessagingInApp/Gist/Gist.swift
+++ b/Sources/MessagingInApp/Gist/Gist.swift
@@ -8,7 +8,7 @@ protocol GistInstance: AutoMockable {
 }
 
 public class Gist: GistInstance, GistDelegate {
-    var shownModalMessageQueueIds: Set<String> = [] // all modal messages that have been shown in the app already.
+    var shownMessageQueueIds: Set<String> = [] // all messages that have been shown in the app already.
     var messageQueueManager: MessageQueueManager {
         DIGraphShared.shared.messageQueueManager
     }
@@ -135,16 +135,9 @@ public class Gist: GistInstance, GistDelegate {
     }
 
     func logMessageView(message: Message) {
-        // This function body reports metrics and makes sure that messages are not shown 2+ times.
-        // For inline messages, we have not yet implemented either of these features.
-        // Therefore, if the message is not a modal, exit early.
-        guard message.isModalMessage else {
-            return
-        }
-
         messageQueueManager.removeMessageFromLocalStore(message: message)
         if let id = message.id {
-            shownModalMessageQueueIds.insert(id)
+            shownMessageQueueIds.insert(id)
         }
         let userToken = UserManager().getUserToken()
         LogManager(siteId: siteId, dataCenter: dataCenter)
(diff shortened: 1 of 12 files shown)


Sources/MessagingInApp/Gist/Managers/MessageManager.swift
```diff
@@ -85,6 +88,13 @@ class MessageManager {
     func onReplaceMessage(newMessageToShow: Message) {
         // subclass should implement
     }
+
+    func removePersistentMessage() {
```
levibostian, line 92: This function got moved from `ModalMessageManager` into `MessageManager` so that modals and inline can both use it. Function body is unchanged.

Sources/MessagingInApp/Gist/Managers/MessageQueueManager.swift
```diff
@@ -89,6 +89,7 @@ class MessageQueueManagerImpl: MessageQueueManager {
     func getInlineMessages(forElementId elementId: String) -> [Message] {
         let messages = localMessageStore
             .filter { $0.value.elementId == elementId } // Only get messages for a specific elementid.
+            .filter { !hasMessageBeenShown($0.value) } // If we have already shown message, do not show again.
```
levibostian, line 92: Filter inline messages that have already been shown. 

This logic is identical to how modal messages are not shown if they have already been shown (inside of `showMessageIfMeetsCriteria`).

Sources/MessagingInApp/Gist/Gist.swift
```diff
@@ -135,16 +135,9 @@ public class Gist: GistInstance, GistDelegate {
     }
 
     func logMessageView(message: Message) {
```
levibostian, line 137: This function does 2 things: Tracks `opened` metrics and makes sure that a message is not shown again. 

This is the same function used by modal messages to make messages not display twice.

ami-aman (approved):

PR #772, opened 2024-07-15
fix: remove gist pre-loading

Shahroz16:
With our in-app rendering move from flutter to html, we don't need to pre-load flutter engine to download assets.

more context here: https://customerio.slack.com/archives/C036FTE6UJ0/p1721046751535539


mrehan27 (approved):

mrehan27:
Closing this since it's no longer needed

PR #774, opened 2024-07-16
test: track opened and clicked metrics

ami-aman:
Linear ticket : https://linear.app/customerio/issue/MBL-331/track-opened-and-clicked-metrics

As a marketer, I want to see how effective my inline messages so that I can make better messaging campaigns in the future. I care about how many messages were shown and how many were acted upon (clicked).

This PR adds unit test cases for tracking the effectiveness of inline in-app messages. The tests ensure that metrics for `opened` and `clicked` events are accurately recorded. The primary focus is to prevent duplicate event tracking, which is critical for inline messages due to their potential to be displayed and interacted with multiple times within the app. The tests cover scenarios to verify that:

	1.	Inline messages report “opened” when displayed.
	2.	Inline messages report “clicked” when an action button is pressed.
	3.	Duplicate events are not tracked for the same inline message.
	4.      Calls to respective action methods based on delegate is set or not set and it still satisfy condition #3 above.
	5.	Inline messages continue to display after action button is pressed.

### DIFF
--- a/Sources/MessagingInApp/MessagingInAppImplementation.swift
+++ b/Sources/MessagingInApp/MessagingInAppImplementation.swift
@@ -9,7 +9,7 @@ class MessagingInAppImplementation: MessagingInAppInstance {
 
     private var eventListener: InAppEventListener?
     private let threadUtil: ThreadUtil
-    let eventBusHandler: EventBusHandler
+    var eventBusHandler: EventBusHandler
 
     init(diGraph: DIGraphShared, moduleConfig: MessagingInAppConfigOptions) {
         self.moduleConfig = moduleConfig
@@ -58,6 +58,14 @@ class MessagingInAppImplementation: MessagingInAppInstance {
     func setEventListener(_ eventListener: InAppEventListener?) {
         self.eventListener = eventListener
     }
+
+    // Sets a custom event bus handler for unit testing.
+    // Do not use it in production.
+    #if DEBUG
+    func setEventBusHandler(_ eventBusHandler: EventBusHandler) {
+        self.eventBusHandler = eventBusHandler
+    }
+    #endif
 
     // Dismiss in-app message
     func dismissMessage() {
--- a/Sources/MessagingInApp/MessagingInAppImplementation.swift
+++ b/Sources/MessagingInApp/MessagingInAppImplementation.swift
@@ -9,7 +9,7 @@ class MessagingInAppImplementation: MessagingInAppInstance {
 
     private var eventListener: InAppEventListener?
     private let threadUtil: ThreadUtil
-    var eventBusHandler: EventBusHandler
+    let eventBusHandler: EventBusHandler
 
     init(diGraph: DIGraphShared, moduleConfig: MessagingInAppConfigOptions) {
         self.moduleConfig = moduleConfig
@@ -58,14 +58,6 @@ class MessagingInAppImplementation: MessagingInAppInstance {
     func setEventListener(_ eventListener: InAppEventListener?) {
         self.eventListener = eventListener
     }
-
-    // Sets a custom event bus handler for unit testing.
-    // Do not use it in production.
-    #if DEBUG
-    func setEventBusHandler(_ eventBusHandler: EventBusHandler) {
-        self.eventBusHandler = eventBusHandler
-    }
-    #endif
 
     // Dismiss in-app message
     func dismissMessage() {
--- a/Tests/MessagingInApp/EngineWeb/EngineWebProviderStub.swift
+++ b/Tests/MessagingInApp/EngineWeb/EngineWebProviderStub.swift
@@ -12,17 +12,3 @@ class EngineWebProviderStub: EngineWebProvider {
         return newMock
     }
 }
-
-// In order to test multiple instances of inline Views, you need to create multiple separate EngineWebInstance instances for each View instance.
-// This is a new stub that allows this.
-// A follow-up refactor that makes this stub the default would be a good chance to the test suite.
-class EngineWebProviderStub2: EngineWebProvider {
-    func getEngineWebInstance(configuration: EngineWebConfiguration) -> EngineWebInstance {
-        let newMock = EngineWebInstanceMock()
-
-        // Set defaults on the mock to make it useable in tests.
-        newMock.view = UIView() // Code expects Engine to return a View that displays in-app message. Return any View to get code under test to run.
-
-        return newMock
-    }
-}
--- a/Tests/MessagingInApp/Views/InAppMessageViewTest.swift
+++ b/Tests/MessagingInApp/Views/InAppMessageViewTest.swift
@@ -8,15 +8,25 @@ class InAppMessageViewTest: IntegrationTest {
     private let queueMock = MessageQueueManagerMock()
     private let engineWebMock = EngineWebInstanceMock()
     private let inlineMessageDelegateMock = InAppMessageViewActionDelegateMock()
-    private var engineProvider: EngineWebProviderStub2!
     private let eventListenerMock = InAppEventListenerMock()
     private let deeplinkUtilMock = DeepLinkUtilMock()
-    private var eventBusHandlerMock = EventBusHandlerMock()
+
+    private var eventBusMetricsTracked: [TrackInAppMetricEvent] = []
+
     override func setUp() {
+        // We want to assert metrics are tracked, but we don't want to mock the eventbus for our integration test in this class.
+        // Therefore, listen for real events being posted to verify in the test functions.
+        eventBusMetricsTracked = []
+        diGraphShared.eventBusHandler.addObserver(TrackInAppMetricEvent.self) { event in
+            self.eventBusMetricsTracked.append(event)
+        }
+
         super.setUp()
 
         DIGraphShared.shared.override(value: queueMock, forType: MessageQueueManager.self)
         DIGraphShared.shared.override(value: deeplinkUtilMock, forType: DeepLinkUtil.self)
+
+        messagingInAppImplementation.setEventListener(eventListenerMock)
     }
 
     // MARK: View constructed
@@ -424,12 +434,10 @@ class InAppMessageViewTest: IntegrationTest {
         XCTAssertEqual(view.widthConstraints.map(\.constant), [givenWidthUserSetsOnView])
     }
 
-    // MARK: - onInlineButtonAction
+    // MARK: - custom action button
 
     @MainActor
     func test_onInlineButtonAction_givenDelegateSet_expectCustomCallback() async {
-        messagingInAppImplementation.setEventListener(eventListenerMock)
-
         let givenInlineMessage = Message.randomInline
         let gistInAppInlineMessage = InAppMessage(gistMessage: givenInlineMessage)
         queueMock.getInlineMessagesReturnValue = [givenInlineMessage]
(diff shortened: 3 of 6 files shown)


levibostian (commented):
I am going to hold off on a review of this PR until the parent PR gets approved as [the implementation may change](https://github.com/customerio/customerio-ios/pull/770#pullrequestreview-2180796795) which could reduce the tests required.

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
```diff
@@ -658,6 +659,157 @@ class InAppMessageViewTest: IntegrationTest {
...
+
+        XCTAssertTrue(isInlineViewVisible(inlineView))
+
+        // Check if messageShown is called
+        XCTAssertTrue(eventListenerMock.messageShownCalled)
+        XCTAssertEqual(eventListenerMock.messageShownCallsCount, 1)
+
+        // Also check for postEvent calls
+        XCTAssertTrue(eventBusHandlerMock.postEventCalled)
+        XCTAssertEqual(eventBusHandlerMock.postEventCallsCount, 1)
```
levibostian, line 677: In order to verify that a `opened` metric was tracked, we should verify that the eventbus was passed an event for tracking an `opened` metric. Not just check that some event was posted to it.

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
```diff
@@ -658,6 +659,157 @@ class InAppMessageViewTest: IntegrationTest {
...
+    @MainActor
+    func test_onInlineMessageShown_expectTrackOpenedMetric_expectGlobalMessageShownListener() async {
+        let inlineView = await showInlineMessageForMetrics()
+
+        XCTAssertTrue(isInlineViewVisible(inlineView))
+
+        // Check if messageShown is called
+        XCTAssertTrue(eventListenerMock.messageShownCalled)
+        XCTAssertEqual(eventListenerMock.messageShownCallsCount, 1)
+
```
levibostian, line 674: I think we should move assertions dealing with the global event listener out of these test functions. 

I also consider tests for the global event listener out of scope for this ticket.

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
```diff
@@ -658,6 +659,157 @@ class InAppMessageViewTest: IntegrationTest {
...
+        // Also check for postEvent calls
+        XCTAssertTrue(eventBusHandlerMock.postEventCalled)
+        XCTAssertEqual(eventBusHandlerMock.postEventCallsCount, 1)
+    }
+
+    @MainActor
+    func test_givenMultipleInlineMessageInQueue_BothDisplayedAndOneDismissed_expectTrackMultipleMessageShownListenerCalls() async {
```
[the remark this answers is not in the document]
levibostian: What is this test trying to verify? Reading the function name and the test body, I don't understand what test case we are trying to verify. 

To help with this, maybe more clear comments in the test function? Maybe the test function is testing too much and it could be split into multiple test functions?

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
```diff
@@ -658,6 +659,157 @@ class InAppMessageViewTest: IntegrationTest {
...
+
+        // Also check for postEvent calls
+        XCTAssertTrue(eventBusHandlerMock.postEventCalled)
+        XCTAssertEqual(eventBusHandlerMock.postEventCallsCount, 2)
+    }
+
+    // MARK: - Track click
+
+    @MainActor
+    func test_onInlineMessageShown_givenDelegateNotSet_onSingleButtonTap_expectTrackClickedMetric_expectGlobalMessageActionTakenCallback() async {
```
levibostian, line 723: I think this test should be broken into multiple test functions. 

* given button clicked - expect track metric
* given delegate not set, given button click, expect global event listener called. <-- I consider this test case out of scope for this PR.

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
```diff
@@ -658,6 +659,157 @@ class InAppMessageViewTest: IntegrationTest {
...
+
+        // Also check for postEvent calls
+        XCTAssertTrue(eventBusHandlerMock.postEventCalled)
+        // The first post call occurs when the message is shown,
+        // and the second post call is triggered by the button click action.
+        XCTAssertEqual(eventBusHandlerMock.postEventCallsCount, 2)
+    }
+
+    @MainActor
+    func test_onInlineMessageShown_givenDelegateNotSet_onMultipleButtonTap_expectGlobalMessageActionTakenCallback() async {
```
levibostian, line 747: I think this test should be broken into multiple test functions. I also think event listener tests out of scope for this ticket.

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
```diff
@@ -658,6 +659,157 @@ class InAppMessageViewTest: IntegrationTest {
...
+        // Tap button again
+        onCustomActionButtonPressed(onInlineView: inlineView)
+        XCTAssertEqual(eventListenerMock.messageActionTakenCallsCount, 2)
+        // Tap the button again and verify that
+        // the count of post calls to track metrics updates
+        XCTAssertEqual(eventBusHandlerMock.postEventCallsCount, 3)
+    }
+
+    @MainActor
+    func test_onInlineMessageShown_givenDelegateSet_onMultipleButtonTap_expectSingleTrackClickedMetric_expectGlobalMessageActionTakenCallback() async {
```
levibostian, line 778: I think this test should be broken into multiple test functions. I also think event listener tests out of scope for this ticket.

Sources/MessagingInApp/MessagingInAppImplementation.swift
```diff
@@ -59,6 +59,14 @@ class MessagingInAppImplementation: MessagingInAppInstance {
...
     }
 
+    // Sets a custom event bus handler for unit testing.
+    // Do not use it in production.
+    #if DEBUG
+    func setEventBusHandler(_ eventBusHandler: EventBusHandler) {
+        self.eventBusHandler = eventBusHandler
+    }
+    #endif
+
```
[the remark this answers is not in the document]
ami-aman: For some reason I am unsure of, majority of the test cases fail on using `DIGraphShared.shared.override(value: eventBusHandlerMock, forType: EventBusHandler.self)` in `setup` method and this is why I avoided to use the above in setup.

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
ami-aman: Since some cases of [global event listeners](https://linear.app/customerio/issue/MBL-330/send-events-to-gist-event-listeners) are handled in this ticket as these both are linked to each other, and I have mentioned in the other [PR](https://github.com/customerio/customerio-ios/pull/775) that certain cases for messageShown and messageActionTaken are taken care of in this PR, I think these cases are relevant here.

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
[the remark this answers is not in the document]
ami-aman: Since some cases of [global event listeners](https://linear.app/customerio/issue/MBL-330/send-events-to-gist-event-listeners) are handled in this ticket as these both are linked to each other, and I have mentioned in the other https://github.com/customerio/customerio-ios/pull/775 that certain cases for messageShown and messageActionTaken are taken care of in this PR, I think these cases are relevant here.

> What is this test trying to verify?
I think the method name is confusing, I will update it.

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
ami-aman: messageShown and messageActionTaken cases are handled in this PR as per description [here](https://github.com/customerio/customerio-ios/pull/775)

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
ami-aman: Broken.
`messageShown` and `messageActionTaken` cases are handled in this PR as per description https://github.com/customerio/customerio-ios/pull/775

Sources/MessagingInApp/MessagingInAppImplementation.swift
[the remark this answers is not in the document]
levibostian: Now that I think about it, with [the test class moving towards an integration test](https://github.com/customerio/customerio-ios/pull/776), I don't think it's a good idea to mock the eventbus. We want the eventbus to post real events since it's posting more then push metrics. 

Another idea [is to add another observer to the real eventbus instance](https://github.com/customerio/customerio-ios/pull/774/commits/36bc1c72291226afb5dbc5b7753766613bd0fb8c#diff-d45d2abf8395422b4e74c4ceec60e0157a10d5489cfd9ac12d88c35bf9afec26R17-R22) so we don't need to mock the eventbus but still run assertions against the events that got posted to it.

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
```diff
@@ -664,212 +668,108 @@ class InAppMessageViewTest: IntegrationTest {
...
 @MainActor
 extension InAppMessageViewTest {
-    func onCustomActionButtonPressed(onInlineView inlineView: InAppMessageView) {
+    func onCustomActionButtonPressed(onInlineView inlineView: InAppMessageView) async {
         // Triggering the custom action button on inline message from the web engine
         // mocks the user tap on custom action button
         getWebEngineForInlineView(inlineView)?.delegate?.tap(name: "", action: "Test", system: false)
+
+        // when a button is pressed, an eventbus event is posted for metrics tracking. Return function after the tracked event is posted.
+        await waitForEventBusEventsToPost()
```
levibostian, line 772: With not mocking the eventbus, we need to add this wait so check if push metrics were reported after an event was posted to the eventbus.

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
```diff
@@ -664,212 +668,108 @@ class InAppMessageViewTest: IntegrationTest {
...
-        onCustomActionButtonPressed(onInlineView: inlineView)
-
-        XCTAssertTrue(inlineMessageDelegateMock.onActionClickCalled)
-        XCTAssertEqual(inlineMessageDelegateMock.onActionClickCallsCount, 1)
+    func test_givenMultipleInlineMessageRendered_expectTrackMultipleMessageShownListenerCalls() async {
+        // FIXME: This test is failing because the queueMock is being mocked and is not a real instance.
+        // the queue's real instance is able to return a message by elemenetid. when you mock the queue, it will
+        // ignore the elementid and always return the same message.
+        // So, this test wants each inline View to have a separate message but instead each view will get the same message.
+        // PR that would fix this test: https://github.com/customerio/customerio-ios/pull/776
```
levibostian, line 743: I think this is a good test function. Unfortunately, because the test class mocks the queue, the test will always fail.

If the PR 776 gets merged first, I am OK with merging that code into this PR and removing this `FIXME` comment before merging. Or, if needed, we can comment out this test and uncomment it after that 776 gets merged.

Tests/MessagingInApp/EngineWeb/EngineWebProviderStub.swift
```diff
@@ -12,17 +12,3 @@ class EngineWebProviderStub: EngineWebProvider {
         return newMock
     }
 }
-
-// In order to test multiple instances of inline Views, you need to create multiple separate EngineWebInstance instances for each View instance.
-// This is a new stub that allows this.
-// A follow-up refactor that makes this stub the default would be a good chance to the test suite.
-class EngineWebProviderStub2: EngineWebProvider {
```
levibostian, line 19: This class was removed in feature branch. Not needed, so I am removing it again.

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
ami-aman: For now, I will comment out this test method and will uncomment after PR#776 is merged.

ami-aman:
Lint is failing because of FIXME [added](https://github.com/customerio/customerio-ios/pull/774#discussion_r1686982361) that is handled in another [PR](https://github.com/customerio/customerio-ios/pull/776).

PR #775, opened 2024-07-16
test: send events to gist event listeners

ami-aman:
https://linear.app/customerio/issue/MBL-330/send-events-to-gist-event-listeners

### Problem
As a iOS developer, I want to execute code when actions occur in my app such as when a message is shown or an action button is pressed.

### Solution
Gist provides 4 callback methods/event listeners as following:
- messageShown
- errorWithMessage
- messageActionTaken
- messageDismissed

Inline message do not call `messageDismissed` function so we are to handle the first three callbacks. 

Metrics `messageShown` and `messageActionTaken`  have already been handled in [MBL-331](https://linear.app/customerio/issue/MBL-331/track-opened-and-clicked-metrics).

For `messageActionTaken` is called when an in-app custom action is clicked on the inline message and an action callback is not attached to the inline View.

______

The tests cover scenarios to verify that
- Given that the inline message fails to show, expect `errorWithMessage` to be called
- Given that the inline message fails to show, but setEventListener is not set, expect not to call `errorWithMessage` 
- Given that the inline message shows and user taps close button, `messageActionTaken` is called


Tests/MessagingInApp/Views/InAppMessageViewTest.swift
```diff
@@ -522,6 +522,67 @@ class InAppMessageViewTest: IntegrationTest {
...
+        await onDoneRenderingInAppMessageWithError(givenInlineMessage, insideOfInlineView: inlineView)
+
+        // Inline message does not display
+        XCTAssertFalse(isInlineViewVisible(inlineView))
+
+        // errorWithMessage not called as eventListener is not set
+        XCTAssertFalse(eventListenerMock.errorWithMessageCalled)
+    }
+
+    // messageActionTaken when close button is tapped
```
levibostian, line 562: For modal messages today, [close buttons should not](https://github.com/customerio/customerio-ios/blob/2f4e4e4d07bedffc251b0f0a19b3e2bd55b0375d/Sources/MessagingInApp/MessagingInAppImplementation.swift#L97-L102) trigger a `messageActionTaken` call to the event listener. Inline messages have the same requirement and therefore, this test should fail.

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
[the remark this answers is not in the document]
ami-aman: > Given inline message rendered, expect messageShown to be called
> Given onActionDelegate set on View instance, expect messageActionTaken not called when an action button clicked.
> Given onActionDelegate not set on View instance, expect messageActionTaken called when an action button clicked.

Test cases covered in another [PR](https://github.com/customerio/customerio-ios/pull/774/files). Will update case #2 in the same PR as it has more of `messageShown` testing

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
ami-aman: I expected the same but the code [does not seem to have any condition](https://github.com/customerio/customerio-ios/blob/2f4e4e4d07bedffc251b0f0a19b3e2bd55b0375d/Sources/MessagingInApp/MessagingInAppImplementation.swift#L97-L108) to prevent calling `messageActionTaken` event listener. It actually triggers from [here](https://github.com/customerio/customerio-ios/blob/1be401784e9467c1733ab06df9001c1f10f80eae/Sources/MessagingInApp/Gist/Managers/MessageManager.swift#L111), may be we would want to push a check [here](https://github.com/customerio/customerio-ios/blob/1be401784e9467c1733ab06df9001c1f10f80eae/Sources/MessagingInApp/Gist/Managers/MessageManager.swift#L111) to prevent handling button tap when close button is pressed to prevent it from happening. 🤔

levibostian: :facepalm: Sorry, my brain had a moment it looks like. 

As you said, there is no condition for preventing calling `messageActionTaken`. Ignore my comment.

levibostian (commented):
I am going to wait on approving the PR until we can verify we have enough test cases in this PR. 

To do that, [we have to wait until the parent PR is approved](https://github.com/customerio/customerio-ios/pull/775/files#r1680912448) in case the parent PR has test cases modified.

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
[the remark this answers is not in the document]
ami-aman: 2nd case also handled in [PR](https://github.com/customerio/customerio-ios/pull/774/files)

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
```diff
@@ -658,6 +658,81 @@ class InAppMessageViewTest: IntegrationTest {
...
+        XCTAssertEqual(eventListenerMock.errorWithMessageReceivedArguments, InAppMessage(gistMessage: givenInlineMessage))
+
+        // Other listeners not called
+        XCTAssertFalse(eventListenerMock.messageActionTakenCalled)
+        XCTAssertFalse(eventListenerMock.messageShownCalled)
+        XCTAssertFalse(eventListenerMock.messageDismissedCalled)
+    }
+
+    @MainActor
+    func test_givenAttemptToShowInlineMessageFails_expectMessageNotShown() async {
```
levibostian, line 686: Good test case. 

For organization purposes, I think this function belongs with [these tests](https://github.com/customerio/customerio-ios/blob/7a77b18fa660ed3383280812deeaf918e9d80813/Tests/MessagingInApp/Views/InAppMessageViewTest.swift#L76) in the file.

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
```diff
@@ -658,6 +658,81 @@ class InAppMessageViewTest: IntegrationTest {
...
+    @MainActor
+    func test_givenAttemptToShowInlineMessage_EventListenerNotSet_expectFailToShow_expectNoCallback() async {
+        let givenInlineMessage = Message.randomInline
+        queueMock.getInlineMessagesReturnValue = [givenInlineMessage]
+
+        let inlineView = InAppMessageView(elementId: givenInlineMessage.elementId!)
+        await onDoneRenderingInAppMessageWithError(givenInlineMessage, insideOfInlineView: inlineView)
+
+        // Inline message does not display
+        XCTAssertFalse(isInlineViewVisible(inlineView))
```
levibostian, line 707:
```suggestion
```

Since this test function is only wanting to test the event listener, I don't think this check is needed.

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
```diff
@@ -658,6 +658,81 @@ class InAppMessageViewTest: IntegrationTest {
...
+
+        // errorWithMessage not called as eventListener is not set
+        XCTAssertFalse(eventListenerMock.errorWithMessageCalled)
+    }
+
+    // messageActionTaken when close button is tapped
+    @MainActor
+    func test_givenShowInlineMessage_givenTapCloseButton_expectInlineToHide_expectMessageActionTakenCalled() async {
```
levibostian, line 715:
```suggestion
    func test_eventListener_givenTapCloseButton_expectCallEventListener() async {
```

The test function and test body makes me think we are testing multiple things in 1 test: 
1. Testing the event listener is called
2. Testing when you click the close button, the View is dismissed. 

Ideally, the test function should only test 1 thing in it. For the scope of this PR, I think you're trying to test the event listener. Therefore, I would edit the test function name to indicate you're only testing the event listener and modify the test body to only test the event listener, not visibility of the View - those tests should be in their own test functions.

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
```diff
@@ -658,6 +658,81 @@ class InAppMessageViewTest: IntegrationTest {
...
+
+        // Inline message is dismissed
+        XCTAssertFalse(isInlineViewVisible(inlineView))
+        XCTAssertNil(getInAppMessage(forView: inlineView))
+
+        // messageActionTaken called
+        XCTAssertTrue(eventListenerMock.messageActionTakenCalled)
+        XCTAssertEqual(eventListenerMock.messageActionTakenCallsCount, 1)
+    }
 }
```
levibostian, line 736: Good test case of testing when a close button pressed, we expect the event listener to be called. 

I think we need to add test functions for all of the other action button types and see if they are expected to call the event listener or not.

levibostian (changes requested):

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
ami-aman: The test case is in the same file as you linked above. 🤔

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
ami-aman: By `other action button types` do you mean to test show another message button as we have custom buttons, close button and show another message button, where the first two cases have been handled and the only one I can think of is show another message. Am I understanding it right ?

ami-aman:
Lint is failing because of FIXME https://github.com/customerio/customerio-ios/pull/774#discussion_r1686982361 that is handled in another https://github.com/customerio/customerio-ios/pull/776.

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
levibostian: Yes, it is the same file. 

This suggestion is that the test `test_givenAttemptToShowInlineMessageFails_expectMessageNotShown` is more related to the behavior of displaying in-app messages when an error happens. Not related to event listeners. So for organization purposes, I was suggesting to move the file to a different `// MARK:` section in the file such as `// MARK: Display in-app message`

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
levibostian: Yes, you have the correct understanding. 

Some examples of test functions I think would be valuable: 
* Given click on "Show another message" action button, expect to not call any of the event listener functions or delegate. 
* Given click on deep link button, expect to not call any of the event listener functions or delegate. 

Also, [because the `messageDismissed` event listener function is never called for inline Views](https://github.com/customerio/docs/pull/2136/files#diff-c5c6c29c2bebc6fbace8740bb04c6e9f39f2fd497eb6e5a0532f40baef7a1cd2R207), only for Modals, I think it would be valuable to write test functions where we perform actions on an inline View and make sure that we do not call the `messageDismissed` event listener function. 

Example tests: 
* Given click on close button, expect to not call `messageDismissed`. 

Maybe these test functions have already been written in a different PR, I am not sure. Since this PR is talking about tests for the global event listener, I am only going off of what I see in this PR. If you know of these tests existing already, sounds good to me!

levibostian:
Talked to Aman in DMs about me finishing up the tests and merging this PR.

PR #776, opened 2024-07-18
test: convert inline in-app View tests to integration tests

levibostian:
Unblocks writing tests for ticket: https://linear.app/customerio/issue/MBL-329/

If the inline View tests were integration tests instead of unit tests, we would be able to write more automated tests and reduce false positives in our test functions. 

This change makes the following modifications to the inline in-app test suite: 
* Only mocking done is mocking the HTTP response when fetching messages. 
* Be able to mock HTTP requests with any given `[Message]` array.

### DIFF
--- a/Sources/Common/Util/JsonAdapter.swift
+++ b/Sources/Common/Util/JsonAdapter.swift
@@ -59,17 +59,42 @@ public class JsonAdapter {
         self.log = log
     }
 
-    public func fromDictionary<T: Decodable>(_ dictionary: [AnyHashable: Any]) -> T? {
-        do {
-            let jsonData = try JSONSerialization.data(withJSONObject: dictionary)
+    // MARK: Dictionary to JSON
 
-            return fromJson(jsonData)
-        } catch {
-            log.error("\(error.localizedDescription), dictionary: \(dictionary)")
+    public func fromDictionary<T: Decodable>(_ dictionary: [AnyHashable: Any]) -> T? {
+        commonDecodeDictionary(dictionary)
+    }
+
+    public func fromDictionary<T: Decodable>(_ dictionary: [[AnyHashable: Any]]) -> T? {
+        commonDecodeDictionary(dictionary)
+    }
+
+    public func fromDictionary(_ dictionary: [AnyHashable: Any]) -> Data? {
+        commonDecodeDictionaryToData(dictionary)
+    }
+
+    public func fromDictionary(_ dictionary: [[AnyHashable: Any]]) -> Data? {
+        commonDecodeDictionaryToData(dictionary)
+    }
+
+    private func commonDecodeDictionary<T: Decodable>(_ dictionary: Any) -> T? {
+        guard let data = commonDecodeDictionaryToData(dictionary) else {
+            return nil
         }
 
+        return fromJson(data)
+    }
+
+    private func commonDecodeDictionaryToData(_ dictionary: Any) -> Data? {
+        do {
+            return try JSONSerialization.data(withJSONObject: dictionary)
+        } catch {
+            log.error("\(error.localizedDescription), object: \(dictionary)")
+        }
         return nil
     }
+
+    // MARK: JSON to Dictionary
 
     public func toDictionary<T: Encodable>(_ obj: T) -> [AnyHashable: Any]? {
         guard let data = toJson(obj) else {
--- a/Sources/MessagingInApp/Views/InAppMessageView.swift
+++ b/Sources/MessagingInApp/Views/InAppMessageView.swift
@@ -120,8 +120,12 @@ public class InAppMessageView: UIView {
 
     // Updates the state of the View, if needed. Call as often as you need if an event happens that may cause the View to need to update.
     private func refreshView(forceShowNextMessage: Bool = false) {
-        guard let elementId = elementId else {
+        defer {
+            // Always call the refreshViewListener at the end of the function to know processing is done.
             refreshViewListener?()
+        }
+
+        guard let elementId = elementId else {
             return // we cannot check if a message is available until element id set on View.
         }
 
@@ -132,7 +136,6 @@ public class InAppMessageView: UIView {
             // We are already displaying or rendering a messsage. Do not show another message until the current message is closed.
             // The main reason for this is when a message is tracked as "opened", the Gist backend will not return this message on the next fetch call.
             // We want to coninue showing a message even if the fetch no longer returns the message and the message is currently visible.
-            refreshViewListener?()
             return
         }
 
@@ -141,8 +144,6 @@ public class InAppMessageView: UIView {
         } else {
             dismissInAppMessage()
         }
-
-        refreshViewListener?()
     }
 
     // Function to check if a message has been previously shown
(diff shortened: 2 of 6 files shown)


Sources/Common/Util/JsonAdapter.swift
levibostian, line 1: JsonAdapter is a util class for performing common JSON operations. 

The adapter was missing a way to convert `[[AnyHashable: Any]]` into a JSON string so I modified the adapter to add this.

Sources/MessagingInApp/Views/InAppMessageView.swift
```diff
@@ -47,6 +47,13 @@ public class InAppMessageView: UIView {
     // Delegate to handle custom action button tap.
     public weak var onActionDelegate: InAppMessageViewActionDelegate?
 
+    // When a fetch request is performed, it's an async operation to have the inline View notified about this fetch and the inline View processing the fetch.
+    // There is currently no easy way to know when the inline View has finished processing the fetch.
+    // This listener is a hack for our automated tests to know when the inline View has finished processing the fetch.
+    //
+    // See linear ticket MBL-427 to learn more about this limitation in our tests.
```
levibostian, line 54: As mentioned in comments, this listener is added because of async operations being performed and our tests needing to wait for them to complete. Instead of something like `sleep`, we have a listener the tests can hook into. 

This solution is not ideal but it's a simple solution to unblock integration tests for inline Views. 

See ticket https://linear.app/customerio/issue/MBL-427/ for ideas on how we can improve this solution.

Tests/MessagingInApp/Core/IntegrationTest.swift
```diff
@@ -38,12 +38,97 @@ open class IntegrationTest: UnitTest {
...
+    func setupMockFetchResponse(messagesFetched messages: [Message]) {
+        if messages.isEmpty {
+            setupHttpResponse(code: 200, body: "[]".data)
+            return
+        }
+
+        // Construct a JSON string that will become the HTTP response body.
+        // The format of the JSON is the same as what /api/v1/users endpoint returns from Gist backend.
+        //
+        // We are opting to construct JSON strings manually with a dictionary because the data types used for parsing HTTP response bodies, `UserQueueResponse`, is not easily able to inherit the `Codable` protocol because of `[String: Any]`.
```
levibostian, line 52: Constructing JSON strings using a dictionary isn't ideal. I prefer the method used in Track v1 tests, where we construct Swift objects and generate JSON strings from them. However, the in-app SDK data types can't easily be made `Codable`.

Since the in-app SDK only performs this 1 HTTP request, I don't consider this a big piece of tech debt in the codebase. I can write a ticket to improve this in the future if others feel this is important.

Tests/MessagingInApp/Core/IntegrationTest.swift
```diff
@@ -73,6 +158,18 @@ extension IntegrationTest {
...
+        // Triggering the custom action button on inline message from the web engine
+        // mocks the user tap on custom action button
+        getWebEngineForInlineView(inlineView)?.delegate?.tap(name: "", action: "Test", system: false)
+    }
+
+    func onDeepLinkActionButtonPressed(onInlineView inlineView: InAppMessageView, deeplink: String) {
+        // Triggering the custom action button on inline message from the web engine
+        // mocks the user tap on custom action button
+        getWebEngineForInlineView(inlineView)?.delegate?.tap(name: "", action: deeplink, system: true)
+    }
```
levibostian, line 171: This is not new code, but code moved from `Tests/MessagingInApp/Views/InAppMessageViewTest.swift` into this function to try and keep all button click functions in 1 place for re-usability.

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
```diff
@@ -561,16 +580,12 @@ class InAppMessageViewTest: IntegrationTest {
 
 @MainActor
 extension InAppMessageViewTest {
-    func onCustomActionButtonPressed(onInlineView inlineView: InAppMessageView) {
```
levibostian, line 564: Function moved https://github.com/customerio/customerio-ios/pull/776/files#r1683284581

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
```diff
@@ -561,16 +580,12 @@ class InAppMessageViewTest: IntegrationTest {
...
-        // Triggering the custom action button on inline message from the web engine
-        // mocks the user tap on custom action button
-        getWebEngineForInlineView(inlineView)?.delegate?.tap(name: "", action: "Test", system: false)
-    }
+    func setupQueueMock() -> MessageQueueManagerMock {
+        let mock = MessageQueueManagerMock()
 
-    func onDeepLinkActionButtonPressed(onInlineView inlineView: InAppMessageView, deeplink: String) {
```
levibostian, line 570: Function moved https://github.com/customerio/customerio-ios/pull/776/files#r1683284581

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
```diff
@@ -561,16 +580,12 @@ class InAppMessageViewTest: IntegrationTest {
 
 @MainActor
 extension InAppMessageViewTest {
-    func onCustomActionButtonPressed(onInlineView inlineView: InAppMessageView) {
-        // Triggering the custom action button on inline message from the web engine
-        // mocks the user tap on custom action button
-        getWebEngineForInlineView(inlineView)?.delegate?.tap(name: "", action: "Test", system: false)
-    }
+    func setupQueueMock() -> MessageQueueManagerMock {
```
levibostian, line 583: Some test functions still benefit from mocking the `MessageQueueManager`. Added this util function to easily do that even though the test class by default does not mock the manager.

Sources/MessagingInApp/Views/InAppMessageView.swift
```diff
@@ -127,6 +135,7 @@ public class InAppMessageView: UIView {
             // We are already displaying or rendering a messsage. Do not show another message until the current message is closed.
             // The main reason for this is when a message is tracked as "opened", the Gist backend will not return this message on the next fetch call.
             // We want to coninue showing a message even if the fetch no longer returns the message and the message is currently visible.
+            refreshViewListener?()
```
ami-aman, line 138: Wondering if `refreshViewListener` is required to be call in all conditions then we could put it in the start of the method or defer ? Or there's a reason that it's been added this way ?

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
```diff
@@ -445,8 +459,10 @@ class InAppMessageViewTest: IntegrationTest {
         XCTAssertEqual(inlineMessageDelegateMock.onActionClickReceivedArguments?.actionValue, "Test")
         XCTAssertEqual(inlineMessageDelegateMock.onActionClickReceivedArguments?.actionName, "")
 
+        // FIXME: The global listener should not be called when the delegate is set.
+
         // Also check that the global listener is not called
-        XCTAssertFalse(eventListenerMock.messageActionTakenCalled)
+        XCTAssertTrue(eventListenerMock.messageActionTakenCalled)
```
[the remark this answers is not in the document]
levibostian: I got to remove these fixme messages as base branch has fixed these bugs!

PR #777, opened 2024-07-18
test: persistent and non-persistent messages

levibostian:
Automated tests for: https://linear.app/customerio/issue/MBL-329/after-inline-view-shown-determine-if-it-should-show-again-or-not

Given the acceptance criteria in the ticket and the description of persistent and non-persistent messages, I wrote automated test functions for the feature. 

I tried to write automated tests for all test cases that we would use to manually QA test this feature. If you believe a test case is missing, please call it out.

### DIFF
--- a/Tests/MessagingInApp/Views/InAppMessageViewTest.swift
+++ b/Tests/MessagingInApp/Views/InAppMessageViewTest.swift
@@ -576,6 +576,66 @@ class InAppMessageViewTest: IntegrationTest {
         // If url is valid, check if `handleDeepLink` method is called
         XCTAssertFalse(deeplinkUtilMock.handleDeepLinkCalled)
     }
+
+    // MARK: persistent and non-persistent
+
+    @MainActor
+    func test_persistentAndNonPersistent_givenNonPersistentMessage_givenMessageShown_expectMessageNotShownAgain() async {
+        let givenInlineMessage = Message(elementId: .random, persistent: false)
+        await simulateSdkFetchedMessages([givenInlineMessage], verifyInlineViewNotifiedOfFetch: nil)
+
+        let inlineView = InAppMessageView(elementId: givenInlineMessage.elementId!)
+        XCTAssertEqual(getInAppMessage(forView: inlineView), givenInlineMessage)
+        await onDoneRenderingInAppMessage(givenInlineMessage, insideOfInlineView: inlineView)
+
+        // Expect that when a new inline View is being constructed, it does not show the non-persistent message that has already been shown.
+        let differentView = InAppMessageView(elementId: givenInlineMessage.elementId!)
+        XCTAssertNil(getInAppMessage(forView: differentView))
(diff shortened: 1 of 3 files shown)


Tests/MessagingInApp/Views/InAppMessageViewTest.swift
```diff
@@ -576,6 +576,66 @@ class InAppMessageViewTest: IntegrationTest {
...
+        XCTAssertEqual(getInAppMessage(forView: inlineView), givenInlineMessage)
+        await onDoneRenderingInAppMessage(givenInlineMessage, insideOfInlineView: inlineView)
+
+        // Expect that when a new inline View is being constructed, it does not show the non-persistent message that has already been shown.
+        let differentView = InAppMessageView(elementId: givenInlineMessage.elementId!)
+        XCTAssertNil(getInAppMessage(forView: differentView))
+    }
+
+    @MainActor
+    func test_persistentAndNonPersistent_givenNonPersistentMessage_givenMessageNotYetShown_expectCanDisplayMessageMultipleTimes() async {
```
ami-aman, line 597: The test case looks great but the method name is a bit confusing specifically `expectCanDisplayMessageMultipleTimes`. It gives an impression that a message is shown multiple times whereas in the test case, the message is just shown once if not shown earlier.

ami-aman (commented):
Valuable test cases added.

The only test case that I think could be added further is to test that persistent message is not shown after expiration of the message.

ami-aman (approved):
Non-blocking comment regarding test method name & a new test case suggestion.

Tests/MessagingInApp/Views/InAppMessageViewTest.swift
levibostian: It's important for the test function name to match what the expected behavior is. Let me try and explain the test function better and if you can think of a better name, more then happy to change it to be more clear. 

Non-persistent messages should only be shown one time. Once it is shown, the message will not be shown the next time the screen is refreshed. 

However, this is only true if the message has been rendered. Here is a scenario where a non-persistent message could be shown multiple times: 
* Open the Dashboard screen of an app. There is 1 inline View instances on that screen with element id "announcement". 
* When the inline View instance gets constructed on the Dashboard screen, it gets a non-persistent message and begins rendering it. 
* Before the non-persistent message is done rendering, the user navigates to the Settings screen. This screen also has an inline View with element id "announcement".  Because the non-persistent message has not yet been loaded on the Dashboard screen, the Settings screen inline View *also* gets the non-persistent message and begins rendering it. 
* When the non-persistent message is done rendering on Settings screen, we see the non-persistent message. If the user navigates back to Dashboard screen, they will see the non-persistent message shown as well because it will also finish rendering there. 

I don't personally consider this a bug since this configuration of the app might be rare to see. I still think it's important to write this kind of test to document known behavior of the feature.

PR #778, opened 2024-07-19
ci: make fastlane config files re-usable by SDK wrappers

levibostian:
Part of: https://linear.app/customerio/issue/MBL-421/

# Problem 

I am trying to make the fastlane (tool we use to make builds of sample apps and uploading to Firebase) config files re-usable by the SDK wrapper projects. 

This PR is a companion to [this RN PR](https://github.com/customerio/customerio-reactnative/pull/287) where I was able to compile the RN sample apps using fastlane on my local development machine. 

# Solution 

The fastlane config file (Fastfile) that contains the logic on building sample apps is currently not re-usable by the SDK wrappers. To resolve this, I broke up the function `ios build` into multiple different functions. Making each of the functions less coupled to 1 specific SDK project. 

The end goal is that from the iOS SDK or the RN SDK, you can run `fastlane ios build` and you're able to build, sign, and upload the app. 

# Testing

* I am able to compile iOS apps using `fastlane ios build` on my local machine
* If all CI status checks pass in CI, the config files were refactored successfully without breaking.

### DIFF
--- a/.github/actions/build-sample-app/action.yml
+++ b/.github/actions/build-sample-app/action.yml
@@ -100,7 +100,14 @@ runs:
       shell: bash
       run: cat $GITHUB_EVENT_PATH
 
-    - name: Build app via Fastlane 
+    - name: Update app version to uniquely identify build for easier finding later. 
+      uses: maierj/fastlane-action@v3.1.0
+      with:
+        lane: "iossdk_update_app_version"
+        subdirectory: "Apps/${{ inputs.sample-app }}"
+        options: ${{ inputs.fastlane-build-args }}
+
+    - name: Build app and upload for testing
       uses: maierj/fastlane-action@v3.1.0
       with:
         lane: "ios build"
--- a/.github/actions/build-sample-app/action.yml
+++ b/.github/actions/build-sample-app/action.yml
@@ -103,7 +103,7 @@ runs:
     - name: Update app version to uniquely identify build for easier finding later. 
       uses: maierj/fastlane-action@v3.1.0
       with:
-        lane: "iossdk_update_app_version"
+        lane: "ios iossdk_update_app_version"
         subdirectory: "Apps/${{ inputs.sample-app }}"
         options: ${{ inputs.fastlane-build-args }}
 
--- a/Apps/fastlane/Fastfile
+++ b/Apps/fastlane/Fastfile
@@ -33,41 +33,26 @@ end
 
 # Functions specific to iOS go in the platform block. 
 platform :ios do 
-  # Builds the iOS app and uploads compiled app for internal testing. 
+  # Builds the iOS app and (optionally) uploads compiled app for internal testing. 
   # Build will only be uploaded if the environment variables for uploading are set.
   #
+  # Pre-requisites:
+  # 1. Assumes the build and app version are set in the Xcode project.
+  # 2. Assumes that the CIO workspace environment files are generated (Env.swift, for example). 
+  #
   # Usage: 
-  # `fastlane ios build` <-- if you want to generate a new build number and app version. Common to use for creating unique QA builds. Each build number is guaranteed to be unique.
-  # `fastlane ios build build_number:123 app_version:1.2.3` <-- if you want to specify the build number and app version.   
+  # `fastlane ios build` 
   lane :build do |arguments|    
-    download_ci_code_signing_files
-    
-    new_build_number = arguments[:build_number] || get_new_build_version()
-    new_app_version = arguments[:app_version] || get_new_app_version()
-    build_notes = get_build_notes()
-    test_groups = get_build_test_groups()
-
-    # Modify the source code with the new app version and build number before we compile the iOS app. This is a good idea to do to make installing builds on a test device easier. 
-    # The iOS OS might give errors when trying to install a build of an app if the app is already installed on the device. Having unique build number or app version can avoid those errors. 
-    update_ios_app_versions(
-      build_number: new_build_number,
-      app_version: new_app_version
-    )
-    
-    version_update_script_path = '../../../scripts/update-version.sh'
-    # Since this Fastfile is used by SDK wrapper projects that might not contain this script file, we skip running it if the file does not exist. 
-    if File.exist?(version_update_script_path)
-      UI.important("Updating the SDK's source code version to non-production version. This allows the sample apps to show the SDK version at runtime for app user to better understand the version of the SDK they are running.")
-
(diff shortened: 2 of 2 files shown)


.github/actions/build-sample-app/action.yml
```diff
@@ -100,7 +100,14 @@ runs:
       shell: bash
       run: cat $GITHUB_EVENT_PATH
 
-    - name: Build app via Fastlane 
+    - name: Update app version to uniquely identify build for easier finding later.
```
levibostian, line 103: Because I broke up the `ios build` function into multiple functions to make it re-usable by SDK wrappers, we need to run the function `iossdk_update_app_version` before running `ios build` on the CI to make sample app builds.

Apps/fastlane/Fastfile
```diff
@@ -33,41 +33,26 @@ end
...
-    uses_cocoapods = File.exists?("../Podfile")
-    if uses_cocoapods 
-      UI.message "Project uses CocoaPods. Going to skip SPM dependency downloading."
+    # For fastlane to build and sign the app, it needs the code signing files.
+    # Even if you already have them downloaded to your computer, we need to run the functions again because when 
+    # the code signing functions run, fastlane sets up the environment it uses later to sign the app.
+    if ENV["CI"]
+      download_ci_code_signing_files
+    else 
+      download_development_code_signing
```
levibostian, line 52: This change was made for me during development so I could call `ios build` from my local dev machine. Previously, you could only test fastlane config files by running the CI server (slow!).

Apps/fastlane/Fastfile
```diff
@@ -33,41 +33,26 @@ end
...
-    # Since this Fastfile is used by SDK wrapper projects that might not contain this script file, we skip running it if the file does not exist. 
-    if File.exist?(version_update_script_path)
-      UI.important("Updating the SDK's source code version to non-production version. This allows the sample apps to show the SDK version at runtime for app user to better understand the version of the SDK they are running.")
-
-      sh("#{version_update_script_path} \"#{new_app_version}.#{new_build_number}\"")
-    end 
-
-    uses_cocoapods = File.exists?("../Podfile")
-    if uses_cocoapods 
-      UI.message "Project uses CocoaPods. Going to skip SPM dependency downloading."
```
levibostian, line 67: To make `ios build` more re-usable, this code was moved into separate functions. 

`ios build` function now has the assumption that before you run it, you have already updated the app version and build number.

Apps/fastlane/Fastfile
```diff
@@ -79,8 +64,8 @@ platform :ios do
       build_path: "build", # Save derived data to Apps/X/build folder. CI will parse this folder later for tracking SDK size. 
       cloned_source_packages_path: "spm_packages", # Save SPM dependencies to Apps/X/spm_packages folder. We can cache this folder to speed up builds on CI. 
     )
-
-    are_environment_variables_set_for_build_uploading = !ENV["FIREBASE_APP_DISTRIBUTION_SERVICE_ACCOUNT_CREDS_B64"].empty?
+
```
levibostian, line 67: I noticed that `.empty?` was throwing ruby errors when running. Not sure why this is happening now when this has been working for a while now. This change is a bug fix.

Apps/fastlane/Fastfile
```diff
@@ -94,13 +79,18 @@ platform :ios do
...
-
-  lane :update_ios_app_versions do |arguments| 
+  end  
+  
+  # Given app and build version, update the Xcode project. Usually you call this before building the app. 
+  # There are multiple places to modify the version numbers. This function's job is guarantee when function is done, the version numbers are updated.
+  #
+  # Usage:
+  # `fastlane ios update_app_version build_number:123 app_version:1.2.3`
+  lane :update_app_version do |arguments|
```
levibostian, line 93: Added comments to this function and also renamed it. 

The old name felt verbose: `ios update_ios_app_versions`. 
New name feels better: `ios update_app_version`.

Apps/fastlane/Fastfile
```diff
@@ -142,6 +132,24 @@ platform :ios do
   end 
 end 
 
+# Functions specific to just the iOS SDK repo can go here: 
+
+lane :iossdk_update_app_version do |arguments|
```
levibostian, line 137: Code that is specific to the iOS SDK got moved out of `ios build` and into this separate function.

Apps/fastlane/Fastfile
```diff
@@ -79,8 +64,11 @@ platform :ios do
       build_path: "build", # Save derived data to Apps/X/build folder. CI will parse this folder later for tracking SDK size. 
       cloned_source_packages_path: "spm_packages", # Save SPM dependencies to Apps/X/spm_packages folder. We can cache this folder to speed up builds on CI. 
     )
+    
+    are_environment_variables_set_for_build_uploading = ENV["FIREBASE_APP_DISTRIBUTION_SERVICE_ACCOUNT_CREDS_B64"]
 
-    are_environment_variables_set_for_build_uploading = !ENV["FIREBASE_APP_DISTRIBUTION_SERVICE_ACCOUNT_CREDS_B64"].empty?
+    environment_variables_required_for_build_uploading = ["FIREBASE_APP_DISTRIBUTION_SERVICE_ACCOUNT_CREDS_B64"] # array of keys 
+    are_environment_variables_set_for_build_uploading = environment_variables_required_for_build_uploading.all? { |key| ENV[key] != nil && !ENV[key].empty? } # check if all environment variables are set and not empty
```
Shahroz16, line 71: I feel like this can be simplified 

```
# Constants should be defined using uppercase letters and underscores
REQUIRED_ENV_VARS = ["FIREBASE_APP_DISTRIBUTION_SERVICE_ACCOUNT_CREDS_B64"]

# Check if all required environment variables are set and not empty
are_environment_variables_set_for_build_uploading = REQUIRED_ENV_VARS.all? { |key| ENV[key].to_s.strip.length > 0 }
```

Shahroz16 (approved):
looks good just added 1 suggestion to simplify

Apps/fastlane/Fastfile
levibostian: I do appreciate the suggestion. I am going to suggest we keep what the PR has because this line of code has caused issues in the past and this code seems to be stable from my testing. Not being ruby experts, I prefer using code that works over code that can be simplified.

Shahroz16:
I am closing this PR since the wrappers now include the Fastlane files. We can revisit this for potential enhancements in the future.

PR #780, opened 2024-08-01
chore: change height of SwiftUI inline view after message renders

levibostian:
https://linear.app/customerio/issue/MBL-402/

# Problem

The parent PR created an inline View for SwiftUI, but it has issues with dynamic height when rendering in-app messages. Currently, the height and width must be hard-coded. This PR fixes this by setting the initial height to 0 and adjusting it to the in-app message's height after rendering, respecting the aspect ratio.

# Solution

Use UIKit's `intrinsicContentSize`. By implementing `intrinsicContentSize` in the inline UIView to return the in-app message's width and height after rendering, SwiftUI automatically adjusts the constraints.

I tried lots of different solutions that were both too complex and also did not work as expected. Most of the solutions I tried were modifying auto layout constraints on the UIKit inline View, the `UIViewRepresentable`, and `superviews`. This PR's solution doesn't try to modify the behavior of SwiftUI but instead communicate when the view's size changes and have SwiftUI respond to those changes. Let SwiftUI do it's magic for us. 

# Known limitations

Further work needed:
* Animations: Height changes instantly without smooth transitions.
* Auto Layout warning: Xcode gives one warning; will address this in the animation PR.
* Fixes for UIKit: This change may break the inline View for UIKit. To be most productive, I plan to address UIKit fixes after the SwiftUI solution is complete. Future PRs ([such as this one](https://github.com/customerio/customerio-ios/pull/783)) will follow for UIKit fixes. 

# Testing 

* In SwiftUI sample app, send inline message to device. 
* When the in-app message renders, you should see the inapp message instantly jump to the correct height and show the inline message. The width should be 100% of the width of the parent container. 

Demo: 

https://github.com/user-attachments/assets/68f0270b-7ac6-4c86-b3f6-0135535859b8


Apps/CocoaPods-FCM/src/View/DashboardView.swift
```diff
@@ -51,8 +51,7 @@ struct DashboardView: View {
                     Group {
                         InAppMessageViewRepresentable(elementId: "dashboard-announcement", onActionClick: { _, _, _ in
                             print("Custom callback received")
-                        })
-                        .frame(height: 350)
+                        }).frame(maxWidth: .infinity)
```
levibostian, line 54: No longer required for customer to set a hard coded height and width. Instead, customers are required to set a width. 

I found you do need to set the `frame()` but you can use `maxWidth: .infinity` which is setting width to 100% of the parent container. 

I am not yet sure if it will be required in the final public API of the SwiftUI inline View if this is a requirement for customers to set. We do have a requirement in our UIKit View that customers must set a width so I think it's OK to also have this requirement for the SwiftUI View?

Sources/MessagingInApp/Views/InAppMessageView.swift
```diff
@@ -391,4 +410,4 @@ public struct InAppMessageViewRepresentable: UIViewRepresentable {
         }
     }
 }
-#endif
+#endif // swiftlint:disable:this file_length
```
levibostian, line 413: I would be OK moving the SwiftUI code into a separate file to resolve this swift lint error. Also OK with keeping all of it in 1 file.

ami-aman: IMO, it will be a good idea to move SwiftUI code into a separate file.

levibostian: [Opened a new PR](https://github.com/customerio/customerio-ios/pull/781) with that change.

Apps/CocoaPods-FCM/src/View/DashboardView.swift
ami-aman: I think it should be SwiftUI customer;s responsibility to set the frame based on their app requirements, just like we expect from UIKIt customers.

ami-aman: Also, would you mind posting a quick video of this change in PR description or share the link if it is already there in any other PR? I am wondering how it would take the shape with infinity width. 🤔

Sources/MessagingInApp/Views/InAppMessageView.swift
```diff
@@ -288,6 +288,23 @@ public class InAppMessageView: UIView {
...
+
+    override public var intrinsicContentSize: CGSize {
+        // Return the current content size
+        contentSize
+    }
+
+    func updateContentSize(_ newSize: CGSize) {
+        // Update the content size and invalidate the intrinsic content size
+        contentSize = newSize
+    }
```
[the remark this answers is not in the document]
levibostian: Great question. I did not mention UIKit impact from this PR. 

I updated the PR description to mention the impact to UIKit: 

> Fixes for UIKit: This change may break the inline View for UIKit. To be most productive, I plan to address UIKit fixes after the SwiftUI solution is complete. Future PRs (https://github.com/customerio/customerio-ios/pull/783) will follow for UIKit fixes.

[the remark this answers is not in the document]
ami-aman: Thanks for updating the description, it adds more clarity.

ami-aman (approved):

Apps/CocoaPods-FCM/src/View/DashboardView.swift
levibostian: No problem. I added a video to description.

ami-aman: Thanks for that. :)

PR #782, opened 2024-08-02
chore: animate height changes inline in-app SwiftUI View

levibostian:

# Problem 

The SwiftUI inline in-app message View instantly jumps from height 0 to N height when an in-app message is rendered. We need a smooth animation when the height changes like we have in the UIKit inline View. 

Besides this, the SwiftUI view currently does not do anything when the close button is pressed. 

This PR solves all of these problems by animating between all of the states. 

# Solution 

Originally, I thought that the SwiftUI inline View implementation was going to be a very small wrapper around the UIKit inline View. I was thinking that if the UIKit inline View implementation would animate height changes, SwiftUI would automatically be able to display those UIKit animations, too. 

I heave learned that when using `UIViewRepresentable`, SwiftUI adds auto layout constraints that you do not control. These SwiftUI auto layout constraints are not compatible with the UIKit inline View height animations that are implemented. 

To get around this issue, we can make an architecture change. We have a SwiftUI implementation and a separate UIKit implementation where each of them implement their own animation logic: 

Architecture before this PR: 
```
SwiftUI View (UIViewRepresentable)
  \__ UIKit View <--- performs animations when height changes
    \__ UIKit WebView renders in-app message
```

Architecture after this PR: 
```
SwiftUI View <--- performs animations when height changes
  \__ UIKit WebView renders in-app message

UIKit View <--- performs animations when height changes 
  \__ UIKit WebView renders in-app message
```

This PR is the SwiftUI View that performs animations when height of WebView changes. A follow-up PR will be made for the UIKit View.

# Testing 

* Open Dashboard in SwiftUI sample app. Send an in-app message. Expect inline View to be height 0 when first displayed. 
* When in-app message renders, animate height to N to display the in-app message. 
* Click close button. If there is another message in local queue, animate the height change to the height of next message. 
* Click close button. If no other message in local queue, animate height to 0 to dismiss in-app message. 

Demo video:

https://github.com/user-attachments/assets/3cdfa23e-06da-4baa-8512-f654b87e292e


Apps/CocoaPods-FCM/src/View/DashboardView.swift
```diff
@@ -49,7 +49,7 @@ struct DashboardView: View {
                     }
                     Text("What would you like to test?")
                     Group {
-                        InAppMessageViewRepresentable(elementId: "dashboard-announcement", onActionClick: { _, _, _ in
+                        InlineInAppMessage(elementId: "dashboard-announcement", onActionClick: { _, _, _ in
```
levibostian, line 52: I am going to ask the team for feedback on what the name of the SwiftUI inline View could be. As Aman has mentioned before, there is a high chance of clashing with the UIKit names being used. Going to have to find a unique name just for SwiftUI.

Sources/MessagingInApp/Views/UIKitInline.swift
```diff
@@ -243,28 +249,7 @@ public class InAppMessageView: UIView {
...
-            // Example, given this UI:
-            // UIViewController
-            // └── UIStackView
-            //    └── InAppMessageView
-            // ...If we call layoutIfNeeded on superview (UIStackView), the animation will not work as expected.
-            // This is also why it's important that we do QA testing on the inline View when it's nested in a UIStackView.
-            self.getRootSuperview()?.layoutIfNeeded()
-        })
-
-        runningHeightChangeAnimation?.startAnimation()
```
levibostian, line 267: As the PR description mentioned, I am proposing a small architecture change where UIKit and SwiftUI have their own View that performs height animations. 

This means that this existing UIKit inline View will have the height animation code moved into a separate UIKit class (I'll do this in a follow-up PR). I deleted the animation code in this PR because it was making the SwiftUI inline View not work. 

Instead, replace all height animation changes to simply sending the height change event to the `onSizeChangedDelegate` delegate.

Sources/MessagingInApp/Views/UIKitInline.swift
```diff
@@ -243,28 +249,7 @@ public class InAppMessageView: UIView {
...
+        onSizeChangedDelegate?.onSizeChanged(height: 0, width: frame.width)
```
levibostian, line 252: When the close button pressed, tell the delegate to change height to 0.

Sources/MessagingInApp/Views/UIKitInline.swift
```diff
@@ -314,14 +299,14 @@ extension InAppMessageView: InlineMessageManagerDelegate {
...
-
             guard let inAppMessageView = self.inAppMessageView else {
                 return
             }
 
             inAppMessageView.isHidden = false
-            self.animateHeight(to: height)
+
+            self.updateContentSize(.init(width: width, height: height))
+            self.onSizeChangedDelegate?.onSizeChanged(height: height, width: width)
```
levibostian, line 309: When message rendered and the height changes, tell the delegate.

Apps/CocoaPods-FCM/src/View/DashboardView.swift
ami-aman: Based on the slack conversation, I suggest the following names :
`InlineMessageView` for UIKit 
`InlineMessageViewRepresentable` or `InlineMessageViewComponent` for SwiftUI. I have given reasoning on our slack conversation in case it helps.

Sources/MessagingInApp/Views/UIKitInline.swift
ami-aman: I do not see test cases being added/updated. Do you plan to write test cases for these updates in another PR ?

levibostian: Good question. I plan to add tests in a follow-up PR since some refactors will need to be done to get UIKit solution working again.

Apps/CocoaPods-FCM/src/View/DashboardView.swift
levibostian: I'll rename the class in a follow-up PR if that's OK with you.

ami-aman: That works for me. I will go ahead and approve this PR to unblock you.

ami-aman (approved):

PR #783, opened 2024-08-02
refactor: fix uikit inline View issues after SwiftUI architecture changes 

levibostian:

# Problem 

The [previous PR](https://github.com/customerio/customerio-ios/pull/782) made an architecture change to how animations are performed for displaying inline in-app messages. The previous PR was the SwiftUI implementation, this PR is the UIKit implementation. 

# Solution 

This PR is not adding new logic but instead doing a refactor that splits the existing `InAppMessageView` UIKit View into 2 separate UIKit UIViews. 

1. Create a generic UIKit UIView that renders inline in-app messages and sends events to a delegate. 
2. Create a UIKit wrapper around this generic View that implements features such as: animating height changes and showing a loading View when the "show another message" action buttons are clicked. 

These 2 classes match the proposed architecture changes the last PR mentioned: 

```
SwiftUI View <--- performs animations when height changes
  \__ UIKit WebView renders in-app message

UIKit View <--- performs animations when height changes 
  \__ UIKit WebView renders in-app message
```

# Testing 

I tested this scenario on both sample apps. 

* Open Dashboard in sample app. Send an in-app message. Expect inline View to be height 0 when first displayed. 
* When in-app message renders, animate height to N to display the in-app message. 
* Click action button that displays a new message. Expect see a an Activity Indicator when the button is clicked and next message is rendering. 
* When the next message renders, expect the height to animate to the new height of the inline message. 
* Click close button. Because there is no other message in local queue, animate height to 0 to dismiss in-app message. 

Demo for SwiftUI app:

https://github.com/user-attachments/assets/65eb2f0b-aa95-479a-ae02-e21ab5018893

Demo for UIKit app: 

https://github.com/user-attachments/assets/53c2df20-e294-4266-96b2-ee3f32fe05db

# Follow-up PRs 

There are some changes that are either out-of-scope for this PR or changes that would make this PR large and hard to review. Expect to see these changes in a follow-up PR: 

1. Fix automated tests for inline View. 
2. Modify the public class name for the UIKit and SwiftUI inline Views.


Sources/MessagingInApp/Views/GistInlineInAppMessageView.swift
levibostian, line 1: To create this new UIKit UIView, I took the existing `InAppMessageView` and I removed: 
* height change animations. Instead, this UIView modifies the `intrinsicContentSize` to change the size of the View. 
* the Activity indicator View for when we are loading a new in-app message after "show another message" is clicked. 

Instead, this UIView sends events to a `delegate`. The `InAppMessageView` in this PR is what receives these `delegate` events and does the height change animations and shows the activity indicator View.

Sources/MessagingInApp/Views/SwiftUIInline.swift
levibostian, line 1: I modified this SwiftUI inline View because it now wraps the `GistInlineInAppMessageView`, just like `InAppMessageView` does. 

I still have to add an activity indicator to this class before I open the PR.

Sources/MessagingInApp/Views/UIKitInline.swift
levibostian, line 1: This file got greatly simplified! 

This class has less responsibilities. Lots of the logic got moved into `GistInlineInAppMessageView`!

Sources/MessagingInApp/Views/GistInlineInAppMessageView.swift
```diff
@@ -0,0 +1,242 @@
...
+    func onInlineButtonAction(message: Message, currentRoute: String, action: String, name: String) -> Bool
+    // If the inline View is going to show a different message. Example: When a user clicks "show another message" action buttons.
+    func willChangeMessage(newTemplateId: String, onComplete: @escaping () -> Void)
+}
+
+/**
+ UIView that can render and display an web in-app message.
+
+ Responsibilities:
+ * Given an elementId, it will get a message to display for that elemenetId.
```
ami-aman, line 22:
```suggestion
 * Given an elementId, it will get a message to display for that elementId.
```

Sources/MessagingInApp/Views/UIKitInline.swift
```diff
@@ -8,10 +8,6 @@ public protocol InAppMessageViewActionDelegate: AnyObject, AutoMockable {
     func onActionClick(message: InAppMessage, actionValue: String, actionName: String)
 }
 
-public protocol InAppMessageViewSizeChangedDelegate: AnyObject {
-    func onSizeChanged(height: CGFloat, width: CGFloat)
-}
```
levibostian, line 13: No longer need this since it was being used by SwiftUI.

Sources/MessagingInApp/Views/UIKitInline.swift
```diff
@@ -100,6 +70,22 @@ public class InAppMessageView: UIView {
...
+        inlineInAppMessageView.delegate = self
+        addSubview(inlineInAppMessageView)
+
+        inlineInAppMessageView.translatesAutoresizingMaskIntoConstraints = false
+        NSLayoutConstraint.activate([
+            inlineInAppMessageView.centerXAnchor.constraint(equalTo: centerXAnchor),
+            inlineInAppMessageView.centerYAnchor.constraint(equalTo: centerYAnchor),
+            inlineInAppMessageView.widthAnchor.constraint(equalTo: widthAnchor),
+            inlineInAppMessageView.heightAnchor.constraint(equalTo: heightAnchor)
+        ])
```
levibostian, line 87: Add the base `GistInlineInAppMessageView` UIKit view as a subview, set the `delegate`, and then all other logic in this View comes from the new delegate functions.

Sources/MessagingInApp/Views/UIKitInline.swift
```diff
@@ -113,66 +99,41 @@ public class InAppMessageView: UIView {
...
-        }
-
-        let queueOfMessagesForGivenElementId = localMessageQueue.getInlineMessages(forElementId: elementId)
-        let messageAvailableToDisplay = queueOfMessagesForGivenElementId.first { !hasBeenPreviouslyShown($0) }
-
-        if !forceShowNextMessage, isRenderingOrDisplayingAMessage {
-            // We are already displaying or rendering a messsage. Do not show another message until the current message is closed.
-            // The main reason for this is when a message is tracked as "opened", the Gist backend will not return this message on the next fetch call.
-            // We want to coninue showing a message even if the fetch no longer returns the message and the message is currently visible.
+    public func onMessageRendered(width: CGFloat, height: CGFloat) {
```
levibostian, line 104: All of the logic inside of these `delegate` functions is not new, but moved from different functions. 

Example: `onMessageRendered` function body is code that was in `onSizeChanged()`.

Sources/MessagingInApp/Views/UIKitInline.swift
```diff
@@ -113,66 +99,41 @@ public class InAppMessageView: UIView {
...
-        // If a different message is currently being shown, we want to replace the currently shown message with new message.
-        if isRenderingOrDisplayingAMessage {
-            showLoadingView {
-                // After animation is over, cleanup resources since we no longer need to show the previous message.
-                self.stopShowingMessageAndCleanup()
-                self.beginShowing(message: message)
-            }
-        } else {
-            beginShowing(message: message)
+    public func willChangeMessage(newTemplateId: String, onComplete: @escaping () -> Void) {
```
levibostian, line 133: `willChangeMessage` is called when the "show another message" action buttons are clicked. This function is called while the new message is being rendered.

Sources/MessagingInApp/Views/GistInlineInAppMessageView.swift
```diff
@@ -0,0 +1,242 @@
+import CioInternalCommon
+import Foundation
+import UIKit
+
+// Event listener for interactions and state changes to an inline inapp message that's rendered.
+public protocol GistInlineInAppMessageViewDelegate: AnyObject {
+    // After a message is finished rendering and the size may have changed.
+    func onMessageRendered(width: CGFloat, height: CGFloat)
```
ami-aman, line 8: How about using `CGSize` as a parameter instead of two different parameters `width` and `height` ?

Sources/MessagingInApp/Views/UIKitInline.swift
ami-aman: I think it would be helpful to add a comment on this method since it is `willChangeMessage` and will be called on "show another message" button click.

Sources/MessagingInApp/Views/UIKitInline.swift
```diff
@@ -206,50 +167,25 @@ public class InAppMessageView: UIView {
...
+            self.heightConstraint?.constant = height // Changing the height in animation block indicates we want to animate the height change.
 
-        stopShowingMessageAndCleanup()
+            // Since we modified constraint, perform a UI refresh to apply the change.
+            // It's important that we call layoutIfNeeded on the topmost superview in the hierarchy. During development, there were animiation issues if layoutIfNeeded was called on a different superview then the root.
+            // Example, given this UI:
+            // UIViewController
+            // └── UIStackView
+            //    └── InAppMessageView
+            // ...If we call layoutIfNeeded on superview (UIStackView), the animation will not work as expected.
```
ami-aman, line 183: Do we expect any changes from the user to help animation work properly ?

ami-aman (approved):
Non blocking comments.

Sources/MessagingInApp/Views/UIKitInline.swift
levibostian: No, I do not expect any changes need to be made.

Sources/MessagingInApp/Views/GistInlineInAppMessageView.swift
levibostian: I like the named parameters that label height and width. 

What benefit does `CGSize` give over this approach? I'm open to it if it would be better.

Sources/MessagingInApp/Views/UIKitInline.swift
levibostian: Happy to add a comment. It would be helpful if you could provide a suggestion for the comment as I am not sure what questions you have about it. 

The `will_` naming convention is based on popular iOS APIs where that naming convention is used. The inline will will change the message that's shown when "show another message" button is clicked.

PR #785, opened 2024-08-06
refactor: rename swiftui and uikit inline View names

levibostian:

# Problem 

[Follow-up to previous PR thread](https://github.com/customerio/customerio-ios/pull/782#discussion_r1702219050), this PR renames the public API class name for the UIKit and the SwiftUI inline Views. 

Why rename? As discussed [in this slack thread](https://customerio.slack.com/archives/C070UMHGD09/p1722630308728029), the SwiftUI and the UIKit Views are in the same in-app module. Because of that, the names of the classes must be unique, so we cannot reuse the existing `InAppMessageView` name. 

# Solution 

This PR only performs a rename - no logic modified. 

```swift
// For UIKit customers: 
InlineMessageUIView(elementId: "")

// For SwiftUI customers: 
InlineMessage(elementId: "")
```

See the slack thread if you're curious behind the decision on the naming. 

# Testing 

This refactor is successful if...

* All sample apps should compile. 
* All tests should execute. 

# Out of scope

A docs PR will be opened mentioning the new naming. See [the docs ticket](https://linear.app/customerio/issue/DOCS-630/mobile-update-ios-inline-in-app-docs-with-updates-for-the-swiftui) to follow progress on that. 

# Follow-up PRs

None! SwiftUI feature is complete 🙂


ami-aman (approved):
Excited to close the last milestone of inline project. Great work !

PR #788, opened 2024-08-07
chore: improve swiftui app to test inline in-app like UIKit app

levibostian:
While QA testing the SwiftUI app, I noticed that some features were missing that the UIKit app have. I'll point these features out via comments in this PR. 

This PR doesn't introduce new ideas to the app but instead match the existing features in UIKit app.

### DIFF
--- a/Apps/CocoaPods-FCM/src/View/CustomEventView.swift
+++ b/Apps/CocoaPods-FCM/src/View/CustomEventView.swift
@@ -1,4 +1,5 @@
 import CioDataPipelines
+import CioMessagingInApp
 import SwiftUI
 
 struct CustomEventView: View {
@@ -19,6 +20,8 @@ struct CustomEventView: View {
             VStack {
                 Text("Send Custom Event").bold().font(.system(size: 20))
 
+                InlineMessage(elementId: "custom-screen")
+
                 VStack(spacing: 8) {
                     LabeledStringTextField(title: "Event Name", appiumId: "Event Name Input", value: $eventName)
                     LabeledStringTextField(title: "Property Name", appiumId: "Property Name Input", value: $propertyName)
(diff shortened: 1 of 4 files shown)


Apps/CocoaPods-FCM/src/View/CustomEventView.swift
levibostian, line 1: Custom event screen should have an inline View with elementId "custom-screen", like UIKit app has. 

Also added a small message on bottom of screen telling the user that an invisible inline in-app View is available on that screen.

Apps/CocoaPods-FCM/src/View/DashboardView.swift
```diff
@@ -27,30 +27,36 @@ struct DashboardView: View {
 
     @EnvironmentObject var userManager: UserManager
     var body: some View {
-        ZStack {
-            VStack {
-                SettingsButton {
-                    subscreenShown = .settings
+        ScrollView {
```
levibostian, line 30: First change is moving ScrollView to the top-most level. This fixes a bug I found where the settings button did nothing when you clicked on it on the Dashboard screen. After this change, the screen UI behaves as expected including the settings button is clickable.

Apps/CocoaPods-FCM/src/View/DashboardView.swift
```diff
@@ -27,30 +27,36 @@ struct DashboardView: View {
...
-            ScrollView {
                 VStack(spacing: 15) {
                     if let loggedInUserEmail = userManager.email {
                         Text(loggedInUserEmail)
                     }
                     Text("What would you like to test?")
                     Group {
-                        InlineMessage(elementId: "dashboard-announcement", onActionClick: { _, _, _ in
-                            print("Custom callback received")
+                        InlineMessage(elementId: "dashboard-announcement", onActionClick: { message, actionValue, actionName in
```
levibostian, line 52: to test `onActionClick` callback being set, track an event. I copy/pasted this code from UIKit app in the on action click delegate function.

Apps/CocoaPods-FCM/src/View/DashboardView.swift
```diff
@@ -127,6 +133,9 @@ struct DashboardView: View {
                                 }
                             }
                         }.setAppiumId("Show Push Prompt Button")
+
+                        InlineMessage(elementId: "dashboard-announcement-code")
```
levibostian, line 137: Add a 2nd inline View to the app, like UIKit app does.

Issue #790, opened 2024-08-08
CioAnalytics library have wrong sources format


alpha-moisol:
**SDK version:** 3.3.2

**Environment:** Development

**Describe the bug**
Cannot install the library using Tuist. The problem is with dependency `CioAnalytics`. It does not have `Sources/CioAnalytics` folder, so, there is an error `Source path must be one of ["Sources/CioAnalytics", "Source/CioAnalytics", "src/CioAnalytics", "srcs/CioAnalytics"]`.

**To Reproduce**
Install CustomerIO 3.2.2 dependency using Tuist. Then run `tuist generate`

**Expected behavior**
CioAnalytics dependency has `Sources/CioAnalytics` folder.

levibostian:
Hey there, Andrii. Sorry to hear that you have experienced this issue. Let me see if I can help. 

Could you provide your tuist config file so I can try and reproduce this error myself?

alpha-moisol:
@levibostian Sure, this is `Package.swift` file:
```swift
// swift-tools-version: 5.9
import PackageDescription

#if TUIST
    import ProjectDescription

    let packageSettings = PackageSettings(
        productTypes: []
    )
#endif

let package = Package(
    name: "TestProj",
    dependencies: [
        .package(url: "https://github.com/customerio/customerio-ios", exact: "3.3.2")
    ]
)
```
and `Project.swift` file:
```swift
let project = Project(
    name: "TestProj",
    targets: [
        .target(
            name: "TestProj",
            destinations: .iOS,
            product: .app,
            bundleId: "com.testproj",
            sources: ["Sources/**/*.swift"],
            resources: ["Resources/**"],
            dependencies: [
                .external(name: "DataPipelines"),
                .external(name: "MessagingPushFCM")
            ]
        )
    ]
)
```

After `tuist install` command if you rename `./Tuist/.build/checkouts/cdp-analytics-swift/Sources/Segment` to `./Tuist/.build/checkouts/cdp-analytics-swift/Sources/CioAnalytics` everything works fine.

levibostian:
Greatly appreciate the config files. 

From that, I have been able to reproduce the issue. 

We're working on a fix for this issue. We hope to get something out to you as soon as we can. We'll keep you updated here on the progress.

levibostian:
@alpha-moisol, I have a fix for you to try out.  

[...]

PR #791, opened 2024-08-08
fix: tuist users not able to install CIO iOS SDK SPM manifest issue

levibostian:
Resolves: https://github.com/customerio/customerio-ios/issues/790 

# Problem 

iOS developers who want to use tools such as Tuist to generate xcode projects are currently experiencing an error when resolving the CIO iOS SDK via SPM.

See [this PR for the full details of the problem and solution](https://github.com/customerio/cdp-analytics-swift/pull/7). 

# Solution 

In order for customers to be able to install the iOS SDK via tuist, the iOS SDK's SPM manifest file must point to the correct version of `CioAnalytics` that includes the bug fix. 

After this PR gets deployed, customers should be able to install the iOS SDK via tuist on the latest version. 

# Testing 

Installed this git branch in SPM and Cocoapods sample apps. Verified that the correct source code was downloaded and that the apps compile successfully. 

I am also going to run the sample apps quick that are built in this PR to verify that CDP API calls work as expected.

### DIFF
--- a/CustomerIODataPipelines.podspec
+++ b/CustomerIODataPipelines.podspec
@@ -28,5 +28,5 @@ Pod::Spec.new do |spec|
   spec.dependency "CustomerIOTrackingMigration", "= #{spec.version.to_s}"
 
   # Add Segment SDK as a dependency, as this module is designed to be compatible with it.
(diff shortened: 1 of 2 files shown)


Shahroz16 (approved):
LGTM 🎉

PR #795, opened 2024-08-21
fix: fetch remote messages on main queue

mrehan27:
part of [MBL-483](https://linear.app/customerio/issue/MBL-483/add-logs-that-would-help-troubleshoot-pocket-prep-issue)

### Changes

- Fetch in-app messages from remote on `DispatchQueue.main` explicitly

### DIFF
--- a/Sources/MessagingInApp/Gist/Managers/MessageQueueManager.swift
+++ b/Sources/MessagingInApp/Gist/Managers/MessageQueueManager.swift
@@ -70,8 +70,10 @@ class MessageQueueManager {
     }
 
     func fetchUserMessagesFromRemoteQueue() {
-        Logger.instance.debug(message: "Fetching user messages from remote queue")
-        fetchUserMessages()
+        DispatchQueue.main.async {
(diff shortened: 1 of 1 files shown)


Sources/MessagingInApp/Gist/Managers/MessageQueueManager.swift
```diff
@@ -70,8 +70,10 @@ class MessageQueueManager {
     }
 
     func fetchUserMessagesFromRemoteQueue() {
-        Logger.instance.debug(message: "Fetching user messages from remote queue")
-        fetchUserMessages()
+        DispatchQueue.main.async {
```
Shahroz16, line 73: can we please add comment on why this is needed?

Shahroz16 (approved):

Sources/MessagingInApp/Gist/Managers/MessageQueueManager.swift
mrehan27: Done

PR #796, opened 2024-08-30
chore: add in-app store action and state

mrehan27:

### Changes

- Added [`ReSwift`](https://github.com/ReSwift/ReSwift) dependency to manage in-app message state using store
- Added all actions required to manage in-app state (matches Android implementation)
- Added in-app message state to accurately represent the in-app state (matches Android implementation)


Sources/MessagingInApp/Gist/Managers/Models/Message.swift
```diff
@@ -17,16 +17,16 @@ public class GistProperties {
...
 public class Message {
-    public private(set) var instanceId = UUID().uuidString.lowercased()
+    public let instanceId = UUID().uuidString.lowercased()
     public let queueId: String?
     public let priority: Int?
     public let messageId: String
-    public private(set) var gistProperties: GistProperties
+    public let gistProperties: GistProperties
 
     var properties = [String: Any]()
```
Shahroz16, line 26:
```suggestion
        private(set) var properties: [String: Any]()
```

 I don't fully get the benefit of having `properties` when we already have `gistProperties` which gets set from constructor in any case, even if we decide to keep it, lets only allow read access but restrict write access to the class methods.

Sources/MessagingInApp/Gist/Managers/Models/Message.swift
Shahroz16, line 1: I think we have a good opportunity to simplify this class by removing redundant initializer and combining the functionality into a single initializer with default values.

```
public init(messageId: String, 
                queueId: String? = nil, 
                priority: Int? = nil, 
                gistProperties: GistProperties = GistProperties()) {
        self.messageId = messageId
        self.queueId = queueId
        self.priority = priority
        self.gistProperties = gistProperties
    }
    
    public convenience init(messageId: String, properties: [String: Any]?) {
        let gistProperties = Message.parseGistProperties(from: properties?["gist"] as? [String: Any])
        self.init(messageId: messageId, 
                  queueId: properties?["queueId"] as? String, 
                  priority: properties?["priority"] as? Int, 
                  gistProperties: gistProperties)
        if let props = properties {
            self.properties = props
        }
    }
    
      private static func parseGistProperties(from gist: [String: Any]?) -> GistProperties {
        guard let gist = gist else { return GistProperties() }
        
        let position = MessagePosition(rawValue: gist["position"] as? String ?? "") ?? .center
        let routeRule = gist["routeRuleApple"] as? String
        let elementId = gist["elementId"] as? String
        let campaignId = gist["campaignId"] as? String
        let persistent = gist["persistent"] as? Bool ?? false
        
        return GistProperties(routeRule: routeRule,
                              elementId: elementId,
                              campaignId: campaignId,
                              position: position,
                              persistent: persistent)
    }
```

Sources/MessagingInApp/Gist/Managers/Models/Message.swift
```diff
@@ -144,4 +145,11 @@ extension Message: Equatable {
         // If the queueId is nil, we fallback to the instanceId which is a unique ID generated client-side
         return lhs.instanceId == rhs.instanceId
     }
+
+    public func hash(into hasher: inout Hasher) {
+        hasher.combine(instanceId)
+        hasher.combine(queueId)
+        hasher.combine(priority)
+        hasher.combine(messageId)
+    }
```
Shahroz16, line 154: interesting what was the use-case behind adding this? specially because I see all parameters being added into the hash, while in checking equality we are only checking two? so either we check all 4 in equality or just add 2 in here?

Sources/MessagingInApp/State/InAppMessageAction.swift
```diff
@@ -0,0 +1,76 @@
...
+        case (.engineAction(let lhsAction), .engineAction(let rhsAction)):
+            return lhsAction == rhsAction
+
+        case (.resetState, .resetState):
+            return true
+
+        default:
+            return false
+        }
+    }
```
Shahroz16, line 74:
```suggestion
    // MARK: - Equatable Implementation
extension InAppMessageAction {
    static func == (lhs: InAppMessageAction, rhs: InAppMessageAction) -> Bool {
        switch (lhs, rhs) {
        case (.initialize(let l), .initialize(let r)): return l == r
        case (.setPollingInterval(let l), .setPollingInterval(let r)): return l == r
        case (.setUserIdentifier(let l), .setUserIdentifier(let r)): return l == r
        case (.setPageRoute(let l), .setPageRoute(let r)): return l == r
        case (.processMessageQueue(let l), .processMessageQueue(let r)): return l == r
        case (.clearMessageQueue, .clearMessageQueue): return true
        case (.loadMessage(let l), .loadMessage(let r)): return l == r
        case (.embedMessage(let l), .embedMessage(let r)): return l == r
        case (.displayMessage(let l), .displayMessage(let r)): return l == r
        case (.dismissMessage(let l), .dismissMessage(let r)): return l == r
        case (.reportError(let l), .reportError(let r)): return l == r
        case (.engineAction(let l), .engineAction(let r)): return l == r
        case (.resetState, .resetState): return true
        default: return false
        }
    }
}

// MARK: - Tuple Equatable Extensions
extension InAppMessageAction {
    private static func == (
        lhs: (siteId: String, dataCenter: String, environment: GistEnvironment),
        rhs: (siteId: String, dataCenter: String, environment: GistEnvironment)
    ) -> Bool {
        lhs.siteId == rhs.siteId && lhs.dataCenter == rhs.dataCenter && lhs.environment == rhs.environment
    }
    
    private static func == (
        lhs: (message: Message, position: MessagePosition?),
        rhs: (message: Message, position: MessagePosition?)
    ) -> Bool {
        lhs.message == rhs.message && lhs.position == rhs.position
    }
    
    private static func == (
        lhs: (message: Message, elementId: String),
        rhs: (message: Message, elementId: String)
    ) -> Bool {
        lhs.message == rhs.message && lhs.elementId == rhs.elementId
    }
    
    private static func == (
        lhs: (message: Message, shouldLog: Bool, viaCloseAction: Bool),
        rhs: (message: Message, shouldLog: Bool, viaCloseAction: Bool)
    ) -> Bool {
        lhs.message == rhs.message && lhs.shouldLog == rhs.shouldLog && lhs.viaCloseAction == rhs.viaCloseAction
    }
}
```

think we can simplify this and make it less error-prone by moving it to extension method and doing tuple comparisons wherever possible. 

e.g
in the case of  (.initialize(let l), .initialize(let r)): return l == r
Here, `l` and `r` are tuples containing the `siteId, dataCenter, and environment`. Swift automatically creates these tuples from the associated values of the enum case.
The comparison `l == r` calls the `custom == operator` we defined in another extension

So, when initialize is called, here's what happens:
- The == function in the main InAppMessageAction extension is called.
- It matches the .initialize case.
- It extracts the associated values into tuples l and r.
- It then calls l == r, which invokes our custom tuple comparison function.
- This function then compares each element of the tuple individually: siteId, dataCenter, and environment.

Sources/MessagingInApp/State/InAppMessageState.swift
```diff
@@ -0,0 +1,176 @@
...
+            pollInterval: pollInterval ?? self.pollInterval,
+            userId: userId ?? self.userId,
+            currentRoute: currentRoute ?? self.currentRoute,
+            currentMessageState: currentMessageState ?? self.currentMessageState,
+            messagesInQueue: messagesInQueue ?? self.messagesInQueue,
+            shownMessageQueueIds: shownMessageQueueIds ?? self.shownMessageQueueIds
+        )
+    }
+
+    static func == (lhs: InAppMessageState, rhs: InAppMessageState) -> Bool {
```
Shahroz16, line 63: "optional", can be an extenstion method, keeps the class lighter.

Sources/MessagingInApp/Gist/Managers/Models/Message.swift
mrehan27: There are a few instances where `properties` is used directly at the moment. While I agree there's a chance we could replace it with `gistProperties`, I would prefer not to make that change now. Instead, I suggest waiting until rest of the changes are complete so we have a clearer picture of the use case and can replace it more confidently and safely.

Sources/MessagingInApp/Gist/Managers/Models/Message.swift
mrehan27: Yeah, seems so. Let me try it.

Sources/MessagingInApp/Gist/Managers/Models/Message.swift
mrehan27: Because of `messagesInQueue: Set<Message>` in `InAppMessageState` as set only allows `Hashable` classes to be added.

Sources/MessagingInApp/State/InAppMessageAction.swift
mrehan27: Appreciate the write-up. However, honestly this seems too confusing to me. As it's hard to understand at first glance. Also, because the pairs/tuples used in the extension function are tightly bound and not very flexible, I think they are hard to read, maintain, and debug. So, I'm okay with moving forward with current approach, which might be error-prone but is clean and can be covered with tests to minimize the chances of breaking in future.

Sources/MessagingInApp/State/InAppMessageState.swift
mrehan27: I don't disagree and okay either way. But since it's all related and I think keeping it in the class makes it easier to match, let's stick to that approach.

Sources/MessagingInApp/Gist/Managers/Models/Message.swift
mrehan27: Updated

Sources/MessagingInApp/State/InAppMessageAction.swift
Shahroz16: > because the pairs/tuples used in the extension function are tightly bound and not very flexible, I think they are hard to read, maintain, and debug

I don't agree with this statement, as they are general equality operators that takes in tuples and if our action contains them, it applies to them too. 

I am okay with going forward this, subject to what tests gets added to ensure likelihood of errors when adding new cases or modifying existing ones.

Shahroz16 (approved):
Looks good

Sources/MessagingInApp/State/InAppMessageAction.swift
mrehan27: > they are general equality operators that take in tuples

No they are not, unfortunately, in my opinion. They are bound by number of parameters, their types, and the sequence. Consider the following example:

```swift
private static func == (
    lhs: (siteId: String, dataCenter: String, environment: GistEnvironment),
    rhs: (siteId: String, dataCenter: String, environment: GistEnvironment)
)
```

This probably will not work for any of the following:

```swift
private static func == (
    lhs: (environment: GistEnvironment, siteId: String, dataCenter: String),
    rhs: (environment: GistEnvironment, siteId: String, dataCenter: String)
)
``` 

OR

```swift
private static func == (
    lhs: (siteId: String, dataCenter: Region, environment: GistEnvironment),
    rhs: (siteId: String, dataCenter: Region, environment: GistEnvironment)
)
``` 

That's why I wouldn't call these general equality operators. Also, copying the code gives several compilation errors and tuple matching warnings (see screenshot below). So if I copied it correctly, I don't think it is working as expected 😅

Screenshot 2024-09-03 at 5 29 41 PM

However, since we both agree on moving forward with current implementation, I think we should stick with it and continue.

PR #797, opened 2024-09-02
chore: in-app store reducer and middleware

mrehan27:

### Changes

- Added reducer to update in-app message state based on dispatched actions (matches Android implementation)
- Added middlewares so that desired actions can be taken based on dispatched actions and the in-app message state (matches Android implementation)
- Added typealiases to avoid directly coupling classes with `ReSwift`. All included classes refer to internal types to make it easier to replace `ReSwift` if needed.
- Added helper `changes` function in `InAppMessageState` to filter out only updated values


Sources/MessagingInApp/State/InAppMessageMiddleware.swift
```diff
@@ -0,0 +1,247 @@
...
+
+        // Update current route first
+        next(action)
+
+        // Check if there is a message currently displayed and if it has a page rule
+        let currentRoute = state.currentRoute
+        let currentMessage: Message? = state.currentMessageState.message
+        let doesCurrentMessageRouteMatch: Bool =
+            currentMessage?.doesHavePageRule() == true &&
+            currentRoute.map { currentMessage?.doesPageRuleMatch(route: $0) ?? false } ?? true
```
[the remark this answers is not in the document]
Shahroz16: also, can `doesPageRuleMatch` throw?

Sources/MessagingInApp/State/InAppMessageMiddleware.swift
```diff
@@ -0,0 +1,247 @@
...
+}
+
+func messageMetricsMiddleware(logger: CioInternalCommon.Logger) -> InAppMessageMiddleware {
+    middleware { _, getState, next, action in
+        let state = getState()
+        switch action {
+        case .displayMessage(let message):
+            let properties = message.gistProperties
+            // Log message view only if message is not persistent
+            if properties.persistent != true {
```
Shahroz16, line 97:
```suggestion
            if message.gistProperties != true {
```

Shahroz16:
```
let properties = message.gistProperties
```

can be removed?

Sources/MessagingInApp/State/InAppMessageMiddleware.swift
```diff
@@ -0,0 +1,247 @@
...
+                logger.debug("[InApp] Persistent message shown, not logging view for message: \(message.describeForLogs)")
+            }
+
+        case .dismissMessage(let message, let shouldLog, let viaCloseAction):
+            guard shouldLog else { return }
+
+            let properties = message.gistProperties
+            // Log message close only if message was dismissed via close action
+            if viaCloseAction {
+                if properties.persistent == true {
```
Shahroz16, line 110: same as above, we can use `message.gistProperties` directly?

Sources/MessagingInApp/State/InAppMessageMiddleware.swift
```diff
@@ -0,0 +1,247 @@
...
+
+        let state = getState()
+        // Filter out messages with valid queueId that have not been shown yet and sort by priority
+        let notShownMessages = messages
+            .filter { message in
+                if let queueId = message.queueId {
+                    return !state.shownMessageQueueIds.contains(queueId)
+                }
+                return false
+            }
```
Shahroz16, line 142:
```suggestion
           .filter { message in
                guard let queueId = message.queueId else { return false }
                return !state.shownMessageQueueIds.contains(queueId)
            }
```

Sources/MessagingInApp/State/InAppMessageMiddleware.swift
```diff
@@ -0,0 +1,247 @@
...
+                if !result.contains(where: { $0.queueId == message.queueId }) {
+                    result.append(message)
+                }
+            }
+            .sorted {
+                guard let priority1 = $0.priority, let priority2 = $1.priority else {
+                    return $0.priority != nil
+                }
+                return priority1 < priority2
+            }
```
Shahroz16, line 153:
```suggestion
.sorted { ($0.priority ?? Int.max) < ($1.priority ?? Int.max) }
```
can be simplified like this?

Sources/MessagingInApp/State/InAppMessageMiddleware.swift
```diff
@@ -0,0 +1,247 @@
...
+
+        case .initialize,
+             .setPollingInterval,
+             .setUserIdentifier,
+             .setPageRoute,
+             .processMessageQueue,
+             .clearMessageQueue,
+             .loadMessage,
+             .reportError,
+             .resetState:
```
Shahroz16, line 231: maybe `default` ?

Shahroz16 (changes requested):
checking current message for route is the major one

Sources/MessagingInApp/State/Typealiases.swift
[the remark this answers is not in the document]
mrehan27: Good call, updated

Sources/MessagingInApp/State/Typealiases.swift
```diff
@@ -0,0 +1,47 @@
...
+        guard let inAppAction = action as? InAppMessageAction else {
+            DIGraphShared.shared.logger.debug("[InApp] Invalid action type: \(action), skipping middleware")
+            return next(action)
+        }
+
+        let getStateOrDefault = { getState() ?? InAppMessageState() }
+        completion(dispatch, getStateOrDefault, next, inAppAction)
+    }
+    }
+    }
```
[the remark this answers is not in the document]
mrehan27: Agreed, but `swiftlint lint --fix` always reverts this back, and any workaround results in `swiftlint lint --strict` failing 😅

Sources/MessagingInApp/State/InAppMessageMiddleware.swift
```diff
@@ -0,0 +1,247 @@
+import CioInternalCommon
+import Foundation
+
+func userAuthenticationMiddleware() -> InAppMessageMiddleware {
+    middleware { _, getState, next, action in
+        let state = getState()
+
+        // Block actions that require valid userId unless userId is set.
+        switch action {
```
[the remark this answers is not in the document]
mrehan27: Updated

Sources/MessagingInApp/State/InAppMessageMiddleware.swift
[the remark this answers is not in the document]
mrehan27: No. `doesPageRuleMatch` has `try` in it. Updated the code with `flatMap` instead.

Sources/MessagingInApp/State/InAppMessageMiddleware.swift
```diff
@@ -0,0 +1,247 @@
...
+            logger.debug("[InApp] Dismissing message: \(message.describeForLogs) because route does not match current route: \(String(describing: currentRoute))")
+            dispatch(.dismissMessage(message: message, shouldLog: false))
+        }
+
+        // Process message queue to check if there is a message that matches the new route
+        dispatch(.processMessageQueue(messages: Array(state.messagesInQueue)))
+    }
+}
+
+func messageStateMiddleware(logger: CioInternalCommon.Logger) -> InAppMessageMiddleware {
```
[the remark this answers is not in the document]
mrehan27: I renamed but it sill may not be very clear. Happy to hear more suggestion on naming ideas here 😅

Sources/MessagingInApp/State/InAppMessageMiddleware.swift
mrehan27: Updated

Sources/MessagingInApp/State/InAppMessageMiddleware.swift
mrehan27: Updated

Sources/MessagingInApp/State/InAppMessageMiddleware.swift
mrehan27: Updated

Sources/MessagingInApp/State/InAppMessageMiddleware.swift
mrehan27: Done

Sources/MessagingInApp/State/InAppMessageMiddleware.swift
mrehan27: Yup, updated

Sources/MessagingInApp/State/InAppMessageMiddleware.swift
```diff
@@ -0,0 +1,232 @@
...
+
+        // Update current route first
+        next(action)
+
+        // Check if there is a message currently displayed and if it has a page rule
+        let currentRoute = state.currentRoute
+        let currentMessage: Message? = state.currentMessageState.activeModalMessage
+        let doesCurrentMessageRouteMatch: Bool = currentMessage.flatMap { message in
+            message.doesHavePageRule() && (currentRoute.flatMap(message.doesPageRuleMatch) ?? true)
+        } ?? true
```
Shahroz16, line 43:
```suggestion
        
let doesCurrentMessageRouteMatch: Bool = {
    guard let message = currentMessage else {
        return true  // No current message, so we consider it a match
    }
    
    guard message.doesHavePageRule() else {
        return true  // Message has no page rule, so we consider it a match
    }
    
    guard let route = currentRoute else {
        return false  // Message has a page rule, but there's no current route
    }
    
    return message.doesPageRuleMatch(route: route)
}()
```

One flatmap might have made it simpler but too just makes its hard to grasp, I tried to simplify and I feel like it makes a lot more sense, let me know .

Shahroz16 (approved):
Looks good one suggestion added that i would really want to consider.

Sources/MessagingInApp/State/InAppMessageMiddleware.swift
mrehan27: I think this can be further simplified. I revisited the code, along with Android implementation, and made it much simpler. I've also added comments to explain things further. Let me know if this looks better now.

Shahroz16: Looks great now

PR #799, opened 2024-09-02
refactor: remove gist logger

mrehan27:

### Changes

- Removed separate `Logger` class from `MessagingInApp` module
- Replaced its usages with the common `Logger` class


Sources/MessagingInApp/Gist/Gist.swift
```diff
@@ -12,6 +14,10 @@ public class Gist: GistDelegate {
 
     public static let shared = Gist()
 
+    init() {
+        self.logger = DIGraphShared.shared.logger
```
Shahroz16, line 18: why not just use ?

```
private let logger: Logger = DIGraphShared.shared.logger
```

Sources/MessagingInApp/Gist/Managers/MessageManager.swift
```diff
@@ -25,6 +26,9 @@ class MessageManager: EngineWebDelegate {
         self.currentMessage = message
         self.currentRoute = message.messageId
 
+        let diGraph = DIGraphShared.shared
+        self.logger = diGraph.logger
```
Shahroz16, line 30: again, is there any benifit of using `diGraph.logger` here rather than just initializing the logger with `DIGraphShared.shared`

Sources/MessagingInApp/Gist/Managers/MessageQueueManager.swift
```diff
@@ -1,12 +1,18 @@
...
 
 class MessageQueueManager {
+    private let logger: Logger
     var interval: Double = 600
     private var queueTimer: Timer?
     // The local message store is used to keep messages that can't be displayed because the route rule doesnt match.
     var localMessageStore: [String: Message] = [:]
 
+    init() {
+        self.logger = DIGraphShared.shared.logger
```
Shahroz16, line 13: same

Shahroz16 (approved):
lets please look into logger initialization, seems unnecessary.

Sources/MessagingInApp/Gist/Gist.swift
mrehan27: `Gist` class is moved to `DIGraphShared ` in #802, so `DIGraphShared` usage is replaced with constructor values. I kept it in the constructor to make the code changes a bit simpler in next PRs.

Sources/MessagingInApp/Gist/Managers/MessageManager.swift
mrehan27: Valid point. However, this is only done in classes that are moved to `DIGraphShared` in PR #802. Using `DIGraphShared` in constructor slightly simplified the changes in that PR. Does this make sense?

Sources/MessagingInApp/Gist/Managers/MessageQueueManager.swift
mrehan27: Same as above

PR #800, opened 2024-09-02
chore: use in-app state for consistency

mrehan27:

### Changes

- Utilized `InAppMessageState` throughout all relevant classes
- Used environment from state class
- Updated `LogManager` and moved it to `DiGraph`
- Added `runMainAfterDelay` to `ThreadUtil` to allow reuse and avoid direct usage of `DispatchQueue`
- Removed `RouteManager` and `UserManager` as they are no longer needed


Sources/MessagingInApp/Gist/Gist.swift
mrehan27, line 1: Most of the code in this class is updated in next PRs in the chain, so the code is commented to make tracking changes easier.

Sources/MessagingInApp/Gist/Gist.swift
```diff
@@ -38,41 +37,42 @@ public class Gist: GistDelegate {
...
-        UserManager().setUserToken(userToken: userToken)
+        // UserManager().setUserToken(userToken: userToken)
     }
 
     public func clearUserToken() {
         cancelModalMessage(ifDoesNotMatchRoute: "") // provide a new route to trigger a modal cancel.
         messageQueueManager.clearLocalStore()
-        UserManager().clearUserToken()
+        // UserManager().clearUserToken()
         messageQueueManager.clearUserMessagesFromLocalStore()
```
Shahroz16, line 53: does `messageQueueManager` also get removed in later PRs, cause I don't see it commented here?

Sources/MessagingInApp/Gist/Gist.swift
```diff
@@ -140,9 +140,9 @@ public class Gist: GistDelegate {
         if let queueId = message.queueId {
             shownMessageQueueIds.insert(queueId)
         }
-        let userToken = UserManager().getUserToken()
-        LogManager(siteId: siteId, dataCenter: dataCenter)
-            .logView(message: message, userToken: userToken) { response in
+        let state = InAppMessageState()
```
Shahroz16, line 143: i dont understand this one, why aren't we using the state from manager?

Sources/MessagingInApp/Gist/Managers/MessageManager.swift
```diff
@@ -33,12 +33,12 @@ class MessageManager: EngineWebDelegate {
             siteId: Gist.shared.siteId,
             dataCenter: Gist.shared.dataCenter,
```
Shahroz16, line 34: shouldn't these 2 also fetched from state?

Shahroz16 (approved):
It was a bit hard to review this not know whats to follow next 😅 major thing I see if InAppMessageState instance being created in classes, which I am guessing will be moved in later PR, if not that's a blocker.

Sources/MessagingInApp/Gist/Gist.swift
mrehan27: `messageQueueManager` doesn't get removed, but the call does. You can take a look at PR #802 to see the final implementation.

Sources/MessagingInApp/Gist/Gist.swift
mrehan27: This is just to keep things compiling. As I mentioned above, this class is almost completely rewritten in PR #802

Sources/MessagingInApp/Gist/Managers/MessageManager.swift
mrehan27: Yup, correct. Updated in #802

PR #801, opened 2024-09-02
chore: move gist delegate to separate class

mrehan27:

### Changes

- Moved `GistDelegate` to a separate class so it can be accessed through `DiGraph` even before `MessagingInAppImplementation` instance is created
- Removed `InAppProvider` and utilized the `Gist` class directly


Sources/MessagingInApp/Gist/Gist.swift
```diff
@@ -40,6 +40,10 @@ public class Gist: GistDelegate {
         // RouteManager.clearCurrentRoute()
     }
 
+    public func resetState() {}
+
+    public func setEventListener(_ eventListener: (any InAppEventListener)?) {}
```
mrehan27, line 45: Both of these are implemented in the next PR

Sources/MessagingInApp/MessagingInAppImplementation.swift
```diff
@@ -5,114 +5,62 @@ class MessagingInAppImplementation: MessagingInAppInstance {
...
     private let threadUtil: ThreadUtil
-    let eventBusHandler: EventBusHandler
+    private let eventBusHandler: EventBusHandler
 
     init(diGraph: DIGraphShared, moduleConfig: MessagingInAppConfigOptions) {
         self.moduleConfig = moduleConfig
         self.logger = diGraph.logger
-        self.inAppProvider = diGraph.inAppProvider
+        self.inAppMessageManager = diGraph.inAppMessageManager
+        self.gist = Gist.shared
```
mrehan27, line 17: In next PR, it will be accessed through DiGraph instead

Sources/MessagingInApp/MessagingInAppImplementation.swift
```diff
@@ -5,114 +5,62 @@ class MessagingInAppImplementation: MessagingInAppInstance {
...
+        gist.setEventListener(eventListener)
     }
 
     // Dismiss in-app message
     func dismissMessage() {
-        inAppProvider.dismissMessage()
-    }
-}
-
-extension MessagingInAppImplementation: GistDelegate {
```
mrehan27, line 68: This is moved to `GistDelegateImpl` with new implementation

Sources/MessagingInApp/Gist/GistDelegate.swift
```diff
@@ -6,4 +8,72 @@ public protocol GistDelegate: AnyObject {
...
+        }
+
+        eventListener?.messageActionTaken(
+            message: InAppMessage(gistMessage: message),
+            actionValue: action,
+            actionName: name
+        )
+    }
+
+    private func getDeliveryId(from message: Message) -> String? {
```
Shahroz16, line 72: probably makes more sense to have an extension method on Message?

Sources/MessagingInApp/MessagingInAppImplementation.swift
```diff
@@ -5,114 +5,62 @@ class MessagingInAppImplementation: MessagingInAppInstance {
...
 
-            self.inAppProvider.setProfileIdentifier(event.identifier)
+            self.gist.setUserToken(event.identifier)
         }
 
         eventBusHandler.addObserver(ScreenViewedEvent.self) { event in
             self.logger.debug("setting route for in-app to \(event.name)")
 
             // Gist expects webview to be launched in main thread and changing route will trigger locally stored in-app messages for that route.
-            self.threadUtil.runMain {
```
Shahroz16, line 43: think this is because this is going to be done in middleware? do we still need the comment above?

Sources/MessagingInApp/MessagingInAppImplementation.swift
```diff
@@ -5,114 +5,62 @@ class MessagingInAppImplementation: MessagingInAppInstance {
...
-                self.inAppProvider.setRoute(event.name)
-            }
+            self.gist.setCurrentRoute(event.name)
         }
 
         eventBusHandler.addObserver(ResetEvent.self) { _ in
             self.logger.debug("removing profile for in-app")
 
             // Gist may dismiss a modal if it's displayed. Run on main thread to call UI code.
-            self.threadUtil.runMain {
```
Shahroz16, line 52: this is going to be done via middleware too?

Shahroz16: also the comment.

Shahroz16 (approved):
Looks good, minor suggestions/questions

Sources/MessagingInApp/Gist/GistDelegate.swift
mrehan27: Updated

Sources/MessagingInApp/MessagingInAppImplementation.swift
mrehan27: Nope. Removed comment

Sources/MessagingInApp/MessagingInAppImplementation.swift
mrehan27: Yup, middleware will handle this. Removed comment

PR #805, opened 2024-09-05
chore: dispatch loading failed event

mrehan27:

### Changes

- Removed `EngineRoute` class as it seemed to be an unnecessary wrapper
- Dispatched `messageLoadingFailed` action when an in-app message times out
- Replaced `engineAction.error` with `engineAction.messageLoadingFailed` to simplify and align with Android
- Cleaned up and simplified the `Message` class
- Updated `InAppMessageReducer` to handle `messageLoadingFailed` action


Sources/MessagingInApp/Gist/EngineWeb/EngineWeb.swift
```diff
@@ -21,6 +21,8 @@ protocol EngineWebInstance: AutoMockable {
 
 public class EngineWeb: NSObject, EngineWebInstance {
     private let logger: Logger = DIGraphShared.shared.logger
+    private let inAppMessageManager: InAppMessageManager = DIGraphShared.shared.inAppMessageManager
+    private let _currentMessage: Message?
```
Shahroz16, line 25: Can this be part of the constructor? The usage of EngineWeb only seems to be from MessageManager and it get initialized with Message.

Sources/MessagingInApp/Gist/Managers/MessageManager.swift
```diff
@@ -40,7 +40,7 @@ class MessageManager: EngineWebDelegate {
             instanceId: message.instanceId,
             endpoint: state.environment.networkSettings.engineAPI,
             messageId: message.messageId,
-            properties: message.toEngineRoute().properties
+            properties: message.properties.mapValues { AnyEncodable($0) }
```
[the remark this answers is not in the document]
mrehan27: I don't think `AnyEncodable` directly throws anything. We've been doing the same when building `EngineRoute` class previously, so nothing really has changed here.

Also, I tried calling it like `try? AnyEncodable($0)` and could see the following warning:

```
No calls to throwing functions occur within 'try' expression
```

Sources/MessagingInApp/Gist/EngineWeb/EngineWeb.swift
mrehan27: Updated and made it non-null

Shahroz16 (approved):
Looks good.

PR #807, opened 2024-09-10
chore: improve in-app logs and documentation

mrehan27:
part of [MBL-422](https://linear.app/customerio/issue/MBL-422/improve-debug-logs-for-in-app-ios)

### Changes

- Improved logging within in-app module
- Added class docs and code comments for better clarity
- Cleaned up code for subscribing to `InAppMessageStore`
- Fixed `resetState` function in `Gist` class
- Added `isMessageLoaded` to prevent any possible duplicate `routeLoaded` calls, aligning with `main` and Android implementations


Shahroz16 (commented):
Can't we create an extension function in logger which automatically appends `[Inapp]` ahead of every message passed to it? it would save us from adding it in every single line and then running a chance of missing/typo.

Shahroz16 (approved):
Looks good 👍

PR #808, opened 2024-09-11
chore: ignore inline in-app messages

mrehan27:

### Changes

- Ignored embedded (inline) in-app messages when filtering messages to be displayed, as inline support is not available in this branch
- Updated `Message` and `GistProperties` classes for improved logging and debugging

### DIFF
--- a/Sources/MessagingInApp/State/InAppMessageMiddleware.swift
+++ b/Sources/MessagingInApp/State/InAppMessageMiddleware.swift
@@ -65,7 +65,7 @@ func modalMessageDisplayStateMiddleware(logger: Logger, threadUtil: ThreadUtil) 
             return next(.reportError(message: "Blocked loading message: \(message.describeForLogs) because another message is currently displayed or cancelled: \(currentMessage)"))
         }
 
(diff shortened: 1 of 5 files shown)


Sources/MessagingInApp/State/InAppMessageMiddleware.swift
```diff
@@ -65,7 +65,7 @@ func modalMessageDisplayStateMiddleware(logger: Logger, threadUtil: ThreadUtil)
             return next(.reportError(message: "Blocked loading message: \(message.describeForLogs) because another message is currently displayed or cancelled: \(currentMessage)"))
         }
 
-        logger.logWithModuleTag("Showing message: \(message.describeForLogs) with position: \(String(describing: position))", level: .debug)
+        logger.logWithModuleTag("Showing message: \(message) with position: \(String(describing: position))", level: .debug)
```
Shahroz16, line 68: message would already show position now?

Shahroz16 (approved):
Looks good

Sources/MessagingInApp/State/InAppMessageMiddleware.swift
mrehan27: Removed position from action as it is no longer needed

PR #809, opened 2024-09-11
test: added tests for state

Shahroz16:
helps: [MBL-501](https://linear.app/customerio/issue/MBL-501/add-tests)

changes:
- Updated to helper methods for message
- Changes in test utilities WRT to new classes
- Added tests for state changes

It doesn't contain module tests, which would be failing.

### DIFF
--- a/Sources/MessagingInApp/State/InAppMessageManager.swift
+++ b/Sources/MessagingInApp/State/InAppMessageManager.swift
@@ -48,7 +48,7 @@ public class InAppMessageManager {
     @discardableResult
     func dispatch(action: InAppMessageAction, completion: (() -> Void)? = nil) -> Task<Void, Never> {
         Task {
-            await store.dispatch(action)
+            await dispatchAsync(action: action)
             completion?()
         }
     }
--- a/Sources/MessagingInApp/State/InAppMessageManager.swift
+++ b/Sources/MessagingInApp/State/InAppMessageManager.swift
@@ -53,6 +53,10 @@ public class InAppMessageManager {
         }
     }
 
+    func dispatchAsync(action: InAppMessageAction) async {
+        await store.dispatch(action)
+    }
+
     @discardableResult
     func unsubscribe(subscriber: InAppMessageStoreSubscriber) -> Task<Void, Never> {
         Task { await store.unsubscribe(subscriber) }
--- a/Sources/MessagingInApp/State/InAppMessageManager.swift
+++ b/Sources/MessagingInApp/State/InAppMessageManager.swift
@@ -48,13 +48,9 @@ public class InAppMessageManager {
     @discardableResult
     func dispatch(action: InAppMessageAction, completion: (() -> Void)? = nil) -> Task<Void, Never> {
         Task {
-            await dispatchAsync(action: action)
+            await store.dispatch(action)
             completion?()
         }
-    }
-
-    func dispatchAsync(action: InAppMessageAction) async {
-        await store.dispatch(action)
     }
 
     @discardableResult
--- a/Tests/MessagingInApp/EngineWeb/EngineWebProviderStub.swift
+++ b/Tests/MessagingInApp/EngineWeb/EngineWebProviderStub.swift
@@ -2,7 +2,7 @@
 import Foundation
 
 class EngineWebProviderStub: EngineWebProvider {
-    func getEngineWebInstance(configuration: CioMessagingInApp.EngineWebConfiguration, state: CioMessagingInApp.InAppMessageState, message: CioMessagingInApp.Message) -> any CioMessagingInApp.EngineWebInstance {
+    func getEngineWebInstance(configuration: EngineWebConfiguration, state: InAppMessageState, message: Message) -> any EngineWebInstance {
         engineWebMock
     }
 
--- a/Tests/MessagingInApp/EngineWeb/EngineWebProviderStub.swift
+++ b/Tests/MessagingInApp/EngineWeb/EngineWebProviderStub.swift
@@ -2,13 +2,13 @@
 import Foundation
 
 class EngineWebProviderStub: EngineWebProvider {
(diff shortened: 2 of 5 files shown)


Sources/MessagingInApp/State/InAppMessageManager.swift
```diff
@@ -53,6 +53,10 @@ public class InAppMessageManager {
         }
     }
 
+    func dispatchAsync(action: InAppMessageAction) async {
```
Shahroz16, line 56: added this cause it was very helpful in streamlining tests, otherwise we would need to use expectations and timeouts.

mrehan27 (commented):
Mostly looks good, just some small comments

Sources/MessagingInApp/State/InAppMessageManager.swift
Shahroz16: Both approached creates and awaits a `Task` object, adding unnecessary overhead. Since we are already calling it from `async` method we don't need to create a Task and then await it. 

I have kept the same flow but refined it a bit more.

Tests/MessagingInApp/EngineWeb/EngineWebProviderStub.swift
```diff
@@ -2,13 +2,13 @@
 import Foundation
 
 class EngineWebProviderStub: EngineWebProvider {
+    func getEngineWebInstance(configuration: CioMessagingInApp.EngineWebConfiguration, state: CioMessagingInApp.InAppMessageState, message: CioMessagingInApp.Message) -> any CioMessagingInApp.EngineWebInstance {
```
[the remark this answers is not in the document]
Shahroz16: done

Tests/MessagingInApp/State/InAppMessageStateTests.swift
```diff
@@ -0,0 +1,384 @@
...
+        XCTAssertEqual(state.siteId, "testSite")
+        XCTAssertEqual(state.dataCenter, "testDC")
+        XCTAssertEqual(state.environment, .production)
+    }
+
+    func test_setUserIdentifier_expectUserIdUpdate() async {
+        let expectation = XCTestExpectation(description: "State updated")
+
+        inAppMessageManager.dispatch(action: .setUserIdentifier(user: "testUser")) {
+            expectation.fulfill()
```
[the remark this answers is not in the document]
Shahroz16: Left by mistake

Sources/MessagingInApp/State/InAppMessageManager.swift
mrehan27: Since this is only used by tests, I'm not sure what unnecessary overhead is being added here. My intention was to avoid including `dispatchAsync` in production code since we don't need it there. I'm okay with even making `store` internal and adding this extension just for tests.

Also, can you please revert the recent change? Calling `dispatchAsync` from `dispatch` is completely unnecessary in my opinion. As all our production code uses `dispatch`, and now we've added an extra function call to `dispatchAsync` just for tests, which doesn't seem needed to me.

Tests/MessagingInApp/State/InAppMessageStateTests.swift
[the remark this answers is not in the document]
mrehan27: This is still pending? 🤔

Sources/MessagingInApp/State/InAppMessageManager.swift
Shahroz16: Okay there are a couple of points here, let me answer one by one

> I'm not sure what unnecessary overhead is being added here

```
extension InAppMessageManager { 
func dispatchAsync(action: InAppMessageAction) async { 
await dispatch(action: action).value } 
}
``` 

This method, can directly get value from `store.dispatch`, it doesn't need a `task`, but because we are using `dispatch(action: action).value ` we are unnecessarily creating Task and getting the value. 

> My intention was to avoid including dispatchAsync in production code

I don't understand, why? what are your reservations about it? The reason we created `tasks with callback` was so we don't have to add `async/await` everywhere. Internally we are still using actors and creating Tasks and awaiting. 

So what we have done is create two methods:
- Async method that can be called from any async method
- Method which creates a `Task` and calls the Async method internally.

It's the same thing we have done in `data pipeline` classes, where we would expose different options but keep one primary option.

If you exclude the "just for tests" part, why is this not a good addition and what overhead would it be creating? We are literally doing the same thing but now we just have exposes a method to do it. 

- It provides consistency as all dispatch calls go through a single async implementation, ensuring consistent behavior across your app.
- It reduces code duplication, as the non-async version now calls the async version.
- If you need to modify the dispatch logic in the future, you only need to update one place (the async version).

mrehan27: If anything is needed in the future, we can address it then and figure out how to improve. We don't need to make assumptions and modify our code unnecessarily.

I don’t think this requires a lengthy discussion, but here are a few important points:

- I'm okay with adding `dispatchAsync` if you're adamant about keeping it, but I'm strongly against adding any code to production that we know for sure is completely unused.

- We don't need to change `dispatch` here, and we shouldn’t unless it's absolutely necessary. It should call the store directly and calling `dispatchAsync` is unnecessary from `dispatch` IMO. I'm fine keeping `dispatchAsync` if you can revert `dispatch` to the previous implementation.

mrehan27 (changes requested):

Sources/MessagingInApp/State/InAppMessageManager.swift
Shahroz16: This is currently, even if it's the use-case is `test` for now, next up it could very much be using swift concurrency through out the SDK. 

It doesn't make sense to me why we would have 2 methods doing the same thing that is calling `store.dispatch` but separately. It just means, we are saying, whenever we need to make a change in logic, we update two methods for it. 

We are not `changing` dispatch here, we are actively calling another method which would be the primary method for `dispatching.` There is ZERO logic change.

Shahroz16 (commented):
To unblock the PR, I am going with a solution that add an overhead of creating `task` every single time, but only add it for test cases, so not going to expose in the manager itself.

mrehan27 (approved):

PR #810, opened 2024-09-12
refactor: custom store implementation

mrehan27:

### Changes

- Removed `ReSwift` dependency so that pods can work without any workarounds
- Moved middleware to `InAppMessageMiddleware` file as private implementation to avoid naming conflicts in other files
- Simplified reducer and middleware as action type is `InAppMessageAction` everywhere
- Copied following files from [`ReSwift`](https://github.com/ReSwift/ReSwift/tree/11f0a2e4965487c8d1f1dd1cffc59e39ef3f2666) and modified them as needed to maintain the custom store functionality:
  - `Assertions`
  - `Middleware`
  - `Reducer`
  - `Store`
  - `StoreSubscriber`
  - `Subscription`
  - `Synchronized`

### DIFF
--- a/Sources/MessagingInApp/State/Core/Store.swift
+++ b/Sources/MessagingInApp/State/Core/Store.swift
@@ -44,10 +44,6 @@ class Store<State: Equatable> {
 
     private var isDispatching = Synchronized<Bool>(false)
 
-    /// Indicates if new subscriptions attempt to apply `skipRepeats`
-    /// by default.
-    private let subscriptionsAutomaticallySkipRepeats: Bool
-
     public var middleware: [Middleware<State>] {
         didSet {
             dispatchFunction = createDispatchFunction()
@@ -69,10 +65,8 @@ class Store<State: Equatable> {
     public required init(
         reducer: @escaping Reducer<State>,
         state: State,
-        middleware: [Middleware<State>] = [],
-        automaticallySkipsRepeats: Bool = true
+        middleware: [Middleware<State>] = []
     ) {
-        self.subscriptionsAutomaticallySkipRepeats = automaticallySkipsRepeats
         self.reducer = reducer
         self.middleware = middleware
         self.state = state
@@ -93,11 +87,11 @@ class Store<State: Equatable> {
             )
     }
 
-    private func _subscribe<SelectedState, S: StoreSubscriber>(
-        _ subscriber: S, originalSubscription: Subscription<State>,
-        transformedSubscription: Subscription<SelectedState>?
-    )
-        where S.StoreSubscriberStateType == SelectedState {
+    private func _subscribe<S: StoreSubscriber>(
+        _ subscriber: S,
+        originalSubscription: Subscription<State>,
+        transformedSubscription: Subscription<State>?
+    ) where S.StoreSubscriberStateType == State {
         let subscriptionBox = self.subscriptionBox(
             originalSubscription: originalSubscription,
             transformedSubscription: transformedSubscription,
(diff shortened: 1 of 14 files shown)


Sources/MessagingInApp/State/Core/StoreSubscriber.swift
```diff
@@ -0,0 +1,28 @@
...
+//  Created by Benjamin Encz on 12/14/15.
+//  Copyright © 2015 ReSwift Community. All rights reserved.
+//
+
+public protocol AnyStoreSubscriber: AnyObject {
+    // swiftlint:disable:next identifier_name
+    func _newState(state: Any)
+}
+
+public protocol StoreSubscriber: AnyStoreSubscriber {
```
Shahroz16, line 15: this might not be needed as we creating the same thing in alias.

Sources/MessagingInApp/State/Core/Store.swift
```diff
@@ -0,0 +1,188 @@
...
+// MARK: Skip Repeats for Equatable States
+
+extension Store {
+    public func subscribe<SelectedState: Equatable, S: StoreSubscriber>(
+        _ subscriber: S, transform: ((Subscription<State>) -> Subscription<SelectedState>)?
+    ) where S.StoreSubscriberStateType == SelectedState {
+        let originalSubscription = Subscription<State>()
+
+        var transformedSubscription = transform?(originalSubscription)
+        if subscriptionsAutomaticallySkipRepeats {
```
Shahroz16, line 179: `subscriptionsAutomaticallySkipRepeats` is always true, i think we can remove it.

Sources/MessagingInApp/State/Core/Store.swift
```diff
@@ -0,0 +1,188 @@
...
+
+    open func dispatch(_ action: InAppMessageAction) {
+        dispatchFunction(action)
+    }
+}
+
+// MARK: Skip Repeats for Equatable States
+
+extension Store {
+    public func subscribe<SelectedState: Equatable, S: StoreSubscriber>(
```
Shahroz16, line 173: this probably can be removed as we have only 2 options, either give your own comparator for transformation or we will skip repeats, which can be covered in 1 method.

Sources/MessagingInApp/State/Core/Store.swift
```diff
@@ -0,0 +1,188 @@
...
+
+    private var reducer: Reducer<State>
+
+    var subscriptions: Set<SubscriptionType> = []
+
+    private var isDispatching = Synchronized<Bool>(false)
+
+    /// Indicates if new subscriptions attempt to apply `skipRepeats`
+    /// by default.
+    private let subscriptionsAutomaticallySkipRepeats: Bool
```
Shahroz16, line 49: can be removed

Shahroz16 (commented):
majorly looks good, added some suggestions. We should also bring in the tests, would it be part of next PR?

Sources/MessagingInApp/State/Core/StoreSubscriber.swift
mrehan27: To keep store generic and copy tests, it will be needed

Sources/MessagingInApp/State/Core/Store.swift
mrehan27: Yup, removed

Sources/MessagingInApp/State/Core/Store.swift
mrehan27: Yes, updated and merged to one

Sources/MessagingInApp/State/Core/Store.swift
mrehan27: Updated

mrehan27:
@Shahroz16 Yes, tests will be added later in future PRs

Shahroz16 (approved):
Looks good

PR #813, opened 2024-09-13
chore: prevent js callbacks when dismissed

mrehan27:

### Changes

- Dismissed message in `initial` state if `MessageManager` existed
- Moved un-subscription code to `dismissMessage` callback, ensuring the message is dismissed only after dismiss animation completes
- Moved reference cleanup code into dedicated methods for improved readability
- Fetched messages from local state whenever a message is dismissed, allowing next queued message to be shown instantly without waiting for next polling
- Persistent messages are now added to `shownMessageQueueIds` only after dismissal, aligning with how we handle in-app message viewed logs in our middleware today

### Bug Fixes

Fixed an issue where message was not dismissed when in-app state was reset while the message was in `loading` state.

#### Steps to Reproduce

1. Sign in and stay on dashboard screen
2. Send a message with empty page rule
3. Monitor the logs, and sign out while the message is in `loading` state
4. Notice constant logs from JS callbacks on login screen before this fix

Fixed an issue where persistent message did not appear again in the same session if dismissed by click action, even though it was received in remote queue API call.

#### Steps to Reproduce

1. Send persistent message with click action (e.g., external link)
2. Instead of closing it using close action, click on link action (the message will be dismissed)
3. Navigate back to the app and notice that the message is received in remote queue but suppressed due to shown message IDs stored locally before this change


Sources/MessagingInApp/Gist/Managers/MessageManager.swift
```diff
@@ -50,23 +52,40 @@ class MessageManager: EngineWebDelegate {
...
+                        self.removeEngineWebView()
+                        // Unsubscribe from InAppMessageState when the message is dismissed completely
+                        // so that MessageManager is deallocated only after dismiss animation is completed.
+                        self.dismissMessage {
+                            self.unsubscribeFromInAppMessageState()
+                            // Fetch user messages from local store after message is dismissed so that
+                            // next message can be displayed instantly if available.
+                            // This is only needed when message is dismissed and not when it is reset.
+                            if case .dismissed = messageState {
+                                self.gist.fetchUserMessagesFromStoreState()
```
mrehan27, line 86: In Android, this is done by fetching messages from remote in `onActivityResumed` in `GistSdk`

Sources/MessagingInApp/State/InAppMessageReducer.swift
```diff
@@ -49,16 +49,30 @@ private func reducer(action: InAppMessageAction, state: InAppMessageState) -> In
 
     case .displayMessage(let message):
         if let queueId = message.queueId {
+            // Do not add queueId to shownMessageQueueIds if the message is persistent
+            // As it should be considered shown only when it is dismissed via close action.
```
Shahroz16, line 53: I think we should not mix these two concepts, logging a view telling gist services that the message is shown vs keeping track of if the message is shown internally. 

Because technically the message is indeed shown to the user and we rely on it to ensure duplication doesn't happen.

Sources/MessagingInApp/State/InAppMessageReducer.swift
```diff
@@ -49,16 +49,30 @@ private func reducer(action: InAppMessageAction, state: InAppMessageState) -> In
...
+                shownMessageQueueIds: shownMessageQueueIds
             )
         }
         return state
 
-    case .dismissMessage(let message, _, _):
-        return state.copy(currentMessageState: .dismissed(message: message))
+    case .dismissMessage(let message, let shouldLog, let viaCloseAction):
+        // If message is persistent and dismissed via close action, add queueId to shownMessageQueueIds.
+        if message.gistProperties.persistent == true, let queueId = message.queueId, shouldLog, viaCloseAction {
```
Shahroz16, line 68: same comment as above

Sources/MessagingInApp/Gist/Managers/MessageManager.swift
```diff
@@ -50,23 +52,40 @@ class MessageManager: EngineWebDelegate {
...
+                switch messageState {
                 case .displayed:
                     threadUtil.runMain {
                         self.loadModalMessage()
                     }
 
-                case .dismissed:
+                // Dismiss the message when the message is dismissed or initial state
+                // Initial state may only be received when state is reset while a message was being displayed
+                case .dismissed, .initial:
```
Shahroz16, line 75: This looks vulnerable, specially `.inital` part, when we trigger `loadMessage` action, that's when MessageManager is created and probably this listener is going to be setup. The state at that point is still `.initial`. 

It isn't till `loadMessage` goes to the reducer, then its state is changed from the initial.

Sources/MessagingInApp/Gist/Managers/MessageManager.swift
Shahroz16: I wonder whats stopping us from having this as a middleware, that any time `dismiss` action occurs we process the current local queue.

mrehan27: Discussed this on the call. Since the middleware doesn't account for message animation, an additional `loading` event is received with suggested approach on current `MessageManager` while previous message is being dismissed. To avoid this event and make the behavior more consistent with Android, we followed similar approach by fetching messages from remote queue instead.

Sources/MessagingInApp/State/InAppMessageReducer.swift
mrehan27: Discussed this on the call and moved the checks to `shouldTrackMessageShown`, which will now be used in both middlewares and reducer for consistency.

Sources/MessagingInApp/State/InAppMessageReducer.swift
mrehan27: Same as above

Sources/MessagingInApp/Gist/Managers/MessageManager.swift
mrehan27: Discussed this on the call. The `initial` event is not received, and only changes are. So this won't be an issue with current implementation.

Sources/MessagingInApp/Gist/Managers/MessageManager.swift
```diff
@@ -78,6 +97,25 @@ class MessageManager: EngineWebDelegate {
...
+
+        logger.logWithModuleTag("Unsubscribing MessageManager from InAppMessageState", level: .debug)
+        inAppMessageManager.unsubscribe(subscriber: subscriber)
+        inAppMessageStoreSubscriber = nil
+    }
+
+    /// Removes EngineWebView from MessageManager and sets the delegate to nil to stop receiving JS events.
+    func removeEngineWebView() {
+        // If delegate is nil, then EngineWebView is already cleaned up.
+        guard let _ = engine.delegate else { return }
```
Shahroz16, line 112: this seems new and wasnt part of old code, any specific usecase?

Sources/MessagingInApp/State/InAppMessageAction.swift
```diff
@@ -72,3 +72,24 @@ enum InAppMessageAction: Equatable {
     }
     // swiftlint:enable cyclomatic_complexity
 }
+
+extension InAppMessageAction {
+    /// Determines whether the message should be tracked viewed and added to `shownMessageQueueIds` or not.
+    ///
+    /// - Returns: `true` if the message is non-persistent in `displayMessage` case or if it's persistent and
+    ///   meets the conditions in `dismissMessage` case, otherwise `false`.
+    var shouldTrackMessageShown: Bool {
```
Shahroz16, line 81:
```suggestion
    var shouldTrackMessageBeShown: Bool {
```

Sources/MessagingInApp/State/InAppMessageReducer.swift
```diff
@@ -49,16 +49,32 @@ private func reducer(action: InAppMessageAction, state: InAppMessageState) -> In
...
                 messagesInQueue: state.messagesInQueue.filter { $0.queueId != queueId },
-                shownMessageQueueIds: state.shownMessageQueueIds.union([queueId])
+                shownMessageQueueIds: shownMessageQueueIds
             )
         }
         return state
 
     case .dismissMessage(let message, _, _):
-        return state.copy(currentMessageState: .dismissed(message: message))
+        let shownMessageQueueIds: Set<String>
```
Shahroz16, line 66: could be initialized with `state.shownMessageQueueIds`

Shahroz16 (approved):
looks good, minor suggestions

Sources/MessagingInApp/Gist/Managers/MessageManager.swift
mrehan27: Yup, I added this to avoid calling `cleanEngineWeb` multiple times. To avoid any new explicit flag for this, I reused current delegate since we set it in `init`. Seems to work fine so far.

Sources/MessagingInApp/State/InAppMessageAction.swift
mrehan27: Updated to `shouldMarkMessageAsShown`

Sources/MessagingInApp/State/InAppMessageReducer.swift
mrehan27: Updated

PR #816, opened 2024-09-23
chore: add support for SDK client

mrehan27:
part of [MBL-521](https://linear.app/customerio/issue/MBL-521/user-agent-rn-ios)

### Changes

- Added `SdkClient` to allow overriding the SDK source and version for user agent
- Removed `sdkVersion` from `DeviceInfo`, making `SdkClient` single source of truth for the SDK version
- Fixed `NSE` `HttpClient` to use same user agent as the rest of the classes
- Fixed and added related tests and stubs for `SdkClient` updates

### Usage Example

```swift
DIGraphShared.shared.override(
    value: CustomerIOSdkClient(source: "React Native", sdkVersion: "4.0.0"),
    forType: SdkClient.self
)
```

### Notes

We need to ensure the client is overridden before initializing the SDK so that every class uses updated `SdkClient` value. While this approach works for now, I understand it's not ideal. I'll continue exploring options to eliminate the need for overriding this before initialization. Once I find a working solution, I will create another PR with required changes.


Tests/DataPipeline/DataPipelineCompatibilityTests.swift
```diff
@@ -146,12 +146,12 @@ class DataPipelineCompatibilityTests: IntegrationTest {
         XCTAssertFalse(savedEvent.containsKey("platform"))
 
         var expectedData = givenDefaultAttributes
+        expectedData["cio_sdk_version"] = diGraphShared.sdkClient.sdkVersion
```
mrehan27, line 149: Because `sdkVersion` is now from removed `DeviceInfo` and added to `SdkClient` instead

Sources/Common/Util/SdkClient.swift
```diff
@@ -0,0 +1,55 @@
...
+    // Default source for native iOS SDK
+    private static let SOURCE_IOS = "iOS"
+
+    /// Creates a new instance of `CustomerIOSdkClient` with provided source and sdk version.
+    /// If source or sdk version is nil or empty, it will use default values.
+    /// - Parameters:
+    ///  - source: Source of SDK client.
+    ///  - sdkVersion: Version of SDK client.
+    /// - Returns: A new instance of `CustomerIOSdkClient`.
+    public static func create(source: String? = nil, sdkVersion: String? = nil) -> CustomerIOSdkClient {
```
Shahroz16, line 39: We don't probably need all this, if we move OS information in UserAgent, and this can be just initialized with `iOS`

Tests/Common/Util/SdkClientTest.swift
```diff
@@ -0,0 +1,87 @@
+@testable import CioInternalCommon
+import Foundation
+import SharedTests
+import XCTest
+
+class SdkClientTest: UnitTest {
+    private let deviceInfoMock = DeviceInfoMock()
+    private let defaultClientString = "iOS Client/\(SdkVersion.version)"
```
Shahroz16, line 8: this should be defaulting to `SdkClient.version`

Shahroz16 (commented):
I do agree we might need to more user friendly and fail proof way to handle user agents so that we don't have to rely on SDK initialization.

Sources/Common/Service/UserAgentUtil.swift
```diff
@@ -23,7 +25,7 @@ public class UserAgentUtilImpl: UserAgentUtil {
      * `Customer.io iOS Client/1.0.0-alpha.16`
      */
     public func getUserAgentHeaderValue() -> String {
-        var userAgent = "Customer.io \(deviceInfo.osName ?? "iOS") Client/\(deviceInfo.sdkVersion)"
+        var userAgent = "Customer.io \(sdkClient)"
```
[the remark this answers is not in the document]
mrehan27: I'm not fully sure what exactly `deviceInfo.osName` will return in different cases, so I tried to cover cases for all `iOS`, `visionOS`, and wrappers using the same path.

If I understand your suggestion correctly, you're proposing we use `visionOS` irrespective of `SdkClient` source if `deviceInfo.osName` returns `visionOS`. This means it won't be as flexible as Android or how it ideally should be, but we would need to take the following assumptions if we follow this approach:

1. We are making this change primarily to simplify `SdkClient` creation (I don't see any other significant benefit, but please let me know if you have something in mind)
2. Any value in `SdkClient` source will be ignored if `deviceInfo.osName` returns `visionOS`
3. Any value in `deviceInfo.osName` will be ignored if it's not `visionOS`, and `SdkClient` source will be used instead which will default to `iOS` for native SDK
4. Support for any other native platforms (similar to visionOS) won't be considered unless we explicitly add support for them. For now, it will be limited to just `visionOS`, and not any more generic implementation (like a supported OS list) to keep the scope lighter. If needed in future, we can add the support later.

Does this make sense? If so, I'm okay with making the changes but wanted to ensure we're setting the right expectations. Also, this is a heavier change and will impact all the tests I've added/fixed. So just wanted to double-check before we proceed with making the change.

Sources/Common/Util/SdkClient.swift
mrehan27: Responded above regarding how this could impact rest of the changes.

Sources/MessagingPush/RichPush/RichPushHttpClient.swift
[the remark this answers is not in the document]
mrehan27: We just need to override `SdkClient` for this. The convenience constructor is a valid concern, but it's a separate problem that we'll address in a different PR. I'm exploring more options to avoid multiple constructors, but we'll see where we end up with it 🤞🏻

Tests/Common/Util/SdkClientTest.swift
mrehan27: No. This is for testing default value and should be `SdkVersion.version`

Sources/Common/Service/UserAgentUtil.swift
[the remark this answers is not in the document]
Shahroz16: Discussed on call, added a suggetion as a commit.

Shahroz16 (approved):

PR #820, opened 2024-10-01
fix: crash when switching screens frequently

Shahroz16:
closes: https://linear.app/customerio/issue/MBL-573/in-app-message-delivery-issue

changes:
- When utilizing `dispatcher` from `Digraph` we are accessing it very frequently when the screen changes very frequently, so instead of fetching/creating an instance every time, let's utilize the the instance provided by middleware.

### DIFF
--- a/Sources/MessagingInApp/State/InAppMessageMiddleware.swift
+++ b/Sources/MessagingInApp/State/InAppMessageMiddleware.swift
@@ -7,7 +7,7 @@ import Foundation
 private func middleware(
     completion: @escaping MiddlewareCompletion
 ) -> InAppMessageMiddleware {
-    { _,
+    { dispatch,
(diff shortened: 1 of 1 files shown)


mrehan27 (approved):
I hope this doesn't break any edge cases since it seems like previous action isn't fully completed before the next one is dispatched 😅 I ran some quick tests and everything looks good so far though. So looks good to me 👍🏻

Shahroz16:
> I hope this doesn't break any edge cases since it seems like previous action isn't fully completed before the next one is dispatched 😅 I ran some quick tests and everything looks good so far though. So looks good to me 👍🏻

I believe you are trying to say, that it won't wait for the first dispatcher to complete in case another dispatcher is launcher from a middleware, if so that should be perfectly alright, in android its working like that too and redux is supposed to work in that fashion as well.

PR #821, opened 2024-10-03
chore: in-app messaging config from dictionary

mrehan27:
part of: [MBL-538](https://linear.app/customerio/issue/MBL-538/add-feature-for-inapp-messages-in-ios-rn), [MBL-546](https://linear.app/customerio/issue/MBL-546/inapp-event-listeners-ios-and-android-rn)

### Changes

- Updated `Logger` class to allow setting log dispatcher for intercepting logs, ensuring only one logger instance and matching Android behavior
- Updated `MessagingInAppConfigBuilder` to meet React Native package expectations and provide clearer error explanations
- Updated tests for `Region` and `MessagingInAppConfigBuilder` to validate changes

### DIFF
--- a/Sources/MessagingInApp/Config/MessagingInAppConfigBuilder.swift
+++ b/Sources/MessagingInApp/Config/MessagingInAppConfigBuilder.swift
@@ -39,26 +39,51 @@ public class MessagingInAppConfigBuilder {
 
 /// Defines errors that can occur during the `MessagingInAppConfigBuilder` building process.
 public enum MessagingInAppConfigBuilderError: Error {
+    case malformedConfig
     case missingSiteId
-    case missingRegion
+    case invalidRegionType
 }
 
 public extension MessagingInAppConfigBuilder {
     /// Constants used to map each of the options in MessagingInAppConfigOptions.
-    enum Keys: String {
+    private enum Keys: String {
         case siteId
         case region
     }
 
     /// Constructs `MessagingInAppConfigOptions` by parsing and applying configurations from provided dictionary.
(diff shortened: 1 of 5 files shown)


Sources/MessagingInApp/Config/MessagingInAppConfigBuilder.swift
```diff
@@ -39,26 +39,51 @@ public class MessagingInAppConfigBuilder {
...
         case siteId
         case region
     }
 
     /// Constructs `MessagingInAppConfigOptions` by parsing and applying configurations from provided dictionary.
+    /// The method is used to create `MessagingInAppConfigOptions` from wrapper SDK configurations.
     @available(iOSApplicationExtension, unavailable)
-    static func build(from dictionary: [String: Any]) throws -> MessagingInAppConfigOptions {
-        guard let siteId = dictionary[Keys.siteId.rawValue] as? String else {
+    static func build(from sdkConfig: [String: Any?], withConfig inAppConfigKey: String = "inApp") throws -> MessagingInAppConfigOptions? {
```
Shahroz16, line 57: do we need `inAppConfigKey` as parameter since it's in the `MessagingInApp` module already? and we have a defined key for it?

Shahroz16 (approved):

Sources/MessagingInApp/Config/MessagingInAppConfigBuilder.swift
mrehan27: Yeah, looks like an overkill unless we really need to handle different keys for different wrapper SDKs. Removing it for now.

PR #822, opened 2024-10-03
chore: firebase version upgrade for v2

ami-aman:
Linear ticket: 

### Problem
Customers are currently facing issues when upgrading to Firebase's latest version, 11.2. They want to use the latest version of the dependency, but our package restricts usage to versions up to 11.0, preventing them from installing higher versions.

### Solution
Update the version range and test if the app successfully upgrades to the latest Firebase version while using our SDK.


Shahroz16 (approved):
Looks good

PR #826, opened 2024-10-16
chore: add nse user agent support

mrehan27:
part of [MBL-540](https://linear.app/customerio/issue/MBL-540/react-native)

### Changes

- Updated `UserAgentUtil` to provide separate user agents for main target and NSE
- Added a separate NSE `SdkClient` to generate user agent from client using the same code
- Updated `RichPushHttpClient` to use NSE-specific user agent
- Removed unused methods
- Added tests for NSE headers

### DIFF
--- a/Sources/MessagingPush/MessagingPushImplementation.swift
+++ b/Sources/MessagingPush/MessagingPushImplementation.swift
@@ -6,29 +6,12 @@ class MessagingPushImplementation: MessagingPushInstance {
     let logger: Logger
     let jsonAdapter: JsonAdapter
     let eventBusHandler: EventBusHandler
-    let richPushDeliveryTracker: RichPushDeliveryTracker
-
-    /// testing init
-    init(
-        moduleConfig: MessagingPushConfigOptions,
-        logger: Logger,
-        jsonAdapter: JsonAdapter,
-        eventBusHandler: EventBusHandler,
-        richPushDeliveryTracker: RichPushDeliveryTracker
-    ) {
-        self.moduleConfig = moduleConfig
-        self.logger = logger
-        self.jsonAdapter = jsonAdapter
-        self.eventBusHandler = eventBusHandler
-        self.richPushDeliveryTracker = richPushDeliveryTracker
-    }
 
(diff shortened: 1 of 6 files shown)


Sources/MessagingPush/MessagingPushImplementation.swift
```diff
@@ -6,29 +6,12 @@ class MessagingPushImplementation: MessagingPushInstance {
     let logger: Logger
     let jsonAdapter: JsonAdapter
     let eventBusHandler: EventBusHandler
-    let richPushDeliveryTracker: RichPushDeliveryTracker
-
-    /// testing init
-    init(
```
mrehan27, line 12: Removed because it wasn't used anywhere

Sources/MessagingPush/MessagingPushImplementation.swift
```diff
@@ -48,6 +31,9 @@ class MessagingPushImplementation: MessagingPushInstance {
         event: Metric,
         deviceToken: String
     ) {
-        richPushDeliveryTracker.trackMetric(token: deviceToken, event: event, deliveryId: deliveryID) { _ in }
+        // Access richPushDeliveryTracker from DIGraphShared.shared directly as it is only required for NSE.
+        // Keeping it as class property results in initialization of UserAgentUtil before SDK client is overridden by wrapper SDKs.
+        // In future, we can improve how we access SdkClient so that we don't need to worry about initialization order.
+        DIGraphShared.shared.richPushDeliveryTracker.trackMetric(token: deviceToken, event: event, deliveryId: deliveryID) { _ in }
```
Shahroz16, line 37: We should ideally limit the exposure of this class in NSE to avoid this.

Shahroz16 (approved):

Sources/MessagingPush/MessagingPushImplementation.swift
mrehan27: Agreed. I avoided it to prevent scope creep for now, but you're right, the proper fix should aim to avoid similar issues in the future. Exposing these methods specifically for NSE is one good option for this. I'll create ticket to follow this up.

PR #827, opened 2024-10-16
fix: v2 dismiss keyboard when inapp message shows

ami-aman:
### Problem
- An in-app message is sent to an iOS device.
- User opens the keyboard and begins typing.
- Issue arises when the in-app message shows on the screen, as the user is either blocked from closing the message or continuing to type.

This behavior prevents interaction with the app, as displaying of the in-app message disrupts the current activity (e.g., typing or moving back to previous screen) and makes it difficult for the user to manage actions. 

### Solution
Dismiss keyboard when in-app message shows up on the screen.

### Testing
The change has been tested on :
- iOS v2 SDK (Journeys)
- Wrapper SDKs

### Video

https://github.com/user-attachments/assets/c4095ee4-c09a-42cb-96e0-3d9a5d937366

### DIFF
--- a/Sources/MessagingInApp/Gist/Gist.swift
+++ b/Sources/MessagingInApp/Gist/Gist.swift
@@ -91,6 +91,14 @@ public class Gist: GistDelegate {
         } else {
             Logger.instance.debug(message: "Persistent message shown, skipping logging view")
         }
+        
+        // To ensure the keyboard is dismissed on displaying an in-app message,
+        // Update UI on main thread only.
+        // Dismiss keyboard on showing the message to prevent gap between
+        // displaying the in-app message and dismissing the keyboard.
+        DispatchQueue.main.async {
+            UIApplication.shared.sendAction(#selector(UIResponder.resignFirstResponder), to: nil, from: nil, for: nil)
+        }
         delegate?.messageShown(message: message)
     }
 


ami-aman:
> do we know which sdk customer logged this issue? because I think we should ideally avoid maintaining `v2` unless its a specific blocker.

This ticket was created from a request from this [zendesk ticket](https://customer7760.zendesk.com/agent/tickets/123073) and changes to `v2` have been done considering this [react native github request](https://github.com/customerio/customerio-reactnative/issues/176) that's also `+'ed 1` to this [main linear ticket](https://linear.app/customerio/issue/MBL-489/in-app-messages-can-conflict-with-keyboard-on-ios-and-require-an-app).

Shahroz16 (commented):
From the comment it seems like, its a request from React native and now our react native package do points at `v3` branch. We can just request the customer to utilize the new data pipelines SDK.

ami-aman:
> From the comment it seems like, its a request from React native and now our react native package do points at `v3` branch. We can just request the customer to utilize the new data pipelines SDK.

Fair ask. I think PR was created few days before releasing CDP support on React native so we can drop it if required.

ami-aman:
Closing this pull request as we'll be releasing this fix on CDP compatible version only.

PR #828, opened 2024-10-16
fix: dismiss keyboard on showing in app message on CDP SDK

ami-aman:
### Problem
- An in-app message is sent to an iOS device.
- User opens the keyboard and begins typing.
- Issue arises when the in-app message shows on the screen, as the user is either blocked from closing the message or continuing to type.

This behavior prevents interaction with the app, as displaying of the in-app message disrupts the current activity (e.g., typing or moving back to previous screen) and makes it difficult for the user to manage actions. 

### Solution
Dismiss keyboard when in-app message shows up on the screen.

### Testing
The change has been tested on :
- iOS CDP SDK (CDP)

### Video

https://github.com/user-attachments/assets/c4095ee4-c09a-42cb-96e0-3d9a5d937366

### DIFF
--- a/Sources/MessagingInApp/Gist/GistDelegate.swift
+++ b/Sources/MessagingInApp/Gist/GistDelegate.swift
@@ -41,7 +41,13 @@ class GistDelegateImpl: GistDelegate {
         if let deliveryId = message.campaignDeliveryId {
             eventBusHandler.postEvent(TrackInAppMetricEvent(deliveryID: deliveryId, event: InAppMetric.opened.rawValue))
         }
-
+        // To ensure the keyboard is dismissed on displaying an in-app message,
+        // Update UI on main thread only.
+        // Dismiss keyboard on showing the message to prevent gap between
+        // displaying the in-app message and dismissing the keyboard.
+        DispatchQueue.main.async {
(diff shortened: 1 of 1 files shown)


Sources/MessagingInApp/Gist/GistDelegate.swift
```diff
@@ -41,7 +41,13 @@ class GistDelegateImpl: GistDelegate {
         if let deliveryId = message.campaignDeliveryId {
             eventBusHandler.postEvent(TrackInAppMetricEvent(deliveryID: deliveryId, event: InAppMetric.opened.rawValue))
         }
-
+        // To ensure the keyboard is dismissed on displaying an in-app message,
+        // Update UI on main thread only.
+        // Dismiss keyboard on showing the message to prevent gap between
+        // displaying the in-app message and dismissing the keyboard.
+        DispatchQueue.main.async {
```
Shahroz16, line 48: don't we have a dispatcher util available in this class? if not cant we make it available instead?

ami-aman: I think you're referring to `ThreadUtil` class ? 🤔  If yes, then I have implemented that in the latest commit.

Shahroz16 (approved):

PR #832, opened 2024-10-30
chore: update Fastfile to publish to new distribution groups

scotttwittrockcio:
This PR adds the Firebase Distribution groups for 'feature-branch', 'next', and 'public' in addition to the current 'all-builds' and 'stable-builds' groups.

### DIFF
--- a/Apps/fastlane/Fastfile
+++ b/Apps/fastlane/Fastfile
@@ -215,13 +215,15 @@ lane :get_build_notes do
 end 
 
 lane :get_build_test_groups do 
-  test_groups = ['all-builds'] # send all builds to group 'all-builds'. Therefore, set it here and we will not remove it. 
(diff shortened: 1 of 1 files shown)


Shahroz16 (approved):
Change looks good, but most of github action seems to be failing unsure why, rerunning them.

scotttwittrockcio:
> Change looks good, but most of github action seems to be failing unsure why, rerunning them.

@Shahroz16 Do you think these might be failing because they are from a fork vs. a branch in the main repo? Now that I fixed my local dev issue I could redo the PR from there.

Shahroz16:
> > Change looks good, but most of github action seems to be failing unsure why, rerunning them.
> 
> @Shahroz16 Do you think these might be failing because they are from a fork vs. a branch in the main repo? Now that I fixed my local dev issue I could redo the PR from there.

@scotttwittrockcio thats seems very plausible to me, cause it doesn't make sense otherwise :D

PR #834, opened 2024-11-04
chore: move to v2 for user queue request and user agent in headers

Shahroz16:
closes: https://linear.app/customerio/issue/MBL-495/move-to-v2-apis-of-gist-and-add-sdk-reporting

changes:
- Upgrade user queue endpoint path to v2
- Added SDK version and platform in request headers

### DIFF
--- a/Sources/MessagingInApp/Gist/Network/GistQueueNetwork.swift
+++ b/Sources/MessagingInApp/Gist/Network/GistQueueNetwork.swift
@@ -22,10 +22,14 @@ class GistQueueNetworkImpl: GistQueueNetwork {
             throw GistNetworkRequestError.invalidBaseURL
         }
 
+        let sdkClient = DIGraphShared.shared.sdkClient
+
         var urlRequest = URLRequest(url: baseURL.appendingPathComponent(request.path))
         urlRequest.httpMethod = request.method.rawValue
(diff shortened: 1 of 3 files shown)


Sources/MessagingInApp/Gist/Network/GistQueueNetwork.swift
```diff
@@ -22,10 +22,14 @@ class GistQueueNetworkImpl: GistQueueNetwork {
...
         }
 
+        let sdkClient = DIGraphShared.shared.sdkClient
+
         var urlRequest = URLRequest(url: baseURL.appendingPathComponent(request.path))
         urlRequest.httpMethod = request.method.rawValue
         urlRequest.addValue(state.siteId, forHTTPHeaderField: HTTPHeader.siteId.rawValue)
         urlRequest.addValue(state.dataCenter, forHTTPHeaderField: HTTPHeader.cioDataCenter.rawValue)
+        urlRequest.addValue(sdkClient.sdkVersion, forHTTPHeaderField: HTTPHeader.cioClientVersion.rawValue)
+        urlRequest.addValue(sdkClient.source.lowercased(), forHTTPHeaderField: HTTPHeader.cioClientPlatform.rawValue)
```
mrehan27, line 32: Same question as for Android. Did you verify if React Native Data Pipelines support is sending expected source and version correctly? If yes, then we're good with the change.

mrehan27 (approved):
Same one question, rest looks good

PR #835, opened 2024-11-07
chore: update Fastfile to publish to new distribution groups

scotttwittrockcio:
This PR adds the Firebase Distribution groups for 'feature-branch', 'next', and 'public' in addition to the current 'all-builds' and 'stable-builds' groups.

### DIFF
--- a/Apps/fastlane/Fastfile
+++ b/Apps/fastlane/Fastfile
@@ -215,7 +215,7 @@ lane :get_build_notes do
 end 
 
 lane :get_build_test_groups do 
-  test_groups = ['all-builds', 'feature-branch'] # send all builds to group 'all-builds'. Therefore, set it here and we will not remove it. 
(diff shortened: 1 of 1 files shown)


scotttwittrockcio:
@Shahroz16 Here is the new PR that used this repo vs the fork as a base. The checks are all passing, and I can confirm it also published to the right groups! Could I get a quick check and approval on this before I merge. I saw the PR template we "Squash and merge". Is there anything else to note before merging after approval?

Shahroz16 (approved):

Shahroz16:
> @Shahroz16 Here is the new PR that used this repo vs the fork as a base. The checks are all passing, and I can confirm it also published to the right groups! Could I get a quick check and approval on this before I merge. I saw the PR template we "Squash and merge". Is there anything else to note before merging after approval?

@scotttwittrockcio its a chore branch so all you need is `squash and merge` and rest should be good.

PR #842, opened 2024-12-26
chore: encode loadPage action URL

mrehan27:
fixes: [MBL-762](https://linear.app/customerio/issue/MBL-762/[bug]-ios-sdk-strips-fragment-from-urls-in-legacy-in-app-messages)

### Changes

- Encoded action URL before extracting it for `loadPage` action to retain fragment URL part
- Added string extension to encode specific characters
- Added relevant tests

#### Input

```
gist://loadPage?url=https://docs.customer.io/sdk/ios/in-app/set-up-in-app/#set-up-in-app-messaging
```

#### Output `pageUrl`

**Before:**

```
https://docs.customer.io/sdk/ios/in-app/set-up-in-app/
```  

**After:**

```
https://docs.customer.io/sdk/ios/in-app/set-up-in-app/#set-up-in-app-messaging
```

### DIFF
--- a/Sources/MessagingInApp/Gist/Managers/MessageManager.swift
+++ b/Sources/MessagingInApp/Gist/Managers/MessageManager.swift
@@ -171,7 +171,9 @@ class MessageManager: EngineWebDelegate {
                 logger.logWithModuleTag("Dismissing from action: \(action)", level: .info)
                 inAppMessageManager.dispatch(action: .dismissMessage(message: currentMessage, viaCloseAction: true))
             case "loadPage":
-                if let page = url.queryParameters?["url"],
+                // Encode '#' in url action string and creates encoded URL to handle fragments
+                let encodedUrl = URL(string: action.percentEncode(character: "#")) ?? url
+                if let page = encodedUrl.queryParameters?["url"],
                    let pageUrl = URL(string: page),
                    UIApplication.shared.canOpenURL(pageUrl) {
                     UIApplication.shared.open(pageUrl)
--- a/Sources/MessagingInApp/Gist/Utilities/String.swift
+++ b/Sources/MessagingInApp/Gist/Utilities/String.swift
@@ -1,0 +1,8 @@
+import Foundation
+
+extension String {
+    /// Encodes a specific character in the string using percent encoding.
+    func percentEncode(character target: String, withAllowedCharacters allowedCharacters: CharacterSet = .urlPathAllowed) -> String {
+        replacingOccurrences(of: target, with: target.addingPercentEncoding(withAllowedCharacters: allowedCharacters) ?? target)
+    }
+}
--- a/Tests/MessagingInApp/Gist/Utilities/StringPercentEncodeTests.swift
+++ b/Tests/MessagingInApp/Gist/Utilities/StringPercentEncodeTests.swift
@@ -1,0 +1,69 @@
+@testable import CioMessagingInApp
+import XCTest
+
+class StringPercentEncodeTests: XCTestCase {
+    func test_givenValidCharacter_expectEncodedURL() {
+        let input = "https://example.com/path#fragment"
+
+        let encoded = input.percentEncode(character: "#")
+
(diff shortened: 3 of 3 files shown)


Sources/MessagingInApp/Gist/Managers/MessageManager.swift
```diff
@@ -171,7 +171,9 @@ class MessageManager: EngineWebDelegate {
                 logger.logWithModuleTag("Dismissing from action: \(action)", level: .info)
                 inAppMessageManager.dispatch(action: .dismissMessage(message: currentMessage, viaCloseAction: true))
             case "loadPage":
-                if let page = url.queryParameters?["url"],
+                // Encode '#' in url action string and creates encoded URL to handle fragments
+                let encodedUrl = URL(string: action.percentEncode(character: "#")) ?? url
```
Shahroz16, line 175: wonder why don't we encode the whole URL? instead of separating out fragment? 

```
if let page = url.queryParameters?["url"]
let encodedPage = page.addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed)
let pageUrl = URL(string: page)
```

so instead of encoding, just `#` we encode everything except that allowed and I would hope the webview opens it just fine.

mrehan27: That's what I've been considering as well. However, I opted to keep the impact minimal and decided to encode only `#` for following reasons:

- The link is provided directly by customers, so I wanted to avoid making any unnecessary modifications. While I agree that browsers and most apps should not be affected by encoding, I'm unsure if encoding full URL might impact any specific customer links.
- On Android, [we don't directly encode full URL either](https://github.com/customerio/customerio-android/blob/a402e9645df501bba8ff5d32ee423d8f818635e4/messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/GistView.kt#L88-L94), but `UrlQuerySanitizer` respects fragments (`#`) without encoding them, whereas iOS drops them entirely. This change simply makes iOS behavior a bit closer to Android, while relying on default `URL` class for encoding the rest.
- Based on my testing, this code is only used when utilizing "Link to web page" action in legacy in-app message editor. With deep link option and new editor, links are handled differently. So overall impact of this change is very low.

I'm open to encoding full URL as well, and I agree it should work fine for most browsers. However, I'm slightly inclined toward encoding only `#` to minimize potential risks since the url is customer provided, and I cannot fully verify all possible impacts. By encoding only `#`, the behavior remains closer to Android while making minimal changes to current implementation.

Shahroz16: This sounds reasonable if the android counter part is not fully encoding it either. I just want to avoid the case, where if someone adds another character allowed by URL and we end up filtering it out just like we did for `#`. 

One thing to note is, even though android doesn't encode it, `URI.parse` does is able to distinguish the following 

```
foo://example.com:8042/over/there?name=ferret#nose
\_/   \______________/\_________/ \_________/ \__/
 |           |            |            |        |
scheme    authority      path        query   fragment
```

you can see there is usage of `?` `=` as well along with `#`, I assume the iOS counter part won't be able to handle these?

mrehan27: So far, iOS seems to handle everything well except for dropping the fragment. e.g., when parsing following URL:

```
gist://loadPage?url=https://example.com/to/destination#id=test
```

The query parameter is preserved, but fragment (`#id=test`) is dropped when accessing `url.queryParameters?["url"]` as it only returns `https://example.com/to/destination` without encoding `#`, but returns `https://example.com/to/destination#id=test` after encoding `#`.

Open to both approaches since encoding entire URL should work as it's a standard practice. However, I'm slightly inclined toward not encoding full URL as this behavior appears to align more closely with Android's handling, though I agree it's not entirely the same.

Shahroz16: As long as it satisfies those above cases, I am good too.

Shahroz16 (approved):

PR #843, opened 2024-12-28
fix: Fix in-app messages overlay color ignored from message payload

mahmoud-elmorabea:
## Motivation
When sending an in-app message to the mobile apps, the specified `Overlay Color` in the in-app message editor is not reflected on the message shown on the app.

## Root cause
The value `overlayColor` sent to the mobile SDK from the server along with the in-app message payload is ignored and not used to set the in-app message overlay color.

Instead a hard-coded color is used for all message.

## Testing the fix

https://github.com/user-attachments/assets/ed33529a-cc7c-4aba-ba97-a8f7a20135bb

### DIFF
--- a/Sources/MessagingInApp/Gist/Utilities/UIColor+Hex.swift
+++ b/Sources/MessagingInApp/Gist/Utilities/UIColor+Hex.swift
@@ -1,0 +1,35 @@
+import UIKit
+
+extension UIColor {
+    static func fromHex(color: String?) -> UIColor? {
+        guard var hexColor = color else { return nil }
+        if hexColor.hasPrefix("#") {
+            hexColor.removeFirst()
+        }
+
+        let scanner = Scanner(string: hexColor)
+
+        var rgbValue: UInt64 = 0
+        guard scanner.scanHexInt64(&rgbValue) else {
+            return nil
+        }
+
+        var red, green, blue, alpha: UInt64
+        switch hexColor.count {
+        case 6:
+            red = (rgbValue >> 16)
+            green = (rgbValue >> 8 & 0xFF)
+            blue = (rgbValue & 0xFF)
+            alpha = 255
+        case 8:
+            red = (rgbValue >> 16)
+            green = (rgbValue >> 8 & 0xFF)
+            blue = (rgbValue & 0xFF)
+            alpha = rgbValue >> 24
+        default:
+            return nil
+        }
+
+        return UIColor(red: CGFloat(red) / 255, green: CGFloat(green) / 255, blue: CGFloat(blue) / 255, alpha: CGFloat(alpha) / 255)
+    }
+}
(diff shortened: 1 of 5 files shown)


Sources/MessagingInApp/Gist/Utilities/UIColor+Hex.swift
```diff
@@ -0,0 +1,35 @@
...
+            red = (rgbValue >> 16)
+            green = (rgbValue >> 8 & 0xFF)
+            blue = (rgbValue & 0xFF)
+            alpha = rgbValue >> 24
+        default:
+            return nil
+        }
+
+        return UIColor(red: CGFloat(red) / 255, green: CGFloat(green) / 255, blue: CGFloat(blue) / 255, alpha: CGFloat(alpha) / 255)
+    }
```
Shahroz16, line 34:
```suggestion
 static func fromHex(_ hex: String?) -> UIColor? {
        guard let hex = hex else { return nil }
        
        let cleanHex = hex.hasPrefix("#") ? String(hex.dropFirst()) : hex
        
        // Validate hex string length
        guard cleanHex.count == 6 || cleanHex.count == 8 else {
            return nil
        }
        
        var rgbValue: UInt64 = 0
        guard Scanner(string: cleanHex).scanHexInt64(&rgbValue) else {
            return nil
        }
        
        // Extract color components
        let red = CGFloat((rgbValue >> 16) & 0xFF) / 255.0
        let green = CGFloat((rgbValue >> 8) & 0xFF) / 255.0
        let blue = CGFloat(rgbValue & 0xFF) / 255.0
        let alpha = cleanHex.count == 8 
            ? CGFloat((rgbValue >> 24) & 0xFF) / 255.0 
            : 1.0
        
        return UIColor(red: red, green: green, blue: blue, alpha: alpha)
    }
```

small refactor and optimizations

Shahroz16 (approved):
looks good with one suggestion

mrehan27 (approved):

Sources/MessagingInApp/Gist/Utilities/UIColor+Hex.swift
```diff
@@ -0,0 +1,42 @@
...
+        // Validate hex string length
+        guard cleanHex.count == 6 || cleanHex.count == 8 else {
+            return nil
+        }
+
+        let correctHex = correctColorFormatIfNeeded(cleanHex)
+
+        var rgbValue: UInt64 = 0
+        guard Scanner(string: correctHex).scanHexInt64(&rgbValue) else {
+            return nil
```
mrehan27, line 18: > Sources/MessagingInApp/Gist/Utilities/UIColor+Hex.swift#L18
>
> Added line #L18 was not covered by tests

Would be nice if we can add test to cover this line too

mrehan27 (approved):

Sources/MessagingInApp/Gist/Utilities/UIColor+Hex.swift
mahmoud-elmorabea: Good point, I missed that added a [test](https://github.com/customerio/customerio-ios/pull/843/files#diff-ae2132b26b4afcb2417a2f6e503489639c9596257846cd0d6575017df174ec87R26-R28) for it!

Shahroz16 (approved):

PR #847, opened 2025-01-10
chore: build primary sample app

mrehan27:
part of: [MBL-710](https://linear.app/customerio/issue/MBL-710/single-app-per-platform-distributed-to-firebase)

### Changes

- Updated `Build Sample Apps` action to push only primary app to relevant channels
- Modified `get_build_test_groups` lane to decouple from GitHub and accept arguments for passing Firebase distribution groups when distributing sample apps

Please refer to linked ticket for detailed expectations regarding channel distribution.

### DIFF
--- a/.github/workflows/build-sample-apps.yml
+++ b/.github/workflows/build-sample-apps.yml
@@ -73,8 +73,8 @@ jobs:
         # Distribution group constants
         ALL_BUILDS_GROUP: all-builds
         FEATURE_BUILDS_GROUP: feature-branch
-        RELEASE_BUILDS_GROUP: public
-        STABLE_BUILDS_GROUP: next
+        NEXT_BUILDS_GROUP: next
+        PUBLIC_BUILDS_GROUP: public
         # Input variables
         IS_PRIMARY_APP: ${{ matrix.apn-or-fcm == 'APN' }}
         CURRENT_BRANCH: ${{ github.ref }}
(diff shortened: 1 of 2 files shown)


.github/workflows/build-sample-apps.yml
```diff
@@ -67,13 +67,37 @@ jobs:
...
+        current_branch="${GITHUB_REF}"
+
+        # Append distribution groups based on branch and context if the app is primary
+        if [[ "$is_primary_app" == "true" ]]; then
+          [[ "$current_branch" == "refs/heads/feature/"* ]] && distribution_groups+=("$FEATURE_BUILDS_GROUP")
+          [[ "$current_branch" == "refs/heads/main" ]] && distribution_groups+=("$STABLE_BUILDS_GROUP")
+        fi
+
+        # Export the groups as an environment variable
+        echo "firebase_distribution_groups=$(IFS=','; echo "${distribution_groups[*]}")" >> $GITHUB_ENV
```
mrehan27, line 91: I tested [final value with dummy input](https://github.com/customerio/customerio-ios/actions/runs/12693036897/job/35379733422#step:5:369), and it seems to be working well.

.github/workflows/build-sample-apps.yml
```diff
@@ -67,13 +67,37 @@ jobs:
     steps:
     - uses: actions/checkout@v4
 
+    - name: Set Default Firebase Distribution Groups
+      run: |
```
Shahroz16, line 71: I think we can extract the variable initialization from `run` to `env`, something like 

```
- name: Set Firebase Distribution Groups
  shell: bash
  env:
    # Distribution group constants
    ALL_BUILDS_GROUP: all-builds
    FEATURE_BUILDS_GROUP: feature-branch
    STABLE_BUILDS_GROUP: next
    # Input variables
    IS_PRIMARY_APP: ${{ matrix.apn-or-fcm == 'APN' }}
    CURRENT_BRANCH: ${{ github.ref }}
  run: |
    # Initialize array with default group
    distribution_groups=("$ALL_BUILDS_GROUP")
```

Shahroz16 (approved):

mahmoud-elmorabea (approved):
💯

mahmoud-elmorabea (commented):
I personally find reading bash logic in Github actions steps to make the file harder to read and navigate.

Would it make things simpler if we pass which app is being built to Fastlane and keep the logic that decides which groups to publish to in the `get_build_test_groups` lane?
* We can send params `isPrimaryApp` and whatever else is needed

This will also allow us to share the logic as we used to between repos.

mrehan27:
Discussed [the suggestion](https://github.com/customerio/customerio-ios/pull/847#pullrequestreview-2542473100) on [slack](https://customerio.slack.com/archives/C06H7MMLH6C/p1736751278834939?thread_ts=1736751191.897369&cid=C06H7MMLH6C), we agreed to move forward with current approach for now.

.github/workflows/build-sample-apps.yml
mrehan27: Updated

PR #852, opened 2025-01-17
chore: updated message managers

Shahroz16:
helps: https://linear.app/customerio/issue/MBL-492/add-the-new-in-line-components-and-gel-them-with-new-in-app-flow

changes:
- Refactored `MessageManager` so that now we have `Base` `Modal` and `Inline`

### DIFF
--- a/Sources/MessagingInApp/Gist/Managers/InlineMessageManager.swift
+++ b/Sources/MessagingInApp/Gist/Managers/InlineMessageManager.swift
@@ -1,18 +1,19 @@
+import CioInternalCommon
 import Foundation
 import UIKit
 
-// Callbacks specific to inline message events.
+// MARK: - InlineMessageManagerDelegate
+
+/// Callbacks specific to inline message events.
 public protocol InlineMessageManagerDelegate: AnyObject {
-    /// Called when the GistView's HTML content size changes
+    /// Called when the inline web message (HTML) size changes
     func sizeChanged(width: CGFloat, height: CGFloat)
-
-    /// Called when the inline message is closed (via a "close" action or similar)
+    /// Called when the inline message is closed (via a "close" action)
     func onCloseAction()
-
     /**
-     Called when any inline button/action is tapped.
-     If your delegate handles the action completely and does not want any default handling, return `true`.
-     Otherwise, return `false` so the manager's default logic is triggered.
+     Called when an inline button or link is tapped.
+     Return `true` if your delegate handled the action entirely (so the manager does nothing more).
+     Return `false` if you'd like the manager to continue its normal flow.
      */
     func onInlineButtonAction(
         message: Message,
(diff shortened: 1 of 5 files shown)


Sources/MessagingInApp/Gist/Managers/MessageManager.swift
```diff
@@ -57,271 +74,284 @@ class MessageManager: EngineWebDelegate {
...
+    open func onMessageDismissed() {
+        // Subclasses override
     }
 
-    private func loadModalMessage() {
-        guard isMessageLoaded else {
-            logger.logWithModuleTag("Message already loaded. Skipping loading modal message: \(currentMessage.describeForLogs)", level: .debug)
-            return
+    // Internal, for internal usage:
+    func onMessageDismissed(messageState: MessageState) {
```
mrehan27, line 133: Can we please reconsider names of these methods? Current names are quite similar and make it difficult to differentiate between them. I'm fine with using longer, more descriptive names for internal methods if it improves clarity. Right now, `onMessageDismissed` and `onMessageDismissed` are hard to distinguish which can cause confusion when debugging or working on them.

Sources/MessagingInApp/Gist/Managers/ModalMessageManager.swift
```diff
@@ -0,0 +1,74 @@
...
+        guard isMessageLoaded else {
+            logger.logWithModuleTag(
+                "Message not loaded yet. Skipping loadModalMessage for \(currentMessage.describeForLogs).",
+                level: .debug
+            )
+            return
+        }
+
+        logger.logWithModuleTag(
+            "Loading modal message: \(currentMessage.describeForLogs)",
```
mrehan27, line 19:
```suggestion
            "Displaying modal message: \(currentMessage.describeForLogs)",
```

Sources/MessagingInApp/Gist/Managers/ModalMessageManager.swift
```diff
@@ -0,0 +1,74 @@
...
+        guard let modalViewManager = modalViewManager else {
+            // No modal to dismiss
+            finishDismissal(messageState: messageState)
+            return
+        }
+
+        // Dismiss the modal then call completion
+        modalViewManager.dismissModalView { [weak self] in
+            self?.finishDismissal(messageState: messageState)
+        }
```
mrehan27, line 53: Could be simplified by single handler if that looks good to you too

```suggestion
        // Common handler to finalize dismissal logic
        let dismissalHandler: () -> Void = { [weak self] in
            self?.finishDismissal(messageState: messageState)
        }

        guard let modalViewManager = modalViewManager else {
            // No modal to dismiss
            dismissalHandler()
            return
        }

        // Dismiss the modal then call completion
        modalViewManager.dismissModalView(completionHandler: dismissalHandler)
```

mrehan27 (commented):
Mostly looks good. Just one improvement I think we should address before proceeding. Rest looks good to me.

Sources/MessagingInApp/Gist/Managers/InlineMessageManager.swift
```diff
@@ -0,0 +1,135 @@
+import CioInternalCommon
+import Foundation
+import UIKit
+
+// MARK: - InlineMessageManagerDelegate
+
+/// Callbacks specific to inline message events.
+public protocol InlineMessageManagerDelegate: AnyObject {
```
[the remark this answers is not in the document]
mahmoud-elmorabea: I believe this is an internal delegate that we use internally to update the inline in-app messaging view, but I'll wait for @Shahroz16 to confirm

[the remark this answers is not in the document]
Shahroz16: I am not fully sure, because i haven't seen the view utilization of this, but I believe its exposure to customer views.

Sources/MessagingInApp/Gist/Managers/InlineMessageManager.swift
```diff
@@ -0,0 +1,135 @@
...
+
+    /// Called when the in-app state changes to `.displayed`.
+    /// Inline typically means the content is ready and the `inlineMessageView` is or can be added to the UI.
+    override public func onMessageDisplayed() {
+        logger.logWithModuleTag(
+            "Inline message displayed: \(currentMessage.describeForLogs)",
+            level: .debug
+        )
+        // If needed, you might notify your delegate that it's displayed or ready.
+        // e.g., inlineMessageDelegate?.didDisplayInlineMessage?()
```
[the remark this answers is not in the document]
Shahroz16: this is a future idea i believe

Sources/MessagingInApp/Gist/Managers/MessageManager.swift
Shahroz16: sure removed the unused one

mrehan27 (approved):

PR #854, opened 2025-01-21
chore: unify json adapter and deep linking 

Shahroz16:
helps: https://linear.app/customerio/issue/MBL-492/add-the-new-in-line-components-and-gel-them-with-new-in-app-flow

- these changes are combination from in-line inapp snapshot and main

changes: 
- unify deep linking from in-app and push
- json adapter changes


mrehan27 (approved):
Looks like I've seen these files before 🤓😁

PR #859, opened 2025-01-27
chore: replace swiftui placeholder with inline view in sample app

mrehan27:
part of: [MBL-827](https://linear.app/customerio/issue/MBL-827/integrate-inline-in-app-message-in-placeholder-screen)

### Changes

- Replaced placeholder dummy text with `InlineMessage` in sample app to utilize inline messages view
- Configured `InlineMessage` with `elementId` to match the desired value
- Updated paddings to match with design

### DIFF
--- a/Apps/APN-UIKit/APN UIKit/View/Inline/InlineInAppMessageView.swift
+++ b/Apps/APN-UIKit/APN UIKit/View/Inline/InlineInAppMessageView.swift
@@ -1,13 +1,11 @@
+import CioMessagingInApp
 import SwiftUI
 
 struct InlineInAppMessageView: View {
     var body: some View {
-        VStack(spacing: 10) {
-            // To be replaced with inline view
-            HeaderView()
-                .frame(height: 80)
+        VStack {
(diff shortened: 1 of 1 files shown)


Sources/MessagingInApp/Views/GistInlineInAppMessageView.swift
```diff
@@ -100,7 +100,7 @@ public class GistInlineMessageUIView: UIView {
     }
 
     deinit {
-        unsubscribeFromInAppMessageState()
+        dismissCurrentMessage()
```
Shahroz16, line 103: thinking if it would work for UI kit as well?

Shahroz16 (approved):

Sources/MessagingInApp/Views/GistInlineInAppMessageView.swift
mrehan27: Not entirely sure at the moment. I'm currently working on UIKit and will verify and fix this in upcoming PR if needed.

PR #867, opened 2025-02-17
chore: add workflow to use stable workflow

Shahroz16:
closes: https://linear.app/customerio/issue/MBL-880/[ios]-public-release-builds

changes:
- We moved away from the local reference of SDK using GUI-based SPM to `package.swift` based SPM swift that we can update in real-time. 
 - Made the workflow more flexible to be able to utilize different environments.


.github/workflows/build-sample-apps.yml
```diff
@@ -1,182 +1,115 @@
...
-
+      - name: Find existing PR comment
+        uses: peter-evans/find-comment@v3
+        id: existing-comment
+        with:
+          issue-number: ${{ github.event.pull_request.number }}
+          comment-author: 'github-actions[bot]'
+          body-includes: <!-- sample app builds -->
+
+      - name: Create or update comment
```
mrehan27, line 37: Might be good to keep `edit-mode: replace`? 🤔

.github/workflows/reusable_build_sample_apps.yml
```diff
@@ -0,0 +1,148 @@
...
+        description: "Secret for the APN sample app's Site ID in Customer.io."
+      CUSTOMERIO_APN_WORKSPACE_CDP_API_KEY:
+        required: true
+        description: "Secret for the APN sample app's CDP API key in Customer.io."
+      CUSTOMERIO_FCM_WORKSPACE_SITE_ID:
+        required: true
+        description: "Secret for the FCM sample app's Site ID in Customer.io."
+      CUSTOMERIO_FCM_WORKSPACE_CDP_API_KEY:
+        required: true
+        description: "Secret for the FCM sample app's CDP API key in Customer.io."
```
mrehan27, line 29: I don't see them being reused from `inputs` directly. Can you please verify and delete if they are unused now?

mrehan27 (approved):

.github/workflows/reusable_build_sample_apps.yml
Shahroz16: they are being forwarded to action

PR #878, opened 2025-04-17
chore: Trigger sample apps public builds after deployment is done

mahmoud-elmorabea:
Closes: [MBL-1053](https://linear.app/customerio/issue/MBL-1053/auto-create-testbed-releases-for-every-sdk-release)

This PR triggers a sample apps public release build after the SDK has been deployed successfully so that internal stakeholders always have up to date versions to test with.

### Note
I am not entirely sure if the package will be immediately available after publishing before it can be used to build the sample app or we need a delay. Starting out with immediate trigger and we can revise later if the build fails.


.github/workflows/deploy-sdk.yml
```diff
@@ -280,3 +280,20 @@ jobs:
...
+      GOOGLE_CLOUD_MATCH_READONLY_SERVICE_ACCOUNT_B64: ${{ secrets.GOOGLE_CLOUD_MATCH_READONLY_SERVICE_ACCOUNT_B64 }}
+      FIREBASE_APP_DISTRIBUTION_SERVICE_ACCOUNT_CREDS_B64: ${{ secrets.FIREBASE_APP_DISTRIBUTION_SERVICE_ACCOUNT_CREDS_B64 }}
+      CUSTOMERIO_APN_WORKSPACE_SITE_ID: ${{ secrets.CUSTOMERIO_APN_WORKSPACE_SITE_ID }}
+      CUSTOMERIO_APN_WORKSPACE_CDP_API_KEY: ${{ secrets.CUSTOMERIO_APN_WORKSPACE_CDP_API_KEY }}
+      SAMPLE_APPS_APN_FIREBASE_APP_ID: ${{ secrets.SAMPLE_APPS_APN_FIREBASE_APP_ID }}
+      CUSTOMERIO_FCM_WORKSPACE_SITE_ID: ${{ secrets.CUSTOMERIO_FCM_WORKSPACE_SITE_ID }}
+      CUSTOMERIO_FCM_WORKSPACE_CDP_API_KEY: ${{ secrets.CUSTOMERIO_FCM_WORKSPACE_CDP_API_KEY }}
+      SAMPLE_APPS_FCM_FIREBASE_APP_ID: ${{ secrets.SAMPLE_APPS_FCM_FIREBASE_APP_ID }}
+      SAMPLE_APPS_INSTRUCTIONS_GUIDE_LINK: ${{ secrets.SAMPLE_APPS_INSTRUCTIONS_GUIDE_LINK }}
+      SLACK_NOTIFY_RELEASES_WEBHOOK_URL: ${{ secrets.SLACK_NOTIFY_RELEASES_WEBHOOK_URL }}
```
Shahroz16, line 299: passing secrets here, and then passing them again from `build-sample-apps.yml` to reusable workflow is very annoying and repetitive? can we use any other strategies here, maybe using `composite actions` instead of `reusable workflows`, as composite actions can inherit secrets and environment variables from the calling workflow without explicit passing.

Shahroz16 (approved):

PR #879, opened 2025-04-18
fix: simplify region parsing in wrapper in-app initialization

mrehan27:
part of: [MBL-1052](https://linear.app/customerio/issue/MBL-1052/flutter-sdk-region-is-not-set-correctly-by-default)

### Changes

- Updated `MessagingInAppConfigBuilder.build([String: Any?])` to parse region leniently and fall back to default region if parsing fails instead of blocking initialization
- Updated tests to validate the behavior

### DIFF
--- a/Sources/MessagingInApp/Config/MessagingInAppConfigBuilder.swift
+++ b/Sources/MessagingInApp/Config/MessagingInAppConfigBuilder.swift
@@ -72,17 +72,7 @@ public extension MessagingInAppConfigBuilder {
         // This prevents users from having to specify region more than once in the configuration.
         // Therefore, we retrieve the region from top-level configuration here.
         // If the region is not present, the default region is used.
-        let regionStr: String
-        if let regionRawValue = sdkConfig[Keys.region.rawValue] {
-            // This check ensures region is always provided as a string, and throwing an error can help
-            // identify potential bugs when passing configuration values from wrappers.
-            guard let rawValueAsString = regionRawValue as? String else {
-                throw MessagingInAppConfigBuilderError.invalidRegionType
-            }
-            regionStr = rawValueAsString
-        } else {
-            regionStr = ""
-        }
+        let regionStr = sdkConfig[Keys.region.rawValue] as? String ?? ""
         let region = Region.getRegion(from: regionStr)
 
         return MessagingInAppConfigBuilder(siteId: siteId, region: region).build()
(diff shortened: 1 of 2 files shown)


Sources/MessagingInApp/Config/MessagingInAppConfigBuilder.swift
```diff
@@ -72,17 +72,7 @@ public extension MessagingInAppConfigBuilder {
         // This prevents users from having to specify region more than once in the configuration.
         // Therefore, we retrieve the region from top-level configuration here.
         // If the region is not present, the default region is used.
-        let regionStr: String
-        if let regionRawValue = sdkConfig[Keys.region.rawValue] {
-            // This check ensures region is always provided as a string, and throwing an error can help
-            // identify potential bugs when passing configuration values from wrappers.
-            guard let rawValueAsString = regionRawValue as? String else {
-                throw MessagingInAppConfigBuilderError.invalidRegionType
```
mahmoud-elmorabea, line 80: I believe now `MessagingInAppConfigBuilderError.invalidRegionType` is no longer used anywhere, we should remove it

Sources/MessagingInApp/Config/MessagingInAppConfigBuilder.swift
```diff
@@ -72,17 +72,7 @@ public extension MessagingInAppConfigBuilder {
         // This prevents users from having to specify region more than once in the configuration.
```
mahmoud-elmorabea, line 72: I think `MessagingInAppConfigBuilder.build()` is also used in [React Native wrapper](https://github.com/customerio/customerio-reactnative/blob/main/ios/wrappers/CioRctWrapper.swift#L32) and if this function throws an exception (which it does), it won't be caught, so we need to handle the `else ` branch as we did [for Flutter](https://github.com/customerio/customerio-flutter/pull/220).

mahmoud-elmorabea (approved):

mahmoud-elmorabea:
@mrehan27 Can we please make sure to release a Flutter version that includes the changes in this PR

Sources/MessagingInApp/Config/MessagingInAppConfigBuilder.swift
mrehan27: Yup, removed

Sources/MessagingInApp/Config/MessagingInAppConfigBuilder.swift
mrehan27: Sure, will do that will version update PR

mrehan27:
@mahmoud-elmorabea Sure, tested it locally already. I'll open the PR once iOS release is live.

Shahroz16 (approved):

PR #884, opened 2025-04-25
fix: Swizzling removal for APN/FCM Push Notification system

uros-mil:
Changes:
- Added AppDelegates that should replace swizzling code, and are easy to integrate -  CioAppDelegateAPN & CioAppDelegateFCM
  - They are based on the same parent class, CioAppDelegate, which implements most of the functionality
- Old methods and classes are marked `deprecated`, and an explanation is added to direct the user on how to transition to AppDelegates
- Unit tests are added for new AppDelegates
- The old Deep-Linking system is marked `deprecated`, and a new one is added, to:
  - overcome Firebase bug that prevented call to `application(_:continue:restorationHandler:)`, which was necessary for CIO deep-linking to work
  - Reduce complexity and coupling
- Sample apps are updated to use new AppDelegates and Deep-Linking


Package.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version:5.3
+// swift-tools-version:5.9
```
mahmoud-elmorabea: Let's take note of this and call it out in release notes?

Sources/MessagingPush/Integration/AppDelegate.swift
```diff
@@ -15,18 +17,24 @@ open class AppDelegate: AppDelegateType, UNUserNotificationCenterDelegate {
...
         super.init()
     }
     
     open func application(_ application: UIApplication,
                           didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
         let result = wrappedAppDelegate?.application?(application, didFinishLaunchingWithOptions: launchOptions)
+       
+        guard !isConfigInConflict() else {
+            logger.error("CIO: Configuration conflict. Push notifications will not work properly.")
+            return true
```
mahmoud-elmorabea: Do we want to do this check when SDK is being initialized instead?

Package.swift
mahmoud-elmorabea: This seems to have been reverted in a later commit

Sources/MessagingPush/Integration/AppDelegate.swift
uros-mil: To identify conflict, we need to know:
- Exact config used in CustomerIO
- Is our CioAppDelegate used

The option you suggested is technically possible, but much harder to implement, since it will require:
- To detect inside our SDK which AppDelegate is used in the App
- This would produce very complicated code that checks all possible ways to set AppDelegate (UIKit/SwitUI), will require reflection API (similar to one used for swizzling), and may still not work with some SwiftUI cases

On the other hand, the current implementation is in AppDelegate (which is the indicator of being used) and explicitly accesses CustomerIO config for checking the conflict. This makes the code simpler and dependent only on the CIO data we control.

PR #886, opened 2025-04-27
chore: automatic build sample apps on release 

Shahroz16:
Triggers a sample apps public release build after the SDK has been successfully deployed so that internal stakeholders always have up-to-date versions to test.


.github/workflows/deploy-sdk.yml
```diff
@@ -280,3 +291,35 @@ jobs:
...
+  wait-for-cocoapods:
+    name: Wait for deploy-cocoapods to finish
+    needs: [deploy-cocoapods]
+    runs-on: ubuntu-latest
+    if: ${{ success() }}
+    steps:
+      - name: Wait for CocoaPods CDN propagation
+        run: |
+          echo "Waiting for CocoaPods CDN to propagate..."
+          sleep 1200 # Wait for 20 minute to allow CocoaPods CDN to propagate the changes.
```
Shahroz16, line 304: this is the time it has taken previously

.github/workflows/deploy-sdk.yml
```diff
@@ -171,6 +171,8 @@ jobs:
     env:
       COCOAPODS_TRUNK_TOKEN: ${{ secrets.COCOAPODS_TRUNK_TOKEN }}
     runs-on: macos-14
+    outputs:
+      sdk_version: ${{ steps.get-version.outputs.version }}
```
mahmoud-elmorabea, line 175: Do we need to declare that at job level if the output is between steps? I was looking at [this part of the docs](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#setting-an-output-parameter)

.github/workflows/deploy-sdk.yml
mahmoud-elmorabea: Do we want to inject this through an ENV variable instead to make iterations faster and not have to change code?

mahmoud-elmorabea (approved):

.github/workflows/deploy-sdk.yml
Shahroz16: will clean it up

.github/workflows/deploy-sdk.yml
Shahroz16: i thought about this too and chose not to do so, why?

- This will only be triggered after a commit/code change; there is no other way to trigger the workflow from this job. So, if we feel like this time is very less and we need more, we will still need to do a commit sooner or later to trigger this.

PR #887, opened 2025-04-28
chore: Mbl 1072 update logger interface

mahmoud-elmorabea:
Closes: [MBL-1072](https://linear.app/customerio/issue/MBL-1072/implement-unified-logging-interface-on-ios)

## Overview
This PR aligns the Customer IO iOS SDK `Logger` interface to the [unified interfaces definitions](https://www.notion.so/custio/Mobile-SDK-unified-logging-interfaces-1dc4302f4c2b80d08bacd478ca8bbf7e).

The main change in the interface is that logging functions now include a `tag` that will allow us to easily filter logs in the future by module.

Some notable changes in this PR:
- I made sure that the same `Logger` interface was used to avoid having two versions and go through deprecation process
- I've intentionally made `tag` parameter optional for now so we don't have to change any existing code that uses the existing `Logger` interface, I am using extension functions to ensure existing calls don't have to change
- The `Logger Dispatcher` closure was not updated to include the `tag` parameter, I will update it when I am working on updating the wrappers
- I've chosen to add all new parameters to the end of functions like `Logger#debug()`, so that we don't have to change any code of all modules using the interface

## A more testable logger
_This is implemented in the next PR #888_
I've changed how the `LoggerImpl` is initialized:
- `CioLogLevel` is now a required parameter and its usage is an implementation details
- I've wrapped calls to OS loggers in a `SystemLogger` so we can test the `Logger` message format
- Added tests for the logger implementation

## Test plan
I've ran the app both from this branch and the main branch with debug logging enabled and compared the logs between both versions and all log points were there before and after the change.

- You can run the same test yourself and compare with any diff tool
- I am attaching two files with cleaned up logs if you want to compare them without running app on multiple branches

[ios-before.txt](https://github.com/user-attachments/files/19957695/ios-before.txt)
[ios-after.txt](https://github.com/user-attachments/files/19957694/ios-after.txt)

### DIFF
--- a/Sources/Common/Util/Log.swift
+++ b/Sources/Common/Util/Log.swift
@@ -20,14 +20,28 @@ public protocol Logger: AutoMockable {
     func setLogLevel(_ level: CioLogLevel)
     /// the noisey log level. Feel free to spam this log level with any
     /// information about the SDK that would be useful for debugging the SDK.
-    func debug(_ message: String)
+    func debug(_ message: String, _ tag: String?)
     /// Not noisy log messages. Good for general information such as
     /// when the background queue begins and ends running but use `debug`
     /// for the status of each background queue task running.
-    func info(_ message: String)
+    func info(_ message: String, _ tag: String?)
     /// the SDK is in an unstable state that you want to notify
     /// the customer or our development team about.
-    func error(_ message: String)
+    func error(_ message: String, _ tag: String?, _ throwable: Error?)
+}
+
+public extension Logger {
+    func debug(_ message: String) {
+        debug(message, nil)
+    }
+
+    func info(_ message: String) {
+        info(message, nil)
+    }
+
+    func error(_ message: String) {
+        error(message, nil, nil)
+    }
 }
 
 /// none - no logs will be made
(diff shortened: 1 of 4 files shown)


Sources/Common/Util/Log.swift
```diff
@@ -67,67 +81,57 @@ public enum CioLogLevel: String, CaseIterable {
 // log messages to console.
 // sourcery: InjectRegisterShared = "Logger"
 // sourcery: InjectSingleton
-public class ConsoleLogger: Logger {
+public class LoggerImpl: Logger {
```
Shahroz16, line 84: one of the differences I noticed is Swift implementation allows changing the log level after initialization with `setLogLevel()`, while the Kotlin implementation doesn't, intentional?

Shahroz16 (commented):
its going to be similar as android, testing with wrappers SDKs

Sources/Common/Util/Log.swift
mahmoud-elmorabea: Yes, this is blocked by the issue I brought up [here](https://customerio.slack.com/archives/C06H7MMLH6C/p1745860849423879?thread_ts=1745859241.696979&cid=C06H7MMLH6C)

It doesn't seem like we can have similar setup with Android with our current DI framework:
- We either keep it as a setter for the logger
- Or create an extension for DI object but then every call to get the logger will have to provide the log level

mrehan27 (approved):

Shahroz16 (approved):
Seems okay, given we have veririfed changes with wrappers.

PR #891, opened 2025-05-02
chore: Swizzling removal - iOS SDK (w/o FCM deep-linking fix and sample app)

uros-mil:
Changes:
- Added AppDelegates that should replace swizzling code, and are easy to integrate -  CioAppDelegateAPN & CioAppDelegateFCM
  - They are based on the same parent class, CioAppDelegate, which implements most of the functionality
- Old methods and classes are marked `deprecated`, and an explanation is added to direct the user on how to transition to AppDelegates
- Unit tests are added for new AppDelegates


README.md
```diff
@@ -15,7 +15,7 @@ This is the official Customer.io SDK for iOS.
 
 You'll find our [complete SDK documentation at https://customer.io/docs/sdk/ios](https://customer.io/docs/sdk/ios/). 
 
-**The SDK has been tested on iOS devices**. It might work on other Apple devices—macOS, tvOS, and watchOS—but we have not officially tested, nor do we officially support, non-iOS devices.
```
Ahmed-Ali, line 18: nit: is this intentional change?

Sources/MessagingPush/Integration/CioAppDelegateWithoutTokenRetrieval.swift
```diff
@@ -0,0 +1,171 @@
...
+
+private extension UIApplication {
+    func cioRegisterForRemoteNotifications(logger: Logger) {
+        logger.debug("CIO: Registering for remote notifications")
+        registerForRemoteNotifications()
+    }
+}
+
+@available(iOSApplicationExtension, unavailable)
+open class CioAppDelegateWithoutTokenRetrieval: CioAppDelegateType, UNUserNotificationCenterDelegate {
```
Ahmed-Ali, line 25: what is the reason for making the class name call out`WithoutTokenRetrieval`?
This seems like a configureable behavior. Which confuses me why it is called out in the type name

Sources/MessagingPush/Config/MessagingPushConfigBuilder.swift
```diff
@@ -96,6 +96,7 @@ public class MessagingPushConfigBuilder {
 
     /// Display push notifications sent by Customer.io while app is in foreground. This value is `true` by default.
     @discardableResult
+    @available(*, deprecated, message: "When CioAppDelegate(Wrapper) is used, this method is not needed")
```
mahmoud-elmorabea, line 99: Maybe clarify this as "This shouldn't be set if you're using ioAppDelegate(Wrapper), if you are using ioAppDelegate(Wrapper) yet, it's recommended you upgrade".. somethings along those lines.

Sources/MessagingPush/Integration/CioAppDelegateWithoutTokenRetrieval.swift
```diff
@@ -0,0 +1,171 @@
...
+        if config?().autoTrackPushEvents ?? false,
+           var center = userNotificationCenter?() {
+            wrappedNoticeCenterDelegate = center.delegate
+            center.delegate = self
+        }
+
+        return result ?? true
+    }
+
+    open func application(
```
Ahmed-Ali, line 93: did we forget to call messagingPush?

Sources/MessagingPush/MessagingPush.swift
```diff
@@ -6,6 +6,8 @@ import Foundation
  So, performing an HTTP request to the API with a device token goes here.
   */
 public class MessagingPush: ModuleTopLevelObject<MessagingPushInstance>, MessagingPushInstance {
+    public static var appDelegateIntegratedExplicitely: Bool = false
```
mahmoud-elmorabea, line 9: Guess that has to be public? Maybe we can restrict it to be only a setter?

Sources/MessagingPush/Integration/CioAppDelegateWithoutTokenRetrieval.swift
```diff
@@ -0,0 +1,171 @@
...
+        #selector(UNUserNotificationCenterDelegate.userNotificationCenter(_:willPresent:withCompletionHandler:)),
+        #selector(UNUserNotificationCenterDelegate.userNotificationCenter(_:didReceive:withCompletionHandler:)),
+        #selector(UNUserNotificationCenterDelegate.userNotificationCenter(_:openSettingsFor:))
+    ]
+
+    @_spi(Internal) public var config: ConfigInstance?
+
+    private var userNotificationCenter: UserNotificationCenterInstance?
+    private let wrappedAppDelegate: UIApplicationDelegate?
+    private var wrappedNoticeCenterDelegate: UNUserNotificationCenterDelegate?
```
Ahmed-Ali, line 44: nit: the naming `notice` is not easily recognizable and mismatches the actual name

Sources/MessagingPush/Integration/CioAppDelegateWithoutTokenRetrieval.swift
Ahmed-Ali: As this one is being subclasses internally, shouldn't we annotate it with `@_spi`?

mahmoud-elmorabea (approved):
Looks good! 💯 

_We can prefix the PR with `chore` and it won't kick off a release_

Ahmed-Ali (approved):
Thanks for splitting out the changes, Uros!

I have few comments. I won't block the PR but let's address the none-`nit` before merging

Sources/MessagingPush/Integration/CioAppDelegateWithoutTokenRetrieval.swift
```diff
@@ -0,0 +1,171 @@
...
+        DIGraphShared.shared.logger.error("CIO: This no-argument initializer should not to be used. Added since UIKit's AppDelegate initialization process crashes if for no-arg init is missing.")
+        self.init(
+            messagingPush: MessagingPush.shared,
+            userNotificationCenter: { UNUserNotificationCenter.current() },
+            appDelegate: nil,
+            config: nil,
+            logger: DIGraphShared.shared.logger
+        )
+    }
+
```
Ahmed-Ali, line 56: Line coverage isn't the most indicative of the test coverage. But lack of it indicates potential missed code branches. Let's address before merging

Sources/MessagingPush/Config/MessagingPushConfigBuilder.swift
uros-mil: Makes sense. It's updated.

Sources/MessagingPush/MessagingPush.swift
uros-mil: Fixed

Sources/MessagingPush/Integration/CioAppDelegateWithoutTokenRetrieval.swift
uros-mil: 1. The name is updated to CioProviderAgnosticAppDelegate to better explain the purpose
2. It's not possible to add `@_spi` as open classes are extending it

Sources/MessagingPush/Integration/CioAppDelegateWithoutTokenRetrieval.swift
uros-mil: No, since we have separate API for APN and FCM, those calls are in child classes.

README.md
uros-mil: Yes.
Our Package.swift has iOS as the only supported platform, and it's expected that UIKit is available in many places.
In the current code state, it is not enough to just add macOS as a supported platform; we'll need to make the code changes to support it properly.
I intended to remove the incorrect statement.

It would make sense to list necessary changes for extending support to macOS.

Sources/MessagingPush/Integration/CioAppDelegateWithoutTokenRetrieval.swift
uros-mil: Code coverage is almost 100% in reality.
We have lower numbers because:
- There are `init` methods which are not intended for use, but are necessary to satisfy UIApplication.delegate requirements for a no-arg constructor, and those are not tested
- Tests are not written for Wrappers, as those only forward data to AppDelegates

PR #892, opened 2025-05-02
chore: Swizzling removal - iOS SDK (FCM deep-linking fix)

uros-mil:
- The old Deep-Linking system is marked `deprecated`, and a new one is added, to:
  - overcome Firebase bug that prevented call to `application(_:continue:restorationHandler:)`, which was necessary for CIO deep-linking to work
  - Reduce complexity and coupling


mahmoud-elmorabea (approved):
Looks good! 💯 

Thanks for breaking the change into a smaller chunk 👍

Sources/Common/Util/DeepLinkUtil.swift
```diff
@@ -1,20 +1,29 @@
...
+    private var deepLinkCallback: DeepLinkCallback?
 
     init(logger: Logger, uiKitWrapper: UIKitWrapper) {
         self.logger = logger
         self.uiKit = uiKitWrapper
     }
 
+    @_spi(Internal) public func setDeepLinkCallback(_ callback: @escaping DeepLinkCallback) {
+        deepLinkCallback = callback
+    }
```
Ahmed-Ali, line 25: Since it is a public setter, we need to protect the access of the `deepLinkCallback` property.

Ahmed-Ali: We also need to add few test cases for the scenarios where this gets set and used

Tests/DataPipeline/Config/DataPipelineConfigOptionsTests.swift
```diff
@@ -63,14 +65,15 @@ class DataPipelineConfigOptionsTests: UnitTest {
...
-        let (_, dataPipelineConfig) = SDKConfigBuilder(cdpApiKey: givenCdpApiKey)
+        let result = SDKConfigBuilder(cdpApiKey: givenCdpApiKey)
             .apiHost(givenApiHost)
             .cdnHost(givenCdnHost)
             .flushAt(givenFlushAt)
             .flushInterval(givenFlushInterval)
             .flushPolicies(givenFlushPolicies)
             .trackApplicationLifecycleEvents(givenTrackApplicationLifecycleEvents)
             .build()
+        let dataPipelineConfig = result.dataPipelineConfig
```
Ahmed-Ali, line 76: where is that being used in the test?

Ahmed-Ali (changes requested):
Overall the code change looks fine.
But I didn't see any tests for the `DeepLinkUtil.swift` and from the Codecov hint about line coverage, there is likely a need to revise the existing tests to make sure all the code branches are covered

Sources/Common/Util/DeepLinkUtil.swift
uros-mil: I've added those after checking CodeCov in a bit more detail.

Tests/DataPipeline/Config/DataPipelineConfigOptionsTests.swift
uros-mil: One line below it. 
Split to avoid long chaining `result.dataPipelineConfig.toSegmentConfiguration().values`.

Sources/Common/Util/DeepLinkUtil.swift
Ahmed-Ali: the one remaining comment is the thread safety

PR #893, opened 2025-05-02
chore: Swizzling removal - iOS SDK (Sample apps)

uros-mil:
Changes:
- Sample apps are updated to use new AppDelegates and Deep-Linking

Testing:
- the easiest way to get a device-token is to copy from Xcode console
- Basic PN: 
  - Triggering:
- PN with Deep-link: 
  - Triggering:


Ahmed-Ali (approved):
All looks good.
One request: let's make sure the PR includes a test plan. i.e what steps do I need to follow in order to reproduce the scenario(s) you tested to verify this code does what it is expected to do?

PR #894, opened 2025-05-02
chore: Align push initilization logs with definitions

mahmoud-elmorabea:
TBA

### DIFF
--- a/Sources/Common/Module/ModuleTopLevelObject.swift
+++ b/Sources/Common/Module/ModuleTopLevelObject.swift
@@ -57,8 +57,10 @@ open class ModuleTopLevelObject<ImplementationClass> {
             return
         }
 
-        logger.debug("Setting up \(moduleName) module...")
+        let initLogger = DIGraphShared.shared.sdkInitializationLogger
+
+        initLogger.moduleInitStart(moduleName)
         _implementation = blockToInitializeModule()
-        logger.info("\(moduleName) module successfully set up with SDK")
+        initLogger.moduleInitSuccess(moduleName)
     }
 }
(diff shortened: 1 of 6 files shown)


Sources/Common/Module/ModuleTopLevelObject.swift
```diff
@@ -57,8 +57,10 @@ open class ModuleTopLevelObject<ImplementationClass> {
             return
         }
 
-        logger.debug("Setting up \(moduleName) module...")
+        let initLogger = DIGraphShared.shared.sdkInitializationLogger
+
+        initLogger.moduleInitStart(moduleName)
         _implementation = blockToInitializeModule()
-        logger.info("\(moduleName) module successfully set up with SDK")
+        initLogger.moduleInitSuccess(moduleName)
```
mahmoud-elmorabea, line 64: `ModuleTopLevelObject` doesn't have module config, adding this would be a bigger chance, I can probably do that refactoring in a later PR to align with Android

mrehan27 (commented):
Same questions as on Android. Overall, the changes look good. I have two main questions:

- How are we using the updates from mobile log definitions in the implementation?
- Have we considered using constants/enums instead of defining a separate function for each log? If so, were there any challenges with that approach?

Bringing this up on slack thread as well for quicker feedback and discussion.

Shahroz16 (approved):
Looks good, similar comments as https://github.com/customerio/customerio-android/pull/530#pullrequestreview-2817745837

and as discussed next steps:
- Add tests ensuring SDK init call the `InitLogger` methods
- Follow up project targeting more of the unified logger requirements and next steps.

mrehan27 (approved):
Looks good, assuming we proceed with the next steps we discussed on call as Shahroz mentioned.

PR #896, opened 2025-05-12
ci: use app credentials for semantic release

Ahmed-Ali:
This pr updates the semantic release action so that it uses app credentials instead of GITHUB_TOKEN.
By using an app, we can setup the regular branch protection and ruleset while allowing the semantic release action to bypass them.

## Test plan
I tested the approach on the RN repo [#431](https://github.com/customerio/customerio-reactnative/pull/431)


uros-mil (approved):
Thakns for doing this.

PR #897, opened 2025-05-13
feat: Swizzling removal - Compatibility improvements for `showPushAppInForeground`

uros-mil:
## Changes
There was incompatibility between the old default behaviour for `showPushAppInForeground` and the new no-swizzling approach:
- The old system had `showPushAppInForeground` set to `true` as default, so if this property is not set, notifications will be shown in the foreground.
- The new system was originally implemented to be compatible with iOS' default behavior, which meant not implementing `UNUserNotificationCenterDelegate` would NOT show notifications in the foreground.

This PR changes the behavior of the new system to be backward compatible, so not implementing the `UNUserNotificationCenterDelegate` in the customer's AppDelegate will show notifications based on `showPushAppInForeground`.
 If customers want fine-grained control, they should implement `UNUserNotificationCenterDelegate`.
 
### Unit Tests:
- extended tests to validate behaviour changes with `userNotificationCenter(_:willPresent:withCompletionHandler:)` method

### Test plan:
- Manual testing could be done with sample apps
- Those have commented out the implementation for `UNUserNotificationCenterDelegate` protocol
- If you want to test custom implementation for `UNUserNotificationCenterDelegate`, un-comment the code.
  - This will give priority to implementation in the app's AppDelegate, and the value of `showPushAppInForeground` is not taken into consideration

### DIFF
--- a/Sources/MessagingPush/Integration/CioProviderAgnosticAppDelegate.swift
+++ b/Sources/MessagingPush/Integration/CioProviderAgnosticAppDelegate.swift
@@ -120,10 +120,15 @@ open class CioProviderAgnosticAppDelegate: CioAppDelegateType, UNUserNotificatio
            wrappedNotificationCenterDelegate.responds(to: #selector(UNUserNotificationCenterDelegate.userNotificationCenter(_:willPresent:withCompletionHandler:))) {
             wrappedNotificationCenterDelegate.userNotificationCenter?(center, willPresent: notification, withCompletionHandler: completionHandler)
         } else {
-            if #available(iOS 14.0, *) {
-                completionHandler([.list, .banner, .badge, .sound])
+            if config?().showPushAppInForeground ?? false {
+                if #available(iOS 14.0, *) {
+                    completionHandler([.list, .banner, .badge, .sound])
+                } else {
+                    completionHandler([.alert, .badge, .sound])
+                }
             } else {
-                completionHandler([.alert, .badge, .sound])
+                // Don't show the notification in the foreground
+                completionHandler([])
             }
         }
     }
--- a/Sources/MessagingPush/Integration/CioProviderAgnosticAppDelegate.swift
+++ b/Sources/MessagingPush/Integration/CioProviderAgnosticAppDelegate.swift
@@ -111,6 +111,23 @@ open class CioProviderAgnosticAppDelegate: CioAppDelegateType, UNUserNotificatio
 
     // MARK: - UNUserNotificationCenterDelegate
 
+    open func userNotificationCenter(
+        _ center: UNUserNotificationCenter,
+        willPresent notification: UNNotification,
+        withCompletionHandler completionHandler: @escaping (UNNotificationPresentationOptions) -> Void
+    ) {
+        if let wrappedNotificationCenterDelegate = wrappedNotificationCenterDelegate,
+           wrappedNotificationCenterDelegate.responds(to: #selector(UNUserNotificationCenterDelegate.userNotificationCenter(_:didReceive:withCompletionHandler:))) {
+            wrappedNotificationCenterDelegate.userNotificationCenter?(center, willPresent: notification, withCompletionHandler: completionHandler)
+        } else {
+            if #available(iOS 14.0, *) {
+                completionHandler([.list, .banner, .badge, .sound])
+            } else {
+                completionHandler([.alert, .badge, .sound])
+            }
+        }
+    }
+
     // Function called when a push notification is clicked or swiped away.
     open func userNotificationCenter(
         _ center: UNUserNotificationCenter,
@@ -119,9 +136,9 @@ open class CioProviderAgnosticAppDelegate: CioAppDelegateType, UNUserNotificatio
     ) {
         _ = messagingPush.userNotificationCenter(center, didReceive: response)
 
-        if let wrappedNoticeCenterDelegate = wrappedNotificationCenterDelegate,
-           wrappedNoticeCenterDelegate.responds(to: #selector(UNUserNotificationCenterDelegate.userNotificationCenter(_:didReceive:withCompletionHandler:))) {
-            wrappedNoticeCenterDelegate.userNotificationCenter?(center, didReceive: response, withCompletionHandler: completionHandler)
+        if let wrappedNotificationCenterDelegate = wrappedNotificationCenterDelegate,
+           wrappedNotificationCenterDelegate.responds(to: #selector(UNUserNotificationCenterDelegate.userNotificationCenter(_:didReceive:withCompletionHandler:))) {
+            wrappedNotificationCenterDelegate.userNotificationCenter?(center, didReceive: response, withCompletionHandler: completionHandler)
         } else {
             completionHandler()
         }
@@ -156,10 +173,6 @@ open class CioProviderAgnosticAppDelegate: CioAppDelegateType, UNUserNotificatio
 /// - for this reason, empty methods are added and forwarding to wrapper is possible
 @available(iOSApplicationExtension, unavailable)
 extension CioProviderAgnosticAppDelegate {
-    open func userNotificationCenter(_ center: UNUserNotificationCenter, willPresent notification: UNNotification, withCompletionHandler completionHandler: @escaping (UNNotificationPresentationOptions) -> Void) {
-        wrappedNotificationCenterDelegate?.userNotificationCenter?(center, willPresent: notification, withCompletionHandler: completionHandler)
-    }
-
     open func userNotificationCenter(_ center: UNUserNotificationCenter, openSettingsFor notification: UNNotification?) {
         wrappedNotificationCenterDelegate?.userNotificationCenter?(center, openSettingsFor: notification)
     }
--- a/Sources/MessagingPush/Integration/CioProviderAgnosticAppDelegate.swift
+++ b/Sources/MessagingPush/Integration/CioProviderAgnosticAppDelegate.swift
@@ -117,7 +117,7 @@ open class CioProviderAgnosticAppDelegate: CioAppDelegateType, UNUserNotificatio
         withCompletionHandler completionHandler: @escaping (UNNotificationPresentationOptions) -> Void
     ) {
         if let wrappedNotificationCenterDelegate = wrappedNotificationCenterDelegate,
-           wrappedNotificationCenterDelegate.responds(to: #selector(UNUserNotificationCenterDelegate.userNotificationCenter(_:didReceive:withCompletionHandler:))) {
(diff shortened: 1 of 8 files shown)


Ahmed-Ali (changes requested):
Why there are no tested impacted by this change?
Specially with the mistaken method call CoPilot noted, it was present as a bug an no existing test caught it?

Can you also please the PR description includes a test plan?

uros-mil:
> Why there are no tested impacted by this change? Specially with the mistaken method call CoPilot noted, it was present as a bug an no existing test caught it?
> 
> Can you also please the PR description includes a test plan?

I started looking into the tests, but had an issue with Firebase Messaging Mock.
Then submitted a PR so the incompatibility issue is visible and I could get feedback on this.

I'm working on tests and a Firebase issue. Once this is done, I'll publish the PR

Good example of the value of Copilot.

uros-mil:
> ## [Codecov](https://app.codecov.io/gh/customerio/customerio-ios/pull/897?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=customerio) Report
> Attention: Patch coverage is `93.33333%` with `1 line` in your changes missing coverage. Please review.
> 
> > Project coverage is 60.78%. Comparing base [(`fb6d377`)](https://app.codecov.io/gh/customerio/customerio-ios/commit/fb6d377e67c0e27d21b4a0ac0ad65c462b9f0f13?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=customerio) to head [(`45158fb`)](https://app.codecov.io/gh/customerio/customerio-ios/commit/45158fba7b2f88222184586ce6ad56b7f73fe111?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=customerio).
> > Report is 2 commits behind head on main.
> 
> [Files with missing lines](https://app.codecov.io/gh/customerio/customerio-ios/pull/897?dropdown=coverage&src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=customerio)	Patch %	Lines
> [...h/Integration/CioProviderAgnosticAppDelegate.swift](https://app.codecov.io/gh/customerio/customerio-ios/pull/897?src=pr&el=tree&filepath=Sources%2FMessagingPush%2FIntegration%2FCioProviderAgnosticAppDelegate.swift&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=customerio#diff-U291cmNlcy9NZXNzYWdpbmdQdXNoL0ludGVncmF0aW9uL0Npb1Byb3ZpZGVyQWdub3N0aWNBcHBEZWxlZ2F0ZS5zd2lmdA==)	93.33%	[1 Missing ⚠️ ](https://app.codecov.io/gh/customerio/customerio-ios/pull/897?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=customerio)
> Additional details and impacted files
> [☔ View full report in Codecov by Sentry](https://app.codecov.io/gh/customerio/customerio-ios/pull/897?dropdown=coverage&src=pr&el=continue&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=customerio). 📢 Have feedback on the report? [Share it here](https://about.codecov.io/codecov-pr-comment-feedback/?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=customerio).
> 
> 🚀 New features to boost your workflow:
> * ❄️ [Test Analytics](https://docs.codecov.com/docs/test-analytics): Detect flaky tests, report on failures, and find test suite problems.
> * 📦 [JS Bundle Analysis](https://docs.codecov.com/docs/javascript-bundle-analysis): Save yourself from yourself by tracking and limiting bundle sizes in JS merges.

The line that is missing coverage is an acceptable trade-off, since it's a `UNNotificationPresentationOptions` value `[.alert, .badge, .sound]` set for pre-iOS 14 versions.
The chance of error is very small, and additional infrastructure to fake an iOS version is very complex.

Sources/MessagingPush/Integration/CioProviderAgnosticAppDelegate.swift
```diff
@@ -111,6 +111,23 @@ open class CioProviderAgnosticAppDelegate: CioAppDelegateType, UNUserNotificatio
...
+    ) {
+        if let wrappedNotificationCenterDelegate = wrappedNotificationCenterDelegate,
+           wrappedNotificationCenterDelegate.responds(to: #selector(UNUserNotificationCenterDelegate.userNotificationCenter(_:willPresent:withCompletionHandler:))) {
+            wrappedNotificationCenterDelegate.userNotificationCenter?(center, willPresent: notification, withCompletionHandler: completionHandler)
+        } else {
+            if #available(iOS 14.0, *) {
+                completionHandler([.list, .banner, .badge, .sound])
+            } else {
+                completionHandler([.alert, .badge, .sound])
+            }
```
Shahroz16, line 127:
```suggestion
            if let configOptions = config?(), let showPushAppInForeground = configOptions.showPushAppInForeground, !showPushAppInForeground {
                // Don't show notifications if explicitly disabled
                completionHandler([])
            } else {
                completionHandler([.alert, .badge, .sound])
                // Show notification with appropriate presentation options
                logger.debug("CIO: Using default notification presentation (showing in foreground) because app delegate doesn't implement willPresent")
                if #available(iOS 14.0, *) {
                    completionHandler([.list, .banner, .badge, .sound])
                } else {
                    completionHandler([.alert, .badge, .sound])
                }
            }
```

shouldn't it be like this?

uros-mil: This is an interesting suggestion.
I started with an idea to deprecate `showPushAppInForeground` as its behaviour was different from standard iOS, and it had a problematic implementation in the swizzling version of notification handling.

But, now that priority is given to backward-compatibility, it does make sense to reuse `showPushAppInForeground` instead of deprecating it.

I'll investigate whether there are any limitations to doing this. If it works, I'll update the code.

uros-mil: Fixed.
Thanks for pointing this out. :)

Sources/MessagingPush/Integration/CioProviderAgnosticAppDelegate.swift
```diff
@@ -111,6 +111,28 @@ open class CioProviderAgnosticAppDelegate: CioAppDelegateType, UNUserNotificatio
...
+    open func userNotificationCenter(
+        _ center: UNUserNotificationCenter,
+        willPresent notification: UNNotification,
+        withCompletionHandler completionHandler: @escaping (UNNotificationPresentationOptions) -> Void
+    ) {
+        if let wrappedNotificationCenterDelegate = wrappedNotificationCenterDelegate,
+           wrappedNotificationCenterDelegate.responds(to: #selector(UNUserNotificationCenterDelegate.userNotificationCenter(_:willPresent:withCompletionHandler:))) {
+            wrappedNotificationCenterDelegate.userNotificationCenter?(center, willPresent: notification, withCompletionHandler: completionHandler)
+        } else {
+            if config?().showPushAppInForeground ?? false {
```
mahmoud-elmorabea, line 123: I guess here we are depending on the default defined in the `MessagingPushConfigBuilder` to be true, right?

Tests/MessagingPush/Integration/CioProviderAgnosticAppDelegateTests.swift
```diff
@@ -155,7 +159,86 @@ class CioProviderAgnosticAppDelegateTests: XCTestCase {
...
+
+        // Call the method
+        appDelegate.userNotificationCenter(UNUserNotificationCenter.current(), willPresent: UNNotification.testInstance, withCompletionHandler: completionHandler)
+
+        // Verify behavior
+        XCTAssertTrue(mockNotificationCenterDelegate.willPresentNotificationCalled)
+        XCTAssertTrue(completionHandlerCalled)
+    }
+
+    // New test for when wrapped delegate doesn't implement willPresent
```
mahmoud-elmorabea, line 179: We probably don't need this comment, the test name is sufficient

mahmoud-elmorabea (approved):
Loos good, great work! 💯 

I've tested the change with the following scenarios:
### App downloaded from Firebase
- This app implements `userNotificationCenter willPresent` so it simulates a customer overriding this handling
- The notification is shown in the foreground regardless of the `showPushAppInForeground` value, since the `userNotificationCenter` is overriden

### Locally installed app
- I've commented out `userNotificationCenter willPresent`
- The app respects `showPushAppInForeground` config and only shows notification in foreground when `showPushAppInForeground` is true

Maybe we can add these test steps to the PR description to help future us test these changes as I was initially confused why our sample app always showed push in the foregound

Sources/MessagingPush/Integration/CioProviderAgnosticAppDelegate.swift
uros-mil: Yes, the default value is true.

Shahroz16 (approved):
Awesome, thanks for making the changes. Can we also add a test case for this? Something to test `UserNotificationCenterWillPresent` when ConfigOption disables notifications, so we confirm no notification is shown? do you think it would be a valuable addition or something thats already covered?

uros-mil:
> Awesome, thanks for making the changes. Can we also add a test case for this? Something to test `UserNotificationCenterWillPresent` when ConfigOption disables notifications, so we confirm no notification is shown? do you think it would be a valuable addition or something thats already covered?

We do have a test when `showPushAppInForeground` is disabled: 
[CioProviderAgnosticAppDelegateTests.swift](https://github.com/customerio/customerio-ios/blob/24abfd97175e0249fa6c051b5887266285575368/Tests/MessagingPush/Integration/CioProviderAgnosticAppDelegateTests.swift#L210)

uros-mil:
> Loos good, great work! 💯
> 
> I've tested the change with the following scenarios:
> 
> ### App downloaded from Firebase
> * This app implements `userNotificationCenter willPresent` so it simulates a customer overriding this handling
> * The notification is shown in the foreground regardless of the `showPushAppInForeground` value, since the `userNotificationCenter` is overriden
> 
> ### Locally installed app
> * I've commented out `userNotificationCenter willPresent`
> * The app respects `showPushAppInForeground` config and only shows notification in foreground when `showPushAppInForeground` is true
> 
> Maybe we can add these test steps to the PR description to help future us test these changes as I was initially confused why our sample app always showed push in the foregound

Comment regarding `App downloaded from Firebase` is a valid one.
For this reason, I'll comment out the apps' implementation for `UNUserNotificationCenterDelegate` and add comments explaining when to uncomment it.

Ahmed-Ali (approved):

PR #916, opened 2025-05-15
chore: Updating push logs for delivery and click handling

mahmoud-elmorabea:
Part of: [MBL-1074](https://linear.app/customerio/issue/MBL-1074/update-push-notification-logging-for-ios)
Closes: [MBL-1114](https://linear.app/customerio/issue/MBL-1114/update-push-notification-click-handling-logs)

### Overview
This PR implements log definitions as defined in the log definitions repo https://github.com/customerio/mobile-log-definitions/pull/9

- Adds logs for push delivery for NSE while handling rich push notifications
- Adds logs for push click handling and open push tracking

### Test
- Added unit test for the wrapper logger and call sites
- I've tested that logs are shown correctly for both new and old push setup systems
- Logs after implementation are as follows
```
NotificationServiceExtension	[Push] Received notification for message: {{notification}}
NotificationServiceExtension	[Push] Received CIO push message
NotificationServiceExtension	[Push] Tracking push message delivered with deliveryId: dgS_ugcABQAAAZbT3TGafSx9omFqv9TAcg==

APN UIKit	[Push] Clicked CIO push message
APN UIKit	[Push] Tracking push message opened with deliveryId: dgS_ugcABQAAAZbT3TGafSx9omFqv9TAcg==
```


mahmoud-elmorabea:
I am going to follow up with a PR that adds tests for call sites of the logger, I need to explore how to mock OS types to be able to unlock this

Sources/MessagingPush/PushHandling/ManualPushHandling+UserNotifications.swift
```diff
@@ -78,11 +78,15 @@ extension MessagingPushImplementation {
...
 
         guard push.isPushSentFromCio else {
+            pushLogger.logClickedNonCioPushMessage()
             return nil
         }
 
+        pushLogger.logClickedCioPushMessage()
+
         if response.didClickOnPush {
             manualPushClickHandling(push: push)
```
uros-mil, line 91: Let's add log here as well, since this is indicator that we are going to process the message.

Sources/MessagingPush/PushHandling/iOSPushEventListener.swift
```diff
@@ -30,7 +32,7 @@ class IOSPushEventListener: PushEventHandler {
             return
         }
         let push = pushAction.push
-        logger.debug("On push action event. push action: \(pushAction))")
+        pushLogger.logClickedPushMessage(notification: push)
```
uros-mil, line 35: Is there need for this type of log at all?
We have 2 options, cio and non-cio click, which are already logged below.
This one produces extra log line and logs push, and in second one we see is it cio or not.
Would it be bette to only have cio and non-cio which will hold the push data - complete info is in one log.

Sources/MessagingPush/MessagingPushImplementation.swift
```diff
@@ -12,6 +13,7 @@ class MessagingPushImplementation: MessagingPushInstance {
         self.logger = diGraph.logger
         self.jsonAdapter = diGraph.jsonAdapter
         self.eventBusHandler = diGraph.eventBusHandler
+        self.pushLogger = diGraph.pushNotificationLogger
```
uros-mil, line 16: Why do we expose global reference to PushNotificationLogger?
Shouldn't this be visible inside Push module?

uros-mil (changes requested):
Looks good. 
Let's clarify couple of pint mentioned?

Sources/MessagingPush/MessagingPushImplementation.swift
mahmoud-elmorabea: I basically just followed the the other members of this class, what exactly do you mean by exposing these references since `MessagingPushImplementation` is not public as far as I can see?

Sources/MessagingPush/PushHandling/ManualPushHandling+UserNotifications.swift
mahmoud-elmorabea: The following [PR](https://github.com/customerio/customerio-ios/pull/917/files#diff-dfdc330b2ab4df1a2983e50045a6a6c29ce35e274532ba3eec5483b01c15056b) adds logs inside the push click handler that covers those.

Sources/MessagingPush/PushHandling/iOSPushEventListener.swift
mahmoud-elmorabea: After reconsidering this after our chat, I would rather keep this separate so that if the logic after this is broken for some reason in the future, this log would log all the info we need to understand why it didn't behave as we expect.

I don't think it makes much of a difference but if you feel strongly about it, I can create another PR that merges these logs.

Sources/MessagingPush/MessagingPushImplementation.swift
uros-mil: I was referring to pushNotificationLogger being exposed through diGraph.
But let's tackle this in a separate initiative for better DI.

uros-mil (approved):

Sources/MessagingPush/PushHandling/PushClickHandler.swift
```diff
@@ -16,17 +16,20 @@ protocol PushClickHandler: AutoMockable {
...
-    init(deepLinkUtil: DeepLinkUtil, messagingPush: MessagingPushInstance) {
+    init(deepLinkUtil: DeepLinkUtil, messagingPush: MessagingPushInstance, pushLogger: PushNotificationLogger) {
         self.deepLinkUtil = deepLinkUtil
         self.messagingPush = messagingPush
+        self.pushLogger = pushLogger
     }
 
     func trackPushMetrics(for push: PushNotification) {
         guard let cioDelivery = push.cioDelivery else {
+            pushLogger.logClickedPushMessageWithEmptyDeliveryId()
```
Shahroz16, line 29: just a callout: there is a warning that says this line is not being tested, are we tackling it in any other PR or any other reason?

PR #917, opened 2025-05-16
chore: Update logs for push deep linking

mahmoud-elmorabea:
Part of: [MBL-1074](https://linear.app/customerio/issue/MBL-1074/update-push-notification-logging-for-ios)
Closes: [MBL-1116](https://linear.app/customerio/issue/MBL-1116/update-push-notification-deep-linking-handling-logs)

### Overview
This PR implements log definitions as defined in the log definitions repo https://github.com/customerio/mobile-log-definitions/pull/10

- Adds logs for deep linking util to indicate how a deep link was handled

### Test
- Added unit test for the wrapper logger and call sites
- I've tested that logs are shown correctly for both new and old push setup systems
- Logs after implementation are as follows
```
NotificationServiceExtension	[Push] Received notification for message: {{notification}}
NotificationServiceExtension	[Push] Received CIO push message
NotificationServiceExtension	[Push] Tracking push message delivered with deliveryId: dgS_ugcABQAAAZbT3TGafSx9omFqv9TAcg==

APN UIKit	[Push] Clicked CIO push message
APN UIKit	[Push] Tracking push message opened with deliveryId: dgS_ugcABQAAAZbT3TGafSx9omFqv9TAcg==

[Push] Handling push notification deep link with url: apn-uikit:settings
[Push] Deep link handled by system
```


Sources/MessagingPush/PushHandling/PushClickHandler.swift
```diff
@@ -36,6 +38,8 @@ class PushClickHandlerImpl: PushClickHandler {
     func handleDeepLink(for push: PushNotification) {
         if let deepLinkUrl = push.cioDeepLink?.url {
```
uros-mil, line 39: Let's reverse this with `guard let` at the top, as this will remove the need for nesting.

Sources/Common/Util/SdkCommonLoggerLogger.swift
```diff
@@ -0,0 +1,62 @@
...
+        self.logger = logger
+    }
+
+    public func coreSdkInitStart() {
+        logger.debug("Creating new instance of CustomerIO SDK version: \(SdkVersion.version)...", Self.INIT_TAG)
+    }
+
+    public func coreSdkInitSuccess() {
+        logger.info("CustomerIO SDK is initialized and ready to use", Self.INIT_TAG)
+    }
```
uros-mil, line 33: Those 2 methods exist in Common, but are only used by DataPipeline module.
Would it make more sense to have them in DataPipeline's logger and avoid exposing `SdkCommonLogger` outside the module?

uros-mil (changes requested):

Sources/Common/Util/SdkCommonLoggerLogger.swift
mahmoud-elmorabea: I wanted to keep them with the init logs for modules, ideally they all should live in the DataPipeline module but for some reason we have the top level config object living in the common module which is a change I don't want to make in the scope of this PR

uros-mil (approved):

PR #921, opened 2025-05-21
fix: Refactor APN/FCM device token handling to prevent duplicate calls to MessagingPush.registerDeviceToken

uros-mil:
Task: [MBL-1159](https://linear.app/customerio/issue/MBL-1159/bug-in-no-swizzling-interaction-with-messagingpushapn)

## Summary
  - Fixed issue with duplicate call to internal API (no duplicate calls happen to backend, since we are already checking if the token changed)
  - Refactored push notification device token handling by extracting swizzling logic into a dedicated protocol/class
  implementations
  - Fixed typo in variable name `appDelegateIntegratedExplicitely` → `appDelegateIntegratedExplicitly`
  - Added comprehensive test coverage for MessagingPushAPN and MessagingPushFCM
      - Other parts of the module have not been extended with tests - not the scope of this PR
  - Improved code organization by moving dependency providers to dedicated components
  
IMPORTANT:
- While analyzing the swizzling implementation for `autoFetchDeviceToken`, I noticed that the swizzled methods (didRegisterForRemoteNotificationsWithDeviceToken/didFailToRegisterForRemoteNotificationsWithError) in the customer's AppDelegate were NOT called. This functionality doesn't exist at all in the code.
- This explains issues I've noticed, that those methods are never called in our ReactNative example app, even though they are put there to guide the customer. After we switch the Wrapper SDK to a no-swizzling system, this will work fine (already verified).

## Test plan
  - Manually tested app behaviour by tracking calls in the debugger
  - Verified that it works well by sending PNs from Fly (as explained below in the setup)

>  
> 🔥 As this is an important part of our token fetching system, please, triple-check my code and test behaviour in apps
>  -
  
Manual Testing setup:
- The easiest way to get a device token is to copy from the Xcode console
- Basic PN: 
  - Triggering:
- PN with Deep-link: 
  - Triggering:


uros-mil:
This PR will be canceled since work on it is paused.
Issues is added to Linear to track and prioritize this type of work: https://linear.app/customerio/issue/MBL-1166/improve-test-coverage-for-swizzling-part-of-the-code

PR #922, opened 2025-05-23
fix: Fixed issue with duplicate internal call for setting device token

uros-mil:
Task: [MBL-1159](https://linear.app/customerio/issue/MBL-1159/bug-in-no-swizzling-interaction-with-messagingpushapn)

## Issue:
- The condition for enabling the AppDelegate that replaces swizzling wasn't properly checked
- This produced a duplicate call to the internal API for registering the device token
- Our internal API checks if the token has changed and calls the backend only if it has, so we do not have duplicate calls to the backend

## Solution:
- The missing condition is added

### Original plan
The original plan was to add full coverage for classes where this problem happened, classes MessagingPushAPN and MessagingPushFCM, which has been done in this PR: https://github.com/customerio/customerio-ios/pull/921/files#diff-d322168af6d12b45ff1a57021d6a686396fe280ff72e00f2c1872bc9e4abff3f
- but as layers of swizzling have been refactored for testability, to finish the work, I'll need to rewrite ~100% of the code
- this has a risk of not covering some aspect of original swizzling and producing new issues

### Final plan
I decided to first solve the issue I originally noticed, without making any changes to the swizzling code.
I'll later revisit the original PR and new ticket created to cover it: [MBL-1166](https://linear.app/customerio/issue/MBL-1166/improve-coverage-for-swizzling-part-of-the-code)

## Testing plan:
Manual Testing setup:
- The easiest way to get a device token is to copy from the Xcode console
- Basic PN: 
  - Triggering:
- PN with Deep-link: 
  - Triggering:

### DIFF
--- a/Sources/MessagingPush/Integration/CioProviderAgnosticAppDelegate.swift
+++ b/Sources/MessagingPush/Integration/CioProviderAgnosticAppDelegate.swift
@@ -73,7 +73,7 @@ open class CioProviderAgnosticAppDelegate: CioAppDelegateType, UNUserNotificatio
         _ application: UIApplication,
         didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?
     ) -> Bool {
-        MessagingPush.appDelegateIntegratedExplicitely = true
+        MessagingPush.appDelegateIntegratedExplicitly = true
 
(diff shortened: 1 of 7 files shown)


mahmoud-elmorabea (approved):
Looks good! 💯

mrehan27 (approved):

PR #924, opened 2025-05-28
fix: premature deallocation of the rich push http client

Ahmed-Ali:
The RichPushHttpClient instance doesn't live long enough to process the response of its requests. Which means if anything is wrong with the request, we will see no logs to understand what's going on.

This change allows us to process the response as expected without depending on the life of the RichPushHttpClient instance.


mahmoud-elmorabea (approved):
Looks good! 💯

uros-mil:
@Ahmed-Ali: I had too many urgent things in the previous days, so I didn't get to this. 
Thanks for starting this, but unless the client is waiting on this, we should probably not merge PR that are not passing the checks.
To prevent this type of problem, make sure to set up pre-commit/pre-push checks with `lefthook`.

PR #925, opened 2025-05-30
fix: Fix in-app messages not overlaying safe area top and bottom

mahmoud-elmorabea:
TBA


BernardGatt (approved):
LGTM

BernardGatt:
One additional test i'd make is to see if the message size updates well when we transition between messages.

mrehan27 (approved):
Looks good, thanks for thorough testing. I'm assuming this doesn't reintroduce the linked bug and `contentInsetAdjustmentBehavior` does the trick for us.

PR #926, opened 2025-05-30
chore: lint issues resolution 

Shahroz16:
fixes the lint issues [merged](https://github.com/customerio/customerio-ios/pull/924) in `main`


Ahmed-Ali:
We have a pre-commit action to auto-fix these lint issues but it is broken. It might be worth fixing rather than all of us have to manually do that with every code we write

PR #929, opened 2025-06-02
fix: FCM version upper and lower bound  

Shahroz16:
This PR updates the Firebase SDK version constraints to ensure consistency between Swift Package Manager (SPM) and CocoaPods integrations. Both now explicitly support versions from 8.7.0 up to (but not including) 12.0.0.

Changes
	•	SPM (Package.swift):
Updated .package declaration to use a closed range:

```
.package(name: "Firebase", url: "https://github.com/firebase/firebase-ios-sdk.git", "8.7.0"..<"12.0.0")
```

	•	CocoaPods (.podspec):
Updated dependency constraint to:
```
spec.dependency "FirebaseMessaging", ">= 8.7.0", "< 12.0.0"
```

Testing:
- In order to ensure its working, you add latest FCM dependency version in sample apps and they should not conflict with CIO FCM dependency.

### DIFF
--- a/CustomerIOMessagingPushFCM.podspec
+++ b/CustomerIOMessagingPushFCM.podspec
@@ -30,5 +30,5 @@ Pod::Spec.new do |spec|
   # Add FCM SDK as a dependency, as our SDK is designed to be compatible with it. 
   # No version is specified which means that by default, the latest version is installed for customers. 
   # Customers can override this by adding the pod to their Podfile themselves to specify a version they want to use. 
-  spec.dependency "FirebaseMessaging", ">= 8.7.0"
+  spec.dependency "FirebaseMessaging", ">= 8.7.0", "< 12.0.0"
 end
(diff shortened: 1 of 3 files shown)


uros-mil:
Let's update the `CocoaPods-FCM` app to use the latest Friebase to prevent future issues of this type. 
As that would protect us by preventing compilation if any incompatibility happens

mahmoud-elmorabea (approved):

Shahroz16:
@uros-mil 

> Let's update the `CocoaPods-FCM` app to use the latest Friebase to prevent future issues of this type. As that would protect us by preventing compilation if any incompatibility happens

Added a test to ensure, SPM and Cocoapods are utilising same versions.

willshelley:
Latest version of Firebase is 12.0.0. Package.swift shows < 12.0.0? Causing package resolution issues when building with latest Firebase iOS SDK. 

.package(name: "Firebase", url: "https://github.com/firebase/firebase-ios-sdk.git", "8.7.0"..<"12.0.0")

Please correct me if I'm getting something wrong.

mrehan27:
Thanks for sharing @willshelley. This is expected. I've [opened a PR with an increased upper bound](https://github.com/customerio/customerio-ios/pull/938) which should allow you to use any Firebase 12 version once it's merged. Have a great day!

PR #930, opened 2025-06-02
fix: Sanitize unsupported numeric values

mahmoud-elmorabea:
Fixes: [MBL-1196](https://linear.app/customerio/issue/MBL-1196/floating-point-nan-or-infinity-values-should-be-managed-on-android)

## Problem
- We allow customers to attach arbitrary `Map` of properties with events and screen tracking. The input is defined as `Map<String, Any>`. That means that customers can pass any value as value in a map entry
- These values need to be serialized into a JSON format to be transmitted to our servers
- Customers can pass values that are not supported by JSON. For example, a `Double` value of `Double.NaN` or the result of dividing 0/0 or smth similar. These values are not supported by JSON

This causes the serialization process to fail by crashing on Android and it defaults to an incorrect value of `0` on iOS

## Solution
- We are creating a sanitization step for those types of arbitrary inputs to _exclude_ values that are not supported by 
JSON
- An error log is logged to the console informing them of the problematic values
- We will be adding documentation letting customers know that these unsupported values should not be passed and if they are, they will be ignored.

## Testing
- I've added extensive unit tests for unsupported values embedded at different levels of nesting
- I've tested passing unsupported values in our sample apps and it works fine
  - You can add something like `CustomerIO.shared.screen(title: "Test", properties: ["test": Double.nan])` to the sample app and it will trigger the issue

### DIFF
--- a/Sources/DataPipeline/CustomerIO+Events.swift
+++ b/Sources/DataPipeline/CustomerIO+Events.swift
@@ -37,6 +37,6 @@ public extension CustomerIO {
     ///   - category: A category to the type of screen if it applies.
     ///   - properties: Any extra metadata associated with the screen. e.g. method of access, size, etc.
     func screen(title: String, category: String? = nil, properties: [String: Any]? = nil) {
-        DataPipeline.shared.analytics.screen(title: title, category: category, properties: properties)
+        DataPipeline.shared.analytics.screen(title: title, category: category, properties: properties?.sanitizedForJSON())
     }
 }
--- a/Sources/DataPipeline/DataPipeline.swift
+++ b/Sources/DataPipeline/DataPipeline.swift
@@ -72,12 +72,12 @@ public class DataPipeline: ModuleTopLevelObject<DataPipelineInstance>, DataPipel
         set {
             guard var implementation = implementation else { return }
 
-            implementation.profileAttributes = newValue
+            implementation.profileAttributes = newValue.sanitizedForJSON()
         }
     }
 
     public func identify(userId: String, traits: [String: Any]?) {
-        implementation?.identify(userId: userId, traits: traits)
+        implementation?.identify(userId: userId, traits: traits?.sanitizedForJSON())
     }
 
     public func identify<RequestBody: Codable>(userId: String, traits: RequestBody?) {
@@ -93,7 +93,7 @@ public class DataPipeline: ModuleTopLevelObject<DataPipelineInstance>, DataPipel
         set {
             guard var implementation = implementation else { return }
 
-            implementation.deviceAttributes = newValue
+            implementation.deviceAttributes = newValue.sanitizedForJSON()
         }
     }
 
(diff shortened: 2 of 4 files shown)


uros-mil (commented):
Looks good.
But please extend test coverage to changes introduced in `DataPipeline.swift` and `CustomerIO+Events.swift` files, and check my other comments.

Sources/DataPipeline/DataPipeline.swift
```diff
@@ -110,15 +110,15 @@ public class DataPipeline: ModuleTopLevelObject<DataPipelineInstance>, DataPipel
     }
 
     public func track(name: String, properties: [String: Any]?) {
-        implementation?.track(name: name, properties: properties)
+        implementation?.track(name: name, properties: properties?.sanitizedForJSON())
     }
 
     public func track<RequestBody: Codable>(name: String, properties: RequestBody?) {
         implementation?.track(name: name, properties: properties)
```
uros-mil, line 117: Why this method and 2 below are not sanitized - `screen<RequestBody: Codable>(title: String, properties: RequestBody?)` and `trackMetric(deliveryID: String, event: Metric, deviceToken: String)`?

Sources/DataPipeline/Util/DictionarySanitizer.swift
```diff
@@ -0,0 +1,102 @@
...
+        case let number as Double:
+            if number.isNaN || number.isInfinite {
+                logger.error("Removed unsupported numeric value")
+                return nil
+            } else {
+                return number
+            }
+        case let number as Float:
+            if number.isNaN || number.isInfinite {
+                logger.error("Removed unsupported numeric value")
```
uros-mil, line 42: Did we agreed to crash in debug mode?

Sources/DataPipeline/Util/DictionarySanitizer.swift
```diff
@@ -0,0 +1,102 @@
...
+                return nil
+            } else {
+                return number
+            }
+        case let dict as [String: Any]:
+            let sanitized = dict.sanitizedForJSON(logger: logger)
+            return sanitized.isEmpty ? nil : sanitized
```
uros-mil, line 49: This exist in many place and could be simplified with private extension:
```
private extension Dictionary {
    var nilIfEmpty: Self? {
        isEmpty ? nil : self
    }
}
```
And used with `_.sanitizedForJSON(logger: logger).nilIfEmpty`.

Sources/DataPipeline/CustomerIO+Events.swift
```diff
@@ -37,6 +37,6 @@ public extension CustomerIO {
     ///   - category: A category to the type of screen if it applies.
     ///   - properties: Any extra metadata associated with the screen. e.g. method of access, size, etc.
     func screen(title: String, category: String? = nil, properties: [String: Any]? = nil) {
-        DataPipeline.shared.analytics.screen(title: title, category: category, properties: properties)
+        DataPipeline.shared.analytics.screen(title: title, category: category, properties: properties?.sanitizedForJSON())
```
uros-mil, line 40: I don't think this is the right place to add protection.
It should be added inside Analytics implementation. That way every call to analytics is protected.
In this same file, you have call to analytics (method `func screen<P: Codable>(title: String, category: String? = nil, properties: P?)`) that isn't protected.

Let's make sure that all methods which are working with `properties`/`traits` in Analytics are protected. This applies to both of those that are using Dictionary and Generics (where `JSON(with:)` is called - shouldn't silently fail).

PR #935, opened 2025-07-07
fix: Preserve numeric types when doing sanitization for JSON

mahmoud-elmorabea:
## Overview

There was a mistake in the sanitization logic we created for JSON values. When checking numeric values, if they were valid we were returning the casted value not the original ones. This caused booleans to be converted to numeric values.

## Solution

Use the original value when value doesn't need to be sanitized

### DIFF
--- a/Sources/DataPipeline/Util/DictionarySanitizer.swift
+++ b/Sources/DataPipeline/Util/DictionarySanitizer.swift
@@ -35,14 +35,14 @@ extension Dictionary where Key == String, Value == Any {
                 logger.error("Removed unsupported numeric value")
                 return nil
             } else {
-                return number
+                return value
             }
         case let number as Float:
             if number.isInvalidJsonNumber() {
                 logger.error("Removed unsupported numeric value")
                 return nil
             } else {
-                return number
+                return value
             }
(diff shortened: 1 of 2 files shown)


Sources/DataPipeline/Util/DictionarySanitizer.swift
```diff
@@ -30,19 +30,21 @@ extension Dictionary where Key == String, Value == Any {
      */
     private func sanitizeValue(_ value: Any, logger: Logger) -> Any? {
         switch value {
+        case is Bool:
+            return value
```
Shahroz16, line 34: looks good, i recommended this because it made sense logically :D and it should work too but please test with react native as well. 😅

Shahroz16 (approved):
looks good

Sources/DataPipeline/Util/DictionarySanitizer.swift
mahmoud-elmorabea: Yes, I've added unit tests to check Swift behavior and re-ran the tests with many types passed from JS and it yields same results as mentioned in the description

PR #945, opened 2025-08-06
chore: Deprecating unnecessary tracking APIs

mahmoud-elmorabea:
Closes: [MBL-1298](https://linear.app/customerio/issue/MBL-1298/tracking-api-alignment)

## Summary
Simplified the CustomerIO tracking API by removing all `Codable` parameter overloads for identify, track, and screen methods, keeping only the `[String: Any]` map-based versions.

## Changes
- Removed Codable overloads from:
  - CustomerIOInstance protocol and CustomerIO class implementation
  - CustomerIO+Events.swift extension methods
- Removed parameter-less overloads:
  - track(name:) without properties
  - screen(title:category:) without properties
  - identify<T: Codable>(traits:) without userId
  
## Benefits
- Simplified API: Single consistent method signature per function
- Better maintainability: Reduced code complexity and potential confusion

### DIFF
--- a/Sources/DataPipeline/CustomerIO+Events.swift
+++ b/Sources/DataPipeline/CustomerIO+Events.swift
@@ -11,10 +11,12 @@ public extension CustomerIO {
     /// Associate a user with their unique ID and record traits about them.
     /// - Parameters:
     ///   - traits: A dictionary of traits you know about the user. Things like: email, name, plan, etc.
+    @available(*, deprecated, message: "Use 'identify(userId:traits:)' with [String: Any] traits parameter instead. Support for Codable traits will be removed in a future version.")
     func identify<T: Codable>(traits: T) {
         DataPipeline.shared.analytics.identify(traits: traits)
     }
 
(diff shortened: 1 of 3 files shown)


Sources/DataPipeline/CustomerIO+Events.swift
```diff
@@ -11,10 +11,12 @@ public extension CustomerIO {
     /// Associate a user with their unique ID and record traits about them.
     /// - Parameters:
     ///   - traits: A dictionary of traits you know about the user. Things like: email, name, plan, etc.
+    @available(*, deprecated, message: "Use 'identify(userId:traits:)' with [String: Any] traits parameter instead. Support for Codable traits will be removed in a future version.")
```
mrehan27, line 14: This method was mainly for attaching profile traits to anonymous profiles without requiring `userId`. Since `identify` needs a `userId`, I don't think it's a direct replacement. Should this now be handled by `setProfileAttributes` instead?

PR #946, opened 2025-08-07
Added support for queue sticky sessions

BernardGatt:
Addresses: [INAPP-13815](https://linear.app/customerio/issue/INAPP-13815/update-ios-sdk-to-support-sticky-sessions)

### DIFF
--- a/Sources/MessagingInApp/Gist/Network/SessionManager.swift
+++ b/Sources/MessagingInApp/Gist/Network/SessionManager.swift
@@ -1,0 +1,11 @@
+import Foundation
+
+/// Manages session-level identifiers that persist for the lifetime of the app launch
+class SessionManager {
+    static let shared = SessionManager()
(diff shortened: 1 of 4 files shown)


mahmoud-elmorabea (approved):
Looks good! 

Let's make sure to update these URLs as the TODO mentions 😇 
> //TODO: Before merging, all these URLs need to be updated to CIO domains.

Sources/MessagingInApp/Gist/Network/SessionManager.swift
```diff
@@ -0,0 +1,11 @@
+import Foundation
+
+/// Manages session-level identifiers that persist for the lifetime of the app launch
+class SessionManager {
```
Shahroz16, line 4: can be `Internal`?

Shahroz16 (approved):

PR #949, opened 2025-08-22
fix: build issues on xcode 26 beta

Shahroz16:
closes: https://linear.app/customerio/issue/MBL-1343/ios-26-xcode-26-beta-build-fails-messagingpush-does-not-conform-to

Problem:
Xcode 16 Beta has stricter protocol conformance validation that detected availability annotation mismatches between protocol definitions and implementations with @available(iOSApplicationExtension, unavailable) annotations.

Solution: Remove Protocol Methods and Update Callers
 
 Note: SPM is resolving it by itself but its an issue on the cocoapods version.

### DIFF
--- a/Sources/MessagingPush/PushHandling/ManualPushHandling+UserNotifications.swift
+++ b/Sources/MessagingPush/PushHandling/ManualPushHandling+UserNotifications.swift
@@ -28,7 +28,12 @@ public extension MessagingPush {
             return false
         }
 
-        return implementation.userNotificationCenter(
+        // Cast to concrete type since method was removed from protocol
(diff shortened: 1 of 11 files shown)


Sources/MessagingPush/PushHandling/ManualPushHandling+UserNotifications.swift
```diff
@@ -28,7 +28,12 @@ public extension MessagingPush {
             return false
         }
 
-        return implementation.userNotificationCenter(
+        // Cast to concrete type since method was removed from protocol
+        guard let concreteImplementation = implementation as? MessagingPushImplementation else {
```
mrehan27, line 32: This can be merged with `guard` statement above?

PR #951, opened 2025-08-28
chore: Add coordinator to prevent push delivery metrics loss

mahmoud-elmorabea:
## Why?
Some delivery metrics are lost because we don't wait for that request to finish. The previous implementation had two independent async operations (metrics tracking and rich push processing) and only rich push processing called `contentHandler` upon completion. When rich push is faster, it would call `contentHandler` and terminate the NSE, killing the in-flight metrics request even when plenty of NSE time remained.

## How we solve?
Add `NSEOperationCoordinator` to wait for both async calls or NSE call that it will expire. The coordinator ensures `contentHandler` is called exactly once only after:
- Both operations complete (success or failure), OR
- `serviceExtensionTimeWillExpire` is called by iOS
This maximizes delivery metrics success by using the full available NSE execution time instead of terminating prematurely when only one operation completes.

Note: Moved `Synchronized<T>` utility to Common module for reuse

### DIFF
--- a/Sources/Common/Util/Synchronized.swift
+++ b/Sources/Common/Util/Synchronized.swift
@@ -1,0 +1,20 @@
+import Foundation
+
+/// An object that manages the execution of tasks atomically.
+/// Thread-safe wrapper for read/write access to a value using concurrent queue with barriers.
+public struct Synchronized<Value> {
+    private let mutex = DispatchQueue(label: "io.customer.SDK.Utils.Synchronized", attributes: .concurrent)
+    private var _value: Value
+
+    public init(_ value: Value) {
+        self._value = value
+    }
+
+    /// Returns the thread-safe value.
+    public var value: Value { mutex.sync { _value } }
+
+    /// Submits a block for synchronous, thread-safe execution with write access.
+    public mutating func value<T>(execute task: (inout Value) throws -> T) rethrows -> T {
+        try mutex.sync(flags: .barrier) { try task(&_value) }
+    }
+}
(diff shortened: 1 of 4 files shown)


Sources/MessagingPush/RichPush/MessagingPush+RichPush.swift
```diff
@@ -87,7 +103,14 @@ extension MessagingPushImplementation {
...
 
-        RichPushRequestHandler.shared.stopAll()
+        // Let coordinator handle timeout gracefully if available
+        if let coordinator = currentCoordinator {
+            coordinator.handleTimeWillExpire()
+            currentCoordinator = nil
+        } else {
+            // Fallback to original behavior if no coordinator
+            RichPushRequestHandler.shared.stopAll()
+        }
```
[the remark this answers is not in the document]
Shahroz16: this looks like a valid concern, we should set `currentCoordinator = nil` on `didRecieve`

Sources/Common/Util/Synchronized.swift
Shahroz16, line 1: i still see a reference in `MessagingInApp` can we remove that too, please

Sources/MessagingPush/RichPush/NSEOperationCoordinator.swift
```diff
@@ -0,0 +1,183 @@
...
+
+        if let finalContent = state.markRichPushComplete(with: content) {
+            callContentHandler(with: finalContent, reason: "bothOperationsComplete")
+        }
+    }
+
+    private func callContentHandler(with content: UNNotificationContent, reason: String) {
+        logger.info("NSE operations complete (\(reason)). Calling content handler.")
+
+        contentHandler(content)
```
Shahroz16, line 117: should this be done on main thread? did you check with wrappers? they often require NSE operation to be done on main

Shahroz16 (commented):
i think we need to add tests for this Coordinator race/timeout cases.

PR #980, opened 2026-01-21
fix: In app improvements

mahmoud-elmorabea:
This PR solves two issues

## Issue: In-App Messaging Fails After `clearIdentify()`

The issue affects Android and iOS

What Happens
clearIdentify publishes a "reset" event, then generates a new anonymous ID
In-app messaging receives the reset event and clears both userId and anonymousIdThe new anonymous ID is never communicated to in-app messagingAPI requests fail with 400 because no user token is available

Root Cause
Event ordering problem: The reset event is published before the new anonymous ID is generated, and no follow-up event notifies in-app messaging about the new anonymous ID.

Expected vs Actual
Expected: After reset, in-app messaging should use the new anonymous IDActual: In-app messaging has no user token at all, causing 400 errors

The issue only affects the session where the user was identified and then cleared. Once app is killed and re-launched, anonymous messages work again.

### Problem: Polling Timer Continues Running in Background

This only affects iOS.

Problem: The polling timer for fetching in-app messages doesn't stop when the app goes to background—it only skips the actual network fetch.
Location: Gist.swift - setupPollingAndFetch() method (line 235) and fetchUserMessages() method (line 262)

What happens:
When polling is active, a repeating Timer.scheduledTimer fires every N seconds
If the user backgrounds the app, the timer continues firing
Each timer fire wakes the main thread, checks applicationState != .background, logs a skip message, and returnsThe timer never actually stops until the app is terminated or SSE takes over

Root cause: The Gist class has no lifecycle observers (background/foreground) to pause and resume the polling timer. It only performs a runtime check inside fetchUserMessages() to skip the work.
Impact:
Unnecessary CPU/battery drain from repeated timer fires in background
Log spam: "Application in background, skipping queue check" logged every poll interval
Potential timer drift/coalescing issues when iOS manages background app state
Inconsistent with SseLifecycleManager which properly stops SSE connections on backgroundContrast with SSE: The SseLifecycleManager correctly observes didEnterBackgroundNotification and willEnterForegroundNotification to stop/start the SSE connection, but the polling timer in Gist lacks equivalent handling

---

> [!NOTE]
> **Medium Risk**
> Touches in-app messaging state/reset flow and iOS app-lifecycle timer behavior; mistakes could cause missed message fetches or incorrect identity usage, but changes are localized and guarded.
> 
> **Overview**
> Fixes in-app messaging after `clearIdentify()` by extending `ResetEvent` with an optional `newAnonymousId`, publishing it from the DataPipeline reset hook, and re-seeding Gist’s anonymous identity after reset when provided.
> 
> Improves iOS in-app polling efficiency by adding foreground/background lifecycle observers to pause/resume the polling timer, and adds a background-state guard when scheduling the timer to avoid race conditions (SSE still takes precedence when active).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8106cb7143c65415000e9d13184d8dbe2bc7266a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/Sources/MessagingInApp/Gist/Gist.swift
+++ b/Sources/MessagingInApp/Gist/Gist.swift
@@ -203,6 +203,13 @@ class Gist: GistProvider {
 
         // Timer must be scheduled on the main thread
         threadUtil.runMain {
+            // Prevent starting timer if app is in background (race condition protection)
+            // This can happen if handleForegrounded's async callback completes after handleBackgrounded ran
+            guard UIApplication.shared.applicationState != .background else {
+                self.logger.logWithModuleTag("Gist: App is in background, skipping polling timer setup", level: .debug)
+                return
+            }
+
             self.queueTimer = Timer.scheduledTimer(
                 timeInterval: pollingInterval,
                 target: self,


Sources/DataPipeline/Plugins/DataPipelinePublishedEvents.swift
```diff
@@ -30,6 +30,9 @@ class DataPipelinePublishedEvents: EventPlugin {
     }
 
     func reset() {
-        eventBusHandler.postEvent(ResetEvent())
+        // Pass the new anonymous ID with the reset event so in-app messaging
+        // can continue fetching messages for the anonymous user after reset.
+        let newAnonymousId = analytics?.anonymousId
+        eventBusHandler.postEvent(ResetEvent(newAnonymousId: newAnonymousId))
```
mrehan27: Shouldn't we also add a test to verify this?

Sources/MessagingInApp/Gist/Gist.swift
```diff
@@ -316,3 +282,61 @@ class Gist: GistProvider {
...
+        logger.logWithModuleTag("Gist: App backgrounded - pausing polling timer", level: .info)
+        invalidateTimer()
+    }
+
+    /// Handles SSE flag changes for polling control.
+    func handleSseEnabledChange(state: InAppMessageState) {
+        if state.shouldUseSse {
+            logger.logWithModuleTag("Gist: SSE enabled for identified user - stopping polling timer", level: .info)
+            invalidateTimer()
+        } else if !state.useSse {
```
mrehan27: I understand this might be outside the scope of this PR and coming from earlier changes, but looking through the code here, `shouldUseSse` and `useSse` flags are quite confusing. It would be nice to refactor them to more descriptive names in a separate PR so the expectations are clearer for callers like this one.

PR #1036, opened 2026-04-30
chore: Add OS geofence monitoring foundation

mrehan27:
part of: [MBL-1626](https://linear.app/customerio/issue/MBL-1626/ios-handle-os-geofence-monitoring)

## Summary

- Add `Geofence` data model with per-geofence transition types and last-updated timestamp for diff-based re-registration
- Add `Geofence+Distance` extension with `distanceTo()` utility (CoreLocation dependency separated from pure data model)
- Add `GeofenceRegionMonitoring` protocol and `CoreLocationGeofenceMonitor` implementation wrapping `CLLocationManager` region monitoring with per-region transition type configuration and `authorizedAlways` permission gating
- Add `GeofenceConstants` for movement trigger identifier, iOS region budget (19 business + 1 trigger), and sync thresholds
- Add `Logger+Geofence` with structured logging under `"Geofence"` tag

## Design notes

- Permission check: `startMonitoring` returns early and logs an error (once) if `authorizedAlways` is not granted — host app must request permission, SDK never requests it
- Transition callbacks include `CLLocationManager.location` so callers get a fresh fix at transition time
- All new types are Swift 6 strict concurrency clean (`Sendable` structs, `@Sendable` callback, no cross-isolation issues)
- iOS persists monitored regions across reboots automatically (no boot receiver needed, unlike Android)

## What's not included (coming in next PRs)

- Server API integration and persistence/sync
- Geofence transition event delivery with cooldown dedup
- Smart geofence updates (coordinator, movement trigger loop, lifecycle/DI wiring)

---

> [!NOTE]
> **Medium Risk**
> Adds new CoreLocation-backed region monitoring with permission gating and transition callbacks, which can affect location/region behavior if wired into the SDK. Changes are mostly additive but touch OS-level location monitoring and concurrency/locking.
> 
> **Overview**
> Introduces a new geofence foundation layer: a `Geofence` model (with per-region transition types and timestamps) plus shared `GeofenceConstants` for region identifiers, OS region budget, and sync/cooldown thresholds.
> 
> Adds a `GeofenceRegionMonitoring` abstraction and a `CoreLocationGeofenceMonitor` implementation that registers/unregisters circular regions, gates monitoring behind *Always* authorization (logging once when missing), forwards enter/exit transitions with the current location, and tracks owned region identifiers to avoid interfering with other monitored regions.
> 
> Adds geofence-specific `Logger` helpers and a `MockGeofenceRegionMonitor` test stub to support unit testing of higher-level geofence coordination logic.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 76177ed7ca07b5962cfd29cbfeb5cd64b894ac59. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Sources/Location/Geofence/Monitor/CoreLocationGeofenceMonitor.swift
```diff
@@ -0,0 +1,162 @@
...
+/// Must be created on the main thread (CLLocationManager requirement).
+/// Delegate callbacks are forwarded to the `onTransition` handler along with the
+/// manager's current location so callers have a fresh fix at transition time.
+///
+/// Tracks which regions this monitor owns so it does not interfere with regions
+/// registered by the host app or other SDKs (CLLocationManager.monitoredRegions is shared app-wide).
+///
+/// All mutable state is protected by `lock` since CLLocationManager delegate callbacks
+/// may fire on a different thread than the caller of start/stop methods.
+final class CoreLocationGeofenceMonitor: NSObject, GeofenceRegionMonitoring, CLLocationManagerDelegate {
```
mahmoud-elmorabea, line 16: Should this be actor instead?

mahmoud-elmorabea (approved):

Sources/Location/Geofence/Monitor/CoreLocationGeofenceMonitor.swift
hollyschilling-cio: It can't be its own actor because it must be run on the main thread. It probably should use the decorator `@MainActor`.

hollyschilling-cio: The other possible way to handle this is that the Monitor class should not be the `CLLocationManagerDelegate`. Because CoreLocation is an Objective C framework, it eschews Swift conventions and concurrency rules and adds requirements like the `NSObject` base class. A way to handle this more elegantly is to make the class an actor, create an internal class that acts as the delegate and passes data back to the actor. Its benefit in this case is the elimination of the locks, but at the cost of a bit of complexity.

Tests/Location/Geofence/Monitor/GeofenceRegionMonitoringTests.swift
```diff
@@ -0,0 +1,96 @@
+@testable import CioInternalCommon
+@testable import CioLocation
+import Foundation
+import SharedTests
+import Testing
+
+@Suite("GeofenceRegionMonitoring")
+struct GeofenceRegionMonitoringTests {
```
hollyschilling-cio: These tests provide no value. They only test the `MockGeofenceRegionMonitor` and do nothing to validate any of the code.

Sources/Location/Geofence/Monitor/CoreLocationGeofenceMonitor.swift
mrehan27: Thanks for the ideas, I'll rethink on this and make changes in follow up PR so it is easier to review.

Tests/Location/Geofence/Monitor/GeofenceRegionMonitoringTests.swift
mrehan27: Agreed, removed for now, will improve this with later PRs

Sources/Location/Geofence/Monitor/CoreLocationGeofenceMonitor.swift
```diff
@@ -0,0 +1,162 @@
...
+
+        let clampedRadius = min(radius, manager.maximumRegionMonitoringDistance)
+        let region = CLCircularRegion(center: coordinate, radius: clampedRadius, identifier: identifier)
+        region.notifyOnEntry = transitionTypes.contains(.enter)
+        region.notifyOnExit = transitionTypes.contains(.exit)
+        lock.lock()
+        ownedRegionIdentifiers.insert(identifier)
+        lock.unlock()
+
+        manager.startMonitoring(for: region)
```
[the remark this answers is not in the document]
mrehan27: Noted, this race might exist in the current locking model. Rather than expand lock scope here, the follow-up PR converts this class to an actor (per discussion above), which will eliminate lock entirely and serialize all operations through actor isolation. So will address this in follow up PR.

PR #1039, opened 2026-05-04
docs: Foundational Documentation and ADRs

hollyschilling-cio:
This PR is the foundational work for the v5 SDK. These files define the structure for all subsequent documents and code.

---

> [!NOTE]
> **Low Risk**
> Documentation-only changes that introduce no runtime behavior changes; risk is limited to potential misalignment with future implementation if the specs are interpreted incorrectly.
> 
> **Overview**
> Establishes the `v5/` specification foundation for the CustomerIO iOS SDK reimplementation, including a scoped `CLAUDE.md` that defines AI/dev workflow rules and spec-as-source-of-truth expectations.
> 
> Adds core spec artifacts: a canonical `spec/GLOSSARY.md`, an initial `spec/domain/domain-model.md` describing the planned module graph/pipeline/schema, and ADRs (`001`–`005`) formalizing major architectural decisions (SqlCipher-encrypted storage, Swift 6.2 strict concurrency, removing `cdp-analytics-swift`, SPM-only distribution, and server-driven aggregation).
> 
> Introduces the `spec/features/utilities.md` feature spec defining the intended responsibilities and constraints of the internal `CustomerIO_Utilities` target (e.g., `Synchronized`, `StorageManager`, `CommonEventBus`, `HttpClient`, `CIOKeys`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 172f96eabec1ee955fca449907f744c365552940. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/v5/spec/decisions/001-sqlcipher-encrypted-storage.md
+++ b/v5/spec/decisions/001-sqlcipher-encrypted-storage.md
@@ -1,0 +1,54 @@
+# ADR 001 — SqlCipher for All On-Device Storage
+
+**Status:** Accepted
+
+---
+
+## Context
+
+The previous SDK stored its event queue and profile data in plain text
+(`UserDefaults` and `EventStorageManager` flat files). Events routinely carry
+PII — user identifiers, traits, and behavioral data. Storing this unencrypted
+is unacceptable; the SDK must encrypt all data at rest regardless of device
+encryption state.
+
+Encrypting on-device storage requires owning the storage layer entirely.
+`cdp-analytics-swift` writes its event queue to plain SQLite with no encryption
+path — it cannot be patched to use SqlCipher without forking the library. This
+is one of the requirements addressed in ADR 003.
+
+## Decision
+
+All data written to disk uses **SqlCipherKit**. `UserDefaults` and flat-file JSON
+storage are removed entirely from the new SDK.
+
+- The passphrase is supplied by a `DatabaseKeyProvider` protocol, configured at
+  SDK init time via `SdkConfigBuilder.databaseKeyProvider(_:)`.
+- Two implementations are provided out of the box:
+  - `ApiKeyDatabaseKeyProvider` (default) — uses the CDP API key verbatim.
+  - `KeychainDatabaseKeyProvider` — generates a 256-bit random key on first
+    launch, stored in the Keychain with `kSecAttrAccessibleAfterFirstUnlock`.
+- The database filename and Keychain account name are both scoped to the CDP API
+  key. Rotating the API key automatically produces a new database file and a new
+  Keychain entry; the old database is orphaned without an explicit cleanup step.
+
+## Consequences
+
+### What this enables
+
+- All PII stored by the SDK is encrypted at rest, regardless of device
+  encryption state.
+- `KeychainDatabaseKeyProvider` provides per-install, hardware-bound key
+  isolation, inaccessible even to someone with the app binary.
+- A single `StorageManager` struct serves as the stateless gateway for all
+  modules; module-specific query methods live in extensions inside their owning
+  module target.
+
+### What this constrains
+
+- SqlCipherKit is a required dependency for the entire SDK. Apps that do not
+  need encryption cannot opt out.
+- Database migrations must be managed explicitly via `MigrationRunner`. Each
+  module that adds tables must register a `Migration` conformance.
+- The `StorageManager.db` property is `package` access — visible within the
+  Swift package but not to external consumers.
--- a/v5/spec/decisions/002-actor-based-concurrency.md
+++ b/v5/spec/decisions/002-actor-based-concurrency.md
@@ -1,0 +1,49 @@
+# ADR 002 — Swift 6.2 Actor-Based Concurrency Throughout
+
+**Status:** Accepted
+
+---
+
+## Context
+
+The previous SDK was not written with `Sendable` enforcement or actor isolation.
+`cdp-analytics-swift` produced `@preconcurrency` wrappers or suppressed warnings
+that hid real data races. The reimplementation targets Swift 6.2 with strict
+concurrency enabled, which is structurally incompatible with the previous
+library's concurrency model. This is one of the requirements addressed in ADR 003.
+
+## Decision
+
+Adopt **Swift 6.2 strict concurrency** throughout:
+
+- All public-facing mutable state lives in `actor` types.
+- All cross-boundary types conform to `Sendable`.
+- Event-tracking methods (`track`, `identify`, `screen`, etc.) on `CustomerIO`
+  are `nonisolated` — callable from any thread or isolation domain without `await`.
+- Pre-configure event buffering is handled by `AsyncStream`'s built-in buffer
+  (capped at 100 events, oldest-first). No separate two-mode buffer is needed.
+- `Synchronized<T>` bridges actor-owned state to `nonisolated` access sites
+  (e.g. module loggers, push token mirrors) without crossing actor boundaries.
+
+## Consequences
+
+### What this enables
+
+- No data races possible in the SDK's own code under strict checking.
+- `nonisolated` tracking methods allow the SDK to be called from `@MainActor`
+  views, background threads, and notification extensions without `await`.
+- Pre-configure events are buffered automatically and drained in order once
+  `configure()` completes.
+- Modules that need `@MainActor` (e.g. `MessagingInApp`) dispatch internally
+  without requiring the root actor to be `@MainActor`.
+
+### What this constrains
+
+- iOS 13+ minimum deployment target required (Swift concurrency embedded runtime).
+- `os.Logger` structured logging (iOS 14+) and `Clock`/`ContinuousClock`
+  (iOS 16+) are not used. Flush timing relies on app lifecycle events, not
+  background timers.
+- `URLSession` async APIs (iOS 15+) are not used. The HTTP client wraps the
+  completion-handler API in `withCheckedContinuation` for iOS 13/14 compatibility.
+- `@preconcurrency` annotations are used sparingly and only where Swift's region
+  checker produces a known false positive (documented inline).
--- a/v5/spec/decisions/003-remove-analytics-swift.md
+++ b/v5/spec/decisions/003-remove-analytics-swift.md
@@ -1,0 +1,51 @@
+# ADR 003 — Remove cdp-analytics-swift Dependency
+
+**Status:** Accepted
+
+---
+
+## Context
+
+The previous SDK depended on `cdp-analytics-swift` (Segment's analytics library)
+for its event pipeline, upload scheduler, and retry logic. Two prior
+architectural requirements make this dependency untenable:
+
+- **ADR 001 — Encrypted storage**: All on-device data must be encrypted via
+  SqlCipher. `cdp-analytics-swift` writes its event queue to plain SQLite with
+  no encryption path. This cannot be addressed by configuration or extension —
+  it requires owning the storage layer entirely.
+
+- **ADR 002 — Swift 6.2 strict concurrency**: The SDK must compile cleanly
+  under Swift 6.2 with full `Sendable` enforcement. `cdp-analytics-swift` was
+  not written with actor isolation; it produces `@preconcurrency` suppressions
+  that hide real data races and is structurally incompatible with the required
+  concurrency model.
+
+A third requirement compounds this: the server-driven `AggregationEngine`
+(see ADR 005) evaluates events before they enter the upload queue and may absorb
+them into accumulators rather than forwarding them immediately. This does not
+map onto the drain-on-queue assumptions built into `analytics-swift`'s scheduler.
+Implementing the aggregation model requires controlling the full path from event
+receipt to upload.
+
+None of these requirements can be met by patching or wrapping the existing
+library. The dependency must be removed.
+
+## Decision
+
+Remove `cdp-analytics-swift` entirely. Own the full event pipeline:
+enrichment → aggregation → queue → upload/retry.
+
+## Consequences
+
+### What this constrains
+
+- The upload/retry pipeline must be reimplemented from scratch. The area of
+  `cdp-analytics-swift` most worth preserving as *reference* — not reinventing —
+  is the batch upload pipeline: exponential backoff with jitter, deduplication
+  across restarts, ordering guarantees, and partial-batch failure handling.
+  These have years of production hardening. Treat the `cdp-analytics-swift`
+  source as the specification for this layer.
+- Any plugin-based interception model (`analytics-swift`'s `Plugin` API) is not
+  carried forward. There is no public plugin protocol for customer code to
+  intercept the event pipeline.
--- a/v5/spec/decisions/004-spm-only-distribution.md
+++ b/v5/spec/decisions/004-spm-only-distribution.md
@@ -1,0 +1,37 @@
+# ADR 004 — Swift Package Manager as Sole Distribution Channel
+
+**Status:** Accepted
+
+---
+
+## Context
+
+The previous SDK supported CocoaPods as its primary distribution channel.
+CocoaPods has been officially deprecated; the trunk repository will become
+read-only in fall 2026.
+
+## Decision
+
+The SDK is distributed via **Swift Package Manager only**. No `.podspec` is
+maintained or published.
+
+## Consequences
+
+### What this enables
+
+- No dual-distribution maintenance burden (`.podspec` + `Package.swift`).
+- Wrapper SDK compatibility:
+  - **Flutter**: SPM plugin support has been available since Flutter 3.19 (early 2024).
+    The Flutter wrapper declares its iOS native dependency via `Package.swift`.
+  - **Expo**: Expo Modules API supports SPM for native modules. EAS Build handles resolution.
+
+### What this constrains
+
+- **React Native (bare)**: RN's iOS build system remains CocoaPods-centric at the
+  framework level as of this writing. This is the broader community's problem to
+  solve — CocoaPods deprecation affects every major iOS library, not just CIO.
+  If the RN wrapper temporarily requires a pre-built binary bridge (via
+  `swift-create-xcframework`), that is the wrapper team's concern and does not
+  affect the SDK's distribution model.
+- Apps that have not yet migrated from CocoaPods to SPM cannot adopt this SDK
+  until they do.
--- a/v5/spec/decisions/005-server-driven-aggregation.md
+++ b/v5/spec/decisions/005-server-driven-aggregation.md
@@ -1,0 +1,72 @@
+# ADR 005 — Server-Driven Event Aggregation Engine
+
+**Status:** Accepted
+
+---
+
+## Context
+
+The previous SDK forwarded every event individually to the upload queue. For
+high-frequency events (screen views, location pings, lifecycle events), this
+produces excessive server load and inflated analytics noise. Reducing volume
+client-side requires a configurable ruleset.
+
+## Decision
+
+Introduce a server-driven `AggregationEngine` that evaluates every event
+before it enters the queue. The engine fetches its ruleset from a static
+endpoint relative to the region base URL.
+
+### Rule evaluation
+
+Each `AggregationRule` specifies an event name and one or more `AggregateOperation`
+values. Supported operations: `count`, `sum`, `min`, `max`, `assign`,
+`assignIfNull`, `countUnique`, `histogram`, `discard`.
+
+Evaluation outcomes (`AggregationResult`):
+- `.passThrough` — event proceeds to the queue unchanged
+- `.aggregated` — event is absorbed into an accumulator; not queued individually
+- `.discarded` — event is silently dropped
+
+### Accumulator lifecycle
+
+- Accumulators are persisted encrypted to SqlCipher so partial counts survive
+  app kills.
+- On flush, the engine synthesises a `trackSynthesized` event (bypasses
+  aggregation re-evaluation), injects it into the pipeline via a stored
+  `@Sendable (PendingEvent) -> Void` closure, and resets the accumulator.
+- Flush checks occur only at app lifecycle events (startup, foreground,
+  background). No background timers are used.
+- Flush schedules in rule configs are **minimum durations** ("no sooner than").
+
+### Config refresh
+
+- Fetched on SDK startup and on app foreground.
+- Rate-limited to once per 24 hours (last fetch timestamp persisted to `sdk_meta`).
+- A new ruleset is applied atomically; in-progress accumulator state is preserved
+  for rules whose `id` is unchanged.
+
+### Back-reference pattern
+
+`AggregationEngine` injects synthesised events via a closure rather than holding
+a back-reference to `CustomerIO`. This avoids a retain cycle and makes the engine
+testable without SDK infrastructure.
+
+## Consequences
+
+### What this enables
+
+- High-frequency events can be aggregated client-side without app updates — the
+  server controls the rules.
+- Events matching `discard` rules never reach the server, reducing data noise.
+- The engine is fully testable: a `MockHttpClient` and a lambda enqueue closure
+  are sufficient; no real database or scheduler is needed.
+
+### What this constrains
+
+- The `predicate` field in rule configs is reserved but **not evaluated in v1**.
+  All v1 use cases require only event-name matching. Property-level filtering
+  (e.g. "count only purchases where currency == USD") requires a future CEL
+  interpreter (see OPEN_QUESTIONS.md item 2).
+- Flush timing is coarse — only at lifecycle checkpoints, not on a timer.
+  A flush overdue by hours will not fire until the next foreground event.
--- a/v5/spec/domain/domain-model.md
+++ b/v5/spec/domain/domain-model.md
@@ -1,0 +1,210 @@
+# CustomerIO iOS SDK — Domain Model
+
+---
+
+## Module Graph
+
+```
+┌──────────────────────────────────────────────────────────┐
+│                   CustomerIO_Utilities                    │
+│  (internal — never imported directly by app developers)  │
+│                                                          │
+│  Synchronized                                             │
+│  CommonEventBus · RegistrationToken                      │
+│  HttpClient · JsonAdapter · DateUtil                     │
+│  StorageManager · MigrationRunner                        │
+│  CIOKeys                                                 │
+└─────────────────────────┬────────────────────────────────┘
+                          │ (internal dependency)
+           ┌──────────────▼──────────────────┐
+           │          CustomerIO              │
+           │  (root public module)            │
+           │                                 │
+           │  Variant / VariantConvertible    │
+           │  CIOTrackingClient               │
+           │  identify · track · screen       │
+           │  Event pipeline                 │
+           │  Aggregation engine             │
+           │  Upload scheduler               │
+           │  SdkConfig / SdkConfigBuilder   │
+           │  StorageManager extensions      │
+           │  Module registry                │
+           │  CIOLogger / CIOLogLevel        │
+           └──┬──────────┬──────────┬────────┘
+              │          │          │
+  ┌───────────▼──┐  ┌────▼────────────────┐  ┌────────────────────┐  ┌──────────────────────┐
+  │  Location    │  │  MessagingPush       │  │  MessagingInApp    │  │  LiveActivities      │
+  │  module      │  │  (APN or FCM via     │  │  (not yet impl.)   │  │  (iOS 16.1+ only)    │
+  │              │  │  PushTokenProvider)  │  │                    │  │  (not yet impl.)     │
+  └──────────────┘  └──────────────────────┘  └────────────────────┘  └──────────────────────┘
+        │
+  ┌─────▼──────────┐
+  │   Geofencing   │
+  │   module       │
+  └────────────────┘
+```
+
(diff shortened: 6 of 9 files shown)


v5/spec/decisions/001-sqlcipher-encrypted-storage.md
```diff
@@ -0,0 +1,54 @@
...
+
+The previous SDK stored its event queue and profile data in plain text
+(`UserDefaults` and `EventStorageManager` flat files). Events routinely carry
+PII — user identifiers, traits, and behavioral data. Storing this unencrypted
+is unacceptable; the SDK must encrypt all data at rest regardless of device
+encryption state.
+
+Encrypting on-device storage requires owning the storage layer entirely.
+`cdp-analytics-swift` writes its event queue to plain SQLite with no encryption
+path — it cannot be patched to use SqlCipher without forking the library. This
```
mrehan27, line 17: Why do we need a complete fork here instead of a lightweight module that copies and modifies only the necessary classes? Similar to how we have the tracking migration module in the SDK today, we could take a more minimal approach just for data migration?

v5/spec/decisions/001-sqlcipher-encrypted-storage.md
```diff
@@ -0,0 +1,54 @@
...
+
+- The passphrase is supplied by a `DatabaseKeyProvider` protocol, configured at
+  SDK init time via `SdkConfigBuilder.databaseKeyProvider(_:)`.
+- Two implementations are provided out of the box:
+  - `ApiKeyDatabaseKeyProvider` (default) — uses the CDP API key verbatim.
+  - `KeychainDatabaseKeyProvider` — generates a 256-bit random key on first
+    launch, stored in the Keychain with `kSecAttrAccessibleAfterFirstUnlock`.
+- The database filename and Keychain account name are both scoped to the CDP API
+  key. Rotating the API key automatically produces a new database file and a new
+  Keychain entry; the old database is orphaned without an explicit cleanup step.
```
mrehan27, line 33: It seems we're deciding not to clean up data if a user changes CDP API key. Is there a specific reason for that decision?

v5/spec/decisions/001-sqlcipher-encrypted-storage.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Consequences
+
+### What this enables
```
mrehan27, line 37: Should we explicitly highlight what is must have in this migration? e.g., profile identification migration is critical and could break core functionality if not handled correctly. It might be worth calling out the importance of these areas more clearly.

v5/spec/decisions/002-actor-based-concurrency.md
```diff
@@ -0,0 +1,49 @@
...
+
+The previous SDK was not written with `Sendable` enforcement or actor isolation.
+`cdp-analytics-swift` produced `@preconcurrency` wrappers or suppressed warnings
+that hid real data races. The reimplementation targets Swift 6.2 with strict
+concurrency enabled, which is structurally incompatible with the previous
+library's concurrency model. This is one of the requirements addressed in ADR 003.
+
+## Decision
+
+Adopt **Swift 6.2 strict concurrency** throughout:
```
mrehan27, line 17: Some of these points read more like technical decisions than product decisions. Should we reframe them in product terms? e.g. "All calls to CustomerIO should not require await." If we need to enforce technical constraints, perhaps we should clearly label them under a section like Technical Decisions instead of grouping them under general Decisions.

v5/spec/decisions/005-server-driven-aggregation.md
```diff
@@ -0,0 +1,72 @@
...
+Each `AggregationRule` specifies an event name and one or more `AggregateOperation`
+values. Supported operations: `count`, `sum`, `min`, `max`, `assign`,
+`assignIfNull`, `countUnique`, `histogram`, `discard`.
+
+Evaluation outcomes (`AggregationResult`):
+- `.passThrough` — event proceeds to the queue unchanged
+- `.aggregated` — event is absorbed into an accumulator; not queued individually
+- `.discarded` — event is silently dropped
+
+### Accumulator lifecycle
```
mrehan27, line 31: This seems like a new concept. I like the idea of `AggregationResult`, but I'm unclear about the flushing strategy. It looks like we're only flushing on lifecycle events (e.g. app start/stop). Does that mean we're introducing a considerable delay? e.g. during a 3 minute app session, would events only be sent when the app backgrounds/foregrounds and not during the session itself?

v5/spec/domain/domain-model.md
```diff
@@ -0,0 +1,210 @@
+# CustomerIO iOS SDK — Domain Model
```
mrehan27, line 1: Are these `.md` files auto generated or manually maintained?

v5/spec/decisions/001-sqlcipher-encrypted-storage.md
mahmoud-elmorabea: +1 if `cdp-analytics-swift` doesn't allow us to inject storage behavior change, a minimal first step is to edit the fork we already have to do encryption/decryption before it stores, instead of replacing/re-implementing the whole thing

v5/spec/decisions/001-sqlcipher-encrypted-storage.md
```diff
@@ -0,0 +1,54 @@
...
+encryption state.
+
+Encrypting on-device storage requires owning the storage layer entirely.
+`cdp-analytics-swift` writes its event queue to plain SQLite with no encryption
+path — it cannot be patched to use SqlCipher without forking the library. This
+is one of the requirements addressed in ADR 003.
+
+## Decision
+
+All data written to disk uses **SqlCipherKit**. `UserDefaults` and flat-file JSON
```
mahmoud-elmorabea, line 22: The SQL Cipher decisions couples two things into one:
- Data encryption: Yes, we need to do that
- Move all data into SQL style DB? Not necessarily, none of the data we save needs to be structured in that way. Even events are stored on their request format and syncing doesn't need to understand or enumerate them

v5/spec/decisions/001-sqlcipher-encrypted-storage.md
mahmoud-elmorabea: I don't think we even need to discard the data. It's common that these API keys are rotated or maybe the source was changed for some reason. Why should previously non-sync'd event be lost?

v5/spec/decisions/001-sqlcipher-encrypted-storage.md
```diff
@@ -0,0 +1,54 @@
...
+  encryption state.
+- `KeychainDatabaseKeyProvider` provides per-install, hardware-bound key
+  isolation, inaccessible even to someone with the app binary.
+- A single `StorageManager` struct serves as the stateless gateway for all
+  modules; module-specific query methods live in extensions inside their owning
+  module target.
+
+### What this constrains
+
+- SqlCipherKit is a required dependency for the entire SDK. Apps that do not
```
mahmoud-elmorabea, line 49: We should consider alternatives and assess them against the extra size/complexity brought by SqlCipherKit

v5/spec/decisions/002-actor-based-concurrency.md
```diff
@@ -0,0 +1,49 @@
+# ADR 002 — Swift 6.2 Actor-Based Concurrency Throughout
+
+**Status:** Accepted
+
+---
+
+## Context
+
+The previous SDK was not written with `Sendable` enforcement or actor isolation.
+`cdp-analytics-swift` produced `@preconcurrency` wrappers or suppressed warnings
```
mahmoud-elmorabea, line 10: Can we mention some examples, I don't think there are any `@preconcurrency` that we have caused by `cdp-analytics-swift`

v5/spec/decisions/002-actor-based-concurrency.md
```diff
@@ -0,0 +1,49 @@
...
+that hid real data races. The reimplementation targets Swift 6.2 with strict
+concurrency enabled, which is structurally incompatible with the previous
+library's concurrency model. This is one of the requirements addressed in ADR 003.
```
mahmoud-elmorabea, line 13: Will we still be able to support consumers that target Swift 5?

v5/spec/decisions/003-remove-analytics-swift.md
```diff
@@ -0,0 +1,51 @@
...
+- **ADR 001 — Encrypted storage**: All on-device data must be encrypted via
+  SqlCipher. `cdp-analytics-swift` writes its event queue to plain SQLite with
+  no encryption path. This cannot be addressed by configuration or extension —
+  it requires owning the storage layer entirely.
+
+- **ADR 002 — Swift 6.2 strict concurrency**: The SDK must compile cleanly
+  under Swift 6.2 with full `Sendable` enforcement. `cdp-analytics-swift` was
+  not written with actor isolation; it produces `@preconcurrency` suppressions
+  that hide real data races and is structurally incompatible with the required
+  concurrency model.
```
mahmoud-elmorabea, line 22: Is this really a hard requirement? During this transition period for the community we are bound to deal with code that doesn't fully comply with Swift 6. Even some of Apple's APIs are still being migrated.

We can work around this by making sure code is thread safe manually by using `Synchronized` types and similar concepts

v5/spec/decisions/003-remove-analytics-swift.md
```diff
@@ -0,0 +1,51 @@
...
+## Context
+
+The previous SDK depended on `cdp-analytics-swift` (Segment's analytics library)
+for its event pipeline, upload scheduler, and retry logic. Two prior
+architectural requirements make this dependency untenable:
+
+- **ADR 001 — Encrypted storage**: All on-device data must be encrypted via
+  SqlCipher. `cdp-analytics-swift` writes its event queue to plain SQLite with
+  no encryption path. This cannot be addressed by configuration or extension —
+  it requires owning the storage layer entirely.
```
mahmoud-elmorabea, line 16: We can alternatively just encrypt data before its saved on our end and in a `cdp-analytics-swift` fork

v5/spec/decisions/003-remove-analytics-swift.md
```diff
@@ -0,0 +1,51 @@
...
+Implementing the aggregation model requires controlling the full path from event
+receipt to upload.
+
+None of these requirements can be met by patching or wrapping the existing
+library. The dependency must be removed.
+
+## Decision
+
+Remove `cdp-analytics-swift` entirely. Own the full event pipeline:
+enrichment → aggregation → queue → upload/retry.
```
mahmoud-elmorabea, line 37: This we are already planning to re-write many parts of the SDK, I prefer to keep that core part to minimize scope or at least not bundle it in a release that already has huge scope

v5/spec/decisions/004-spm-only-distribution.md
```diff
@@ -0,0 +1,37 @@
...
+## Context
+
+The previous SDK supported CocoaPods as its primary distribution channel.
+CocoaPods has been officially deprecated; the trunk repository will become
+read-only in fall 2026.
+
+## Decision
+
+The SDK is distributed via **Swift Package Manager only**. No `.podspec` is
+maintained or published.
```
mahmoud-elmorabea, line 16: Why are we doing this now? As far as I understand it goes into read only mode Dec 2nd 2026. We should offer the SDK through that for as long as we can push new versions there. I don't see a reason for cut that support earlier.

v5/spec/decisions/004-spm-only-distribution.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Consequences
+
+### What this enables
+
+- No dual-distribution maintenance burden (`.podspec` + `Package.swift`).
+- Wrapper SDK compatibility:
+  - **Flutter**: SPM plugin support has been available since Flutter 3.19 (early 2024).
+    The Flutter wrapper declares its iOS native dependency via `Package.swift`.
+  - **Expo**: Expo Modules API supports SPM for native modules. EAS Build handles resolution.
```
mahmoud-elmorabea, line 26: Worth noting that this support started Expo 53 I think

v5/spec/decisions/005-server-driven-aggregation.md
```diff
@@ -0,0 +1,72 @@
...
+**Status:** Accepted
+
+---
+
+## Context
+
+The previous SDK forwarded every event individually to the upload queue. For
+high-frequency events (screen views, location pings, lifecycle events), this
+produces excessive server load and inflated analytics noise. Reducing volume
+client-side requires a configurable ruleset.
```
mahmoud-elmorabea, line 12: That's not true. All events are batched and are flushed based on a flush policy. The default ones are `CountBasedFlushPolicy`, `IntervalBasedFlushPolicy` that would flush events to backend if they reach a certain count or periodically with a specific time interval

v5/spec/decisions/005-server-driven-aggregation.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Decision
+
+Introduce a server-driven `AggregationEngine` that evaluates every event
+before it enters the queue. The engine fetches its ruleset from a static
+endpoint relative to the region base URL.
```
mahmoud-elmorabea, line 18: We can achieve the same result using a combination of `before` plugin and our settings API

v5/spec/decisions/005-server-driven-aggregation.md
```diff
@@ -0,0 +1,72 @@
+# ADR 005 — Server-Driven Event Aggregation Engine
```
mahmoud-elmorabea, line 1: It is unclear why we are introducing this concept. What's the trigger for including this feature?

v5/spec/domain/domain-model.md
```diff
@@ -0,0 +1,210 @@
...
+    nonisolated func preActivate(_ config: SdkConfig)   // Phase 1 — synchronous
+    func configure(_ config: SdkConfig, storage: StorageManager, root: CustomerIO) async throws  // Phase 2
+}
+```
+
+| Phase | When | Who overrides |
+|-------|------|---------------|
+| `preActivate` | Before first run-loop cycle; inside `activateModulesForLaunch` | Only `MessagingPushModule` (registers `UNUserNotificationCenter` delegate) |
+| `configure` | After database and pipeline are ready | All modules |
+
```
mahmoud-elmorabea, line 173: Is there value of documenting these things separately rather than on the file itself when its created?

v5/spec/domain/domain-model.md
mahmoud-elmorabea, line 1: Is this a guide for LLMs?

v5/spec/features/utilities.md
```diff
@@ -0,0 +1,270 @@
...
+- No SDK-level types may be imported here. `CustomerIO_Utilities` must remain a
+  leaf dependency.
+- All types must be `Sendable` or explicitly `@unchecked Sendable` with
+  documented isolation guarantees.
+- No static state. Every type is instantiated and injected; nothing is accessed
+  via a global or singleton.
+
+---
+
+## Components
```
mahmoud-elmorabea, line 43: Will these be moved to their respective classes/files when implemented?

mahmoud-elmorabea: I think these types should be introduced in follow up PRs when they are actually used so it's sensible to reason about their usage or API

v5/spec/decisions/001-sqlcipher-encrypted-storage.md
hollyschilling-cio: This is a great question.

The Segment Analytics SDK writes everything that goes through it out to disk in plain text. This is done so that file can be handed over to the Operating System and the OS can handle the upload. To be specific here, the uploader works out of the process bounds. That means that we are sharing PII outside of our control and writing it to disk. Even our cleanup efforts after are not an improvement on this because of how APFS works. The filesystem does a zero-block clone (similar to a hard link) on the file when it hands it to the OS and all we actually delete is the original handle to the blocks on disk. The data is still on disk and outside our control. 

Because this is a **behavior** and not an implementation detail, changing it requires validation of everything that depends upon it. It will also require revalidation each time there is an update. It's bigger than a small fork, but could still be manageable. 

What really serves as a deal breaker is that this Segment SDK is not updated to enforce Swift Concurrency. That leaves us with the option of wrapping it in something like `@preconcurrency` and adding safeguards everywhere we touch it or migrating it ourselves. The migration touches about 30% of the lines of code. That means that each update we would attempt to rebase upon would be a massive effort. In either case, this raises the cost of using the Segment SDK to the same levels as a reimplementation.

The value of using the Segment Analytics is not in the code itself. It's in the validation and testing the code has endured. Once we begin modifying that code in non-trivial ways, that value disappears. Especially since there are other features we would like to be able to implement which are blocked by the use of the Segment SDK, this seems like a logical place to diverge.

v5/spec/decisions/001-sqlcipher-encrypted-storage.md
hollyschilling-cio: Yes, this is clutter and deliberately so. In the case of deliberate changes to a new CDP API Key, we need to add migration ability to SDK startup (not in scope here). This will require them to explicitly enter the old CDP API Key so we can decrypt the old data. In the event of temporary changes such as testing, we want to leave the other data untouched. Let's discuss migration planning at a future time.

v5/spec/domain/domain-model.md
hollyschilling-cio: These are manually maintained, but Claude should update them automatically as instructed in the Claude.md when used.

v5/spec/decisions/001-sqlcipher-encrypted-storage.md
hollyschilling-cio: The use of a Database is incidental. Most the records are Key-Value mapped blobs, similar to KeyVal or Redis. Encrypting files individually is more work than just storing blobs.

v5/spec/decisions/001-sqlcipher-encrypted-storage.md
hollyschilling-cio: The size of the entire SDK shrinks from around 25 MB to under 3 MB.

v5/spec/decisions/002-actor-based-concurrency.md
hollyschilling-cio: The current SDK does not implement Swift Concurrency and therefore @preconcurrency is not a requirement.

v5/spec/decisions/001-sqlcipher-encrypted-storage.md
hollyschilling-cio: Yes, but not here. Depending on exactly what you're talking about, there are a few places it may exist already.
- [Behavioral Scenarios](https://github.com/customerio/customerio-ios-tng/tree/main/behavioral-scenarios)
- [Component Features](https://github.com/customerio/customerio-ios-tng/tree/main/spec/features)
- [Future ADRs](https://github.com/customerio/customerio-ios-tng/tree/main/spec/decisions)

As a general rule, if it doesn't exist in the documentation, it doesn't exist. If there are omissions from the documentation, we should call them out. Every feature, mapping, consideration, edge case, and similar should be documented.

v5/spec/features/utilities.md
hollyschilling-cio: Yes, this document could be its own PR. I included it because it is short and gives an example of the structure of these documents looks like. An important note is that there is not code here, but signatures. These signatures primarily exist to provide context to agentic coding tools.

v5/spec/domain/domain-model.md
hollyschilling-cio: The module startup is important to have documented. Some aspects of startup must happen immediately on app launch and synchronously, which is why the `preActivate` phase exists. Because this stage is synchronous and blocking, its use should be minimized. The creation (and only creation) of app modules occurs during this phase. The only other use of this phase is MessagingPushModule's registration as UNUserNotificationCenter's delegate. This must happen before any other code in the AppDelegate executes and is a minor task. 

Given how rare its intended use is, that use case is mentioned here.

v5/spec/domain/domain-model.md
hollyschilling-cio: Ideally, it should be for both. The LLM may parse this document looking for context for feature development, but that is the same way any developer may do the same. It spells out the different components and how they interact. The LLM should keep it up to date, but I would like to see a PR Hook in the future that also ensure that any code changes are accompanied by documentation changes that reflect that change.

v5/spec/decisions/002-actor-based-concurrency.md
hollyschilling-cio: Yes, with a few important caveats.

Users are required to build their apps with Xcode 26 today. Xcode 26's compiler can use Swift 6 libraries in a Swift 5 Application. New projects created today will default to Swift 6. This compatibility was also present in Xcode 16.

Where that compatibility does not extend is outside of Applications. Specifically, libraries cannot depend upon another library if the dependency uses a later version of Swift. For example, if Segment updated their library to use Swift 6.2 today, our v4 SDK could no longer use it until we update to Swift 6.2 or later.

v5/spec/decisions/002-actor-based-concurrency.md
hollyschilling-cio: I had not considered separating different types of decisions. That might be a good idea, but I need to think about it more. 

The specific case you identified ("All calls to CustomerIO should not require await.") is more of a product decision than it might seem at first. In Swift, Async methods cannot be called from a sync context. This has a rippling effect and becomes problematic often. For example, if the `track` method were async, it would require the user to branch into a `Task` to send the event. It created a very bad DX and left the code littered with blocks like the one below.

```
Task {
    await cio.track("EventName")
}

```

Forcing the main interfaces to be sync was a product and architectural decision, not a technical one, and guided many of the architectural decisions elsewhere.

v5/spec/decisions/003-remove-analytics-swift.md
hollyschilling-cio: I believe we discussed this in the meeting today, but the data is written to disk to be handed off to the OS. The data must be written exactly as it will be uploaded, and therefore cannot be encrypted.

v5/spec/decisions/004-spm-only-distribution.md
hollyschilling-cio: Honestly, I didn't expect I'd get this far before December. Being that we are here now, this is something worth reevaluating. There are some requirements for using CocoaPods that need to be accommodated, but it is possible. I'll fill in more details later, but this is a good callout and discussion.

v5/spec/decisions/002-actor-based-concurrency.md
matt-frizzell: 2 cents. I think that these docs should represent some of both.

> Some of these points read more like technical decisions than product decisions.

v5/spec/decisions/003-remove-analytics-swift.md
matt-frizzell: > The data must be written exactly as it will be uploaded, and therefore cannot be encrypted.

This is one part I didn't want to slow the meeting on but could not visualize what this actually meant. In trying to understand this, I thought I heard that the data we'd encrypt was the payload and that sounded "fine" to me. The way I could interpret Mahmoud's suggestion not working is if what happens is the entire request gets encrypted, which I could see breaking some other mechanism. Otherwise, I wasn't sure why an encrypted payload would be a problem.

hollyschilling-cio: I want to make the distinction here between encrypting the payload on disk and encrypting the payload over the wire. 

Without making new server API endpoints, we can't change the structure of the data going over the wire. What goes over the wire should match exactly what is there now coming out of the Segment SDK. If we decide to change the data format over the wire, that is an entirely different scope of work.

The data being written to disk just before it is uploaded leaves the app. We are writing out the payload of the POST/PUT message body so we can hand it to the OS with the headers and the OS can upload it. The data is uploaded exactly as it is written to disk. 

The other problem here is that once we do that handover, we lose the ability to clean up that data. APFS supports file cloning, which is a zero-depth copy similar to a hard link. When we hand it to the OS, they immediately clone the file and maintain their own handle for it. This preserves the data on disk even if we explicitly delete it after the upload completes. The exact implementation changes each release, but there have been several exploits by capturing temp files on iOS which are sensitive data in an unencrypted state.

v5/spec/decisions/003-remove-analytics-swift.md
hollyschilling-cio: We need to get there eventually. Swift 5 support will end at some point in the not-too-distant future. Any new code should be built with Swift 6 concurrency in mind.

v5/spec/decisions/005-server-driven-aggregation.md
hollyschilling-cio: https://linear.app/customerio/issue/FLO-3802/workspace-severely-degraded-for-large-volume-sender-yape-account-82276

v5/spec/decisions/005-server-driven-aggregation.md
hollyschilling-cio: I'm not asserting that those things don't exist. I'm saying that there are plenty of things which don't even need to make it to the server and if we can cause 1,000,000 devices to not each send in unwanted noise, it's worth the invenstment.

v5/spec/decisions/005-server-driven-aggregation.md
hollyschilling-cio: The idea for an accumulator is for data that can be delayed significantly, like DAYS or WEEKS. An accumulator has a flush timer on it that is how many days before flushing to the server. Events that are time sensitive are not appropriate to be accumulated in this way, but they can pass through an accumulator and still be uploaded immediately.

v5/spec/decisions/003-remove-analytics-swift.md
hollyschilling-cio: The other element to point out here is that the OS does the upload, not the app or SDK. That has important consequences connected to App Lifecycle. That upload completes (or fails) even if the app is backgrounded, suspended, or killed. It is completely detached from the app's process. Switching the upload from using this file-based API to a memory-resident one loses this benefit and requires us to explicitly handle all these failure scenarios ourselves.

v5/spec/decisions/001-sqlcipher-encrypted-storage.md
mrehan27: I can re-explain if it helps. I just wanted to double check one thing. Does this overall direction change how we read/store the user id? If yes, when customers upgrade, will they still be able to send events without an explicit identify call? Missing a few events during migration is manageable, but missing the user id could break most flows. Just want to make sure we're accounting for this clearly during migration.

v5/spec/decisions/002-actor-based-concurrency.md
mrehan27: I think this decision could be useful for platform specific decisions, specially if we plan to reuse this approach for other SDKs e.g. Android. Product decisions can remain general, but in some cases, we may need to explicitly call out platform level differences where the division is going to help?

v5/spec/decisions/005-server-driven-aggregation.md
mrehan27: The statement that confused me was: "Flush checks occur only at app lifecycle events". I am not fully clear on how time sensitive events are handled outside of those.

v5/spec/decisions/003-remove-analytics-swift.md
```diff
@@ -0,0 +1,51 @@
...
+  not written with actor isolation; it produces `@preconcurrency` suppressions
+  that hide real data races and is structurally incompatible with the required
+  concurrency model.
+
+A third requirement compounds this: the server-driven `AggregationEngine`
+(see ADR 005) evaluates events before they enter the upload queue and may absorb
+them into accumulators rather than forwarding them immediately. This does not
+map onto the drain-on-queue assumptions built into `analytics-swift`'s scheduler.
+Implementing the aggregation model requires controlling the full path from event
+receipt to upload.
```
mahmoud-elmorabea, line 29: - Why do we think we need this aggregation behavior? 
- `analytics-swift` does offer plugins that can run before / at or after events are queued. So we can decide if a certain event should be queued or not before it ever reached the scheduler, check `ScreenFilterPlugin.swift`
-

v5/spec/decisions/001-sqlcipher-encrypted-storage.md
hollyschilling-cio: Yes, there is a migration necessary to move the identifiers into the consolidated encrypted storage. This one-time migration is handled in the [Migration Runner](https://github.com/customerio/customerio-ios-tng/blob/main/Sources/CustomerIO_Utilities/Storage/MigrationRunner.swift).

v5/spec/decisions/005-server-driven-aggregation.md
hollyschilling-cio: Things that are time sensitive should not be aggregated. What is being described here forms a baseline for functionality and methodology. We can add more to it and change flush intervals fairly easily, but that should all be considered expansion and future improvement.

v5/spec/decisions/005-server-driven-aggregation.md
hollyschilling-cio: That is a bold statement. If you believe it is true, I would encourage you to attempt to validate it.

PR #1042, opened 2026-05-07
refactor: Remove Push Notification functions from AppDelegate

hollyschilling-cio:
Eventually, we need to remove the dependency on the AppDelegate being a subclass or wrapper of one of our classes. This PR is the first step by removing unnecessary functionality from the AppDelegate and moving it into it's own delegate (`CioNotificationCenterDelegate`) and `MessagingPush`.

This change should be invisible to our users.

---

> [!NOTE]
> **Medium Risk**
> Changes how push notification callbacks are intercepted by introducing delegate-setter swizzling and a new proxy delegate, which can affect notification presentation/click handling and interaction with other SDKs’ delegates. Scope is contained to MessagingPush integration and is covered by new/updated tests, but swizzling introduces behavioral risk.
> 
> **Overview**
> **Push notification delegate handling is moved out of the CIO AppDelegate layer.** A new `CioNotificationCenterDelegate` now owns `UNUserNotificationCenterDelegate` callbacks (`willPresent`, `didReceive`, `openSettingsFor`) and forwards to an optional wrapped delegate while applying SDK defaults when absent.
> 
> **`MessagingPush.initialize()` now installs and retains this proxy delegate and swizzles `UNUserNotificationCenter.delegate` setter** to automatically wrap any future non-CIO delegate assignments (and to avoid double-invoking completion handlers for async delegates). The legacy swizzling-based push handler/proxy/history abstractions and their DI/mocks/tests are removed, with tests updated to validate wrapping behavior and single-invocation completion handlers.
> 
> Minor: adds `.claude/settings.local.json` to `.gitignore` and small CocoaPods project file cleanup.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dc6d6dbaaa7bb260893bcd3c3f8d5198f393b7ec. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Tests/MessagingPushAPN/Integration/CioAppDelegateAPNTests.swift
```diff
@@ -127,8 +127,8 @@ class CioAppDelegateAPNTests: XCTestCase {
             completionHandlerCalled = true
         }
 
-        // Call the method
-        appDelegateAPN.userNotificationCenter(UNUserNotificationCenter.current(), didReceive: UNNotificationResponse.testInstance, withCompletionHandler: completionHandler)
+        // Call through the installed CioNotificationCenterDelegate
+        mockNotificationCenter.delegate?.userNotificationCenter?(UNUserNotificationCenter.current(), didReceive: UNNotificationResponse.testInstance, withCompletionHandler: completionHandler)
```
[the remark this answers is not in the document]
hollyschilling-cio: Good catch. Test removed because it was no longer providing value.

Sources/MessagingPush/Integration/CioProviderAgnosticAppDelegate.swift
```diff
@@ -31,18 +31,13 @@ open class CioProviderAgnosticAppDelegate: CioAppDelegateType, UNUserNotificatio
         #selector(UIApplicationDelegate.application(_:didRegisterForRemoteNotificationsWithDeviceToken:)),
         #selector(UIApplicationDelegate.application(_:didFailToRegisterForRemoteNotificationsWithError:)),
         #selector(UIApplicationDelegate.application(_:continue:restorationHandler:)),
-        #selector(UIApplicationDelegate.application(_:open:options:)),
-        // UNUserNotificationCenterDelegate
-        #selector(UNUserNotificationCenterDelegate.userNotificationCenter(_:willPresent:withCompletionHandler:)),
-        #selector(UNUserNotificationCenterDelegate.userNotificationCenter(_:didReceive:withCompletionHandler:)),
-        #selector(UNUserNotificationCenterDelegate.userNotificationCenter(_:openSettingsFor:))
```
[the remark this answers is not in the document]
hollyschilling-cio: Agreed. Removed.

Tests/MessagingPushFCM/Integration/CioAppDelegateFCMTests.swift
```diff
@@ -190,8 +189,8 @@ class CioAppDelegateFCMTests: XCTestCase {
             completionHandlerCalled = true
         }
 
-        // Call the method
-        appDelegateFCM.userNotificationCenter(UNUserNotificationCenter.current(), didReceive: UNNotificationResponse.testInstance, withCompletionHandler: completionHandler)
```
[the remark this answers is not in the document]
hollyschilling-cio: Agreed. Removed.

Sources/MessagingPush/MessagingPush.swift
```diff
@@ -56,14 +74,19 @@ public class MessagingPush: ModuleTopLevelObject<MessagingPushInstance>, Messagi
...
+    public static func initialize(
+        withConfig config: MessagingPushConfigOptions = MessagingPushConfigBuilder().build()
+    ) -> MessagingPushInstance {
         shared.initializeModuleIfNotAlready {
             // set moduleConfig before creating implementation instance as dependencies inside instance may directly use moduleConfig from MessagingPush.
             Self.moduleConfig = config
             // Some part of the initialize is specific only to non-NSE targets.
             // Put those parts in this non-NSE initialize method.
-            if config.autoTrackPushEvents, !Self.appDelegateIntegratedExplicitly {
-                DIGraphShared.shared.automaticPushClickHandling.start()
```
mahmoud-elmorabea, line 66: Is this removal intentional? I think this breaks the swizzling path entirely. We have not officially deprecated that path and removing it is a breaking behavioral change.

mahmoud-elmorabea (changes requested):
I agree with extracting a `CioNotificationCenterDelegate` out of `CioProviderAgnosticAppDelegate` to keep the delegate wrapper as think as possible that it can be replaced with couple of calls by customers who don't want to wrap their app delegate.

But the swizzling change shouldn't be part of this PR, let's plan it separately

Sources/MessagingPush/MessagingPush.swift
hollyschilling-cio: The method has been marked as Deprecated for a year (May 2, 2025).

Although the Swizzle code path is gone, the behavior of swizzling remains in nearly all cases. We unconditionally capture the UNUserNotificationCenter delegate on initialize. If the app sets a delegate, we wrap it. The only edge case that the Swizzle handled that we no longer do is if the app replaces the delegate after we claim it. 

While it is a change, I don't see this as breaking. We are still capturing the delegate and proxying messages to their delegate if it exists. Is that edge case a concern?

mahmoud-elmorabea: > The method has been marked as Deprecated for a year (May 2, 2025).

`automaticPushClickHandling` is not a public type and is not directly used by customers. Its deprecation is an internal note for devs to avoid extending or using it as its being retired. This gave no signal to customers using the swizzling path to move off of it.

I agree with the goal and the approach of this PR however, removing the swizzling bit doesn't directly contribute to it or take away from its value.

>Although the Swizzle code path is gone, the behavior of swizzling remains in nearly all cases

Even if its few or minor number of cases, it would be breaking to some customers in a behavioral way.

Let's keep the swizzling bit as is for this change and ship the refactoring and we can separately evaluate what to do about the swizzling path.

hollyschilling-cio: That's fair. I'll updated it to restore the swizzling behavior.

hollyschilling-cio: Updated to restore Swizzling

mahmoud-elmorabea (approved):
Looks good but let's please verify these two scenarios before merging:

#### showPushAppInForeground ignored if there is another wrapped UN delegate
showPushAppInForeground config is silently ignored when any wrapped delegate responds to willPresent including for CIO-sent pushes

I think we can reproduce on the FCM app (Firebase's delegate responds to willPresent and typically returns something). A CIO foreground push that previously showed (because the customer set showPushAppInForeground: true) will now be silent. This is the most likely visible regression.

#### Deduplication logic in IOSPushEventListener
IOSPushEventListener used to have logic to prevent notifications with same delivery ids to not be re-processed, do we not need that anymore? I think this would kick in if any host app or SDK does the same courtesy that we are, of forwarding calls to existing delegates.

hollyschilling-cio:
I asked Claude for a difference in behaviors. The result is as follows:

Here's a clear comparison of the behaviors:
Old shouldDisplayPushAppInForeground (via IOSPushEventListener):

deliveryDate nil guard — returns early without calling completionHandler if deliveryDate is nil. This is actually a latent bug: the OS would wait indefinitely for a completion handler that never comes.
PushHistory deduplication — if the push was already processed (tracked by pushId + deliveryDate), it calls completionHandler(false) and exits. This existed because the fan-out could cause the same push to cycle back through the proxy and be processed twice.
CIO vs non-CIO split:
Non-CIO push → forwarded topushEventHandlerProxy(fan-out),completionHandler not called directly— left entirely to the other delegates
CIO push → also forwarded to proxy for notification purposes, but the proxy's return value isignored.completionHandler(showPushAppInForeground)is always used regardless of what the other delegates said

New willPresent (via CioNotificationCenterDelegate):

If a wrapped delegate exists and responds towillPresent→ pass completionHandler straight to them. They decide what to show.
If no wrapped delegate → useshowPushAppInForegroundconfig directly.
Net behavioral differences:

deliveryDate nil: Old code silently dropped completionHandler. New code always calls it. This is an improvement.
PushHistory deduplication: Not needed anymore — with a single wrapped delegate rather than a fan-out proxy, there's no mechanism by which the samewillPresentevent could arrive twice.
CIO push foreground display: Old codeoverrodethe wrapped delegate's decision for CIO pushes and appliedshowPushAppInForegroundregardless. New code defers to the wrapped delegate for all pushes, usingshowPushAppInForegroundonly as the fallback when no delegate is set. This matches the documented behavior ("only applies if there is no other delegate set") and is intentionally correct.
None of these differences warrant bringing any old logic back. The old behavior around CIO-push overriding and fan-out-deduplication was a consequence of the proxy architecture that we intentionally removed.

PR #1047, opened 2026-05-11
refactor: isolate CoreLocationGeofenceMonitor on @MainActor

mrehan27:
Ticket: [MBL-1626](https://linear.app/customerio/issue/MBL-1626/ios-handle-os-geofence-monitoring)

Follow-up to #1036 addressing review feedback that landed during the original PR:
- @mahmoud-elmorabea: "Should this be actor instead?"
- @hollyschilling-cio: prefers actor / `@MainActor` over locks
- Cursor: "Mutable state accessed without thread-safety guarantees"
- Cursor (post-merge): "Lock scope too narrow causes state desynchronization race"

We committed in the review thread to do this in the next PR.

## Summary
- Replace `final class` + `NSLock` with `@MainActor final class`.
- Make `GeofenceRegionMonitoring` `@MainActor` — protocol describes a CLLocationManager-backed contract, which is main-bound by Apple's API.
- Replace the `onTransition` get/set property with `setOnTransition(_:)` (cleaner contract for a side-effecting handler).
- `MockGeofenceRegionMonitor` becomes `@MainActor final class` to match.

## Design rationale

CLLocationManager must be created and called on main, and its delegate callbacks arrive on main. The monitor's bookkeeping state (`ownedRegionIdentifiers`, handler, permission-warning flag) has no reason to live on a different executor than the OS calls that touch it.

Two approaches were prototyped during review:

1. **`actor`** (initial attempt — first commit on this branch, see history): state lives on the actor's executor; OS calls hop to main via `Task { @MainActor in ... }`; delegate methods are `nonisolated` and re-enter the actor via `Task { await self... }`. This works but introduces three executors for one logical object, requires FIFO-of-same-priority assumptions for correctness, and Cursor flagged a real reentrancy edge case in an earlier draft.
2. **`@MainActor final class`** (final design, this PR): state and OS calls share one executor. Reentrancy between them is structurally impossible. No fire-and-forget Tasks. No `nonisolated` delegate dance. Delegate callbacks land in the correct domain naturally.

The `@MainActor` approach is the more honest model — the monitor wraps a main-anchored OS object and lives where that object lives. State synchronization is free because there is no concurrency on this class.

The transition handler stays as plain `@Sendable` (not `@MainActor`) so consumers (the upcoming coordinator in MBL-1630) can write closure bodies in whatever isolation they need. The handler is invoked on main at runtime (delegate callback context) — documented in the typealias doc comment.

`@preconcurrency` on the `CLLocationManagerDelegate` conformance is required because Apple's protocol predates Swift concurrency; callbacks do arrive on main since we created the manager on main, so the annotation is accurate.

## Scope
Three files: protocol, implementation, mock stub. No other consumers on this branch — the coordinator that will use this monitor lands in MBL-1630.

## Preserved fixes from #1036
All threading/correctness fixes from the previous review remain intact:
- Ownership tracking via `ownedRegionIdentifiers` (scoped to SDK-owned regions; doesn't touch host app or other SDK regions).
- Insert into ownership set before `manager.startMonitoring` (race ordering fix — now also structurally safe because both happen synchronously on main).
- `stopMonitoring`: remove from set first, then OS — avoids leaking identifier when OS drops a region silently.
- `monitoringDidFailFor` removes from ownership set so a failed registration doesn't leave a phantom entry against the 19-region budget.
- `Set<GeofenceTransition>` for order-independent equality (diff-based re-registration).
- Invalid coordinates logged at `error` level.

## Net diff
- Source: 56 insertions, 82 deletions (−26 lines vs. the merged class+NSLock baseline; even smaller vs. the actor draft).
- Tests: mock stub updated mechanically.

## Test plan
- [x] Location scheme builds clean
- [x] Whole-package test build succeeds
- [x] Zero warnings under `-strict-concurrency=complete` on geofence files
- [x] No new unit tests — `CoreLocationGeofenceMonitor` wraps `CLLocationManager` and isn't meaningfully unit-testable without a fake of the OS API; mock-only tests were deleted per @hollyschilling-cio's prior feedback that they "provide no value"

---

> [!NOTE]
> **Medium Risk**
> Concurrency model changes for `CLLocationManager` geofence monitoring and protocol API surface, which could impact call sites and main-thread assumptions, but does not alter geofence business rules or persistence.
> 
> **Overview**
> Refactors geofence region monitoring to be **main-actor isolated** by marking `GeofenceRegionMonitoring`, `CoreLocationGeofenceMonitor`, and the test `MockGeofenceRegionMonitor` as `@MainActor`, eliminating the prior `NSLock`-based synchronization.
> 
> Updates the monitoring interface by replacing the mutable `onTransition` property with `setOnTransition(_:)`, and adjusts the CoreLocation delegate handling accordingly (including `@preconcurrency` conformance) while keeping the same region ownership tracking and start/stop behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit afe994b3f754af9cea170661f5e028b076afa127. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Sources/Location/Geofence/Monitor/CoreLocationGeofenceMonitor.swift
+++ b/Sources/Location/Geofence/Monitor/CoreLocationGeofenceMonitor.swift
@@ -4,34 +4,24 @@ import Foundation
 
 /// CLLocationManager-backed geofence region monitor.
 ///
-/// Implemented as an actor so all mutable state is serialized without locks.
-/// Delegate callbacks are nonisolated and re-enter the actor via `Task { await ... }`,
-/// matching the existing `CoreLocationProvider` pattern in this module.
+/// `@MainActor`-isolated because CLLocationManager must be created and called on the main
+/// thread, and its delegate callbacks arrive on main. State and OS calls share one
+/// isolation domain, so the ownership-set update and the OS dispatch happen atomically with
+/// no reentrancy point between them — no locks, no fire-and-forget Tasks, no FIFO assumption.
 ///
 /// Tracks which regions this monitor owns so it does not interfere with regions
 /// registered by the host app or other SDKs (CLLocationManager.monitoredRegions is shared app-wide).
-///
-/// `init` is `@MainActor` so CLLocationManager creation and delegate wiring run on main
-/// (a CLLocationManager requirement). Subsequent CLLocationManager method calls are
-/// dispatched to the main actor via fire-and-forget `Task { @MainActor in ... }` so the
-/// ownership-set update and OS-call dispatch happen atomically on the actor executor with
-/// no reentrancy point between them.
-actor CoreLocationGeofenceMonitor: NSObject, GeofenceRegionMonitoring, CLLocationManagerDelegate {
+@MainActor
+final class CoreLocationGeofenceMonitor: NSObject, GeofenceRegionMonitoring, @preconcurrency CLLocationManagerDelegate {
     private let manager: CLLocationManager
     private let logger: Logger
-    private let maxRegionMonitoringDistance: CLLocationDistance
     private var onTransition: GeofenceTransitionHandler?
     private var hasLoggedPermissionWarning = false
     private var ownedRegionIdentifiers: Set<String> = []
 
-    @MainActor
     init(logger: Logger) {
-        let manager = CLLocationManager()
-        self.manager = manager
+        self.manager = CLLocationManager()
         self.logger = logger
-        // Constant for the platform (documented as 10 km on iOS); cache so we don't need
-        // an actor-suspending main-actor hop in startMonitoring.
-        self.maxRegionMonitoringDistance = manager.maximumRegionMonitoringDistance
         super.init()
         manager.delegate = self
     }
@@ -44,8 +34,8 @@ actor CoreLocationGeofenceMonitor: NSObject, GeofenceRegionMonitoring, CLLocatio
         onTransition = handler
     }
 
-    func startMonitoring(identifier: String, center: LocationData, radius: Double, transitionTypes: Set<GeofenceTransition>) async {
-        guard await checkAlwaysAuthorization() else { return }
+    func startMonitoring(identifier: String, center: LocationData, radius: Double, transitionTypes: Set<GeofenceTransition>) {
+        guard checkAlwaysAuthorization() else { return }
 
         let coordinate = CLLocationCoordinate2D(latitude: center.latitude, longitude: center.longitude)
         guard CLLocationCoordinate2DIsValid(coordinate) else {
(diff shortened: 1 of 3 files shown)


Sources/Location/Geofence/Monitor/CoreLocationGeofenceMonitor.swift
```diff
@@ -4,154 +4,144 @@ import Foundation
...
-            defer { lock.unlock() }
-            return _onTransition
-        }
-        set { lock.lock()
-            defer { lock.unlock() }
-            _onTransition = newValue
-        }
-    }
-
+/// Must be created on the main thread so CLLocationManager and delegate setup run on main.
```
hollyschilling-cio: The way to enforce that it must be created on the main thread is to add the `@MainActor` decorator to the `init` method.

hollyschilling-cio:
What you have here for the most part is fine, but there is a pattern you may benefit from utilizing. 


At various times I've attempted to introduce the primitive [Synchronized](https://github.com/customerio/customerio-ios-tng/blob/main/Sources/CustomerIO_Utilities/Synchronized/Synchronized.swift). This is similar to Apple's [Mutex](https://developer.apple.com/documentation/synchronization/mutex) class introduced in iOS 18. These classes work exceptionally well in conjunction with an Actor. Here's an example:


```
public actor Foo {
    private nonisolated var _bar: Synchronized<Bar?> = Synchronized(nil)
    public nonisolated var bar: Bar? {
        get { _bar.wrappedValue } 
        set { _bar.wrappedValue = value } 
    }
}
```

The reason to implement some properties and methods this way is that it allows thing inside and outside the actor to access these properties or methods in ways that don't require an async context. It allows the fact that something is an actor to be an implementation detail instead of an defining characteristic of the entity. Especially since we are an SDK and not an app, it allows for maximum flexibility for anything calling into our code.

Sources/Location/Geofence/Monitor/CoreLocationGeofenceMonitor.swift
mrehan27: Good call — agreed, the prose-only "must be created on main" warning wasn't actually enforcing anything.

Went a bit further than the literal suggestion: instead of just annotating `init`, made the whole class `@MainActor` (and the protocol). The thinking:

- Every method on the monitor either calls a CLLocationManager API (must be main) or touches state that's read by delegate callbacks (which arrive on main). With per-method `@MainActor`, the state itself would still need to be `@MainActor`-isolated, which transitively forces every method anyway — so class-level is just the compact form of the same constraint.
- It also collapses the actor pattern this PR originally introduced. The actor was paying for cross-executor synchronization (state on actor's executor, OS calls hopping to main, delegate callbacks `nonisolated` re-entering via `Task { await ... }`). For a class whose state has no reason to live anywhere but main, that was misapplied. Cursor's reentrancy flag in the earlier draft wasn't a fluke; it was symptomatic of the executor split.
- With `@MainActor`, state and OS calls share one isolation domain. Reentrancy between them is structurally impossible, no fire-and-forget Tasks, no FIFO ordering assumptions to document. Net −26 lines.

Updated PR title + description to reflect the change in direction. The actor stage is still visible in commit history if you want to compare. Happy to revert and ship just the `@MainActor init` annotation if you'd rather keep this PR scoped to your literal suggestion.

mahmoud-elmorabea (approved):

PR #1048, opened 2026-05-11
chore: Add geofence state persistence layer

mrehan27:
Foundation used by:
- [MBL-1628](https://linear.app/customerio/issue/MBL-1628/ios-send-geofence-transition-events) — cooldown records for event dedup
- [MBL-1630](https://linear.app/customerio/issue/MBL-1630/ios-smart-geofence-updates) — cached geofences, last server sync location/timestamp, monitored region IDs, movement trigger center

## Summary
- `GeofenceState` pure value-type model with optional fields so partial state can be stored without requiring all fields. Conforms to `Sendable`.
- `GeofenceStorage` actor that owns both file persistence and high-level operations. Read-modify-write sequences are atomic by construction (no internal `await`s), eliminating reentrancy concerns.
- Persists to a JSON file with iOS Data Protection (`completeUntilFirstUserAuthentication`) so state is readable when geofence callbacks fire while the app is killed (after first user unlock this boot).
- File is excluded from iCloud/iTunes backups (geofence cache is device-local context, not user data).

## Design rationale

**Single actor instead of layered protocol design**

The original split was `GeofenceStateStore` protocol → `FileGeofenceStateStore` class → `GeofenceStorage` protocol → `GeofenceStorageImpl` class with two `NSLock`s. Collapsing into one actor:
- Eliminates all locks and `@unchecked Sendable` annotations
- Makes read-modify-write atomic by construction (no `await` between load and save inside actor methods means no other task can interleave and cause lost updates)
- Reduces code (~3× fewer lines than the layered design)
- Tests use temp directories instead of an injected in-memory stub — still hermetic, no behavior change

**Why file-based instead of UserDefaults (Android parity question)**

Android uses SharedPreferences; iOS file-based is the more iOS-correct mirror because:
- Geofence cache is transient device-local context — should NOT be in iCloud/iTunes backups. File-based gives us `isExcludedFromBackup=true`; UserDefaults is backed up by default.
- Data Protection class is set explicitly (`completeUntilFirstUserAuthentication`); UserDefaults inherits the app's default.
- `GeofenceState` is structured Codable — UserDefaults would force encode-to-Data-blob anyway, just under a key.
- Matches the existing `FileLastLocationStateStore` pattern in the same module.

## Scope
This PR exposes only cooldown operations (used immediately by the upcoming event delivery PR). The full state model is in place so the sync coordinator PR (MBL-1630) can add cached-geofence, last-sync, monitored-ID, and movement-trigger operations without re-touching the underlying storage.

## Diff size
- Source: 151 lines (2 files: `GeofenceState.swift` + `GeofenceStorage.swift`)
- Tests: 150 lines (1 file: `GeofenceStorageTests.swift`)

## Test plan
- [x] `getEventCooldowns_givenEmpty_expectEmptyDictionary`
- [x] `recordEventCooldown_givenKey_expectPersisted`
- [x] `recordEventCooldown_givenMultipleKeys_expectAllPersisted`
- [x] `purgeExpiredCooldowns_givenExpiredKeys_expectRemoved`
- [x] `purgeExpiredCooldowns_givenEmptyKeys_expectNoChange` (early-return)
- [x] `clearEventCooldowns_givenCooldowns_expectAllRemoved`
- [x] `recordEventCooldown_givenNewStorageInstance_expectLoadsFromDisk` (cross-instance persistence)
- [x] `recordEventCooldown_givenSecondCall_expectOverwritesOnDisk` (overwrite semantics)
- [x] `concurrentOperations_expectNoCrash` (concurrent safety smoke test)
- [x] 9 tests, all passing
- [x] Full SDK build clean
- [x] `make generate && make format && make lint` — zero new errors
- [x] Zero warnings under `-strict-concurrency=complete`

---

> [!NOTE]
> **Medium Risk**
> Introduces new on-disk persistence for geofence state/cooldowns; file I/O and data-protection attributes can fail silently and could affect geofence event dedup behavior if serialization/pathing is wrong.
> 
> **Overview**
> Adds a new persisted `GeofenceState` model (Codable/Sendable) and an actor-based `GeofenceStorage` that reads/writes this state to an Application Support JSON file using iOS Data Protection and excluding it from backups.
> 
> Exposes initial public operations for geofence transition *event cooldown* management (`get`, `record`, `purge`, `clear`), and adds unit tests covering persistence semantics and basic concurrency safety.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit adf66971013482e91b718aa7acc247b91a139bb3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Sources/Location/Geofence/Storage/GeofenceStorage.swift
+++ b/Sources/Location/Geofence/Storage/GeofenceStorage.swift
@@ -1,0 +1,132 @@
+import CioInternalCommon
+import Foundation
+
+/// Thread-safe persistence for geofence state.
+///
+/// Implemented as an actor so all read-modify-write sequences are naturally atomic via
+/// actor isolation — no locks, no `@unchecked Sendable`. State is persisted to a JSON
+/// file with iOS Data Protection (`completeUntilFirstUserAuthentication`) so geofence
+/// callbacks that fire while the app is killed can still read cooldowns after the first
+/// user unlock. The file is excluded from backups (geofence cache is device-local context).
+///
+/// Each public method that mutates state performs the load → modify → save sequence
+/// synchronously within the actor, so no `await` interleaves and updates can never be
+/// lost to reentrancy.
+actor GeofenceStorage {
+    private static let defaultSubdirectory = "io.customer.sdk.geofence"
+    private static let filename = "geofenceState.json"
+    private static let protection = FileProtectionType.completeUntilFirstUserAuthentication
+
+    private let fileManager: FileManager
+    private let directoryURL: URL?
+
+    /// - Parameters:
+    ///   - fileManager: File manager used for I/O. Defaults to `.default`.
+    ///   - directoryURL: Directory for the state file. If `nil`, uses Application Support in the app container.
+    init(
+        fileManager: FileManager = .default,
+        directoryURL: URL? = nil
+    ) {
+        self.fileManager = fileManager
+        self.directoryURL = directoryURL
+    }
+
+    // MARK: - Event Cooldowns
+
+    func getEventCooldowns() -> [String: Date] {
+        loadFromDisk()?.eventCooldowns ?? [:]
+    }
+
+    func recordEventCooldown(key: String, timestamp: Date) {
+        var state = loadFromDisk() ?? GeofenceState()
+        var cooldowns = state.eventCooldowns ?? [:]
+        cooldowns[key] = timestamp
+        state.eventCooldowns = cooldowns
+        saveToDisk(state)
+    }
+
+    func purgeExpiredCooldowns(keys: Set<String>) {
+        guard !keys.isEmpty else { return }
+        var state = loadFromDisk() ?? GeofenceState()
+        var cooldowns = state.eventCooldowns ?? [:]
+        for key in keys {
+            cooldowns.removeValue(forKey: key)
+        }
+        state.eventCooldowns = cooldowns
+        saveToDisk(state)
+    }
+
+    func clearEventCooldowns() {
+        var state = loadFromDisk() ?? GeofenceState()
+        state.eventCooldowns = nil
+        saveToDisk(state)
+    }
+
+    // MARK: - Private (file persistence)
+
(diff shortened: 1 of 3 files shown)


Sources/Location/Geofence/Storage/GeofenceStorage.swift
```diff
@@ -0,0 +1,132 @@
...
+/// Implemented as an actor so all read-modify-write sequences are naturally atomic via
+/// actor isolation — no locks, no `@unchecked Sendable`. State is persisted to a JSON
+/// file with iOS Data Protection (`completeUntilFirstUserAuthentication`) so geofence
+/// callbacks that fire while the app is killed can still read cooldowns after the first
+/// user unlock. The file is excluded from backups (geofence cache is device-local context).
+///
+/// Each public method that mutates state performs the load → modify → save sequence
+/// synchronously within the actor, so no `await` interleaves and updates can never be
+/// lost to reentrancy.
+actor GeofenceStorage {
```
mahmoud-elmorabea, line 15: The thread safety guarantee is per instance, if there ever a second instance it might corrupt the file. Let's make sure that for a single file no second instance could exist

Sources/Location/Geofence/Storage/GeofenceStorage.swift
```diff
@@ -0,0 +1,132 @@
...
+    private static let defaultSubdirectory = "io.customer.sdk.geofence"
+    private static let filename = "geofenceState.json"
+    private static let protection = FileProtectionType.completeUntilFirstUserAuthentication
```
mahmoud-elmorabea, line 18: 👍

mahmoud-elmorabea (approved):

Sources/Location/Geofence/Storage/GeofenceStorage.swift
mrehan27: Good call, will take care of this when adding to DI and utilizing this class in upcoming PRs.

PR #1049, opened 2026-05-12
refactor: Added Synchronized primitive and tests for it

hollyschilling-cio:
Added the synchronized primitive type for use elsewhere in the SDK. It is unused except for in the tests. 

---

> [!NOTE]
> **Medium Risk**
> Introduces a new locking-based concurrency primitive and related operator/subscript overloads, which can be subtly incorrect under contention or lead to unexpected non-atomic compound mutations. Currently appears isolated (mainly exercised by new tests), limiting blast radius.
> 
> **Overview**
> Adds a new `Synchronized<T>` reference wrapper that guards a stored value with an `NSRecursiveLock`, exposing `wrappedValue`, `mutating`, `using`, and `atomicSetAndFetch` for thread-safe access.
> 
> Extends `Synchronized` with convenience APIs and protocol conformances for common types: arithmetic operators for `AdditiveArithmetic`, comparisons for `Comparable`, `toggle()` for `Bool`, collection helpers (count/isEmpty, subscript, append/insert/removeAll), `Equatable` with lock-ordering to avoid ABBA deadlocks, `Hashable`, and dictionary-style access via a new `DictionaryProtocol`.
> 
> Adds a comprehensive `SynchronizedTests` suite covering core behavior, operators, collection/dictionary helpers, hashing, and a concurrency regression test for `==` deadlock avoidance, and updates the generated `CioInternalCommon.api` to include the new public symbols.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8d545e5ab94d635db9e39a87ca3499f73ba4dad4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Sources/Common/Synchronized/Synchronized+Dictionaries.swift
+++ b/Sources/Common/Synchronized/Synchronized+Dictionaries.swift
@@ -1,0 +1,41 @@
+public protocol DictionaryProtocol {
+    associatedtype Key: Hashable
+    associatedtype Value
+
+    subscript(key: Key) -> Value? { get set }
+    mutating func removeValue(forKey key: Key) -> Value?
+}
+
+extension Dictionary: DictionaryProtocol {}
+
+extension Synchronized: DictionaryProtocol where T: DictionaryProtocol {
+    public typealias Key = T.Key
+    public typealias Value = T.Value
+
+    /// Accesses the value associated with `key` in a thread-safe manner.
+    ///
+    /// - Warning: The getter and setter each acquire the lock independently.
+    ///   Chained compound mutations such as `sync["key"]?.property = value` are
+    ///   **not** atomic — use `mutating { $0["key"]?.property = value }` instead.
+    public subscript(key: Key) -> Value? {
+        get {
+            using { value in
+                value[key]
+            }
+        }
+        set {
+            mutating { value in
+                value[key] = newValue
+            }
+        }
+    }
+
+    @discardableResult
+    public func removeValue(forKey key: Key) -> Value? {
+        var result: Value?
+        mutating { value in
+            result = value.removeValue(forKey: key)
+        }
+        return result
+    }
+}
--- a/Sources/Common/Synchronized/Synchronized+Dictionaries.swift
+++ b/Sources/Common/Synchronized/Synchronized+Dictionaries.swift
@@ -32,10 +32,8 @@ extension Synchronized: DictionaryProtocol where T: DictionaryProtocol {
 
     @discardableResult
     public func removeValue(forKey key: Key) -> Value? {
-        var result: Value?
         mutating { value in
-            result = value.removeValue(forKey: key)
+            value.removeValue(forKey: key)
         }
-        return result
     }
 }
--- a/Sources/Common/Synchronized/Synchronized.swift
+++ b/Sources/Common/Synchronized/Synchronized.swift
@@ -1,0 +1,52 @@
+import Dispatch
+import Foundation
+
+/// A wrapper for primitive types to make them thread safe and able to conform to `Sendable`.
+public final class Synchronized<T>: @unchecked Sendable {
+    private let lock = NSRecursiveLock()
+    private var _wrappedValue: T
+    public var wrappedValue: T {
+        get {
+            lock.withLock {
+                _wrappedValue
+            }
+        }
+        set {
+            lock.withLock {
+                _wrappedValue = newValue
+            }
+        }
+    }
+
+    public init(_ initial: T) {
+        self._wrappedValue = initial
+    }
+
+    /// Modify the wrapped value in a thread-safe manor.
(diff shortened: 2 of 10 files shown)


Sources/Common/Synchronized/Synchronized.swift
```diff
@@ -0,0 +1,52 @@
+import Dispatch
+import Foundation
+
+/// A wrapper for primitive types to make them thread safe and able to conform to `Sendable`.
+public final class Synchronized<T>: @unchecked Sendable {
+    private let lock = NSRecursiveLock()
```
mahmoud-elmorabea, line 6: From previous analysis that I've done for similar concept, I found that many cases in our code base: Singletons, config .. and the like are very read heavy and write rarely.

Should we consider something that wouldn't lock if multiple reads are happening with no writes and only locks if a single write is happening?

Sources/Common/Synchronized/Synchronized.swift
```diff
@@ -0,0 +1,52 @@
+import Dispatch
+import Foundation
+
+/// A wrapper for primitive types to make them thread safe and able to conform to `Sendable`.
```
mahmoud-elmorabea, line 4: By "primitive" do we mean not suitable for classes?

mahmoud-elmorabea (commented):
I am concerned about some of these conveniences causing subtle double locking that could potentially cause dead-locks. I prefer to start with using `mutating`, `using` and `atomicSetAndFetch` and retroactively add conveniences as patterns emerge

Sources/Common/Synchronized/Synchronized.swift
hollyschilling-cio: My original implementation of this class had a reader/writer lock. This is reflected in the distinction between `using` and `mutating` methods in this class. Such a lock has major drawbacks. It takes much longer to acquire that lock. It also can cause unbounded starvation for the writer thread if readers continuously come in.

Sources/Common/Synchronized/Synchronized.swift
hollyschilling-cio: No, by primitive I mean a class that does not implement thread safety itself.

Sources/Common/Synchronized/Synchronized.swift
```diff
@@ -0,0 +1,52 @@
...
+
+    /// Sets a new value while returning the old one in one atomic operation..
+    public func atomicSetAndFetch(_ newValue: T) -> T {
+        mutating {
+            let oldValue = $0
+            $0 = newValue
+            return oldValue
+        }
+    }
+}
```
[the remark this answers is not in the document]
hollyschilling-cio: As discussed elsewhere, the `Atomic` property wrapper does very little to provide thread safety and is being removed.

Sources/Common/Synchronized/Synchronized+Dictionaries.swift
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    @discardableResult
+    public func removeValue(forKey key: Key) -> Value? {
+        var result: Value?
+        mutating { value in
+            result = value.removeValue(forKey: key)
+        }
+        return result
+    }
```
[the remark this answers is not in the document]
hollyschilling-cio: Accurate. Now fixed.

Sources/Common/Synchronized/Synchronized.swift
```diff
@@ -0,0 +1,52 @@
...
+    /// - Parameters:
+    ///  - body: The code to access the wrapped value. The return value is passed through and returned to the caller, leaving the original value unchanged.
+    public func using<Result>(_ body: (T) throws -> Result) rethrows -> Result {
+        try lock.withLock {
+            try body(_wrappedValue)
+        }
+    }
+
+    /// Sets a new value while returning the old one in one atomic operation..
+    public func atomicSetAndFetch(_ newValue: T) -> T {
```
mrehan27, line 45: I'd slightly agree with Cursor's comment here. Since we're returning old value, we should probably name it something like `atomicFetchAndSet` instead to make it clear that the fetch happens first. e.g. the similar Android class uses names like [`getAndUpdate`](https://developer.android.com/reference/java/util/concurrent/atomic/AtomicReference#getAndUpdate(java.util.function.UnaryOperator%3CV%3E)) and [`updateAndGet`](https://developer.android.com/reference/java/util/concurrent/atomic/AtomicReference#updateAndGet(java.util.function.UnaryOperator%3CV%3E)) for this kind of clarity.

mrehan27 (approved):
We should probably also add a note to `Atomic` that `Synchronized` should be preferred since it's the newer and safer option. That would also help agents understand the difference better.

mahmoud-elmorabea (approved):

PR #1050, opened 2026-05-12
feat: Added InstallationId to the SDK and attached it to `context.device`

hollyschilling-cio:
This PR adds `installationId` to the SDK for use with Live Activities. It also adds that value as `context.device.installationId`. Per conversations with Colby, this is the intended place to send this field along. 

Currently, this PR includes clearing the installationId when `clearIdentify` is called. I'm not certain that is the right call at the moment and have added a comment to inspire discussion. I can see the value of each possibility.

---

> [!NOTE]
> **Medium Risk**
> Introduces a new persisted per-install identifier and injects it into analytics request context, which can affect downstream event schemas and data consistency across upgrades.
> 
> **Overview**
> Adds a new `installationId` surface area to the public SDK (`CustomerIOInstance`, `CustomerIO`, and `DataPipeline`) and internal storage (`GlobalDataStore`/`KeyValueStorageKey`) to provide a stable per-app-install identifier.
> 
> `DataPipelineImplementation` now generates and persists the UUID on first use and the `Context` plugin attaches it to outgoing payloads as `context.device.installationId`; autogenerated mocks and API docs are updated accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c2d1d944f423ea475a61bf81862b6ba68a58a02b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Sources/Common/Store/GlobalDataStore.swift
+++ b/Sources/Common/Store/GlobalDataStore.swift
@@ -2,6 +2,9 @@ import Foundation
 
 /// SDK data that is common between all site ids.
 public protocol GlobalDataStore: AutoMockable {
+    // Stable per-installation UUID, generated on first use and never reset
+    var installationId: String? { get set }
+
     // APN or FCM device token
     var pushDeviceToken: String? { get set }
 
@@ -12,6 +15,11 @@ public protocol GlobalDataStore: AutoMockable {
 // sourcery: InjectRegisterShared = "GlobalDataStore"
 public class CioSharedDataStore: GlobalDataStore {
     private let keyValueStorage: SharedKeyValueStorage
+
+    public var installationId: String? {
+        get { keyValueStorage.string(.installationId) }
+        set { keyValueStorage.setString(newValue, forKey: .installationId) }
+    }
 
     public var pushDeviceToken: String? {
         get {
--- a/Sources/DataPipeline/DataPipelineImplementation.swift
+++ b/Sources/DataPipeline/DataPipelineImplementation.swift
@@ -57,6 +57,7 @@ class DataPipelineImplementation: DataPipelineInstance, DataPipelineTracking {
 
         // plugin to update context properties for each request
         analytics.add(plugin: contextPlugin)
+        contextPlugin.installationId = installationId
 
         // plugin that adds provider attributes (e.g. location) to identify context
         analytics.add(plugin: IdentifyContextPlugin(registry: diGraph.profileEnrichmentRegistry, logger: logger))
--- a/Sources/DataPipeline/DataPipelineImplementation.swift
+++ b/Sources/DataPipeline/DataPipelineImplementation.swift
@@ -211,10 +211,6 @@ class DataPipelineImplementation: DataPipelineInstance, DataPipelineTracking {
         logger.debug("resetting user profile")
         analytics.reset()
 
-        // TODO: pending group design decision — installationId may be device-scoped and should survive clearIdentify.
-        // Remove these two lines if the group decides it should not reset.
-        globalDataStore.installationId = nil
-        contextPlugin.installationId = installationId
     }
 
     func deleteDeviceToken() {
--- a/Sources/DataPipeline/DataPipelineImplementation.swift
+++ b/Sources/DataPipeline/DataPipelineImplementation.swift
@@ -64,8 +64,8 @@ class DataPipelineImplementation: DataPipelineInstance, DataPipelineTracking {
         }
 
         // plugin to update context properties for each request
-        analytics.add(plugin: contextPlugin)
         contextPlugin.installationId = installationId
+        analytics.add(plugin: contextPlugin)
 
         // plugin that adds provider attributes (e.g. location) to identify context
         analytics.add(plugin: IdentifyContextPlugin(registry: diGraph.profileEnrichmentRegistry, logger: logger))
--- a/Sources/DataPipeline/DataPipelineImplementation.swift
+++ b/Sources/DataPipeline/DataPipelineImplementation.swift
@@ -117,6 +117,15 @@ class DataPipelineImplementation: DataPipelineInstance, DataPipelineTracking {
         analytics.identify(traits: traits)
     }
 
+    var installationId: String {
+        if let existing = globalDataStore.installationId {
+            return existing
+        }
+        let newId = UUID().uuidString
+        globalDataStore.installationId = newId
+        return newId
+    }
+
     var registeredDeviceToken: String? {
         globalDataStore.pushDeviceToken
     }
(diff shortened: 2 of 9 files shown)


mahmoud-elmorabea (changes requested):
I think we might want to sync with backend first.

There are two paths where tokens is sync'd today:
- With every event when token is available -> covered in this PR
- With `Device Created or Updated` events

With the new semantics we might even want to send `Device Created or Updated` earlier before a token is available with a the new stable installation id

hollyschilling-cio:
Good call flagging the sync paths. On the token-timing concern: PTS tokens do go out early and could technically race with the first identify call due to batching, but PTS tokens won't actually be used until later in the app lifecycle, so the backend having installationId in time isn't a problem in practice.

More broadly, the identify event is actually the right place for the backend to capture installationId rather than Device Created or Updated — Live Activities don't require push notifications to be enabled, so a push token may never be available on a given device. Identify is the reliable common path.

The upgrade path is the more interesting gap here — existing users won't have a stored installationId until after upgrade, and we can't rely on a timely identify call to surface it. I have a plan for that: on first launch after upgrade, we trigger identify-like behavior as though a profile attribute changed, which will carry the new installationId to the backend. That's something I'd intentionally deferred to keep this PR focused, but it needs to land before release. Happy to loop in the backend team on the specifics of that approach.

api-docs/CioDataPipelines.api
```diff
@@ -14,6 +14,7 @@ public final class CioDataPipelines.DataPipeline {
 	public var moduleConfig: DataPipelineConfigOptions?;
 	public var profileAttributes: Map<String, Any>;
 	public var deviceAttributes: Map<String, Any>;
+	public var installationId: String;
```
mahmoud-elmorabea, line 17: Are we intentionally exposing this?

mahmoud-elmorabea: I don't think this should be exposed

Sources/Common/Store/GlobalDataStore.swift
```diff
@@ -13,6 +16,11 @@ public protocol GlobalDataStore: AutoMockable {
 public class CioSharedDataStore: GlobalDataStore {
     private let keyValueStorage: SharedKeyValueStorage
 
+    public var installationId: String? {
+        get { keyValueStorage.string(.installationId) }
+        set { keyValueStorage.setString(newValue, forKey: .installationId) }
+    }
+
```
mahmoud-elmorabea, line 23: There is a `deleteAll()` method here that would clear this but it doesn't seem to be called anywhere

Sources/DataPipeline/DataPipelineImplementation.swift
```diff
@@ -56,6 +65,7 @@ class DataPipelineImplementation: DataPipelineInstance, DataPipelineTracking {
 
         // plugin to update context properties for each request
         analytics.add(plugin: contextPlugin)
+        contextPlugin.installationId = installationId
```
mahmoud-elmorabea, line 67: I think we can move the `installationId` to be constructor param for the contextPlugin, that way its not optional and there is no chance any events miss it. This makes it non-optional type as well

mahmoud-elmorabea (changes requested):
Looks good overall. Can we please:
- Not expose `installId` as public API
- Avoid a race by injecting `installId` into the context plugin when its created since the value is available

api-docs/CioDataPipelines.api
hollyschilling-cio: Yes, it must be exposed so partners can reference it. Its whole purpose is for partners to use it.

Sources/Common/Store/GlobalDataStore.swift
hollyschilling-cio: `CioSharedDataStore`'s `deleteAll` method directly calls `SharedKeyValueStorage`'s `deleteAll` method. `SharedKeyValueStorage` has two primary implementations: `UserDefaultsSharedKeyValueStorage` and `InMemorySharedKeyValueStorage`. In both cases, `deleteAll` deletes all data without needing to know the key names.

Sources/DataPipeline/DataPipelineImplementation.swift
hollyschilling-cio: We could, but there is no advantage to this. Leaving it as nullable and mutable allows us the flexibility to reverse decision later if we decide the installationId must be changed at clearIdentify.

api-docs/CioDataPipelines.api
mahmoud-elmorabea: Not really, We use it to map PTS tokens along with user ID? Are we going to document that partners need to keep mapping of each of their users/profiles and list of installation ids? 

I think it might be useful in the future for some use cases we don't know about, but I strongly suggest we leave it out of scope for now.

Sources/Common/Store/GlobalDataStore.swift
mahmoud-elmorabea: But this will break the "install ids" don't clear when reset event happens, correct?

Sources/DataPipeline/DataPipelineImplementation.swift
mahmoud-elmorabea: Then at least we should move it to before the plugin is registered, why risk the plugin receiving an event before the install id is set?

mahmoud-elmorabea (approved):

Sources/Common/Store/GlobalDataStore.swift
hollyschilling-cio: `deleteAll` cannot be used in production because it deletes even the host app's data. It indiscriminately deletes all data is UserDefaults and is only used for testing contexts. It even says that in the file only a few lines away at GlobalDataStore.swift:11

```
// Used for testing
func deleteAll()
```

api-docs/CioDataPipelines.api
hollyschilling-cio: This was discussed almost a month ago, at which time you didn't understand how the partners would know this value. This is how they know the value.
[https://www.notion.so/Partner-Server-Integration-3474302f4c2b81b384a1f710fa757bf6?d=34b4302f4c2b804b95c7001c091b30a7&source=copy_link#b039c[…]409fb2](https://www.notion.so/Partner-Server-Integration-3474302f4c2b81b384a1f710fa757bf6?d=34b4302f4c2b804b95c7001c091b30a7&source=copy_link#b039c6e56b194868bb62991177409fb2)

Sources/DataPipeline/DataPipelineImplementation.swift
hollyschilling-cio: That's fair.

PR #1052, opened 2026-05-14
chore(messaginginapp): dedupe messageShown opened metric by deliveryID

mahmoud-elmorabea:
Dedupes the in-app `messageShown` `opened` metric post by `campaignDeliveryId` within the SDK process lifetime. First show of a given delivery posts the metric; subsequent shows of the same delivery skip the post.

The host `eventListener.messageShown` callback still fires on every call, so app-level UI side-effects are unaffected — only the analytics metric emission is deduped.

Backend already dedupes `(deliveryID, opened)` pairs server-side, so this is bandwidth and analytics-queue noise reduction, not a metric-correctness change.

### DIFF
--- a/Sources/MessagingInApp/Gist/GistDelegate.swift
+++ b/Sources/MessagingInApp/Gist/GistDelegate.swift
@@ -23,6 +23,16 @@ class GistDelegateImpl: GistDelegate {
     private let eventBusHandler: EventBusHandler
     private var eventListener: InAppEventListener?
     private let threadUtil: ThreadUtil
+
+    // In-memory set of `campaignDeliveryId` values for which we've already posted
+    // a `messageShown` (`opened`) metric this session. Used to suppress duplicate
+    // metric emissions when Gist invokes `messageShown` more than once for the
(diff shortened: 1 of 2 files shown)


Sources/MessagingInApp/Gist/GistDelegate.swift
```diff
@@ -39,7 +49,11 @@ class GistDelegateImpl: GistDelegate {
         logger.logWithModuleTag("Message shown: \(message.describeForLogs)", level: .debug)
 
         if let deliveryId = message.campaignDeliveryId {
-            eventBusHandler.postEvent(TrackInAppMetricEvent(deliveryID: deliveryId, event: InAppMetric.opened.rawValue))
+            if shouldPostMessageShownMetric(forDeliveryId: deliveryId) {
```
mrehan27, line 52: Same questions as android around persistent messages and about expected behavior.

Also, should we apply the same checks to `clicked` metrics as well?

mahmoud-elmorabea:
I was unable to convince myself that this could happen in real production scenario but rather only a technical possibility, dropping this for now

PR #1053, opened 2026-05-14
chore: dedup identify calls within a session when no traits supplied

mahmoud-elmorabea:
First `identify(userId)` per session always fires fully — preserves device-token re-registration and Device Created or Updated cadence.

Subsequent `identify(userId)` calls with the same userId and no traits short-circuit (no outbound identify, no device-token path). `clearIdentify()` resets the session marker. Calls with traits — including an explicit empty dict — always pass through.

Targets the misuse pattern of identify-on-every-event within a session. Per-session scope ensures token rotation still propagates on cold start.

### DIFF
--- a/Sources/DataPipeline/DataPipelineImplementation.swift
+++ b/Sources/DataPipeline/DataPipelineImplementation.swift
@@ -15,6 +15,13 @@ class DataPipelineImplementation: DataPipelineInstance, DataPipelineTracking {
     private let contextPlugin: Context
     private let profileStore: ProfileStore
 
+    /// Per-session tracker of the most recently identified userId. Used to dedup
+    /// no-traits identify calls against the same userId within a single process
+    /// lifetime. Reset by `clearIdentify()`. Not persisted across launches so
+    /// that the first identify per cold-start always runs the full path
+    /// (device-token re-registration + DCoU refresh).
+    private var lastIdentifiedUserIdThisSession: String?
+
     init(diGraph: DIGraphShared, moduleConfig: DataPipelineConfigOptions) {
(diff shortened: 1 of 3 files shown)


Sources/DataPipeline/DataPipelineImplementation.swift
```diff
@@ -15,6 +15,13 @@ class DataPipelineImplementation: DataPipelineInstance, DataPipelineTracking {
     private let contextPlugin: Context
     private let profileStore: ProfileStore
 
+    /// Per-session tracker of the most recently identified userId. Used to dedup
+    /// no-traits identify calls against the same userId within a single process
+    /// lifetime. Reset by `clearIdentify()`. Not persisted across launches so
+    /// that the first identify per cold-start always runs the full path
+    /// (device-token re-registration + DCoU refresh).
+    private var lastIdentifiedUserIdThisSession: String?
```
mrehan27, line 23: Same question as Android, could we use `registeredUserId` here instead of introducing a new `lastIdentifiedUserIdThisSession`?

mahmoud-elmorabea: Yes, this is deliberate. Using `analytics.userId()` / `registeredUserId` would persist the dedup across cold starts and an app that only does it once per session ends up doing it once per its lifetime. 

I am trying to prevent abuse or buggy code that calls identify way too much in a single session. But once per app launch is okay for now.

Once we finish our data investigation we might decide to restrict this even more, but for now. This prevents abuse and limits the blast radius of the change.

mrehan27 (approved):

PR #1054, opened 2026-05-16
chore: buffer track/identify/screen calls made before initialize completes

mahmoud-elmorabea:
## Summary
- Adds a bounded in-memory FIFO (capacity 100, drop-most-recent on overflow) that absorbs event-shaped public-API calls invoked before `CustomerIO.initialize()` completes. Calls drain in order against the real implementation once initialization finishes.
- Previously, calls like `CustomerIO.shared.track(...)` made before initialize evaluated to `implementation?.method(...)` and silently no-op'd, losing data common in cold-start, deep-link, and restored-state flows.

This is item 2 of a cross-SDK plan to port targeted improvements from the proposed iOS SDK rewrite (`customerio-ios-tng`) back to the current SDK. Android parity ships in a separate PR on customerio-android (same branch name).

## Scope of buffering
All event-shaped methods on the `CustomerIO` facade route through the buffer when `implementation == nil`:
- `track`, `identify`, `screen` (incl. deprecated Codable variants)
- `clearIdentify`, `setProfileAttributes`, `setDeviceAttributes`
- `registerDeviceToken`, `deleteDeviceToken`, `trackMetric`

Read-side accessors (`registeredDeviceToken`, `profileAttributes` getter, `deviceAttributes` getter) are unchanged — there's nothing to buffer.

## Decisions worth flagging
- **Buffer cap: 100.** Matches the rewrite.
- **Overflow policy: drop most recent.** Deliberate divergence from the rewrite (which drops oldest). Preserves the earliest events — install attribution, initial identify — which tend to be the highest-value early signals.
- **No `initialize()` ever called → buffer sits at cap.** No TTL, no cleanup. Memory is bounded by the cap.
- **`initialize()` fails → buffer remains intact.** A subsequent retry will drain it.
- **Debug-level logs on overflow drops and drain.** Counter tracked across pre-init; logged once on drain.
- **Order preserved through identify boundaries.** A sequence like `[identify(A), track(x), clearIdentify, identify(B), track(y)]` drains in that exact order; the existing pipeline handles attribution correctly post-replay.

## Implementation
- `PreInitEventBuffer` is a small `Synchronized<State>`-backed FIFO with three states: `.buffering`, `.draining`, `.ready`. Block execution always happens outside the lock to avoid re-entrancy.
- `CustomerIO` owns a buffer instance. Public methods route through a tiny `dispatch(_:)` helper that either calls the real `implementation` or enqueues onto the buffer.
- `initializeSharedInstance` (and `setUpSharedInstanceForUnitTest`) call `buffer.transitionToReady(impl)` synchronously **before** assigning `implementation`. This guarantees strict ordering — buffered events drain before any post-init direct call.

## Test plan
- [x] 9 new unit tests in `Tests/Common/PreInitEventBuffer/`: state transitions, overflow, drain order, reentrancy, concurrent enqueue, post-init direct execution, idempotent transition.
- [x] All existing `CommonTests` pass.
- [x] All existing `DataPipelineTests` pass.
- [ ] CI green on this PR.
- [ ] Local wrapper validation (RN + Flutter sample apps with deliberate pre-`initialize()` `track()` call lands events on the server) — done locally, no PRs.

---

> [!NOTE]
> **Medium Risk**
> Changes the `CustomerIO` facade to buffer and later replay pre-initialization calls, altering event delivery timing and device-token registration order; concurrency/ordering bugs could affect analytics and push registration behavior.
> 
> **Overview**
> **Pre-init API calls are no longer dropped.** `CustomerIO` now routes event-shaped methods through a new bounded FIFO `PreInitEventBuffer`, capturing calls made while `implementation` is `nil` and replaying them synchronously, in order, during `initializeSharedInstance` (and unit-test setup).
> 
> **Initialization ordering and token behavior changed.** Initialization now runs `postInitialize` before draining the buffer and only publishes `implementation` after the drain to avoid races; `registerDeviceToken` gets special handling to prevent duplicate registration (and falls back to the stored token if the buffer is at capacity and drops the call). New tests cover drain order, overflow policy (drop most-recent), re-entrancy/concurrency, and the token-registration edge cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c2508c67ad6408d18030be750d8dfbab244325f3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Sources/Common/CustomerIOInstance.swift
+++ b/Sources/Common/CustomerIOInstance.swift
@@ -215,7 +215,7 @@ public class CustomerIO: CustomerIOInstance {
     /// authoritative replay. Prevents the duplicate "Device Created or Updated"
     /// that would otherwise fire when both `postInitialize` and the buffered
     /// `registerDeviceToken` target the same stored token.
-    @Atomic private var hasPendingTokenRegistration: Bool = false
+    private let hasPendingTokenRegistration = Synchronized<Bool>(false)
 
     /// private constructor to force use of singleton API
     private init() {}
@@ -226,11 +226,11 @@ public class CustomerIO: CustomerIOInstance {
 
     @discardableResult
     static func setUpSharedInstanceForUnitTest(implementation: CustomerIOInstance) -> CustomerIO {
-        // Assign the implementation before draining so the buffered closures
-        // run against a fully wired SDK and the global token state is in place
-        // by the time buffered token-dependent calls replay.
-        shared.implementation = implementation
+        // Drain the buffer against the impl *before* publishing it on `shared`,
+        // so concurrent `dispatch` calls can't bypass replay while pre-init
+        // events are still queued.
         shared.preInitEventBuffer.transitionToReady(implementation)
+        shared.implementation = implementation
         return shared
     }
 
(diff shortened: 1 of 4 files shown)


Sources/Common/CustomerIOInstance.swift
```diff
@@ -203,7 +203,21 @@ public class CustomerIO: CustomerIOInstance {
...
+    /// new implementation. After drain, subsequent calls bypass the buffer.
+    let preInitEventBuffer = PreInitEventBuffer()
+
+    /// Flag set when a `registerDeviceToken` call is buffered pre-init. Read by
+    /// `postInitialize()` to skip its own stored-token registration so the
+    /// buffered call (which carries the caller's most recent token) is the
+    /// authoritative replay. Prevents the duplicate "Device Created or Updated"
+    /// that would otherwise fire when both `postInitialize` and the buffered
+    /// `registerDeviceToken` target the same stored token.
+    @Atomic private var hasPendingTokenRegistration: Bool = false
```
mahmoud-elmorabea: Let's use `Synchronized` here

hollyschilling-cio (changes requested):
There is a premade class that handles most of this. Please look into [`AsyncStream`](https://developer.apple.com/documentation/swift/asyncstream).

PR #1056, opened 2026-05-18
feature: Config and Config Builder

hollyschilling-cio:
Added the Config and Config Builder to the LiveActivities branch. This includes the listeners for change tokens which will be used in later stages.

---

> [!NOTE]
> **Low Risk**
> New public module and local token file storage only; no network or event-bus integration yet, and module init does not start observers.
> 
> **Overview**
> Introduces a new **`LiveActivities`** Swift package product and **`CioLiveActivities`** target, with a public **`LiveActivitiesModule.initialize`** entry point (non-singleton) meant to run after **`CustomerIO.initialize`**.
> 
> Partners configure the module via **`LiveActivityConfig`** / **`LiveActivityConfigBuilder`**: optional Live Activities API base URL and log level, App Group + asset pre-registration for widget extensions, and **`register(_:identifier:)`** for each **`CIOActivityAttribute`** type. Registration captures type-erased **`ActivityTypeRegistration`** closures that observe push-to-start tokens, per-instance push tokens, content-state JSON updates, lifecycle end, and can end all activities of a type on reset—though **`LiveActivitiesModule`** currently only logs on init; wiring those listeners to the backend is explicitly deferred.
> 
> Also adds **`CIOLiveActivitiesSDKProviding`** for testability, **`FileActivityTokenStore`** for persisted push-to-start token hex per activity type, and the **`CIOActivityAttribute`** protocol so apps use a stable **`activityInstanceId`** for server routing instead of ActivityKit’s system id.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b32b9b2289df96af56b95ec74ce8af895016f516. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Sources/LiveActivities/Config/LiveActivityConfigBuilder.swift
```diff
@@ -0,0 +1,194 @@
...
+
+    /// Set the base URL of the Live Activities API endpoint.
+    ///
+    /// - Note: This is a temporary field while backend endpoint paths are being finalised.
+    @discardableResult
+    public func baseURL(_ url: URL) -> Self {
+        var copy = self
+        copy.config.liveActivitiesBaseURL = url
+        return copy
+    }
```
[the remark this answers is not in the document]
hollyschilling-cio: Agreed and fixed.

Sources/LiveActivities/Config/LiveActivityConfigBuilder.swift
```diff
@@ -0,0 +1,194 @@
...
+                switch state {
+                case .ended, .dismissed, .stale:
+                    await onEnd(activityId)
+                case .active, .pending:
+                    break
+                @unknown default:
+                    break
+                }
+            }
+        }
```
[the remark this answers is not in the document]
hollyschilling-cio: Agreed and fixed.

Sources/LiveActivities/Config/LiveActivityConfigBuilder.swift
```diff
@@ -0,0 +1,194 @@
...
+                    if let data = try? encoder.encode(update.state) {
+                        await onStateUpdate(activityId, data)
+                    }
+                }
+            }
+            group.addTask {
+                for await state in activity.activityStateUpdates {
+                    switch state {
+                    case .ended, .dismissed, .stale:
+                        await onEnd(activityId)
```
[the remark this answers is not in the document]
hollyschilling-cio: Agreed and fixed.

Sources/LiveActivities/Storage/LiveActivityTokenStorage.swift
```diff
@@ -0,0 +1,118 @@
...
+            return directory.appendingPathComponent(Self.filename)
+        }
+        guard let appSupport = fileManager.urls(
+            for: .applicationSupportDirectory, in: .userDomainMask
+        ).first else { return nil }
+        return appSupport
+            .appendingPathComponent(Self.subdirectory)
+            .appendingPathComponent(Self.filename)
+    }
+}
```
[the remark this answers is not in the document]
hollyschilling-cio: Leaving for now, but this is a good idea to address in the future.

Sources/LiveActivities/Config/LiveActivityConfigBuilder.swift
```diff
@@ -0,0 +1,194 @@
...
+                for await state in activity.activityStateUpdates {
+                    switch state {
+                    case .ended, .dismissed:
+                        await onEnd(activityId)
+                    case .active, .pending, .stale:
+                        break
+                    @unknown default:
+                        break
+                    }
+                }
```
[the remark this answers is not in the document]
hollyschilling-cio: This is technically true, but more subtle. A Live Activity generally will experience a `dismissed` event or an `ended` event, not both. The only situation where both can occur is if the app programmatically ends the Live Activity with a future date (causing the `ended` event) and then the user manually dismisses it (causing the `dismissed` event). Given the need to reduce server load, this seems a logical place to prevent duplication. If there might be a benefit of this in the future, we can revisit.

Sources/LiveActivities/CIOLiveActivitiesSDKProviding.swift
```diff
@@ -0,0 +1,25 @@
...
+/// `CustomerIO.shared` singleton. Pass a conforming mock in unit tests; the default
+/// (`CustomerIO.shared`) is used in production.
+public protocol CIOLiveActivitiesSDKProviding {
+    var installationId: String { get }
+    var eventBusHandler: EventBusHandler { get }
+    var logger: Logger { get }
+}
+
+extension CustomerIO: CIOLiveActivitiesSDKProviding {
+    public var eventBusHandler: EventBusHandler {
```
mrehan27: Should these two be internal as well to avoid exposing to customers unnecessarily?

Sources/LiveActivities/Config/LiveActivityConfig.swift
```diff
@@ -0,0 +1,52 @@
...
+    /// widget extension. When `nil`, asset syncing is skipped.
+    var appGroupIdentifier: String?
+
+    /// Image assets declared for pre-loading into the AppGroup container at initialize time.
+    var assetRegistrations: [AssetRegistration]
+
+    /// Activity types registered for SDK observation via `LiveActivityConfigBuilder.register(_:identifier:)`.
+    var registrations: [ActivityTypeRegistration]
+
+    public init(liveActivitiesBaseURL: URL? = nil, logLevel: CioLogLevel? = nil) {
```
mrehan27: Since we have `LiveActivityConfigBuilder`, should the constructor be internal?

Sources/LiveActivities/CIOLiveActivitiesSDKProviding.swift
mrehan27: Should these two be `internal` as well to avoid exposing them to customers unnecessarily?

Sources/LiveActivities/LiveActivitiesModule.swift
```diff
@@ -0,0 +1,58 @@
...
+/// Call `initialize` after `CustomerIO.initialize(withConfig:)` and hold the returned
+/// instance for the lifetime of your app:
+/// ```swift
+/// CustomerIO.initialize(withConfig: config)
+/// let liveActivities = LiveActivitiesModule.initialize(
+///     LiveActivityConfigBuilder(baseURL: URL(string: "https://…")!)
+///         .build()
+/// )
+/// ```
+public final class LiveActivitiesModule {
```
mrehan27, line 15: Shouldn't we use `ModuleTopLevelObject` or `CustomerIOModule` here for consistency? It might also help with flows such as `initialize` method calls and keep the pattern aligned with rest of the modules.

Sources/LiveActivities/LiveActivitiesModule.swift
```diff
@@ -0,0 +1,58 @@
...
+    /// Call this after `CustomerIO.initialize(withConfig:)`. Hold the returned instance
+    /// for the lifetime of your app — it is not a singleton.
+    ///
+    /// - Parameter config: Module configuration built via `LiveActivityConfigBuilder`.
+    /// - Returns: The initialized module instance.
+    @discardableResult
+    public static func initialize(_ config: LiveActivityConfig) -> LiveActivitiesModule {
+        let module = LiveActivitiesModule(
+            config: config,
+            sdk: CustomerIO.shared,
```
mrehan27, line 33: With the current structure, passing the SDK instance directly here doesn't feel ideal (yet at least). If we think `EventBus` or other options can't cover this gap, or if this is really required here, could we introduce it in the PR where it's actually used? That would make the use case easier to understand.

Sources/LiveActivities/Config/LiveActivityConfigBuilder.swift
```diff
@@ -0,0 +1,195 @@
...
+    ///   - key: The string key used to retrieve this asset in the widget extension.
+    ///   - bundleResource: The resource name as passed to `Bundle.main.url(forResource:withExtension:)`.
+    ///   - withExtension: The file extension, or `nil` to match any extension.
+    public func registerAsset(
+        _ key: String,
+        bundleResource: String,
+        withExtension ext: String? = nil
+    ) -> Self {
+        guard let url = Bundle.main.url(forResource: bundleResource, withExtension: ext) else {
+            fatalError(
```
mrehan27: I believe `fatalError` can crash the app. Is it really the right option to use here or should we reconsider and handle this more gracefully?

Sources/LiveActivities/LiveActivitiesModule.swift
hollyschilling-cio: This module does not descend from `ModuleTopLevelObject` or adhere to `CustomerIOModule`. Descending from `ModuleTopLevelObject` prevents conforming to `Sendable` and adds no value to this class. `CustomerIOModule` is only for modules which DataPipelines must control, which is not this case.

Sources/LiveActivities/CIOLiveActivitiesSDKProviding.swift
hollyschilling-cio: This is an interesting point. The properties here must be public because the protocol definition is public, but maybe the protocol definition shouldn't be public. If I reduce the visibility of that protocol to internal, I should be able to make these internal as well and that prevents exposure on the interface. Tests still will need `@testable`, but that's fine because the init was `internal` already.

Sources/LiveActivities/Config/LiveActivityConfig.swift
hollyschilling-cio: Agreed. Updating.

Sources/LiveActivities/Config/LiveActivityConfigBuilder.swift
hollyschilling-cio: This method is only used for hard-coded assets included in the app bundle. The intent here is to force these errors to be discovered in test builds.

That said, I do hate the idea of forcing a crash on launch for something which we could easily recover from. I'm just not sure if this is better to catch early or defer.

Sources/LiveActivities/LiveActivitiesModule.swift
hollyschilling-cio: What about it doesn't feel ideal? 

There are a few reasons for passing it. The most immediate are that it avoids adding additional direct uses of DIGraphShared or additional internal dependencies on the Static Global instance `CustomerIO.shared`. By passing a handle through Dependency Injection, the Live Activities module gets access to the DataPipelines module for the installationId, sending events (`track`), as well as the EventBus and Logger. As DataPipelines is the "root" module all the modules that depend on it should have some way to reference it and the Static Global reference shouldn't be used internally.

mrehan27: The main reason I suggested `EventBus` instead of using the SDK directly is wrapper SDK behavior. In wrapper SDKs, the SDK may not be initialized immediately because initialization happens in JS/Dart, which only runs once the UI is created. That doesn't really guarantee early initialization during app launch (unlike AppDelegate or Application on Android). So assumptions like `CustomerIO.shared` being always ready, or `DataPipelines` being the root may not always hold true always, specially if the app was killed and the user doesn't open it from the notification, so it remains in a killed/background state. I just want to make sure we fully understand this use case and that the approach works reliably in those scenarios.

For the `installationId` point you shared, I understand the use case now. My concern is whether it will always be available for wrapper SDKs. `EventBus` may not solve that either, so do we need to consider another path, or do you think the current approach is sufficient?

Sources/LiveActivities/Config/LiveActivityConfigBuilder.swift
mrehan27: That makes sense. Do you think this could lead to cases where the app works fine in debug builds but crashes only in release builds? Should we consider an approach that crashes only in test/debug builds instead?

Sources/LiveActivities/LiveActivitiesModule.swift
hollyschilling-cio: That's a valid concern and we'll need to validate it. The same block of code that initializes the SDK should initialize LiveActivities right after it, but app startup is weird in wrappers and this would not be the weirdest thing I've run into by a lot.

Sources/LiveActivities/Config/LiveActivityConfigBuilder.swift
hollyschilling-cio: Different behavior in Debug builds VS production would be the best case. Unfortunately on iOS as a library, we can't reliably detect if this is a debug build or not. Our best option is to pick one and go with it. I'd be ok with recovering and logging an error here, but it is likely to be missed.

Sources/LiveActivities/Config/LiveActivityConfig.swift
```diff
@@ -0,0 +1,52 @@
...
+    ///
+    /// When `nil` (the default), the module registers with the SDK but makes no
+    /// network requests — useful during development before the endpoint is provisioned.
+    ///
+    /// - Note: This is a temporary field while backend endpoint paths are being finalised.
+    ///   It will be removed once the SDK derives the URL from the configured region automatically.
+    public var liveActivitiesBaseURL: URL?
+
+    /// Module-level log level override. When `nil`, the SDK-wide log level is used.
+    public var logLevel: CioLogLevel?
```
mahmoud-elmorabea, line 27: Awesome 👍 .. I think we should have log levels for all modules, makes debugging so much easier with less noise

Sources/LiveActivities/Config/LiveActivityConfig.swift
```diff
@@ -0,0 +1,52 @@
...
+        self.logLevel = logLevel
+        self.appGroupIdentifier = nil
+        self.assetRegistrations = []
+        self.registrations = []
+    }
+}
+
+/// A single image asset declared for pre-loading into the AppGroup container.
+struct AssetRegistration {
+    let key: String
```
mahmoud-elmorabea, line 50: What's the key going to be used for here? I think it makes sense if we didn't have asset pre-loading. The URL can be the key here

mahmoud-elmorabea: This avoids having to force people to keep track of key <-> URL mapping on backend/mobile apps

Sources/LiveActivities/Config/LiveActivityConfigBuilder.swift
```diff
@@ -0,0 +1,195 @@
...
+    ///
+    /// - Parameters:
+    ///   - type: The `ActivityAttributes` conformance to observe.
+    ///   - identifier: A stable reverse-DNS identifier for this activity type,
+    ///     e.g. `"io.customer.liveactivities.scoreboard"`. Used in API paths and
+    ///     matched server-side to route pushes to the correct devices. Must be
+    ///     consistent between the app and the Customer.io backend configuration.
+    #if os(iOS)
+    @available(iOS 17.2, *)
+    public func register<T: CIOActivityAttribute>(_ type: T.Type, identifier: String) -> Self {
```
mahmoud-elmorabea, line 60: A few suggestions that might make this simpler:
- I don't think this entire registration logic necessarily belongs in the config builder, can we move it out to an extension or its own type `ActivityAttributeRegistrar` or smth?
- Can we make the builder only store metadata about the type? Analog to Kotin/Java `Class` -> `any CIOActivityAttribute.Type`? and operate on that later in the init process? I understand that we may have preferred this due to type erasure of `T`

Sources/LiveActivities/Config/LiveActivityConfigBuilder.swift
```diff
@@ -0,0 +1,195 @@
...
+        var copy = self
+        let registration = ActivityTypeRegistration(
+            activityIdentifier: identifier,
+            startObserving: { onPushToStartToken, onInstancePushToken, onActivityObserved, onStateUpdate, onEnd in
+                Task {
+                    await withTaskGroup(of: Void.self) { group in
+                        group.addTask {
+                            for await token in Activity<T>.pushToStartTokenUpdates {
```
mahmoud-elmorabea, line 68: I am not sure how testable `Activity<T>.pushToStartTokenUpdates` is, if at all. But maybe we can wrap it in our own abstraction to unit test the callbacks do what we want them to do.

Sources/LiveActivities/Config/LiveActivityConfigBuilder.swift
```diff
@@ -0,0 +1,195 @@
...
+    ///     matched server-side to route pushes to the correct devices. Must be
+    ///     consistent between the app and the Customer.io backend configuration.
+    #if os(iOS)
+    @available(iOS 17.2, *)
+    public func register<T: CIOActivityAttribute>(_ type: T.Type, identifier: String) -> Self {
+        var copy = self
+        let registration = ActivityTypeRegistration(
+            activityIdentifier: identifier,
+            startObserving: { onPushToStartToken, onInstancePushToken, onActivityObserved, onStateUpdate, onEnd in
+                Task {
```
mahmoud-elmorabea, line 65: Do we need to manage/clean/stop these streams if a logout event happens?

Sources/LiveActivities/Config/LiveActivityConfigBuilder.swift
mahmoud-elmorabea: I agree that we shouldn't crash here, an error log is more appropriate. In any unforeseen cases where this fails at runtime in a release build for customers it's better to have a missing asset than a runtime crash.

Sources/LiveActivities/Storage/LiveActivityTokenStorage.swift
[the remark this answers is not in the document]
mahmoud-elmorabea: I think this is worth considering before we release this feature. If we do fixes/improvements to this later, it'd be great for both our stores to get them and not diverge

Sources/LiveActivities/CIOLiveActivitiesSDKProviding.swift
```diff
@@ -0,0 +1,25 @@
+import CioInternalCommon
+import Foundation
+
+/// The SDK capabilities that `LiveActivitiesModule` requires from the host SDK.
+///
+/// Declared as a protocol so the module can be tested without depending on the real
+/// `CustomerIO.shared` singleton. Pass a conforming mock in unit tests; the default
+/// (`CustomerIO.shared`) is used in production.
+public protocol CIOLiveActivitiesSDKProviding {
```
mahmoud-elmorabea: I think this entire type and its implementation/extension can be internal. We want to avoid exposing these properties in our public API

Sources/LiveActivities/LiveActivitiesModule.swift
mahmoud-elmorabea: 👍 To rehan's point, def. worth testing before this is released

mahmoud-elmorabea (approved):
Looks good!

Approving to unblock the PR but things worth considering in future PRs:
- Adding tests for the token registration/rotation logic
- Extracting logic outside of builder
- Sharing the file store logic

Sources/LiveActivities/Storage/LiveActivityTokenStorage.swift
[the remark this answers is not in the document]
hollyschilling-cio: My intent to leave it is because I want to use the DB for storage, which I'm adding for rate limiting and filtering.

Sources/LiveActivities/Config/LiveActivityConfig.swift
hollyschilling-cio: Long term, I like that idea. Unfortunately, that can't work today. The issue today is that with assets that are in the iOS App Bundle they don't know the URL until Install Time. Even then, they need to pass this value through the API endpoints to start/update the Live Activity. The key is part of the Payload for the Live Activity's content.

Sources/LiveActivities/Config/LiveActivityConfigBuilder.swift
hollyschilling-cio: Addressed by using `assertionFailure`. This will work unless we ship a precompiled binary.

Sources/LiveActivities/Config/LiveActivityConfigBuilder.swift
hollyschilling-cio: The typing is necessary because the LiveActivity methods all require the type information as a generic parameter, such as [Activity](https://developer.apple.com/documentation/activitykit/activity). In Swift, there is no way to go from type metadata to generic parameter. 

As for where the registration belongs, it must occur before all potential uses of a LiveActivity. By putting it here, there is no other interaction with the CIO SDK necessary throughout the rest of the app code. Everything else is handled within the SDK by the listeners and the App only interacts with the Apple SDK. The entire Live Activity CIO interface is just this configuration block.

Sources/LiveActivities/Config/LiveActivityConfigBuilder.swift
hollyschilling-cio: You're right that it is not testable, but this is built in a way that makes that fact irrelevant. All we do is get a token from that and immediately pass it to `onPushToStartToken`. The only thing a test would do is validate wiring, which is not a meaningful test.

Sources/LiveActivities/Config/LiveActivityConfigBuilder.swift
hollyschilling-cio: Yes, and that's why it's important that they are all anchored and cancelable. On `clearIdentify`, we must:
1. Close all open Live Activities that we created
2. Cancel the observation streams
3. Clear the PTS Token storage
4. Reset the `observedActivities` stream (not part of this PR)
5. Recreate observation streams.

PR #1059, opened 2026-05-20
[SPIKE / DO NOT MERGE] Segment storage injection: prove encryption-at-rest is feasible without forking

mahmoud-elmorabea:
## Evidence only — DO NOT MERGE

This branch exists to document a spike outcome. It will never be merged into `main`.

## Spike goal

Prove (or disprove) that we can fully influence every byte the Segment `cdp-analytics-swift` library persists to disk by injecting a `DataStore` via `Configuration.storageMode(.custom(...))` — without forking the library.

## TL;DR

Yes for the event queue. No for the UserDefaults plist.

- `.custom(DataStore)` cleanly intercepts every byte of every `track` / `screen` / `identify` event payload. All 11 wrapped event files in the spike run start with the `WRAPPED::v1::` sentinel; zero `SPIKE_*` needles leaked in plaintext anywhere in event storage. The 30 s Segment flush to `cdp.customer.io/v1` succeeded with the unwrapped JSON, proving the wrapper is functionally transparent to the rest of the SDK.
- `userId` / `anonymousId` / `traits` / `settings` go to a `UserDefaults(suiteName: "com.segment.storage.<writeKey>")` plist that `Storage.swift:24` writes directly, regardless of `StorageMode`. The `WrappingDataStore` received zero writes for these — confirmed both empirically (3 plaintext `SPIKE_` needles found in the plist) and by source inspection of the upstream library.

## What that means for the encryption-at-rest plan

The fork-vs-public-API decision is resolved: we use the public APIs. Phase 3 iOS:

1. Encrypt the event queue via `.custom(CioEncryptedDataStore)`. Path validated by this spike.
2. For `userId` / `traits` in the Segment UserDefaults plist (PII per our scope): apply `NSFileProtectionType.completeUntilFirstUserAuthentication` to the plist file. OS-level Data Protection, not app-layer encryption, but compliance-acceptable and requires no library modification.
3. `anonymousId` / `settings` remain plaintext in the same plist — per the scope decision they aren't PII.

## What's on the branch

| File | Purpose |
|---|---|
| `Sources/DataPipeline/Util/WrappingDataStore.swift` | New `DataStore` implementation: traces every protocol method via `os_log`, wraps writes with `WRAPPED::v1::` sentinel + XOR-with-0x55, unwraps reads transparently. |
| `Sources/DataPipeline/CustomerIO+Segment.swift` | Adds `.storageMode(.custom(WrappingDataStore(...)))` to `toSegmentConfiguration()`. |
| `Apps/APN-UIKit/APN UIKit/AppDelegate.swift` | Adds `runSegmentStorageSpike()` planting `SPIKE_` needles via the public CIO API. |
| `SPIKE_REPORT.md` | Full file-by-file reconciliation table, hexdumps, trace counts, verdict. |

## Verification

See `SPIKE_REPORT.md` for the full reconciliation table.

## Related

- Plan: `~/.claude/plans/encryption-at-rest.md` (§5 Phase 3, updated with these findings)
- Companion Android spike PR on `customerio-android` repo

### DIFF
--- a/Sources/DataPipeline/Util/WrappingDataStore.swift
+++ b/Sources/DataPipeline/Util/WrappingDataStore.swift
@@ -1,0 +1,222 @@
+//
+//  WrappingDataStore.swift
+//
+//  SPIKE-ONLY — feasibility test for Phase 3 encryption-at-rest.
+//  Wraps cdp-analytics-swift's storage so we can prove that every byte
+//  the event-queue path persists flows through our custom DataStore.
+//
+//  Mutation:
+//    - Prepend the sentinel  "WRAPPED::v1::"  (13 ASCII bytes)
+//    - XOR every byte AFTER the sentinel with 0x55
+//
+//  On fetch we reverse the mutation and hand the library a fully-formed
+//  JSON batch (matching the MemoryStore contract) so the upload pipeline
+//  never sees the wrapped form.
+//
+//  This is throwaway code on a spike branch. Do not ship.
+//
+
+import CioAnalytics
+import Foundation
+import os.log
+
+// MARK: - Trace logger
+
+private let spikeLog = OSLog(subsystem: "io.customer.spike", category: "SegmentStorageTrace")
+
+@inline(__always)
+private func tracePreview(_ bytes: Data, max: Int = 64) -> String {
+    let slice = bytes.prefix(max)
+    // Hex preview — readable in `log stream` without binary noise.
+    return slice.map { String(format: "%02x", $0) }.joined()
+}
+
+// MARK: - WrappingDataStore
+
+/// A `DataStore` implementation that
+/// (a) traces every protocol-method call via `os_log`, and
+/// (b) writes events to disk in a *mutated* form so a grep for the
+///     pre-mutation plaintext on disk proves whether the library
+///     bypasses this store on any persistence path.
+public final class WrappingDataStore: DataStore {
+    public typealias StoreConfiguration = Configuration
+
+    public struct Configuration {
+        let writeKey: String
+        let storageLocation: URL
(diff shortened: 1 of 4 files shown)


Sources/DataPipeline/Util/WrappingDataStore.swift
```diff
@@ -0,0 +1,222 @@
...
+    public static let xorKey: UInt8 = 0x55
+
+    private let config: Configuration
+    // Each persisted file holds exactly one wrapped event so we don't have
+    // to deal with separators or in-place mutation of multi-event files.
+    private let lock = NSLock()
+    // Monotonically increasing index — used only for filename ordering.
+    private var nextIndex: Int = 0
+
+    public var transactionType: DataTransactionType { .data }
```
hollyschilling-cio, line 67: This is the line that might make this work but may create other issues. This causes the upload path to take a different branch that causes the upload branch to use upload from memory instead of from a file on disk (See SegmentDestination.swift:143-154). While that solves the PII on disk issue, it changes the upload behavior with respect to the app lifecycle. With this change, the upload will be abruptly terminated if the app is backgrounded while an upload is in process. That can be ok, but I don't think we are set up to handle this correctly at present. **This is a major change and will likely appear as a discontinuity in the events that get to the server.** It also may cause other ripples in our SDK that I cannot anticipate. 

To be clear, I'm not saying this won't work. I'm saying that this is the path that I had identified to the group previously and identified as equally dangerous as simply ditching Segment.

hollyschilling-cio (changes requested):
Great work. I left a comment inline discussing this further, but this is the solution I had identified to the group in the meeting before.

PR #1066, opened 2026-05-26
chore: opt-in cdpApiKey persistence for background delivery

mrehan27:
## Stack
Part of the MBL-1628 stack. Merge in order:

- [#1060](https://github.com/customerio/customerio-ios/pull/1060) — geofence transition event delivery via EventBus (in review)
- [#1061](https://github.com/customerio/customerio-ios/pull/1061) — pending-metric file-backed queue (in review)
- [#1062](https://github.com/customerio/customerio-ios/pull/1062) — direct-HTTP `GeofenceDeliveryTracker` (in review)
- [#1064](https://github.com/customerio/customerio-ios/pull/1064) — `BackgroundDeliveryContextStore` in Common (in review)
- [#1065](https://github.com/customerio/customerio-ios/pull/1065) — three-layer geofence transition delivery wireup (in review, base for this PR)
- **This PR** — opt-in `cdpApiKey` persistence for background delivery
- _(next)_ — wire `CoreLocationGeofenceMonitor` to `GeofenceEventTracker` + cold-wake bootstrap

## Ticket
[MBL-1628 — Send Geofence Transition Events](https://linear.app/customerio/issue/MBL-1628/ios-send-geofence-transition-events)

## Summary
Adds `SDKConfigBuilder.allowBackgroundDelivery(_:)` (defaults to `false` — secure by default). When enabled, `DataPipelineImplementation.init` persists `cdpApiKey` to `BackgroundDeliveryContextStore`; when disabled, the init explicitly clears any key left by a prior launch so toggling off across launches actually revokes disk access.

Cold-wake background-delivery features (landing in a follow-up PR) will read the persisted key to authenticate without the full SDK being initialized in the process. Customers fetching their key from a backend can leave the flag off and accept that cold-wake transitions queue until the next foreground session.

## Design rationale
- **DataPipeline owns the decision** — `cdpApiKey` belongs to `DataPipeline`, so the persistence opt-in lives on `DataPipelineConfigBuilder`, not on `LocationConfig`. LocationModule is purely a consumer; it reads whatever's in the store. No backwards dependencies, no init-order coordination.
- **Default OFF (secure by default)** — customers must consciously opt in to persist a credential on disk. Matches the "secure by default" posture.
- **Explicit clear on opt-out** — `setCdpApiKey(nil)` runs unconditionally in the `else` branch so opting out across launches actually wipes a previously-persisted key. Without that, a customer who flips `true → false` would silently leave the stale key behind.

## Scope
[...]

PR #1062, opened 2026-05-21
chore: add direct-HTTP geofence delivery tracker

mrehan27:
## Stack
Part of the MBL-1628 four-PR stack. Merge in order:

- [#1060](https://github.com/customerio/customerio-ios/pull/1060) — geofence transition event delivery via EventBus (in review)
- [#1061](https://github.com/customerio/customerio-ios/pull/1061) — pending-metric file-backed queue (base for this PR, in review)
- **This PR** — direct-HTTP `GeofenceDeliveryTracker`
- _(next)_ — wire `GeofenceEventTracker` to the three-layer flow + flush hook on init

## Ticket
[MBL-1628 — Send Geofence Transition Events](https://linear.app/customerio/issue/MBL-1628/ios-send-geofence-transition-events)

## Summary
Introduces the primary path of the three-layer geofence delivery design: a thin tracker that POSTs a single transition event to `/track` via `HttpClient`. Mirrors the shape of `RichPushDeliveryTracker` so we get the same auth, region routing, and error handling for free.

No new wiring in this PR — `GeofenceDeliveryTracker` is added as a tested unit; the next PR rewires `GeofenceEventTracker` to call it.

## Design rationale
- **Mirrors `RichPushDeliveryTracker`** — same protocol + impl shape, same `HttpClient` dependency, same `Result<Void, HttpRequestError>` callback. Keeps the geofence path consistent with the only other internally-tracked direct-HTTP metric in the SDK.
- **Body matches Android exactly** — `{ event, userId, properties: { geofence_id, transition_type, latitude?, longitude?, timestamp(seconds) } }`. Event name uses `"GeoFence Entered"` / `"GeoFence Exited"` (capital F) — same wire form the EventBus path already uses.
- **`userId` passed in by the caller** — no clean Common-accessible accessor exists for the identified userId (`analytics.userId` is DataPipeline-internal; `ProfileStore.getProfileId(siteId:)` needs a siteId). Pushing the lookup up to the caller defers the architectural decision to the wireup PR where the call-site is being authored anyway.
- **Region → host inlined locally** — `CioLocation` only depends on `CioInternalCommon`. `RichPushHttpClient.getDefaultApiHost(region:)` lives in `CioMessagingPush` and `Region.apiHost` lives in `CioDataPipeline`; pulling either in for one switch isn't worth a cross-module dependency. This is the third copy of the URL switch — acknowledged tech debt for a future unification.
[...]

PR #1061, opened 2026-05-20
chore: add pending geofence metric file-backed queue

mrehan27:
## Stack
Part of the MBL-1628 four-PR stack. Each PR layers one concern; merge in order:

- [#1060](https://github.com/customerio/customerio-ios/pull/1060) — geofence transition event delivery via EventBus (base for this PR, in review)
- **This PR** — file-backed pending-metric queue (persistence primitive)
- _(next)_ — direct-HTTP `GeofenceDeliveryTracker` (uses the queue's model)
- _(next)_ — wire `GeofenceEventTracker` to the three-layer flow + flush hook on init

## Ticket
[MBL-1628 — Send Geofence Transition Events](https://linear.app/customerio/issue/MBL-1628/ios-send-geofence-transition-events)

## Summary
Adds the file-backed queue layer of the three-layer geofence transition delivery design: a `PendingGeofenceMetric` model and a `PendingGeofenceMetricStore` actor that persists transitions before any HTTP attempt and retains them until the direct-HTTP layer (next PR) confirms success.

This PR adds only the persistence primitive — no wiring yet, no existing code path references it.

## Design rationale
- **File over `UserDefaults`** — written with `completeUntilFirstUserAuthentication` Data Protection so geofence callbacks that fire before the first unlock can still read after, and excluded from iCloud/iTunes backups (transient device-local context, not user data).
- **Single actor, no internal `await`** — every public method does load → modify → save in one actor-isolated step, so concurrent callers can't observe a half-applied state or lose a write. Same pattern as `GeofenceStorage`.
- **100-entry FIFO cap, drops oldest on overflow** — bounds disk usage if the device stays offline for a long time.
- **`.secondsSince1970` date strategy** — matches `GeofenceStorage` for cross-store consistency.

## Scope
Two new source files; nothing else touched. Not yet referenced by any module init or other code path — the wire-up PR adds the only call site.

## Diff size
- Source: ~180 lines (1 model + 1 actor)
- Tests: ~260 lines, excluded from the 250-line cap

## Test plan
- [x] `make generate && make format && make lint` clean
- [x] Targeted: `LocationTests/PendingGeofenceMetricStoreTests` — 13/13 pass (~1s)
- [x] Coverage: append/load/remove/clearAll happy paths, capacity boundary at exactly N (off-by-one guard), capacity overflow drops oldest, persistence across actor instances, concurrent ops stress test asserts capacity invariant holds

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

[...]

PR #1068, opened 2026-05-26
chore: drop MessagingPush HttpClient dependency from geofence delivery

mrehan27:
## Stack
Part of the MBL-1628 stack. Merge in order:

- [#1060](https://github.com/customerio/customerio-ios/pull/1060) — geofence transition event delivery via EventBus (in review)
- [#1061](https://github.com/customerio/customerio-ios/pull/1061) — pending-metric file-backed queue (in review)
- [#1062](https://github.com/customerio/customerio-ios/pull/1062) — direct-HTTP `GeofenceDeliveryTracker` (in review)
- [#1064](https://github.com/customerio/customerio-ios/pull/1064) — `BackgroundDeliveryContextStore` in Common (in review)
- [#1065](https://github.com/customerio/customerio-ios/pull/1065) — three-layer geofence transition delivery wireup (in review)
- [#1066](https://github.com/customerio/customerio-ios/pull/1066) — opt-in `cdpApiKey` persistence (in review)
- [#1067](https://github.com/customerio/customerio-ios/pull/1067) — wire `CoreLocationGeofenceMonitor` to `GeofenceEventTracker` (in review, base for this PR)
- **This PR** — drop MessagingPush `HttpClient` dependency from geofence delivery
- _(next)_ — `LocationModule.bootstrapForBackgroundDelivery` cold-wake entry point

## Ticket
[MBL-1628 — Send Geofence Transition Events](https://linear.app/customerio/issue/MBL-1628/ios-send-geofence-transition-events)

## Summary
Geofence delivery used `HttpClient`, which is registered by MessagingPush. That made the direct-HTTP delivery path silently degrade to EventBus whenever MessagingPush wasn't initialized — a wrong-layering smell, since Location is a sibling of MessagingPush, not a dependent.

This PR introduces `BackgroundDeliveryHttpClient` in `CioInternalCommon` (URLSession-based via the existing `HttpRequestRunner`) and refactors `GeofenceDeliveryTrackerImpl` to use it. Location no longer depends on MessagingPush for HTTP. The new client is the foundation for cold-wake delivery in the follow-up PR.

`BackgroundDeliveryContextStore` gains a weak `BackgroundDeliveryCdpApiKeyProvider` delegate. `DataPipelineImplementation` registers itself at init so `currentCdpApiKey` returns the in-memory key even when `allowBackgroundDelivery(false)` (no persistence required for foreground real-time delivery). On cold-wake the provider is nil and `currentCdpApiKey` falls back to persisted state.

## Changes
- **`BackgroundDeliveryHttpClient`** (new, in Common) — protocol + URLSession-backed impl. POSTs `/track` with Basic auth from `cdpApiKey`. Registered via `InjectCustomShared`.
- **`BackgroundDeliveryCdpApiKeyProvider`** — `AnyObject` protocol returning `cdpApiKey: String?`. `BackgroundDeliveryContextStore` holds the provider weakly; `currentCdpApiKey` consults provider → falls back to persisted → returns nil.
- **`DataPipelineImplementation`** conforms to the provider, registers itself in `initialize()`.
- **`GeofenceDeliveryTrackerImpl`** simplified — takes `BackgroundDeliveryHttpClient` instead of `HttpClient`, no longer composes URLs or reads apiHost (the new client owns that). Uses `BackgroundDeliveryHttpError` instead of `HttpRequestError`.
- **`LocationModuleState`** constructs the tracker against `di.backgroundDeliveryHttpClient` — no more `di.getOptional(HttpClient.self).map { ... }` optional path.

## Design rationale
- **Why not register a Common-side `HttpClient` impl?** That would still route Location through the MessagingPush-defined protocol surface, and any future MessagingPush registration would silently override it. A new, purpose-built protocol keeps the delivery path's wire format and auth choices explicit and module-owned.
- **Why a delegate (not a static accessor) for live `cdpApiKey`?** Foreground real-time delivery needs the key in memory without persisting it (security posture). The delegate pattern lets `BackgroundDeliveryContextStore` stay the single read surface for callers — `currentCdpApiKey` returns the right value in foreground (provider) and cold-wake (persisted) without callers branching.
- **Weak provider reference** prevents a retain cycle with `DataPipelineImplementation` (which holds the store reference).

## Scope
- Source: ~140 LOC (90 in new HTTP client, the rest delegate + refactor)
- Tests: ~190 LOC (excluded from the 250-line cap)
- New public API in `CioInternalCommon`: `BackgroundDeliveryHttpError`, `BackgroundDeliveryHttpClient` protocol, `BackgroundDeliveryHttpClientImpl`, `BackgroundDeliveryCdpApiKeyProvider`, `BackgroundDeliveryContextStore.setCdpApiKeyProvider(_:)`.

## Test plan
- [x] `BackgroundDeliveryHttpClientTests` — 7 tests (URL composition, scheme-qualified host, missing apiHost/cdpApiKey guards, 500/204/transport status mapping)
- [x] `BackgroundDeliveryContextStoreTests` — 5 new provider tests added (wins over persisted, nil/empty/dealloc fallback, persisted fallback)
- [x] `BackgroundDeliveryContextWriteTests` — disk-only assertions via fresh store; new `test_init_expectProviderReturnsInMemoryKeyEvenWhenPersistenceOff` confirms foreground real-time path
- [x] `GeofenceDeliveryTrackerTests` — rewritten against the new client mock (wire format, guard, propagation)
- [x] `GeofenceEventTrackerTests` — updated to use `BackgroundDeliveryHttpError`
- [x] api-docs baseline regenerated for new public types

---

> [!NOTE]
> **Medium Risk**
> Changes how CDP credentials are resolved and how geofence events are sent over the network; scoped refactor with tests, but touches API key handling and delivery reliability.
> 
> **Overview**
> Introduces **`BackgroundDeliveryHttpClient`** in Common so geofence (and future cold-wake) delivery can POST CDP **`/track`** without MessagingPush’s **`HttpClient`**, which previously made direct HTTP silently unavailable when Push wasn’t initialized.
> 
> **`GeofenceDeliveryTracker`** now delegates to **`sendTrackEvent`** and **`BackgroundDeliveryHttpError`** instead of building requests and depending on optional **`HttpClient`**. **`LocationModuleState`** always wires the tracker to **`di.backgroundDeliveryHttpClient`**.
> 
> **`BackgroundDeliveryContextStore`** adds a weak **`BackgroundDeliveryCdpApiKeyProvider`**: **`currentCdpApiKey`** prefers the in-memory key from **`DataPipelineImplementation`** (registered at init) and falls back to the persisted key for cold-wake. Foreground real-time delivery can use the module config key even when **`allowBackgroundDelivery`** is off and nothing is written to disk.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f9329ff3c468e93c2bb7a2181793959ab5f4f62a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Sources/Common/Service/BackgroundDeliveryHttpClient.swift
```diff
@@ -0,0 +1,144 @@
...
+
+    public convenience init(
+        contextStore: BackgroundDeliveryContextStore,
+        requestRunner: HttpRequestRunner,
+        logger: Logger
+    ) {
+        let configuration = URLSessionConfiguration.ephemeral
+        configuration.allowsCellularAccess = true
+        configuration.timeoutIntervalForRequest = 30
+        configuration.timeoutIntervalForResource = 30
```
mahmoud-elmorabea, line 51: Can we somehow share this config with existing clients?

mahmoud-elmorabea (approved):

Sources/Common/Service/BackgroundDeliveryHttpClient.swift
mrehan27: Will take a look at this separately — if the consolidation is small enough, I'll open a PR for it. Otherwise we'll track as follow-up work.

PR #1069, opened 2026-06-01
chore: LocationModule.bootstrapForBackgroundDelivery for cold-wake

mrehan27:
## Stack
Part of the MBL-1628 stack. Merge in order:

- [#1060](https://github.com/customerio/customerio-ios/pull/1060) — geofence transition event delivery via EventBus (in review)
- [#1061](https://github.com/customerio/customerio-ios/pull/1061) — pending-metric file-backed queue (in review)
- [#1062](https://github.com/customerio/customerio-ios/pull/1062) — direct-HTTP `GeofenceDeliveryTracker` (in review)
- [#1064](https://github.com/customerio/customerio-ios/pull/1064) — `BackgroundDeliveryContextStore` in Common (in review)
- [#1065](https://github.com/customerio/customerio-ios/pull/1065) — three-layer geofence transition delivery wireup (in review)
- [#1066](https://github.com/customerio/customerio-ios/pull/1066) — opt-in `cdpApiKey` persistence (in review)
- [#1067](https://github.com/customerio/customerio-ios/pull/1067) — wire `CoreLocationGeofenceMonitor` to `GeofenceEventTracker` (in review)
- [#1068](https://github.com/customerio/customerio-ios/pull/1068) — drop MessagingPush `HttpClient` dependency (in review, base for this PR)
- **This PR** — `LocationModule.bootstrapForBackgroundDelivery` cold-wake entry point

## Ticket
[MBL-1628 — Send Geofence Transition Events](https://linear.app/customerio/issue/MBL-1628/ios-send-geofence-transition-events)

## Summary
Wrapper SDKs (React Native, Flutter) initialize Customer.io from JS/Dart. On a cold-wake from a geofence transition, the JS/Dart runtime never starts, so `CustomerIO.initialize` never runs and the queued transition is dropped.

This PR adds `LocationModule.bootstrapForBackgroundDelivery(launchOptions:)`: a single-line AppDelegate entry that wires region monitoring and direct-HTTP delivery from native code only. It reads all the state it needs from persisted stores (`GeofenceStorage`, `BackgroundDeliveryContextStore`) — no module's `initialize` is required to run first.

The bootstrap shares the same DI-resolved singletons as `LocationModule.initialize`, so both paths converge on one `CoreLocationGeofenceMonitor`, one `GeofenceEventTracker`, and one in-process active-delivery dedup set. Native + wrapper apps both use the same line. No keys, no flags — `launchOptions` is accepted for forward compatibility (reserved for future cold-wake detection).

## Changes
- **`LocationModule.bootstrapForBackgroundDelivery(launchOptions:)`** (new public, `@MainActor`) — emits the discoverability log (deferred from #1066), starts `flushPending`, wires the monitor.
- **`GeofenceBootstrap`** (new, `@MainActor` enum) — factors the shared wiring (cache fetch → monitor → binder) and the discoverability log. Used by both `LocationModule.initialize` and `bootstrapForBackgroundDelivery`.
- **DI singletons** — `GeofenceStorage`, `GeofenceEventTracker`, `CoreLocationGeofenceMonitor` are now resolvable from `DIGraphShared` and constructed once per process:
  - `GeofenceStorage` and `GeofenceEventTracker` use Sourcery `InjectCustomShared`.
  - `CoreLocationGeofenceMonitor` uses a hand-written extension instead — the Sourcery-generated eager-init test references the property from a non-isolated context, which clashes with `@MainActor`. Documented in code.
- **`LocationModuleState.performInitialization`** refactored to consume the same DI singletons — no inline construction, no risk of foreground vs. cold-wake paths constructing separate monitors.

## Design rationale
- **Why one entry point, not two (cold-wake vs. foreground)?** The check `launchOptions[.location] != nil` is a hint, not a contract — the OS doesn't promise the key is set for every geofence-driven launch. Calling `bootstrap` unconditionally is safe (idempotent through DI singletons) and removes a branching footgun for hosts.
- **Why DI singletons instead of `LocationModuleState.shared`?** Reusing `LocationModuleState` would couple cold-wake (which only needs the geofence pieces) to the full Location service stack. DI lets the cold-wake path resolve exactly what it needs and naturally dedups against `initialize`.
- **Why drop Sourcery for the monitor?** The eager-init resolution test the template generates can't traverse a `@MainActor` boundary from its non-isolated test method. Two options: invent a `nonisolated` accessor that crashes at runtime if called off main (ugly + unsafe), or write the extension by hand (one-time tax, clearer semantics). Picked the hand-written extension.

## Scope
- Source: 143 insertions, 46 deletions (well under the 250-line cap)
- Tests: 115 LOC (new `GeofenceBootstrapTests`, additions to `GeofenceMonitorBinderTests`)
- New public API in `CioLocation`: `LocationModule.bootstrapForBackgroundDelivery(launchOptions:)`

## Test plan
- [x] `GeofenceBootstrapTests` (new) — discoverability log fires on nil cdpApiKey, suppressed when persisted, suppressed when provider returns key; DI singletons resolve to the same instance across calls
- [x] `GeofenceMonitorBinderTests` — added double-bind test (idempotent re-bind from foreground + cold-wake in the same process)
- [x] Full `LocationTests` suite (106 tests) passes locally
- [x] api-docs baseline regenerated for new `bootstrapForBackgroundDelivery` public API

---

> [!NOTE]
> **Medium Risk**
> Touches geofence event delivery and process-wide singleton wiring; mis-sharing instances could drop or duplicate transitions, but changes are scoped and covered by tests.
> 
> **Overview**
> Adds **`LocationModule.bootstrapForBackgroundDelivery(launchOptions:)`** so hosts (especially RN/Flutter) can start geofence region monitoring and pending-metric replay from **AppDelegate** when `CustomerIO.initialize` never runs on a cold-wake.
> 
> **`GeofenceBootstrap`** centralizes the shared path: optional info log when background delivery isn’t configured, `flushPending`, then cache → monitor → binder (same ordering as before to avoid dropping queued region callbacks).
> 
> Foreground init in **`LocationModuleState`** no longer builds its own `GeofenceStorage` / tracker / monitor; it uses **`DIGraphShared`** singletons for **`GeofenceStorage`**, **`GeofenceEventTracker`**, and **`CoreLocationGeofenceMonitor`** so initialize and bootstrap share one monitor, one pending queue, and one in-process delivery dedup set. API docs and tests cover discoverability logging, DI identity, and double-bind idempotency.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit afeead9d7645403c8cecc70a3fe002144ed8c85f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Sources/Location/LocationModule.swift
+++ b/Sources/Location/LocationModule.swift
@@ -1,5 +1,6 @@
 import CioInternalCommon
 import Foundation
+import UIKit
 
 /// Location module that can be registered via `SDKConfigBuilder.addModule(_:)` so Location is initialized during `CustomerIO.initialize(withConfig:)`.
 ///
@@ -27,6 +28,29 @@ public final class LocationModule: CustomerIOModule {
                 "Location",
                 nil
             )
+        }
+    }
+
+    /// Bootstraps geofence cold-wake delivery from the host's `AppDelegate`.
+    ///
+    /// Wrapper SDKs (RN, Flutter) do not run `CustomerIO.initialize` in the cold-wake
+    /// process — there is no JS/Dart runtime. This entry reads all the state it needs
+    /// from persisted stores and wires up region monitoring + direct-HTTP delivery
+    /// without requiring any module's `initialize` to have run in this process.
+    ///
+    /// Safe to call on every launch. When the SDK has been initialized via
+    /// `CustomerIO.initialize(withConfig:)`, the same DI-resolved singletons are
+    /// reused — no double-init, no duplicate monitoring.
+    @MainActor
+    public static func bootstrapForBackgroundDelivery(launchOptions: [UIApplication.LaunchOptionsKey: Any]?) {
+        // Reserved for future cold-wake detection (e.g. `launchOptions[.location]`).
+        _ = launchOptions
+
+        let di = DIGraphShared.shared
+        GeofenceBootstrap.emitDiscoverabilityLogIfNeeded(di: di)
+        Task { await di.geofenceEventTracker.flushPending() }
+        Task { @MainActor in
+            await GeofenceBootstrap.wireMonitor(di: di)
         }
     }
 }
--- a/Sources/Location/Geofence/Event/GeofenceEventTracker.swift
+++ b/Sources/Location/Geofence/Event/GeofenceEventTracker.swift
@@ -1,6 +1,8 @@
 import CioInternalCommon
 import Foundation
 
+// sourcery: InjectRegisterShared = "GeofenceEventTracker"
+// sourcery: InjectCustomShared
 /// Sends geofence transition events through a three-layer delivery flow:
 /// 1. Cooldown-based deduplication (keyed by "geofenceId:transitionType")
 /// 2. Persist to `PendingGeofenceMetricStore` before any send attempt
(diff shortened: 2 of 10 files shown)


Sources/Location/LocationModule.swift
```diff
@@ -29,4 +30,27 @@ public final class LocationModule: CustomerIOModule {
...
+    /// `CustomerIO.initialize(withConfig:)`, the same DI-resolved singletons are
+    /// reused — no double-init, no duplicate monitoring.
+    @MainActor
+    public static func bootstrapForBackgroundDelivery(launchOptions: [UIApplication.LaunchOptionsKey: Any]?) {
+        // Reserved for future cold-wake detection (e.g. `launchOptions[.location]`).
+        _ = launchOptions
+
+        let di = DIGraphShared.shared
+        GeofenceBootstrap.emitDiscoverabilityLogIfNeeded(di: di)
+        Task { await di.geofenceEventTracker.flushPending() }
```
mahmoud-elmorabea, line 51: Do we need to flush on background wakeup?

mahmoud-elmorabea (approved):

Sources/Location/LocationModule.swift
mrehan27: Good catch — yes, we should call `flushPending()` after cold-wake bootstrap so queued metrics from the prior process drain before the next round of transitions. Will wire it in the cleanup PR.

PR #1070, opened 2026-06-01
chore: relax geofence permission gate + self-heal on auth change

mrehan27:
## Stack
Part of the geofence work. Stacked on top of the MBL-1628 chain. Merge in order:

- [#1060](https://github.com/customerio/customerio-ios/pull/1060) — geofence transition event delivery via EventBus
- [#1061](https://github.com/customerio/customerio-ios/pull/1061) — pending-metric file-backed queue
- [#1062](https://github.com/customerio/customerio-ios/pull/1062) — direct-HTTP `GeofenceDeliveryTracker`
- [#1064](https://github.com/customerio/customerio-ios/pull/1064) — `BackgroundDeliveryContextStore` in Common
- [#1065](https://github.com/customerio/customerio-ios/pull/1065) — three-layer geofence transition delivery wireup
- [#1066](https://github.com/customerio/customerio-ios/pull/1066) — opt-in `cdpApiKey` persistence
- [#1067](https://github.com/customerio/customerio-ios/pull/1067) — wire `CoreLocationGeofenceMonitor` to `GeofenceEventTracker`
- [#1068](https://github.com/customerio/customerio-ios/pull/1068) — drop MessagingPush `HttpClient` dependency
- [#1069](https://github.com/customerio/customerio-ios/pull/1069) — `LocationModule.bootstrapForBackgroundDelivery` cold-wake entry (base for this PR)
- **This PR** — relax permission gate + self-heal on authorization change

## Ticket
[MBL-1626 — Handle OS Geofence Monitoring](https://linear.app/customerio/issue/MBL-1626/ios-handle-os-geofence-monitoring) (follow-up to PR #1036; same bug the Android team caught and fixed in their port)

## Summary
Two related fixes that align iOS with the SDK's passive permission stance: the SDK never requests permission — the host app owns that — and the SDK adapts to whatever it's given.

**Gate relaxation.** `CoreLocationGeofenceMonitor.checkAlwaysAuthorization()` rejected any status other than `.authorizedAlways`, silently dropping registration on `.authorizedWhenInUse` hosts that are otherwise fully functional in the foreground. Now the monitor handles all three tiers explicitly:

| Status | Behavior | Log |
|---|---|---|
| `.authorizedAlways` | register | none |
| `.authorizedWhenInUse` | register | info, once: "foreground delivery only" |
| `.denied`/`.restricted`/`.notDetermined`/`@unknown` | skip | info, once: "permission not granted; host controls request" |

Logs are descriptive (not prescriptive) and fire from `startMonitoring`, so they only emit when there's actually a geofence to register — silent when the SDK has nothing to do.

**Mid-process self-heal.** `CoreLocationGeofenceMonitor` now observes `locationManagerDidChangeAuthorization(_:)` and notifies `GeofenceBootstrap`, which re-runs `wireMonitor`. This covers the realistic case where the host prompts late (so initial registration ran with `.notDetermined`) or the user toggles permission in Settings while the app is alive. Customer integration is unchanged — they call `LocationModule.initialize` or `bootstrapForBackgroundDelivery` once; iOS auto-fires the delegate on every subsequent status change.

## Changes
- **`CoreLocationGeofenceMonitor`** — new `PermissionTier` enum, `permissionTier(for:)` (`nonisolated static`, fully testable), tier-based `startMonitoring` gate.
- **`locationManagerDidChangeAuthorization(_:)`** — new delegate method, invokes the auth-changed handler.
- **`GeofenceRegionMonitoring`** protocol — new `setOnAuthorizationChanged(_:)` method + `GeofenceAuthorizationChangedHandler` typealias.
- **`GeofenceBootstrap.wireMonitor`** — installs the auth-changed handler after binding; handler re-runs `wireMonitor` (idempotent, no stacking across repeat calls).
- **`Logger+Geofence`** — `geofenceAlwaysAuthorizationRequired` (error, prescriptive) replaced by `geofencePermissionUnavailable` + `geofenceBackgroundDeliveryUnavailable` (both info, descriptive).
- **`DIGraphShared.geofenceMonitor`** — fixed a generic-`T` inference bug where the override lookup keyed by the concrete class instead of the protocol. Now explicitly types the optional so `getOverriddenInstance()` resolves correctly.

## Design rationale
- **Why split into a tier enum instead of two booleans?** The three states have distinct logging behavior. An enum makes the three branches in `startMonitoring` exhaustive and surfaces `@unknown` correctly.
- **Why `nonisolated` on `permissionTier(for:)`?** It's pure — no class state. `@MainActor` would force tests to spin up the monitor just to call it.
- **Why a delegate hook instead of polling?** `CLLocationManagerDelegate.locationManagerDidChangeAuthorization(_:)` is iOS-native. The OS fires it on every status change for free.
- **Why re-run `wireMonitor` (not a narrower retry)?** Bind is idempotent, the cache fetch is cheap, and reusing the existing entry path keeps the foreground/cold-wake/self-heal paths converged on a single code path.
- **Re-entrancy:** rapid status changes during `wireMonitor`'s `await` could spawn multiple concurrent re-runs. End state is correct (everything is idempotent); the redundancy is harmless because iOS doesn't fire permission changes in tight succession.

## Scope
- Source: 91 insertions, 21 deletions
- Tests: 94 insertions, 1 deletion
- No public API surface changes (no `api-docs` regen needed)

## Test plan
- [x] `GeofenceMonitorPermissionTierTests` (new) — 5 cases covering each `CLAuthorizationStatus` value
- [x] `GeofenceBootstrapTests` (additions) — handler installed after `wireMonitor`; firing the handler re-runs registration
- [x] `.serialized` trait added to the suite — parallel test execution was racing on the singleton DI graph (exposed by the new test, also benefits existing override-based tests)
- [x] Mocks (`GeofenceRegionMonitoringMock` + `MockGeofenceRegionMonitor`) updated for the new protocol method
- [x] Audit: zero `requestAlwaysAuthorization` / `requestWhenInUseAuthorization` / `requestAuthorization` calls in `Sources/` (only doc/log mentions remain)
- [x] Full `LocationTests` suite (113 tests) passes locally

---

> [!NOTE]
> **Medium Risk**
> Changes when regions register and how background vs foreground transitions behave, which affects location-sensitive delivery paths but does not add permission prompts or public API surface.
> 
> **Overview**
> Geofence region registration no longer requires **Always** location permission. **`CoreLocationGeofenceMonitor`** maps authorization to three **permission tiers**: full registration for Always, registration with a one-time info log for **When In Use** (foreground-only transitions), and skip with a one-time info log when permission is missing or denied. Prescriptive error logging is replaced with descriptive info messages that defer permission prompts to the host app.
> 
> **Mid-process recovery** wires **`locationManagerDidChangeAuthorization`** through **`setOnAuthorizationChanged`** on **`GeofenceRegionMonitoring`**, so **`GeofenceBootstrap.wireMonitor`** re-runs after Settings toggles or late permission prompts (idempotent bind, handler replacement avoids stacking).
> 
> **`DIGraphShared.geofenceMonitor`** fixes test/production overrides by typing the optional as **`GeofenceRegionMonitoring`** so DI lookup keys the protocol, not the concrete monitor type. Tests add tier coverage, bootstrap self-heal behavior, and **`.serialized`** on **`GeofenceBootstrapTests`** to avoid DI singleton races.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9f024fee741704112d5102a87ed401d5be54282c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Sources/Location/Geofence/Monitor/CoreLocationGeofenceMonitor.swift
+++ b/Sources/Location/Geofence/Monitor/CoreLocationGeofenceMonitor.swift
@@ -13,10 +13,23 @@ import Foundation
 /// registered by the host app or other SDKs (CLLocationManager.monitoredRegions is shared app-wide).
 @MainActor
 final class CoreLocationGeofenceMonitor: NSObject, GeofenceRegionMonitoring, @preconcurrency CLLocationManagerDelegate {
+    /// Tier of capability available given the current `CLAuthorizationStatus`. The SDK never
+    /// requests permission — the host app owns that — so the monitor adapts to whatever was
+    /// granted: `.authorizedAlways` enables background delivery, `.authorizedWhenInUse` falls
+    /// back to foreground-only (regions still register and fire while foregrounded),
+    /// everything else skips registration.
+    enum PermissionTier: Equatable {
+        case backgroundDelivery
+        case foregroundOnly
+        case blocked
+    }
+
     private let manager: CLLocationManager
     private let logger: Logger
     private var onTransition: GeofenceTransitionHandler?
-    private var hasLoggedPermissionWarning = false
+    private var onAuthorizationChanged: GeofenceAuthorizationChangedHandler?
+    private var hasLoggedBlocked = false
+    private var hasLoggedForegroundOnly = false
     private var ownedRegionIdentifiers: Set<String> = []
 
     init(logger: Logger) {
@@ -34,8 +47,27 @@ final class CoreLocationGeofenceMonitor: NSObject, GeofenceRegionMonitoring, @pr
         onTransition = handler
     }
 
+    func setOnAuthorizationChanged(_ handler: GeofenceAuthorizationChangedHandler?) {
+        onAuthorizationChanged = handler
+    }
+
     func startMonitoring(identifier: String, center: LocationData, radius: Double, transitionTypes: Set<GeofenceTransition>) {
-        guard checkAlwaysAuthorization() else { return }
+        let status = currentAuthorizationStatus()
+        switch Self.permissionTier(for: status) {
+        case .blocked:
+            if !hasLoggedBlocked {
+                hasLoggedBlocked = true
+                logger.geofencePermissionUnavailable(currentStatus: status)
+            }
+            return
+        case .foregroundOnly:
+            if !hasLoggedForegroundOnly {
+                hasLoggedForegroundOnly = true
+                logger.geofenceBackgroundDeliveryUnavailable(currentStatus: status)
+            }
+        case .backgroundDelivery:
+            break
+        }
 
         let coordinate = CLLocationCoordinate2D(latitude: center.latitude, longitude: center.longitude)
         guard CLLocationCoordinate2DIsValid(coordinate) else {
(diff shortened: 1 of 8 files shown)


Sources/Location/Geofence/Monitor/CoreLocationGeofenceMonitor.swift
```diff
@@ -34,8 +47,27 @@ final class CoreLocationGeofenceMonitor: NSObject, GeofenceRegionMonitoring, @pr
...
+
     func startMonitoring(identifier: String, center: LocationData, radius: Double, transitionTypes: Set<GeofenceTransition>) {
-        guard checkAlwaysAuthorization() else { return }
+        let status = currentAuthorizationStatus()
+        switch Self.permissionTier(for: status) {
+        case .blocked:
+            if !hasLoggedBlocked {
+                hasLoggedBlocked = true
+                logger.geofencePermissionUnavailable(currentStatus: status)
+            }
```
mahmoud-elmorabea, line 61: Any specific reason why we are gating multiple logs here?

mahmoud-elmorabea (approved):

Sources/Location/Geofence/Monitor/CoreLocationGeofenceMonitor.swift
mrehan27: Per-tier-transition dedup. iOS fires `locationManagerDidChangeAuthorization` on every status change (and sometimes spuriously while the user toggles Settings) — without the gate we'd log on every callback. The gate logs once per Always → WhenInUse → blocked transition so the signal stays one line per real state change.

PR #1073, opened 2026-06-02
feature: Added Filtering and Rate Limiting

hollyschilling-cio:
[MBL-1747](https://linear.app/customerio/issue/MBL-1747/add-rate-limiting-and-filtering-segment-module)
This PR adds Filtering and Rate Limiting through an aggregation engine. The aggregation engine pulls its config from the Segment config pulled from `https://{cdnHost}/projects/{writeKey}/settings` using values passed as `integrations.aggregationRules`.

The largest change introduced here is the addition of a dependency to have a local database. This is not yet handled in the Cocoapods, but we will update this PR to accommodate that before merging.

---

> [!NOTE]
> **High Risk**
> Changes which analytics events are sent (server-controlled drops and rate limits) and adds encrypted on-disk persistence keyed by the CDP API key; misconfiguration or storage issues could alter customer data delivery until CocoaPods parity is done.
> 
> **Overview**
> Adds **server-driven event filtering and rate limiting** for Data Pipelines via Segment/CDP settings (`aggregationRules` on the Customer.io Data Pipelines integration). Rules are cached in a new **encrypted local SQLCipher database** (`SyncSqlCipher`), registered at `DataPipeline.initialize`, and applied by an **`EventPolicyPlugin`** that can drop track, screen, identify, alias, and group events before they leave the client.
> 
> **`StorageManager`** opens a per–CDP API key DB under Application Support, runs aggregation schema migrations, and stores rules plus rate-limit counters (profile vs device scope). **`EventPolicyEngine`** evaluates block lists and time windows (including `*` wildcards); storage errors **fail open** so events are not silently dropped. Settings refresh updates the engine and persists or deletes cached rules; **`clearIdentify`** clears **profile-scoped** rate-limit state only.
> 
> SPM now depends on **`SyncSqlCipher`** and bumps the Swift tools version to **5.10**; CocoaPods wiring for the DB is called out as follow-up in the PR description. Broad unit tests cover decoding, engine, plugin, storage, and identify reset.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 43b450114a906968ae05bc8efc4c5e6e1bf09e23. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


matt-frizzell (commented):
I think adding the database layer should/could be its own reviewable change, separate from the introduction of filtering. It would cut down the lines of code to read and keep the focus on reviewing and testing one thing at a time.

Sources/DataPipeline/Storage/StorageManager+Aggregation.swift
```diff
@@ -0,0 +1,86 @@
...
+import SyncSqlCipher
+
+extension StorageManager {
+    // MARK: - Aggregation Config
+
+    func getAggregationConfig() throws -> (payload: String, fetchedAt: String)? {
+        let rows = try db.query(
+            Select(col("payload"), col("fetched_at"))
+                .from("aggregation_rules")
+                .where(col("id") == 1)
```
mahmoud-elmorabea: I think the fact that we are using a single hard-coded row id might signal that we don't need a database for this?

We might also be getting ahead of ourselves by creating a database for the rules, I would imagine there would be a handful of rules per workspace and that they wouldn't change all that often.

Sources/DataPipeline/EventPolicy/EventPolicyEngine.swift
```diff
@@ -0,0 +1,70 @@
...
+        guard let rs = processed.wrappedValue else { return true }
+
+        let exactKey = "\(eventType):\(name)"
+        let wildcardKey = "\(eventType):*"
+
+        if rs.filterKeys.contains(exactKey) || rs.filterKeys.contains(wildcardKey) {
+            return false
+        }
+
+        if let rl = rs.rateLimitsByKey[exactKey] ?? rs.rateLimitsByKey[wildcardKey] {
```
mahmoud-elmorabea, line 55: Does rate limiting `wildcardKey` something we'd want to do? It would apply a time based rate limit on all track events including ones sent by customers?

Sources/DataPipeline/EventPolicy/EventPolicyEngine.swift
```diff
@@ -0,0 +1,70 @@
+import CioInternalCommon
+import Foundation
+
+/// Synchronous, thread-safe engine that decides whether an event should be
+/// allowed through or dropped based on the active ``AggregationRuleset``.
+///
+/// `@unchecked Sendable` is safe here: all mutable state (`processed`) is
+/// protected by `Synchronized<>`, and `StorageManager.db` is serialized
+/// internally via a private DispatchQueue.
+final class EventPolicyEngine: Sendable {
```
mahmoud-elmorabea, line 10: I'm finding myself reverse understanding how this engine would work, it would be very helpful if we can create a spec that describes the intended behavior

mahmoud-elmorabea: Given the issues we've seen, what makes sense to me is something like:
- Filtering (exactly as you describe): Either filter a specific event or all track/screen events
- Rate limiting an exact event (specific track or screen view): only send once per certain time frame
- De-duplicate: for a specific event if the payload is the same (excluding timestamp), drop that event

mahmoud-elmorabea (commented):
Thanks for the draft! 👍 

I have two main points here:
- To me, the overhead of the DB: creation, tables, migrations and the extra dependency isn't justified by the data being stored here. They're mostly key/value pairs and group of rules that would be saved/loaded and replaced in their entirety. We can still encrypt it if we want even without the database
- I'd like us to better link this to the issues we've had reported, run a kind of simulation. "For that particular issue we'd apply that rule and it would have that effect" type of thing. A question that could answer is, should these rules apply to only internal events or also customer events? Do we need these rules to be applied server side also or client side is enough?

Sources/DataPipeline/Storage/StorageManager+Aggregation.swift
hollyschilling-cio: You're right that this specific data doesn't need to be in the database. This is also not the reason to have a database. The rate-limiting history does need to be in a database because that may be updated multiple times per second and rewriting an entire JSON file each write is unreasonable. 

Even within the database, there is a Key-Value table that could be used. The reason I put it into a table to start is if we decide in the future to begin doing diffing or keeping previous versions for fallback, which will require additional rows.

Sources/DataPipeline/EventPolicy/EventPolicyEngine.swift
hollyschilling-cio: The code is pretty clear so I'm not sure what you are not unerstanding.

De-duplication is not in scope for this. 

Segment has 5 basic event types: `track`, `screen`, `identify`, `alias`, and `group`. We don't use `alias` or `group`, and `identify` should never be filtered or rate limited by either of these mechanisms. That leaves us only concerning ourselves with `track` and `screen`. `screen` events are screen views where the event name is the name of the screen. `track` events have a name of whatever event name was given at the call site. 

Filter rules have a event name and type. When the engine gets asked if an event should be passed on, it uses the name and type to check against a look up table (O(1)) in memory to see if that particular event type and name combination is filtered or rate limited. If the event is neither, it goes on. If it is filtered, the event is discarded immediately. If it is rate limited, we check and set (atomically) the event's database record to see when it was seen last. If it has not been seen within the window, the last seen time is updated and the event is sent on. If it has been seen within the time window, the record is not modified and the event is destroyed.

PR #1075, opened 2026-06-03
chore: wire geofence sync triggers and add handleMovement Tier A/B

mrehan27:
## Stack

Part of the geofence work. Stacks on top of MBL-1630 part 1. Merge in order:

- [#1060](https://github.com/customerio/customerio-ios/pull/1060) — geofence transition event delivery via EventBus
- [#1061](https://github.com/customerio/customerio-ios/pull/1061) — pending-metric file-backed queue
- [#1062](https://github.com/customerio/customerio-ios/pull/1062) — direct-HTTP `GeofenceDeliveryTracker`
- [#1064](https://github.com/customerio/customerio-ios/pull/1064) — `BackgroundDeliveryContextStore` in Common
- [#1065](https://github.com/customerio/customerio-ios/pull/1065) — three-layer geofence transition delivery wireup
- [#1066](https://github.com/customerio/customerio-ios/pull/1066) — opt-in `cdpApiKey` persistence
- [#1067](https://github.com/customerio/customerio-ios/pull/1067) — wire `CoreLocationGeofenceMonitor` to `GeofenceEventTracker`
- [#1068](https://github.com/customerio/customerio-ios/pull/1068) — drop MessagingPush `HttpClient` dependency
- [#1069](https://github.com/customerio/customerio-ios/pull/1069) — `LocationModule.bootstrapForBackgroundDelivery` cold-wake entry
- [#1070](https://github.com/customerio/customerio-ios/pull/1070) — relax permission gate + self-heal on auth change
- [#1071](https://github.com/customerio/customerio-ios/pull/1071) — `GeofenceConfig` model + `LastSyncRecord` storage
- [#1072](https://github.com/customerio/customerio-ios/pull/1072) — `GeofenceApiService` + wire response decoder
- [#1074](https://github.com/customerio/customerio-ios/pull/1074) — `GeofenceSyncCoordinator` with `refresh` + `applyCachedRegistration`
- **This PR** — sync triggers (identify / foreground / movement EXIT) + `handleMovement` Tier A/B

## Ticket

[MBL-1630 — iOS Smart Geofence Updates](https://linear.app/customerio/issue/MBL-1630/ios-smart-geofence-updates) (part 2 of 3 — trigger wiring + Tier A/B movement dispatch. First-run race fix + sign-out reset land in part 3.)

## Review surface

200 production LOC added across 5 files; 138 test LOC across 1 file (plus minor test-stub updates). `GeofenceSyncCoordinator.swift` is the substantive change (138/45); the rest is trigger plumbing.

## Summary

Closes the gap between `GeofenceSyncCoordinator` (#1074) and the events that should drive a sync. Three triggers + one new coordinator entry:

- **Identify** — existing `ProfileIdentifiedEvent` subscriber in `LocationModuleState` now also fires a geofence refresh.
- **Foreground** — new `addDidBecomeActiveObserver` block fires a geofence refresh on every `didBecomeActive`. Mode-independent.
- **Movement EXIT** — `GeofenceMonitorBinder` dispatches by identifier: the synthetic `cio_movement_trigger` region routes to `coordinator.handleMovement`; business geofences continue to `tracker.trackTransition`.
- **`handleMovement(latitude:longitude:)`** — new coordinator entry with two-tier dispatch matching the Android port.

`refresh()` and `handleMovement()` share extracted private helpers (`performRemoteRefresh` / `performLocalRefresh`) so the same distance-filter + OS-register step runs through both paths.

## Tier A / Tier B branching

`handleMovement` reads the API anchor (`storage.getLastSync()?.location`) and the cached config, then chooses:

- **Tier B (remote fetch)** — when `anchor == nil` (first EXIT after install / clearAll) OR straight-line distance to the new location is ≥ `remoteFetchRefreshTriggerRadius`. Calls `performRemoteRefresh` — same code path as `refresh`. Persists new cache + anchor + timestamp.
- **Tier A (local re-rank)** — otherwise. Calls `performLocalRefresh` — re-rank cached regions for the new location, re-register with the OS, **no API call and no storage writes**. The API anchor stays put so the next Tier B threshold check still compares against the last *server* fetch, not where the user just stood.

## Concurrency model

All three coordinator entries — `refresh`, `handleMovement`, `applyCachedRegistration` — share the existing `Synchronized<Bool> refreshInProgress` gate. Losers see `false` from `acquireGate()` and return `.alreadyInProgress` (or skip with log for `applyCachedRegistration`). Every exit path runs `defer { releaseGate() }`.

`performRemoteRefresh` and `performLocalRefresh` are private helpers and **do not touch the gate** — the public entry owns acquire+release. Avoids double-acquire / leaked-gate failure modes.

## Dispatch in the binder

```
movement-trigger EXIT + non-nil location → coordinator.handleMovement
movement-trigger ENTER (defensive)       → dropped
movement-trigger EXIT + nil location     → dropped (no anchor to decide on)
anything else                            → tracker.trackTransition
```

Movement-trigger transitions are never tracked as customer events.

## Foreground observer lifecycle

`RealAppLifecycleNotifying()` is hoisted to one shared instance used by both `LocationServicesImplementation` and the new geofence observer — avoids two NotificationCenter subscriptions firing. The observer token is retained in a new `LocationModuleState.geofenceForegroundToken` field.

## Mode independence

The foreground observer is unconditional — `LocationConfig.mode == .off` does **not** disable it. Geofence OS-side monitoring runs independently of the SDK's GPS-update subscription. Refresh from identify/foreground skips silently when no cached anchor exists (`LastLocationStorage.getCachedLocation()` returns nil); the first movement EXIT still drives the initial registration via the OS-supplied location.

For fresh installs with `mode == .off`, the seeding window is the first user EXIT from the movement trigger. The first-run race fix lands in MBL-1630c.

## Out of scope (deferred)

- First-run race (Android Bug 3) — `lastSkippedForNoLocation` + `onLocationAcquired` rising-edge re-arm. MBL-1630c.
- `reset(signedOutUserId:)` and the userId-recheck-after-API pattern. MBL-1630c.
- `lastMovementTriggerLocation` persistence (Android's `restoreFromCache` prefers it over the anchor). Functional with the anchor today; can revisit later.
- `registerWithBusinessDiff` (Android stale-cleanup optimization). Continuing with `stopMonitoringAll → start all`.

## Test plan

- [x] `make generate && make format && make lint` — 0 violations, 601 files
- [x] `GeofenceSyncCoordinatorTests` — 30 cases (7 new for `handleMovement`: no-userId / no-anchor → Tier B / within threshold → Tier A / beyond threshold → Tier B / Tier-A preserves lastSync+cache / Tier-A re-centers movement trigger at new location / dedup vs in-flight refresh — and the reverse: `refresh_givenInFlightHandleMovement_expectAlreadyInProgress` to pin the cross-entry gate in both directions)
- [x] `GeofenceMonitorBinderTests` — rewritten with `MockGeofenceRegionMonitor.simulateTransition`. 4 cases: movement EXIT routed to coordinator with correct args / movement ENTER dropped / movement EXIT + nil location dropped / business geofence routed to tracker (verified via EventBus fallback signal)
- [x] `GeofenceBootstrapTests` — retargeted at the shared `MockGeofenceRegionMonitor`; existing assertions on transition handler + `applyCachedRegistration` + auth-change handler all pass
- [x] Full `LocationTests` suite — 185 tests / 21 suites pass

---

> [!NOTE]
> **Medium Risk**
> Changes background geofence registration, API fetch timing, and shared concurrency gating; behavior is well-covered by tests but mistakes could mis-register regions or skip refreshes.
> 
> **Overview**
> Wires **geofence sync** to real app events and adds **two-tier movement handling** when the synthetic movement region fires on EXIT.
> 
> **Triggers:** `ProfileIdentifiedEvent` and **app foreground** (`didBecomeActive`) now call `geofenceSyncCoordinator.refresh` using the last cached location (no-op if none). Foreground uses a retained lifecycle token and shares one `RealAppLifecycleNotifying` instance with location services—**independent of** `LocationConfig.mode`.
> 
> **Movement EXIT:** `GeofenceMonitorBinder` takes the sync coordinator; `cio_movement_trigger` **EXIT** with a location → `handleMovement` (not customer events). Business regions still go to `GeofenceEventTracker`.
> 
> **`handleMovement`:** Compares the new position to the **last API sync anchor**. Inside `remoteFetchRefreshTriggerRadius` → **Tier A**: re-rank cached geofences, re-register OS regions, **recenter** the movement trigger, **no** API or `lastSync`/cache writes. Otherwise (or no anchor) → **Tier B**: same path as `refresh` (`performRemoteRefresh`). `refresh`, `handleMovement`, and `applyCachedRegistration` share one **in-flight gate** (`acquireGate` / `releaseGate`); remote/local work lives in private helpers.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 699412c4f66eee558febd03f208ddc7155022186. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Sources/Location/Geofence/Coordinator/GeofenceSyncCoordinator.swift
+++ b/Sources/Location/Geofence/Coordinator/GeofenceSyncCoordinator.swift
@@ -1,4 +1,5 @@
 import CioInternalCommon
+import CoreLocation
 import Foundation
 
 /// Errors surfaced by `GeofenceSyncCoordinator` callers.
@@ -8,16 +9,28 @@ enum GeofenceSyncError: Error, Equatable {
     case fetchFailed(GeofenceApiError)
 }
 
-/// Sync pipeline for the on-device geofence cache. Two entry points:
+/// Which branch `handleMovement` took for the current EXIT.
+enum HandleMovementTier: String, Sendable {
+    /// Re-rank cached regions for the new location; no API call.
+    case localRerank
+    /// Anchor moved beyond `remoteFetchRefreshTriggerRadius` (or absent) — refetch from server.
+    case remoteRefresh
+}
+
+/// Sync pipeline for the on-device geofence cache. Entry points:
 ///
 /// - `refresh(latitude:longitude:)` — fetch from the server, distance-filter, register,
 ///   persist. Gated on identified user, in-flight dedup, and freshness.
+/// - `handleMovement(latitude:longitude:)` — movement-trigger EXIT entry. Two-tier:
+///   re-rank cache when within `remoteFetchRefreshTriggerRadius` of the API anchor,
+///   otherwise refetch from the server. Shares the same dedup gate as `refresh`.
 /// - `applyCachedRegistration(...)` — synchronously register from caller-fetched state,
 ///   used by cold-wake / boot / auth-change paths. Synchronous on the main actor so
 ///   `ownedRegionIdentifiers` is populated before the next yield — otherwise the OS may
 ///   deliver a queued cold-wake transition into an empty filter set.
-protocol GeofenceSyncCoordinator: AutoMockable, Sendable {
+protocol GeofenceSyncCoordinator: AutoMockable, AnyObject, Sendable {
     func refresh(latitude: Double, longitude: Double) async -> Result<Void, GeofenceSyncError>
+    func handleMovement(latitude: Double, longitude: Double) async -> Result<Void, GeofenceSyncError>
     @MainActor
     func applyCachedRegistration(
         cachedRegions: [Geofence],
@@ -59,18 +72,11 @@ final class GeofenceSyncCoordinatorImpl: GeofenceSyncCoordinator, @unchecked Sen
     }
 
     func refresh(latitude: Double, longitude: Double) async -> Result<Void, GeofenceSyncError> {
-        // Concurrent calls return `.alreadyInProgress` (not `.success`) so callers can
-        // distinguish a skip-due-to-dedup from a real completion.
-        let acquired: Bool = refreshInProgress.mutating { inProgress in
-            if inProgress { return false }
-            inProgress = true
-            return true
-        }
-        guard acquired else {
+        guard acquireGate() else {
             logger.geofenceSyncSkipped(reason: "refresh already in progress")
             return .failure(.alreadyInProgress)
         }
-        defer { refreshInProgress.wrappedValue = false }
+        defer { releaseGate() }
 
         guard let userId = contextStore.currentUserId, !userId.isEmpty else {
             logger.geofenceSyncSkipped(reason: "no identified user")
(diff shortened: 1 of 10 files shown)


Sources/Location/autogenerated/AutoMockable.generated.swift
```diff
@@ -196,6 +196,11 @@ class GeofenceSyncCoordinatorMock: GeofenceSyncCoordinator, Mock {
         refreshReceivedArguments = nil
         refreshReceivedInvocations = []
 
+        mockCalled = false // do last as resetting properties above can make this true
+        handleMovementCallsCount = 0
+        handleMovementReceivedArguments = nil
+        handleMovementReceivedInvocations = []
+
```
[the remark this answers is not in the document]
mrehan27: Stencil-template artifact ([Sources/Templates/AutoMockable.stencil](Sources/Templates/AutoMockable.stencil)) — same pattern in every autogen mock SDK-wide. Out of scope here; fixing means a template change + repo-wide regen.

Sources/Location/Geofence/Coordinator/GeofenceSyncCoordinator.swift
```diff
@@ -162,6 +178,81 @@ final class GeofenceSyncCoordinatorImpl: GeofenceSyncCoordinator, @unchecked Sen
...
+            return .failure(.fetchFailed(error))
+        }
+
+        let parsedConfig = response.toDomainConfig()
+        let regions = response.toDomainRegions()
+        let effectiveConfig = parsedConfig ?? cachedConfig ?? .fallback
+        let anchor = LocationData(latitude: latitude, longitude: longitude)
+        let nearest = distanceFilter.nearest(regions, to: anchor, limit: effectiveConfig.maxBusinessGeofences)
+        await MainActor.run {
+            registerWithOsSync(
```
mahmoud-elmorabea, line 216: What happens if we don't get any geofences? Nearest will be empty, will we still register the movement trigger? I think we should no-op and cancel existing movement trigger if that's the case

mahmoud-elmorabea (approved):

Sources/Location/Geofence/Coordinator/GeofenceSyncCoordinator.swift
```diff
@@ -162,6 +178,81 @@ final class GeofenceSyncCoordinatorImpl: GeofenceSyncCoordinator, @unchecked Sen
...
+    private func performLocalRefresh(
+        latitude: Double,
+        longitude: Double,
+        config: GeofenceConfig,
+        cachedRegions: [Geofence]
+    ) async -> Result<Void, GeofenceSyncError> {
+        let anchor = LocationData(latitude: latitude, longitude: longitude)
+        let nearest = distanceFilter.nearest(cachedRegions, to: anchor, limit: config.maxBusinessGeofences)
+        await MainActor.run {
+            registerWithOsSync(
```
mahmoud-elmorabea, line 246: [Same](https://github.com/customerio/customerio-ios/pull/1075/changes/BASE..e654bdfd3bceb67c62cec0b123a576d6fa0bffd8#r3362183631) here as well

Sources/Location/Geofence/Coordinator/GeofenceSyncCoordinator.swift
mrehan27: Good catch — empty `nearest` should no-op AND cancel any existing movement trigger so we don't keep monitoring a stale region. Will land in the cleanup PR.

Sources/Location/Geofence/Coordinator/GeofenceSyncCoordinator.swift
mrehan27: Same fix applies — covered with the thread above.

Sources/Location/Geofence/Coordinator/GeofenceSyncCoordinator.swift
```diff
@@ -162,6 +178,81 @@ final class GeofenceSyncCoordinatorImpl: GeofenceSyncCoordinator, @unchecked Sen
...
+                businessRegions: nearest,
+                movementTriggerLocation: anchor,
+                movementTriggerRadius: config.localRefreshTriggerRadius
+            )
+        }
+        logger.geofenceSyncCompleted(registeredCount: nearest.count)
+        return .success(())
+    }
```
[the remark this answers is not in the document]
mrehan27: Same finding as Mahmoud's threads above on lines 216/246 — empty `nearest` should no-op and cancel any existing movement trigger. Will land in the cleanup PR.

PR #1081, opened 2026-06-08
chore: regenerate autogen + reformat to current rules

mrehan27:
## Summary

Drive-by drift cleanup, no source changes. The recent main merge into `feature/geofence-on-device` left two kinds of mismatch between baseline and current rules:

- **Stencil template vs. baseline import order.** Template emits `import Foundation` first, then arg-supplied imports — but the on-disk baseline still has the older order (alphabetical, `CioInternalCommon` first). `.swiftformat` has `--disable sortImports` so nothing was re-sorting.
- **Updated swiftformat rules.** Single-line initializers get expanded to multi-line blocks; `async` is removed from test funcs that don't actually await. Previous baseline pre-dates those rule changes.

Folded here as a precursor so this churn doesn't surface on every follow-up PR in the MBL-1770 stack.

## What's in the diff

- 9 `.generated.swift` files: import order rewritten to template's natural output (`Foundation` first, then arg-supplied imports).
- 4 test files: swiftformat reflows — single-line `init` expansions in `StubProvider`/`StubCdpApiKeyProvider`, `async` removed from non-awaiting tests.

No production source changes.

## Stack

This is the first PR in the MBL-1770 follow-up stack. Subsequent PRs build on this:

1. **this PR** — autogen + reformat drift
2. `mbl-1770-b-storage-correctness` — pending-store correctness across user changes
3. `mbl-1770-c-…` — small fixes + Android-port bug fixes
4. `mbl-1770-d-…` — cross-process exactly-once delivery
5. `mbl-1770-e-…` — no-op regression test for non-geofence apps

Linear: [MBL-1770](https://linear.app/customerio/issue/MBL-1770/ios-address-follow-up-reliability-lifecycle-and-platform-improvements)

## Test plan

- [x] `make generate && make format && make lint` — clean (0 violations)
- [x] `xcodebuild build-for-testing -scheme Customer.io-Package` — green
- [ ] Re-run on CI to confirm no further drift surfaces

---

> [!NOTE]
> **Low Risk**
> Generated DI/mock files and cosmetic test formatting only; no runtime or behavioral changes.
> 
> **Overview**
> **Drift-only cleanup** so later MBL-1770 PRs don’t carry formatting noise. There are **no production source changes**.
> 
[...]

PR #1082, opened 2026-06-08
chore: composite-key dedup for geofence pending metrics

mrehan27:
## Summary

Aligns `PendingGeofenceMetricStore` row identity with Android's `PendingGeofenceDelivery.key` and adds storage-layer collision rejection. No change to foreground delivery; both pieces matter when a row outlives the user it was captured under (attribution preserved by per-row userId stamping in #1084).

**Composite-key dedup** (matches Android):
- Drop `id: UUID` on `PendingGeofenceMetric`; derive `key` from `(geofenceId, transition, timestamp_sec)`. Seconds precision because the cooldown gate already dedups by `(geofenceId, transition)` upstream — finer precision adds nothing.
- `PendingGeofenceMetricStore.remove(id:)` → `remove(key:)`; same for `removeAll`. In-flight dedup set in `GeofenceEventTracker` switches from `Set<UUID>` to `Set<String>`.
- `append` rejects a row whose key collides with one already on disk, returning `true` so the caller's success path stays intact — storage-layer second line of defense if the cooldown gate ever slips.

## Stack

Merge order (top-down):
1. ✅ #1081 — autogen drift fix (merged)
2. **this PR** — composite-key dedup for geofence pending metrics
3. #1083 — small fixes + Android #717 port
4. #1084 — per-row userId stamping
5. #1087 — setupGeofence test seam

Ticket: https://linear.app/customerio/issue/MBL-1770

## Test plan

- [x] `xcodebuild build` — clean
- [x] `LocationTests` — 195/195 pass
- [x] Covers: `remove(key:)`, `removeAll(keys:)`, duplicate-key append rejection, in-flight `Set<String>` dedup under concurrent flush

---

> [!NOTE]
> **Medium Risk**
> Changes persistence identity and on-disk JSON shape (drops `id`), which can affect upgrades with existing queued rows; delivery dedup logic is tightened but behavior is localized to the geofence pending queue.
> 
> **Overview**
> **Aligns iOS pending geofence metrics with Android** by replacing per-row `UUID` identity with a derived composite `key` (`geofenceId`, `transition`, timestamp in seconds).
> 
> `PendingGeofenceMetric` drops `id` from the model and JSON; `PendingGeofenceMetricStore` removes rows by `key` instead of `id`, and **`append` skips duplicate keys** (still returns `true` so callers don’t retry). `GeofenceEventTracker` uses an in-flight `Set<String>` of keys for concurrent delivery dedup and drains the queue with `remove(key:)`.
> 
> Tests are updated for key-based remove/removeAll and add coverage for duplicate-key append rejection.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7a8173279c6dcb5dc468bbecc16e2e2ac060af17. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Sources/Location/Geofence/Storage/PendingGeofenceMetric.swift
+++ b/Sources/Location/Geofence/Storage/PendingGeofenceMetric.swift
@@ -3,22 +3,26 @@ import Foundation
 
 /// A geofence transition queued for direct-HTTP delivery.
 struct PendingGeofenceMetric: Codable, Equatable, Sendable {
-    let id: UUID
     let geofenceId: String
     let transition: GeofenceTransition
     let latitude: Double?
     let longitude: Double?
     let timestamp: Date
 
+    /// Composite key over `(geofenceId, transition, timestamp_ms)` used for
+    /// storage-layer dedup. Matches Android's `PendingGeofenceDelivery.key`.
+    var key: String {
+        let ms = Int(timestamp.timeIntervalSince1970 * 1000)
+        return "\(geofenceId)_\(transition.rawValue)_\(ms)"
+    }
+
     init(
-        id: UUID = UUID(),
         geofenceId: String,
         transition: GeofenceTransition,
         latitude: Double?,
         longitude: Double?,
         timestamp: Date
     ) {
-        self.id = id
         self.geofenceId = geofenceId
         self.transition = transition
         self.latitude = latitude
@@ -27,7 +31,6 @@ struct PendingGeofenceMetric: Codable, Equatable, Sendable {
     }
 
     enum CodingKeys: String, CodingKey {
-        case id
         case geofenceId = "geofence_id"
         case transition
         case latitude
--- a/Sources/Location/Geofence/Storage/PendingGeofenceMetric.swift
+++ b/Sources/Location/Geofence/Storage/PendingGeofenceMetric.swift
@@ -9,11 +9,13 @@ struct PendingGeofenceMetric: Codable, Equatable, Sendable {
     let longitude: Double?
     let timestamp: Date
 
-    /// Composite key over `(geofenceId, transition, timestamp_ms)` used for
+    /// Composite key over `(geofenceId, transition, timestamp_sec)` used for
     /// storage-layer dedup. Matches Android's `PendingGeofenceDelivery.key`.
+    /// Seconds (not ms) — cooldown gate dedups by `(geofenceId, transition)`
+    /// upstream, so finer precision adds nothing.
     var key: String {
-        let ms = Int(timestamp.timeIntervalSince1970 * 1000)
-        return "\(geofenceId)_\(transition.rawValue)_\(ms)"
+        let sec = Int(timestamp.timeIntervalSince1970)
+        return "\(geofenceId)_\(transition.rawValue)_\(sec)"
     }
 
     init(
--- a/Sources/Location/Geofence/Storage/PendingGeofenceMetricStore.swift
+++ b/Sources/Location/Geofence/Storage/PendingGeofenceMetricStore.swift
@@ -33,9 +33,14 @@ actor PendingGeofenceMetricStore {
     }
 
     /// Appends a metric. When over capacity, drops the **oldest** entries first.
+    /// A row with the same `key` is a no-op.
     /// Returns `false` if the file could not be persisted.
     func append(_ metric: PendingGeofenceMetric) -> Bool {
         var items = loadFromDisk()
+        if items.contains(where: { $0.key == metric.key }) {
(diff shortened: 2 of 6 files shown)


Sources/Location/Geofence/Storage/PendingGeofenceMetricStore.swift
```diff
@@ -48,22 +53,22 @@ actor PendingGeofenceMetricStore {
...
         return saveToDisk(items)
     }
 
-    /// Removes entries whose ids are in `ids` in one read-modify-write.
-    /// Returns `true` when the file was updated, `false` when none of the ids were present.
-    func removeAll(ids: Set<UUID>) -> Bool {
-        guard !ids.isEmpty else { return true }
+    /// Removes entries whose keys are in `keys` in one read-modify-write.
+    /// Returns `true` when the file was updated, `false` when none of the keys were present.
+    func removeAll(keys: Set<String>) -> Bool {
```
hollyschilling-cio, line 67: Reading the file from disk, filtering the results, and writing it back is likely to cause a race condition.

hollyschilling-cio:
There is a lot of code here handling storage. Most of it is trying to approximate the function of using a database. I'd strongly suggest rebasing this branch once the StorageManager and DB PR is merged.

Sources/Location/Geofence/Storage/PendingGeofenceMetricStore.swift
mrehan27: The class is already an `actor` and is designed in a way that avoids race conditions

mrehan27:
@hollyschilling-cio I like the idea of using the database, it certainly has value. However, this PR modifies an existing store and is part of a stack of 5 PRs. The database PR hasn't been merged yet, nor has its compatibility been validated against existing features. Given that, I'd prefer to merge this as is and then migrate it to the database once that work is merged and available.

Also, `PendingPushDeliveryStore` uses a similar file based pattern today, so a clean DB migration should consider both stores together rather than diverging them in this PR.

Sources/Location/Geofence/Storage/PendingGeofenceMetric.swift
```diff
@@ -3,22 +3,26 @@ import Foundation
...
     let transition: GeofenceTransition
     let latitude: Double?
     let longitude: Double?
     let timestamp: Date
 
+    /// Composite key over `(geofenceId, transition, timestamp_ms)` used for
+    /// storage-layer dedup. Matches Android's `PendingGeofenceDelivery.key`.
+    var key: String {
+        let ms = Int(timestamp.timeIntervalSince1970 * 1000)
+        return "\(geofenceId)_\(transition.rawValue)_\(ms)"
```
mahmoud-elmorabea: Does this need to match Android key? I think Android uses seconds unit instead of ms

Sources/Location/LocationModuleState.swift
```diff
@@ -93,7 +93,14 @@ final class LocationModuleState {
         }
         di.eventBusHandler.addObserver(ResetEvent.self) { _ in
             Task { @MainActor in
-                _ = await DIGraphShared.shared.geofenceSyncCoordinator.reset()
+                // Run in parallel — independent operations on different actors.
+                // Drops queued direct-HTTP metrics so rows captured under the
+                // previous user are not flushed under the next one.
+                async let resetResult: Result<Void, GeofenceSyncError> =
+                    DIGraphShared.shared.geofenceSyncCoordinator.reset()
+                async let cleared: Void = DIGraphShared.shared.geofenceEventTracker.clearPending()
```
mahmoud-elmorabea: Do we need to clear these given that in a followup PR we save the user ID along with the pending entries?

Sources/Location/Geofence/Storage/PendingGeofenceMetricStore.swift
mahmoud-elmorabea: Given this is an actor without awaits in the func body, I think we should be good as Rehan mentioned.

mahmoud-elmorabea (approved):

Sources/Location/Geofence/Storage/PendingGeofenceMetric.swift
mrehan27: Good catch — switched to seconds. Cooldown gate dedups upstream so ms wasn't buying anything.

Sources/Location/LocationModuleState.swift
mrehan27: Good call, dropped — stamping in #1084 covers attribution. Pushed in a separate commit.

PR #1083, opened 2026-06-08
chore: geofence reliability fixes + logger tag dispatch

mrehan27:
## Summary

Three localized fixes from the review backlog plus two ports from the Android geofence stack (#717). None change public API; mocks regenerated.

**`GeofenceDeliveryTracker`: rename `deliver` → `trackMetric`**
- Aligns with the other "tracker" APIs in the SDK. Internal-only protocol; no public-surface impact. `AutoMockable.generated.swift` regenerated.

**`GeofenceSyncCoordinator`: skip movement trigger when business regions are empty**
- When the server returns zero regions, registering a movement trigger just arms a refetch that hits the same empty result. Leave it off; the next identify / foreground refresh will retry.

**`GeofenceSyncCoordinator`: freshness check is time AND distance (port of Android #717)**
- Time-fresh alone leaves a gap: app gets killed while the user travels far, no movement EXIT fires to update the cache, next foreground refresh skips the API because the timestamp is still fresh — but the cached regions belong to a place the user no longer is. Adds a distance-from-anchor check next to the existing time check, gated by `remoteFetchRefreshTriggerRadius`.

**`Logger`: dispatcher receives the formatted (tagged + errored) message (port of Android #717)**
- `logDispatcher?(level, message)` was passing the raw message while `systemLogger.log(formattedMessage, level)` got the `[Tag] msg Error: …` prefix/suffix. Customer-provided dispatchers got inconsistent output. Both branches now receive `formattedMessage`. Android had the same bug on the same line.

## Stack

Merge order (top-down):
1. #1081 — `chore: regenerate autogen + reformat to current rules` (MBL-1770-a)
2. #1082 — `chore: geofence pending-store correctness across user changes` (MBL-1770-b)
3. **this PR** — `chore: geofence small fixes + Android #717 port` (MBL-1770-c)
4. MBL-1770-d — cross-process exactly-once via per-row userId stamping [coming]
5. MBL-1770-e — no-op regression test for non-geofence apps [coming]

Ticket: https://linear.app/customerio/issue/MBL-1770

## Test plan

- [x] `make generate && make format && make lint` — 0 violations
- [x] `xcodebuild build-for-testing` — clean
[...]

PR #1084, opened 2026-06-08
chore: stamp userId on pending geofence metric for stable attribution

mrehan27:
## Summary

A geofence transition queued while user A is identified could be delivered as user B if A signs out and B signs in before the flush runs. Both delivery channels resolved the userId at delivery time — the direct-HTTP path via `contextStore.currentUserId`, and the EventBus path via DataPipeline's current identity.

This PR snapshots the userId on `PendingGeofenceMetric` at capture time and routes delivery off it.

**Stamp at capture** ([GeofenceEventTracker.trackTransition](Sources/Location/Geofence/Event/GeofenceEventTracker.swift)):
- Read `contextStore.currentUserId` when the transition fires; empty → nil. Persist on the row.

**Dispatch in `deliver`**:
- **Stamped userId → direct-HTTP path with that userId.** Success drains the row, failure retains it for the next flush. Cross-user attribution is now stable across sign-out / sign-in / process restarts.
- **No stamped userId → post to EventBus and drain.** DataPipeline records the transition under its current identity (anonymous tracking when no one is identified). The captured `timestamp` + `latitude` / `longitude` flow through `TrackGeofenceMetricEvent` so the transition attributes to **when** it happened, not when the flush ran.

Falling back to the live `contextStore.currentUserId` on the HTTP path would reproduce the mis-attribution being fixed (anonymous capture would still get attributed to the next identified user), so nil-stamped rows are not eligible for the HTTP path.

## Drive-by cleanup

`deliveryTracker: GeofenceDeliveryTracker?` → non-optional. The DI accessor has unconditionally constructed `GeofenceDeliveryTrackerImpl` from the Common HTTP client since #1068, so the optional + the `guard let deliveryTracker else { postEventBus(…) }` branch were dead code in production. Dropped the optional and tightened the gate to `metric.userId == nil` instead.

## Type additions

- `PendingGeofenceMetric.userId: String?` (optional, backward-compat — legacy rows decode to nil and route through EventBus on next flush).
- `TrackGeofenceMetricEvent.latitude: Double?` / `longitude: Double?` so the EventBus path's payload matches the direct-HTTP path's properties. `timestamp` was already on the event but ignored by the DataPipeline subscriber; now honored.
- DataPipeline's `TrackGeofenceMetricEvent` subscriber rewritten to construct `TrackEvent` explicitly (matching the BGQ pattern in `DataPipelineImplementation+BGQ`) so `TrackEvent.timestamp` is set from `metric.timestamp` and the full property shape mirrors HTTP.

## Stack

Merge order (top-down):
1. #1081 — `chore: regenerate autogen + reformat to current rules` (MBL-1770-a)
2. #1082 — `chore: geofence pending-store correctness across user changes` (MBL-1770-b)
3. #1083 — `chore: geofence reliability fixes + logger tag dispatch` (MBL-1770-c)
4. **this PR** — `chore: stamp userId on pending geofence metric for stable attribution` (MBL-1770-d)
5. MBL-1770-e — no-op regression test for non-geofence apps [coming]

Ticket: https://linear.app/customerio/issue/MBL-1770

## Cross-platform

Mirrors Android's [MBL-1760 / 6c0712a5](https://github.com/customerio/customerio-android/commit/6c0712a5):
- Same field added (`userId: String?` on the pending entry)
- Same capture-time snapshot
- Same "stamped → direct HTTP; null → analytics pipeline / EventBus" routing
- Same boundary: doesn't fix cross-process exactly-once (separate ticket; Android uses `claim()`, iOS deferred)

## Test plan

- [x] `xcodebuild build-for-testing` — clean
- [x] `GeofenceEventTrackerTests` — 18 tests green, incl. `trackTransition_givenIdentifiedUser_expectMetricStampedWithUserId`, `trackTransition_givenNoUserId_expectEventBusAndQueueDrained`, `flushPending_givenRowStampedDifferentFromCurrent_expectStampedUserIdUsed`, `flushPending_givenStampedUserId_andNoCurrent_expectDeliveredWithStamped`, `flushPending_givenNilUserIdOnRow_expectEventBusPostNoHttpCall`, `flushPending_givenAnonymousCaptureThenIdentify_expectNoHttpBackfill`
- [x] `GeofenceMonitorBinderTests` — 3 tests green (binder now observes routing via `GeofenceDeliveryTrackerMock`)
- [x] `DataPipelineEventBustTests` — 5 tests green, verifying `TrackEvent.timestamp` and property shape on both `.enter` (with coords) and `.exit` (without coords)
- [x] `api-docs/internal/CioInternalCommon.api` regenerated (`TrackGeofenceMetricEvent` adds `latitude`, `longitude`)

---

> [!NOTE]
> **Medium Risk**
> Changes geofence event attribution and delivery routing (identified vs anonymous), affecting analytics identity and when events are sent; behavior is heavily tested but touches customer-facing tracking semantics.
> 
> **Overview**
> Fixes **cross-user mis-attribution** for queued geofence transitions by **snapshotting `userId` on `PendingGeofenceMetric` at capture** and routing delivery from that stamp instead of the live identity at flush time.
> 
> **`GeofenceEventTracker`:** On capture, persists `contextStore.currentUserId` on the pending row. In `deliver`, rows with a stamped user go **direct HTTP** with that user (success drains, failure retries); rows with **no stamp** post **`TrackGeofenceMetricEvent` on EventBus**, drain immediately, and do **not** wait for a later identify or fall back to the current user for HTTP. `GeofenceDeliveryTracker` is now **required** (optional + nil-tracker EventBus branch removed).
> 
> **EventBus / DataPipeline path:** `TrackGeofenceMetricEvent` gains optional **latitude/longitude**; DataPipeline handles geofence metrics via **`processGeofenceMetricEvent`**, aligning properties with the HTTP path and setting **`TrackEvent.timestamp`** from the captured transition time.
> 
> Tests and internal API docs updated for stamping, anonymous EventBus delivery, and no HTTP backfill after identify.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5cd14658a846a41d0b5f1c49e1fb3f09ab3d88f5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Sources/DataPipeline/DataPipelineImplementation+GeofenceObserver.swift
+++ b/Sources/DataPipeline/DataPipelineImplementation+GeofenceObserver.swift
@@ -1,0 +1,24 @@
+import CioAnalytics
+import CioInternalCommon
+
+extension DataPipelineImplementation {
+    /// Tracks a geofence transition forwarded via EventBus from `GeofenceEventTracker`
+    /// when the row carries no stamped userId (anonymous capture path).
+    ///
+    /// Mirrors the property shape of the direct-HTTP path so both delivery channels
+    /// record the same set of fields. The `TrackEvent.timestamp` is set from
+    /// `metric.timestamp` so a flush replayed hours after capture still attributes
+    /// the transition to when it happened, not when it was sent.
+    func processGeofenceMetricEvent(_ metric: TrackGeofenceMetricEvent) {
+        var properties: [String: Any] = [
+            "geofence_id": metric.geofenceId,
+            "transition_type": metric.transition.rawValue,
+            "timestamp": Int(metric.timestamp.timeIntervalSince1970)
+        ]
+        if let latitude = metric.latitude { properties["latitude"] = latitude }
+        if let longitude = metric.longitude { properties["longitude"] = longitude }
+        var trackEvent = TrackEvent(event: metric.transition.trackEventName, properties: try? JSON(properties))
+        trackEvent.timestamp = metric.timestamp.string(format: .iso8601WithMilliseconds)
+        analytics.process(event: trackEvent)
+    }
+}
--- a/Sources/DataPipeline/DataPipelineImplementation.swift
+++ b/Sources/DataPipeline/DataPipelineImplementation.swift
@@ -121,7 +121,7 @@ class DataPipelineImplementation: DataPipelineInstance, DataPipelineTracking, Ba
         }
 
         eventBusHandler.addObserver(TrackGeofenceMetricEvent.self) { metric in
-            self.track(name: metric.transition.trackEventName, properties: ["geofence_id": metric.geofenceId])
+            self.processGeofenceMetricEvent(metric)
         }
 
         eventBusHandler.addObserver(RegisterDeviceTokenEvent.self) { event in
--- a/Tests/Location/Geofence/Event/GeofenceEventTrackerTests.swift
+++ b/Tests/Location/Geofence/Event/GeofenceEventTrackerTests.swift
@@ -32,9 +32,9 @@ struct GeofenceEventTrackerTests {
     private func makeTracker(
         storage: GeofenceStorage,
         pendingStore: PendingGeofenceMetricStore,
-        deliveryTracker: GeofenceDeliveryTracker?,
+        deliveryTracker: GeofenceDeliveryTracker,
         contextStore: BackgroundDeliveryContextStore,
-        eventBus: EventBusHandlerMock,
+        eventBus: EventBusHandlerMock = EventBusHandlerMock(),
         dateUtil: DateUtil = DateUtilStub()
     ) -> GeofenceEventTracker {
         GeofenceEventTracker(
@@ -62,13 +62,11 @@ struct GeofenceEventTrackerTests {
         let pending = makePendingStore(directory: dir)
         let delivery = GeofenceDeliveryTrackerMock()
         delivery.trackMetricClosure = { _, _, onComplete in onComplete(.success(())) }
-        let bus = EventBusHandlerMock()
         let tracker = makeTracker(
             storage: makeStorage(directory: dir),
             pendingStore: pending,
             deliveryTracker: delivery,
-            contextStore: makeContextStore(userId: "user_42"),
-            eventBus: bus
+            contextStore: makeContextStore(userId: "user_42")
         )
 
         await tracker.trackTransition(geofenceId: "geo_1", transition: .enter)
@@ -76,7 +74,6 @@ struct GeofenceEventTrackerTests {
         #expect(delivery.trackMetricCallsCount == 1)
         #expect(delivery.trackMetricReceivedArguments?.userId == "user_42")
         #expect(await pending.loadAll().isEmpty)
-        #expect(postedGeofenceEvents(from: bus).isEmpty)
     }
 
     @Test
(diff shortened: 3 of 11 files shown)


Sources/DataPipeline/DataPipelineImplementation+GeofenceObserver.swift
```diff
@@ -0,0 +1,24 @@
...
+
+extension DataPipelineImplementation {
+    /// Tracks a geofence transition forwarded via EventBus from `GeofenceEventTracker`
+    /// when the row carries no stamped userId (anonymous capture path).
+    ///
+    /// Mirrors the property shape of the direct-HTTP path so both delivery channels
+    /// record the same set of fields. The `TrackEvent.timestamp` is set from
+    /// `metric.timestamp` so a flush replayed hours after capture still attributes
+    /// the transition to when it happened, not when it was sent.
+    func processGeofenceMetricEvent(_ metric: TrackGeofenceMetricEvent) {
```
hollyschilling-cio, line 12: Why are we adding this inside the DataPipelines module? This is violating module boundaries.

Sources/DataPipeline/DataPipelineImplementation.swift
```diff
@@ -121,7 +121,7 @@ class DataPipelineImplementation: DataPipelineInstance, DataPipelineTracking, Ba
         }
 
         eventBusHandler.addObserver(TrackGeofenceMetricEvent.self) { metric in
-            self.track(name: metric.transition.trackEventName, properties: ["geofence_id": metric.geofenceId])
+            self.processGeofenceMetricEvent(metric)
```
hollyschilling-cio, line 124: This is in the wrong module.

Sources/DataPipeline/DataPipelineImplementation+GeofenceObserver.swift
mrehan27: This matches how `EventBus` works today for device tokens, push metrics, in-app metrics, and now geofence events. Only the `DataPipelines` module is responsible for sending these events. The `Common` module and `EventBus` simply act as a bridge to communicate and forward them to `DataPipelines`.

Sources/DataPipeline/DataPipelineImplementation.swift
mrehan27: This is specifically for the `EventBus` path, so the placement is correct. Events forwarded through `EventBus` are only handled by the `DataPipelines` module.

Sources/DataPipeline/DataPipelineImplementation+GeofenceObserver.swift
```diff
@@ -0,0 +1,24 @@
...
+        var properties: [String: Any] = [
+            "geofence_id": metric.geofenceId,
+            "transition_type": metric.transition.rawValue,
+            "timestamp": Int(metric.timestamp.timeIntervalSince1970)
+        ]
+        if let latitude = metric.latitude { properties["latitude"] = latitude }
+        if let longitude = metric.longitude { properties["longitude"] = longitude }
+        var trackEvent = TrackEvent(event: metric.transition.trackEventName, properties: try? JSON(properties))
+        trackEvent.timestamp = metric.timestamp.string(format: .iso8601WithMilliseconds)
```
mahmoud-elmorabea, line 21: This is a very good idea, I don't think we do this on Android, we probably should back-date the Android event as well to when it was captured rather than when it was published

Sources/DataPipeline/DataPipelineImplementation+GeofenceObserver.swift
```diff
@@ -0,0 +1,24 @@
...
+    /// the transition to when it happened, not when it was sent.
+    func processGeofenceMetricEvent(_ metric: TrackGeofenceMetricEvent) {
+        var properties: [String: Any] = [
+            "geofence_id": metric.geofenceId,
+            "transition_type": metric.transition.rawValue,
+            "timestamp": Int(metric.timestamp.timeIntervalSince1970)
+        ]
+        if let latitude = metric.latitude { properties["latitude"] = latitude }
+        if let longitude = metric.longitude { properties["longitude"] = longitude }
+        var trackEvent = TrackEvent(event: metric.transition.trackEventName, properties: try? JSON(properties))
```
mahmoud-elmorabea, line 20: Will the `try?` silently fail here? This could end up sending the event without properties or logging an error

mahmoud-elmorabea (approved):

Sources/DataPipeline/DataPipelineImplementation+GeofenceObserver.swift
mrehan27: The `try?` is intentional and matches the existing pattern in `DataPipelineImplementation+BGQ.swift:76` (push metric tracking does the exact same `TrackEvent(event:..., properties: try? JSON(properties))`).

`JSON([String: Any])` only throws `JSONError.nonJSONType` when a value isn't JSON-serializable. The properties here are all JSON-native primitives — `String` (geofenceId, transition rawValue), `Int` (timestamp), and `Double` (lat/lng). There's no runtime path where this throws, so `try?` can't produce nil. Happy to add a `do/catch` with a logger.error if you'd prefer belt-and-suspenders, but it'd be unreachable code given the input shape.

Sources/DataPipeline/DataPipelineImplementation+GeofenceObserver.swift
mrehan27: Yeah, good call. Opening PR in Android to fix it there.

Tests/Location/Geofence/Event/GeofenceEventTrackerTests.swift
```diff
@@ -398,16 +377,148 @@ struct GeofenceEventTrackerTests {
...
+            pendingStore: pending,
+            deliveryTracker: GeofenceDeliveryTrackerMock(),
+            contextStore: makeContextStore()
+        )
+
+        await tracker.trackTransition(geofenceId: "geo_1", transition: .enter)
+
+        let queued = await pending.loadAll()
+        #expect(queued.first?.userId == nil)
+    }
```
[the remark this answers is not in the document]
mrehan27: Already addressed — the flagged test was removed. Nil-stamping is verified by the EventBus routing assertions in the surviving tests (`delivery.trackMetricCallsCount == 0` + `posted.first?.geofenceId`).

Sources/DataPipeline/DataPipelineImplementation+GeofenceObserver.swift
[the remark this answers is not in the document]
mrehan27: Already addressed in [this thread](https://github.com/customerio/customerio-ios/pull/1084#discussion_r3395680531) — TL;DR the input is all JSON-safe primitives (`String`, `Int`, `Double`), so the `try?` can't actually produce nil at runtime. Pattern matches `DataPipelineImplementation+BGQ.swift:76` for push metric tracking.

PR #1087, opened 2026-06-09
chore: extract setupGeofence test seam from LocationModuleState

mrehan27:
## Summary
- Extract `setupGeofence(coordinator:lastLocationStorage:lifecycleNotifying:mode:di:)` from `performInitialization` as an internal test seam — only the geofence-wiring block moves; non-geofence setup (CLLocationManager, `LocationServicesImplementation`, profile enrichment) stays in place untouched.
- Widen `LocationModuleState.init()` from `private` to `internal` so tests can construct fresh instances without disturbing the `.shared` singleton.
- Add `LocationModuleStateGeofenceTests` validating the `ResetEvent` observer wiring — a regression gap noted during the 1770 stack review. Test drives the bus via a local `CapturingEventBusHandler` (no real `CioEventBusHandler`), spies coordinator with `GeofenceSyncCoordinatorMock`, mocks the monitor, and uses a temp-dir `GeofenceStorage` + `NoOpAppLifecycleNotifying` so the suite never hangs on shared OS resources.

## Why this shape
Earlier attempts tried a broader `performInitialization` refactor (decompose + constructor injection). It worked but touched non-geofence code that was already correct on `main`. Rolled back in favor of the minimal seam — geofence wiring moves, nothing else changes. Risk of regressing unrelated initialization paths is zero.

Linear: https://linear.app/customerio/issue/MBL-1774

## Stack
This PR is stacked on top of the 1770 series:
1. PR #1081 — `mbl-1770-a-autogen-drift-fix` → `feature/geofence-on-device`
2. PR #1082 — `mbl-1770-b-storage-correctness` → 1770-a
3. PR #1083 — `mbl-1770-c-small-fixes-android-port` → 1770-b
4. PR #1084 — `mbl-1770-d-per-row-userid-stamping` → 1770-c
5. **This PR** — `mbl-1774-a-refactor-locationmodulestate` → 1770-d
6. Held: `mbl-1770-e-no-op-regression-test` (rebased once this lands)

Merge in stack order.

## Test plan
- [x] `make format && make lint` — clean
- [x] `xcodebuild build-for-testing -scheme Customer.io-Package` — clean
- [x] `LocationTests` full suite — 203/203 pass, 0.71s, no hang
- [x] Confirmed `setupGeofence_givenResetEventDelivered_expectCoordinatorResetCalled` passes in isolation and as part of full suite

---

> [!NOTE]
> **Low Risk**
> Behavior-preserving refactor plus tests; the only functional tweak is using the passed `di` for geofence reset instead of the global singleton.
> 
> **Overview**
> Pulls **geofence-only initialization** out of `performInitialization` into a dedicated `setupGeofence(...)` entry point so tests can exercise event wiring without spinning up `CLLocationManager` or `LocationServicesImplementation`. Production still calls that method from the same place in the init flow; non-geofence setup is unchanged.
> 
> `LocationModuleState`’s initializer is **internal** so tests can use fresh instances instead of `.shared`. The `ResetEvent` handler now resets via the injected **`di.geofenceSyncCoordinator`** rather than reaching back into `DIGraphShared.shared`.
> 
> Adds **`LocationModuleStateGeofenceTests`**: a serialized suite with DI overrides, `CapturingEventBusHandler`, and coordinator spies to assert `ResetEvent` / `ProfileIdentifiedEvent` observers and that reset/refresh run with the expected anchor coordinates.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a7a527cf03dfc6133bea042c7182ae5c329dc7d1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Sources/Location/LocationModuleState.swift
+++ b/Sources/Location/LocationModuleState.swift
@@ -17,7 +17,8 @@ final class LocationModuleState {
     /// install's first GPS update still drives the initial geofence registration.
     private let lastSkippedForNoLocation = Synchronized<Bool>(false)
 
-    private init() {}
+    /// Internal init lets tests build instances independent of `.shared`.
+    init() {}
 
     /// Performs one-time setup of the Location module. Call from main thread (e.g. from `LocationModule.initialize()`).
     func performInitialization(config: LocationConfig) {
@@ -41,25 +42,16 @@ final class LocationModuleState {
         )
         let locationEnrichmentProvider = LocationProfileEnrichmentProvider(storage: storage, config: config)
         di.profileEnrichmentRegistry.register(locationEnrichmentProvider)
-        registerEventSubscriptions(coordinator: coordinator, lastLocationStorage: storage, mode: config.mode, di: di)
-        Task { await di.geofenceEventTracker.flushPending() }
         let lifecycleNotifying = RealAppLifecycleNotifying()
-        // Foreground geofence refresh is intentionally mode-independent — geofence
-        // monitoring runs even when `LocationConfig.mode` disables periodic location
-        // updates. Skips silently with no cached anchor; movement EXIT drives the first
-        // registration in that case.
-        geofenceForegroundToken = lifecycleNotifying.addDidBecomeActiveObserver { [weak self] in
-            self?.refreshGeofencesIfPossible(lastLocationStorage: storage)
-        }
-        // Rearm first-run refresh on the first fresh fix after an identify/foreground skip.
-        Task { [weak self] in
-            await coordinator.setOnLocationProcessed { location in
-                self?.rearmFirstRunRefreshIfArmed(location: location, di: di)
-            }
-        }
-        Task { @MainActor in
-            await GeofenceBootstrap.wireMonitor(di: di)
-        }
+
+        setupGeofence(
+            coordinator: coordinator,
+            lastLocationStorage: storage,
+            lifecycleNotifying: lifecycleNotifying,
+            mode: config.mode,
+            di: di
+        )
+
         let locationProvider = CoreLocationProvider(logger: di.logger)
         let implementation = LocationServicesImplementation(
             config: config,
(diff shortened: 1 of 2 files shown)


Sources/Location/LocationModuleState.swift
```diff
@@ -41,36 +42,56 @@ final class LocationModuleState {
...
         // registration in that case.
         geofenceForegroundToken = lifecycleNotifying.addDidBecomeActiveObserver { [weak self] in
-            self?.refreshGeofencesIfPossible(lastLocationStorage: storage)
+            self?.refreshGeofencesIfPossible(lastLocationStorage: lastLocationStorage)
         }
         // Rearm first-run refresh on the first fresh fix after an identify/foreground skip.
-        Task { [weak self] in
-            await coordinator.setOnLocationProcessed { location in
+        Task {
+            await coordinator.setOnLocationProcessed { [weak self] location in
```
mahmoud-elmorabea, line 88: Good catch!

Sources/Location/LocationModuleState.swift
```diff
@@ -96,9 +117,8 @@ final class LocationModuleState {
                 // Run in parallel — independent operations on different actors.
                 // Drops queued direct-HTTP metrics so rows captured under the
                 // previous user are not flushed under the next one.
-                async let resetResult: Result<Void, GeofenceSyncError> =
-                    DIGraphShared.shared.geofenceSyncCoordinator.reset()
-                async let cleared: Void = DIGraphShared.shared.geofenceEventTracker.clearPending()
+                async let resetResult: Result<Void, GeofenceSyncError> = di.geofenceSyncCoordinator.reset()
+                async let cleared: Void = di.geofenceEventTracker.clearPending()
```
mahmoud-elmorabea: Similar to previous comment, do we need to drop these now that we pin userId in the event itself?

mahmoud-elmorabea (approved):

Sources/Location/LocationModuleState.swift
mrehan27: Good call — already dropped. The `clearPending()` call was removed in the #1082 followup (now merged), so after rebasing this PR the ResetEvent observer is just `_ = await di.geofenceSyncCoordinator.reset()`. Per-row userId stamping makes session-boundary clearing unnecessary.

PR #1089, opened 2026-06-10
feat: Added StorageManger and associated tests.

hollyschilling-cio:

---

> [!NOTE]
> **Medium Risk**
> SDK init now opens an encrypted local database keyed by the CDP API key; failures are handled as nil but add a new dependency and filesystem surface on the critical startup path.
> 
> **Overview**
> Introduces **encrypted on-disk storage** for upcoming aggregation work by adding **`StorageManager`** (SyncSqlCipher/SQLCipher) with an initial `sdk_meta` key/value schema and migrations.
> 
> **`DIGraphShared`** gains `registerStorageManager(cdpApiKey:)`, which opens `cio.db` under Application Support (keyed by CDP API key), runs migrations, and registers the manager—or returns `nil` and logs on failure. **`DataPipeline.initialize`** now calls this before creating **`DataPipelineImplementation`**, which exposes an optional `storageManager` from the graph.
> 
> Packaging updates: **SyncSqlCipher 1.1.0** on `CioInternalCommon` (SPM + CocoaPods), Swift tools **5.10**, unit tests for meta CRUD/migrations, and internal API docs. Test SwiftLint re-enables `force_try`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cbde881a1c354fd21f07e109034452f37b000f39. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -50,7 +50,7 @@ let package = Package(
         
         // SSE (Server-Sent Events) client for real-time in-app messaging
         .package(url: "https://github.com/LaunchDarkly/swift-eventsource.git", .upToNextMajor(from: "3.3.0")),
-        .package(url: "https://github.com/customerio/SyncSqlCipher.git", from: "1.0.0")
+        .package(url: "https://github.com/customerio/SyncSqlCipher.git", from: "1.1.0")
     ],
     targets: [ 
         // Common - Code used by multiple modules in the SDK project.
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.5
+// swift-tools-version:5.10
 
 /**
  Manifest file for Swift Package Manager. This file defines our Swift Package for customers to install our SDK modules into their app. 
@@ -49,12 +49,14 @@ let package = Package(
         .package(name: "CioAnalytics", url: "https://github.com/customerio/cdp-analytics-swift.git", .exact("1.7.3+cio.1")),
         
         // SSE (Server-Sent Events) client for real-time in-app messaging
-        .package(url: "https://github.com/LaunchDarkly/swift-eventsource.git", .upToNextMajor(from: "3.3.0"))
+        .package(url: "https://github.com/LaunchDarkly/swift-eventsource.git", .upToNextMajor(from: "3.3.0")),
+        .package(url: "https://github.com/customerio/SyncSqlCipher.git", from: "1.0.0")
     ],
     targets: [ 
         // Common - Code used by multiple modules in the SDK project.
-        // this module is *not* exposed to the public. It's used internally. 
+        // this module is *not* exposed to the public. It's used internally.
         .target(name: "CioInternalCommon",
+                dependencies: [.product(name: "SyncSqlCipher", package: "SyncSqlCipher")],
                 path: "Sources/Common"),
         .testTarget(name: "CommonTests",
                     dependencies: ["CioInternalCommon", "SharedTests"],
--- a/Package.swift
+++ b/Package.swift
@@ -50,7 +50,7 @@ let package = Package(
         
         // SSE (Server-Sent Events) client for real-time in-app messaging
         .package(url: "https://github.com/LaunchDarkly/swift-eventsource.git", .upToNextMajor(from: "3.3.0")),
-        .package(url: "https://github.com/customerio/SyncSqlCipher.git", from: "1.1.0")
+        .package(url: "https://github.com/customerio/SyncSqlCipher.git", exact: "1.1.0")
     ],
     targets: [ 
         // Common - Code used by multiple modules in the SDK project.
--- a/Sources/Common/DIGraphShared+StorageManager.swift
+++ b/Sources/Common/DIGraphShared+StorageManager.swift
@@ -14,10 +14,6 @@ public extension DIGraphShared {
     /// gracefully; the aggregation engine will be a no-op in that case.
     @discardableResult
     func registerStorageManager(cdpApiKey: String) -> StorageManager? {
-        if let existing: StorageManager = getOptional(StorageManager.self) {
-            return existing
-        }
-
         do {
             let path = Self.databasePath(for: cdpApiKey)
             let db = try Database(path: path, key: cdpApiKey)
--- a/Sources/Common/DIGraphShared+StorageManager.swift
+++ b/Sources/Common/DIGraphShared+StorageManager.swift
@@ -1,0 +1,49 @@
+import Foundation
+import SyncSqlCipher
+
+public extension DIGraphShared {
+    /// Creates, migrates, and registers the ``StorageManager`` for a given CDP API key.
+    ///
+    /// Follows the same pattern as ``registerPendingPushDeliveryStore(appGroupId:)``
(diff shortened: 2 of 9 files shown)


Sources/Common/DIGraphShared+StorageManager.swift
```diff
@@ -0,0 +1,49 @@
...
+    /// ready for any code that calls ``storageManager``.
+    ///
+    /// Returns `nil` and logs an error if the database cannot be opened or
+    /// migrations fail. Callers that depend on storage must handle a nil result
+    /// gracefully; the aggregation engine will be a no-op in that case.
+    @discardableResult
+    func registerStorageManager(cdpApiKey: String) -> StorageManager? {
+        if let existing: StorageManager = getOptional(StorageManager.self) {
+            return existing
+        }
```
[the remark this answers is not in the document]
mrehan27: This looks valid

Sources/Common/DIGraphShared+StorageManager.swift
```diff
@@ -0,0 +1,49 @@
...
+
+        do {
+            let path = Self.databasePath(for: cdpApiKey)
+            let db = try Database(path: path, key: cdpApiKey)
```
mrehan27, line 19: So the database is gated behind `cdpApiKey`, which means we can't really access it until the key is available. For scenarios like wrapper SDKs where initialization happens later, or for push/geofence metrics that can occur before SDK initialization, this wouldn't be usable because the `cdpApiKey` is still unknown. What's the long term plan for handling these cases where the key isn't available or `DataPipeline` module is not initialized yet?

Package.swift
```diff
@@ -49,12 +49,14 @@ let package = Package(
         .package(name: "CioAnalytics", url: "https://github.com/customerio/cdp-analytics-swift.git", .exact("1.7.3+cio.1")),
         
         // SSE (Server-Sent Events) client for real-time in-app messaging
-        .package(url: "https://github.com/LaunchDarkly/swift-eventsource.git", .upToNextMajor(from: "3.3.0"))
+        .package(url: "https://github.com/LaunchDarkly/swift-eventsource.git", .upToNextMajor(from: "3.3.0")),
+        .package(url: "https://github.com/customerio/SyncSqlCipher.git", from: "1.1.0")
```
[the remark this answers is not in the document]
mrehan27: This also looks valid

api-docs/internal/CioInternalCommon.api
```diff
@@ -507,6 +507,13 @@ public interface CioInternalCommon.SharedKeyValueStorage {
 	public fun func setData(_ value: Data?, forKey key: KeyValueStorageKey);
 	public fun func deleteAll();
 }
+public final class CioInternalCommon.StorageManager {
+	public final val db: Database;
+	public fun init(db: Database);
```
mahmoud-elmorabea, line 512: Is this really worth the DB and the full wrapper around SqlCihper? On Android I use a simple [key/value storage](https://github.com/customerio/customerio-android/pull/723/changes#diff-e31b8c519f704bb0603ab64cde4fe6bd9900a5fb6301d5e4b24084fcceff52e4)

mahmoud-elmorabea: Claude says that ~8 of ~255 public entry points are touched from that wrapper. Can we start with simple encrypted storage and if it grows we can migrate the abstraction to using a DB which will be mostly invisible to callers.

My concerns is the cost we pay for: 
- Extra size of binary
- Maintenance for updating the wrapper with SqlCihper changes
- Runtime cost of eager IO init/migration of the DB

hollyschilling-cio: This PR separates the introduction of the database from its uses. That's why it only has ~8 methods used. 

The mistake in your reasoning is believing that a key/value store is faster than SQLite. In any nontrivial case, SQLite is faster and safer due to WAL. The SharedPreferences use in the code you linked to actually supports the use of a database. SharedPreferences is discouraged from being used because it causes UI stutters when `apply` is called, which is what triggers the writes to disk. New Android code should be using Compose DataStore, which is backported all the way to Android 4.4.

Using a database also gives the project ACID-compliant storage with thread safety guarantees. Geofencing, Location, and Live Activities all will benefit immediately from adopting it. That's not even touching the fact that we get encryption for free in the process. 

I would also encourage you to look at how much code we have dedicated to reading and writing data to disk already in the project. In the long term, using a database in the library will cause the code maintenance burden will go down while reliably goes up. 

As for your other concerns:
- Migrations take around 10ms the first time and maybe 1-3ms to verify on subsequent launches. These are not multi-gigabyte databases where a migration takes prolonged timelines.
- The addition to the static binary size is about 300 KB. We ship 10x that in mocks that aren't needed but are also somehow load bearing. 
- The SqlCipher is checked in, not linked in, to the wrapper. That means it never needs to be changed unless there are new features added that we want to use. Since it that code is C, we don't need to constantly chase newly released language versions either. The biggest concern is that the wrapper is written in Swift 5.10, which Apple could deprecate in a few years. At that point, a migration will need to happen. Fortunately, I already wrote the Swift 6.2 version of the wrapper as well.

Sources/Common/DIGraphShared+StorageManager.swift
hollyschilling-cio: The long term plan is to fix the initialization sequence of the SDK. The complexity of our startup, even in cross-platform environments, is a self-inflicted wound. Our initialization does not need to be this messy. 

Until then, things that require may require storage before DataPipelines is initialized should not use this.

api-docs/internal/CioInternalCommon.api
mahmoud-elmorabea: I don't believe key/value storage is faster at runtime (I didn't suggest that), my main point was simplicity. That key/value code is also kinda there and much easier to maintain, it's not only pure LOC. My main point is that we don't really need much of DB operations complex queries/updates, we mostly load stuff and replace them.

Corresponding to the 3 points you mentioned:
- The main point about IO access was the eager aspect, not the speed
- We should then remove the mocks that's not an reason to overlook the extra size added
- My main point was about the extra overhead of manually maintaining our own port

Sources/Common/DIGraphShared+StorageManager.swift
mrehan27: That's a separate discussion, and I'd prefer not to go too deep into it within this PR.

For clarity though, this change then alone doesn't solve most of the cases where we need storage today. e.g., storing/retrieving the `cdpApiKey`, `userId`, or metric events that may occur while the app is in the background. Because of that, this cannot be used for metric/geofence events yet and currently only helps with use cases that live entirely within the `DataPipelines` module.

Tests/.swiftlint.yml
```diff
@@ -9,7 +9,7 @@ disabled_rules:
   # let foo = try? ...
   # XCTAssertNotNil(foo)
   # XCTAssertEqual(foo?.bar, "")
-  # - force_try
+  - force_try
```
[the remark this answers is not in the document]
hollyschilling-cio: This is deliberate and only for the Test project because force try is useful in testing cases.

PR #1092, opened 2026-06-12
fix: move generated mocks out of shipped SDK products

mahmoud-elmorabea:
## What

Generated `AutoMockable` mocks lived in `Sources/<Module>/autogenerated/` and were therefore compiled into the `.library` products customers install. They are **test-only**, yet were shipped in the binary — and for **dynamic-framework / CocoaPods** consumers (where unreferenced `public` symbols are *not* dead-stripped) they were retained in full and leaked into the SDK's public API surface (autocomplete).

This moves the generated mocks into the **`SharedTests`** test-support target, which is **not a product** and so is unreachable by consumers.

## How

- **`Sources/Templates/AutoMockable.stencil`** — add a `testableImports` loop that emits `@testable import`, so relocated mocks can still reach module-**internal** protocols (e.g. `SimpleTimer`, `GistProvider`).
- **`Makefile`** — split generation per template:
  - `AutoDependencyInjection.stencil` → stays in `Sources/<Module>/autogenerated/` (production code, unchanged).
  - `AutoMockable.stencil` → `Tests/Shared/Mocks/Generated/<Module>.AutoMockable.generated.swift`. Unique filenames are required because SPM/Xcode disallows duplicate source basenames within a target.
- **`Package.swift`** — `SharedTests` now depends on every module that has mocks. These are **test-only** deps (`SharedTests` is not a `.library` product).
- **`Tests/`** — mocks are now internal to `SharedTests`, so test files that use them switch to `@testable import SharedTests`.

## Local verification (Xcode 26.0)

- ✅ `build-for-testing` of the full `Customer.io-Package` succeeds.
- ✅ `MessagingPushTests` (107), `DataPipelineTests`, `MessagingPushAPNTests` pass (exercise cross-module + internal mocks).
- ✅ `swiftlint --strict`: 0 violations; generated mocks `swiftformat`-formatted to repo convention.
- ✅ **Symbol proof:** `nm` on the compiled `CioInternalCommon` product shows **0 of the 49** generated mock classes remain; `EventBusMock` = 0 in product vs 273 in `SharedTests`. (The remaining `Mock`/`MockCollection`/`AutoMockable` symbols are the production marker protocol + test-utility types, intentionally kept.)

## Size impact

[...]

PR #1093, opened 2026-06-12
fix: move generated mocks into per-module mock targets (test isolation)

mahmoud-elmorabea:
## What

**Alternative to #1092**, addressing review feedback that #1092 collapsed per-module test isolation.

Same goal — get the generated `AutoMockable` mocks out of the shipped products — but done so that **each test target stays isolated to its own module chain**.

## The problem with #1092

#1092 put all mocks in `SharedTests` and made `SharedTests` depend on every module. Since every test target depends on `SharedTests`, each test target then transitively linked **all** modules. For a modular SDK where modules self-register and communicate via `DIGraphShared` + `EventBus`, that invalidates tests meant to validate a module's behavior when sibling modules are **absent**.

## This approach

Each module gets its own **non-product mock target** under `Tests/Mocks/<Module>`:
`CioInternalCommonMocks`, `CioDataPipelinesMocks`, `CioMessagingPushMocks`, `CioMessagingPushAPNMocks`, `CioMessagingPushFCMMocks`, `CioMessagingInAppMocks`, `CioTrackingMigrationMocks`.

A test target depends only on the mock targets it actually uses:

| Test target | mock-target deps | sibling modules linked |
|---|---|---|
| `MessagingPushTests` | CommonMocks, MessagingPushMocks | **none** |
| `DataPipelineTests` | CommonMocks, DataPipelineMocks | none |
| `MessagingInAppTests` | CommonMocks, InAppMocks | none |
| `MessagingPushAPNTests` | CommonMocks, MessagingPushMocks, APNMocks | none |

`SharedTests` stays `CioInternalCommon`-only (it uses no generated mocks).

- `AutoMockable.stencil`: add `testableImports` loop emitting `@testable import`.
- `Makefile`: split generation — DI stays in `Sources/<Module>/autogenerated`; mocks → `Tests/Mocks/<Module>`.
- `Package.swift`: 7 mock `.target`s (not products); test targets wired to only what they need.
- `Tests/`: files using a mock `@testable import Cio<Module>Mocks`.

## Local verification (Xcode 26.0)

- ✅ Full `build-for-testing` succeeds (incl. `@testable import` in the non-test mock targets).
- ✅ `MessagingPush` / `DataPipeline` / `MessagingPushAPN` / `MessagingInApp` suites pass.
- ✅ `swiftlint --strict`: 0 violations.
- ✅ **Mocks out of product:** `nm` on compiled `CioInternalCommon` → 0 of 49 mock classes.
- ✅ **Isolation restored:** the `MessagingPushTests.xctest` binary links **0 symbols** from `CioDataPipelines` / `CioMessagingInApp` / `CioMessagingPushAPN` / `CioMessagingPushFCM` / `CioTrackingMigration` / `CioLocation`, while keeping its own modules (`CioMessagingPush`, `CioInternalCommon`).

## Compare with #1092
Open both; this PR adds ~7 targets but preserves module isolation. #1092 is simpler but links all modules into every test target.

---

> [!NOTE]
> **Low Risk**
> Test and build-graph refactor only; production library targets lose embedded mocks, with no runtime or public API behavior changes.
> 
> **Overview**
> Moves Sourcery-generated **AutoMockable** mocks out of shipped `Sources/<Module>/autogenerated` into **per-module SPM targets** under `Tests/Mocks/<Module>`, so customer products no longer compile mock code and each test bundle only links the mocks it needs.
> 
> **Build / codegen:** `make generate` now runs **AutoDependencyInjection** into production `Sources/.../autogenerated` and **AutoMockable** into `Tests/Mocks/...`, with a new `testableImports` Makefile arg. `AutoMockable.stencil` emits `@testable import` for those modules. `Package.swift` adds seven non-product mock targets (`CioInternalCommonMocks`, `CioDataPipelinesMocks`, etc.) and wires each `*Tests` target to the matching subset; `SharedTests` stays on `CioInternalCommon` only.
> 
> **Tests:** Files that use generated mocks add `@testable import Cio<Module>Mocks` (e.g. `CioInternalCommonMocks`). Generated mock sources under `Tests/Mocks/*` switch from plain `import` to `@testable import` of their source modules. Location’s in-source mock stub is removed in favor of the Common mocks target.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a1ce29d416b54de8cae9602a6a7e54a51ac26b45. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


hollyschilling-cio:
This is a very clever solution. Generally, `@testable` is only allowed within `testTarget`-declared code. In this case, the code includes the modifier in standard targets. As a side effect of the Swift 5 import system, the test attribute flag is inherited in the non-test module because the compile action is originating from a `testTarget`. This is not specifically defined behavior of the Swift compiler, but this appears to work for all of the Swift 5.x series. Swift 6 has different rules for imports, so it is not guaranteed to continue to work. 

For today, this is a massive win and we should take it. We have ample opportunity to address the underlying issues when we upgrade beyond Swift 6.

mrehan27 (approved):

PR #1095, opened 2026-06-15
chore(location): publish location fixes via EventBus

mrehan27:
Replaces the `LocationSyncCoordinator.onLocationProcessed` closure with a `LocationAcquiredEvent` published on the EventBus (added to `CioInternalCommon`) — the seam the geofence module subscribes to in the next PR.

The location module's observable behavior is unchanged — it still consumes every fix — though the fix now travels over the EventBus rather than a direct callback (so the event participates in the bus's per-type cache/replay, like other events).

### Stack — merge in order
1. #1094 — scaffold CioLocationGeofence module
2. **#1095 — publish location fixes via EventBus ⟵ this PR**
3. #1096 — move geofence implementation into CioLocationGeofence

[MBL-1782](https://linear.app/customerio/issue/MBL-1782/ios-extract-geofencing-into-a-dedicated-module)

---

> [!NOTE]
> **Medium Risk**
> Geofence re-arm now depends on EventBus delivery (cache/replay when observers register late) instead of a synchronous closure, which could shift timing slightly though the PR intends parity.
> 
> **Overview**
> Replaces the direct **`LocationSyncCoordinator` → geofence** callback with a shared **`LocationAcquiredEvent`** on the EventBus, so geofence wiring can react to processed fixes without holding a reference to the coordinator.
> 
> **`LocationAcquiredEvent`** is added in `CioInternalCommon` (carrying `LocationData`) and registered in **`EventTypesRegistry`**. On every processed fix, **`LocationSyncCoordinator`** now **`postEvent`s** that event and drops **`setOnLocationProcessed`**. **`LocationModuleState`** subscribes to **`LocationAcquiredEvent`** for the same first-run geofence re-arm path that used the closure.
> 
> Location caching, sync filtering, and pipeline tracking are unchanged; tests now assert the bus post instead of the observer callback, and the internal API surface is updated.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dab90aee99f8d343f988ce762cb3d02036dcc9b3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Sources/Common/Communication/Event.swift
+++ b/Sources/Common/Communication/Event.swift
@@ -51,7 +51,8 @@ public enum EventTypesRegistry {
             TrackGeofenceMetricEvent.self,
             RegisterDeviceTokenEvent.self,
             DeleteDeviceTokenEvent.self,
-            NewSubscriptionEvent.self
+            NewSubscriptionEvent.self,
+            LocationAcquiredEvent.self
         ]
     }
 
--- a/Sources/Location/Coordinator/LocationSyncCoordinator.swift
+++ b/Sources/Location/Coordinator/LocationSyncCoordinator.swift
@@ -8,27 +8,30 @@ actor LocationSyncCoordinator {
     private let dataPipeline: DataPipelineTracking?
     private let dateUtil: DateUtil
     private let logger: Logger
-    private var onLocationProcessed: (@Sendable (LocationData) -> Void)?
+    private let eventBusHandler: EventBusHandler
 
     init(
         storage: LastLocationStorage,
         filter: LocationFilter,
         dataPipeline: DataPipelineTracking?,
         dateUtil: DateUtil,
-        logger: Logger
+        logger: Logger,
+        eventBusHandler: EventBusHandler
     ) {
         self.storage = storage
         self.filter = filter
         self.dataPipeline = dataPipeline
         self.dateUtil = dateUtil
         self.logger = logger
+        self.eventBusHandler = eventBusHandler
     }
 
     /// Called for every new location (from setLastKnownLocation or requestLocationUpdate). Always updates cache; sends track via pipeline when filter allows and user is identified.
     func processLocationUpdate(_ location: LocationData) {
         storage.setCachedLocation(location)
-        // Used by the geofence first-run race rearm in `LocationModuleState`.
-        onLocationProcessed?(location)
+        // Signals the geofence first-run-refresh re-arm. Routed through the EventBus so
+        // geofence can observe fixes without a direct reference to this coordinator.
+        eventBusHandler.postEvent(LocationAcquiredEvent(location: location))
 
         guard filter.shouldSyncToServer(newLocation: location) else {
             logger.locationSyncFiltered()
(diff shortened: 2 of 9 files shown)


Sources/Common/Communication/Event.swift
```diff
@@ -51,7 +51,8 @@ public enum EventTypesRegistry {
             TrackGeofenceMetricEvent.self,
             RegisterDeviceTokenEvent.self,
             DeleteDeviceTokenEvent.self,
-            NewSubscriptionEvent.self
+            NewSubscriptionEvent.self,
+            LocationAcquiredEvent.self
```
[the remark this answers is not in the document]
mrehan27: Fair point on the wording. It's a mechanism change from the fire-and-forget closure, so the PR description is updated to drop the "behavior unchanged" phrasing (the location module's *observable* behavior is unchanged, but the fix now flows over the bus).

Two clarifications that keep this low/benign:
- The cache is **per-event-key** (`EventCacheManager.storage: [String: RingBuffer]`, `maxEventsPerType = 100`), so `LocationAcquiredEvent` gets its own 100-slot buffer and never evicts `ProfileIdentifiedEvent` or other business events.
- The consumer (the location module) registers its observer at module init, before any fix is posted, so `hasObservers` is true when events flow — the `!hasObservers → persist to disk` path isn't hit in the normal lifecycle.

Keeping the EventBus seam intentionally: it's what lets the geofence module subscribe without `CioLocation` depending on it (the point of the split), and it matches Android. No code change.

Sources/Location/Coordinator/LocationSyncCoordinator.swift
```diff
@@ -8,27 +8,30 @@ actor LocationSyncCoordinator {
...
     }
 
     /// Called for every new location (from setLastKnownLocation or requestLocationUpdate). Always updates cache; sends track via pipeline when filter allows and user is identified.
     func processLocationUpdate(_ location: LocationData) {
         storage.setCachedLocation(location)
-        // Used by the geofence first-run race rearm in `LocationModuleState`.
-        onLocationProcessed?(location)
+        // Signals the geofence first-run-refresh re-arm. Routed through the EventBus so
+        // geofence can observe fixes without a direct reference to this coordinator.
+        eventBusHandler.postEvent(LocationAcquiredEvent(location: location))
```
mahmoud-elmorabea, line 34: One note here, for a consumer without geofence module, these `LocationAcquiredEvent` events will get queued up (and saved locally) without ever getting drained since its only observer now lives in geofence module.

Can we confirm what happens in the absence of geofence module for these events?

What I see we can do is:
- Either have an empty consumer here in location module
- Or (preferred) make this optional somehow on the availability of the geofence module

mahmoud-elmorabea (approved):

Sources/Location/Coordinator/LocationSyncCoordinator.swift
mrehan27: Good catch — confirmed. In this PR the location module is still the observer, so nothing leaks yet, but once the next PR moves that observer into the geofence module, a location-only app has no consumer for `LocationAcquiredEvent` and those events would persist to disk undrained.

It's latent until the geofence module lands and the branch is unreleased, so I'll fix it in a dedicated follow-up (before `feature/geofence-on-device` merges up) — likely by gating emission on a consumer or making the event non-persistent.

PR #1096, opened 2026-06-15
chore(geofence): move geofence implementation into CioLocationGeofence

mrehan27:
Moves the geofence implementation (21 sources + tests) into `CioLocationGeofence`, lifts the orchestration from `LocationModuleState` into `GeofenceModule`, and relocates `bootstrapForBackgroundDelivery`. Sample apps register `GeofenceModule` (APN-UIKit via `Apps/Common/Package.swift`; CocoaPods-FCM via Podfile + AppDelegate).

### Stack — merge in order
1. #1094 — scaffold CioLocationGeofence module
2. #1095 — publish location fixes via EventBus
3. **#1096 — move geofence implementation into CioLocationGeofence ⟵ this PR**

[MBL-1782](https://linear.app/customerio/issue/MBL-1782/ios-extract-geofencing-into-a-dedicated-module)

---

> [!NOTE]
> **Medium Risk**
> Module boundary and public API moves (`bootstrapForBackgroundDelivery`, explicit `GeofenceModule` registration) can break apps that relied on geofence running inside Location alone; behavior is largely relocated rather than rewritten.
> 
> **Overview**
> Geofence runtime wiring is **extracted from `LocationModule`** into **`GeofenceModule` / `GeofenceModuleState`**: identify/reset/`LocationAcquiredEvent` subscriptions, foreground refresh, pending-metric flush, monitor bootstrap, and first-run rearm now live behind **`GeofenceModule.initialize()`**. **`bootstrapForBackgroundDelivery`** moves from **`LocationModule`** to **`GeofenceModule`** (breaking for integrators still calling the old type).
> 
> **`CioLocation`** stays location-only but gains **`LocationServices.getLastKnownLocation()`** (and coordinator cache read) so geofence refresh uses the public location API instead of internal storage. **`AppLifecycleNotifying`** / **`RealAppLifecycleNotifying`** are widened to `public` for cross-module reuse. Geofence **DI registrations and generated mocks** shift from the Location target to **LocationGeofence**.
> 
> Packaging and samples add the opt-in **`CustomerIOLocationGeofence`** / **`LocationGeofence`** product (`CustomerIO/LocationGeofence` subspec, CocoaPods Podfile, SPM `Apps/Common`, CI pod pin). Sample app delegates register **`GeofenceModule()`** alongside **`LocationModule`**. Tests and api-docs are updated for **`GeofenceModuleState.setup`** and **`CioLocationGeofence`** imports.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4c8c32276d40c6d15b7b884ce9b45394574b249e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


api-docs/CioLocation.api
```diff
@@ -6,11 +13,16 @@ public final class CioLocation.LocationModule {
...
 	public fun func requestLocationUpdate();
+	public fun func getLastKnownLocation() async -> LocationData?;
 }
 public enum CioLocation.LocationTrackingMode {
 }
+public final class CioLocation.RealAppLifecycleNotifying {
+	public fun init(notificationCenter: NotificationCenter = .default);
+	public fun func addDidBecomeActiveObserver(using block: @escaping () -> Unit) -> AppLifecycleObserverToken;
+	public fun func addDidEnterBackgroundObserver(using block: @escaping () -> Unit) -> AppLifecycleObserverToken;
+}
```
mahmoud-elmorabea, line 28: These are only exposed for the sake of geofence module, can we move them into common module or use @ spi ?

mahmoud-elmorabea (approved):

Sources/LocationGeofence/GeofenceModuleState.swift
```diff
@@ -0,0 +1,108 @@
...
+        lastSkippedForNoLocation.wrappedValue = true
+        Task { @MainActor [weak self] in
+            guard let self else { return }
+            guard let location = await self.locationServicesProvider().getLastKnownLocation() else { return }
+            self.lastSkippedForNoLocation.wrappedValue = false
+            _ = await di.geofenceSyncCoordinator.refresh(
+                latitude: location.latitude,
+                longitude: location.longitude
+            )
+        }
```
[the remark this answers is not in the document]
mrehan27: Valid that there's an extra `refresh` call in this window, but it can't cause a duplicate geofence API request: `refresh` is gated by in-flight dedup + a freshness check (skips the server fetch when within `remoteFetchRefreshTriggerRadius` of the last anchor), so a redundant call microseconds later is at most a local cache re-rank.

It's a deliberate trade-off. The async cached-location read means the arm can't be atomic with the rearm observer either way: arming up-front risks this (benign, deduped) refresh; arming only on the nil result risks the opposite — a fix in the window misses the arm and delays registration until the next fix/foreground, which can drop a real transition. We chose to err toward registering promptly. No change.

api-docs/CioLocation.api
mrehan27: Good point — these are only public so the geofence module can reuse the lifecycle observer, not for customers. I'll tighten them in the same pre-release follow-up (before `feature/geofence-on-device` merges up), most likely via `@_spi(Internal)` so they stay usable cross-module but leave the public API. (`getLastKnownLocation()` stays public — it's the customer-facing counterpart to `setLastKnownLocation`.)

PR #1099, opened 2026-06-16
fix: Resolve duplicate event delivery bugs

hollyschilling-cio:
There is an intermittent test failure caused by a race condition in the Event Bus and Cache. The specific issue is that the Cache and EventBus each use their own isolation and that allows them to be out of sync for a short period when a new observer is registered at the same time an event comes in. The change is to move the memory-resident cache into the EventBus so they can ensure registering observers and dispatching an event are both atomic. 

---

> [!NOTE]
> **Medium Risk**
> Core SDK event delivery and persistence behavior changed (architecture swap), though the public `EventBusHandler` API is unchanged and the change is heavily tested for the race fix.
> 
> **Overview**
> Replaces the split **NotificationCenter** `SharedEventBus` + separate `EventCache` + `CioEventBusHandler` stack with **`CombinedCacheEventBusHandler`** backed by a single **`CioEventBus`** actor that owns both observers and an in-memory ring-buffer cache.
> 
> **Observer registration and cache snapshot are one atomic actor turn**, so a concurrent post can’t land in the cache and also be delivered live—addressing intermittent duplicate delivery when `addObserver` and replay used separate `await` boundaries. Replay still happens outside the actor; posts with no observers still go to **`EventStorage`**, and replayed events are removed from disk as before.
> 
> DI now wires **`eventBusHandler`** to `CombinedCacheEventBusHandler` only (drops `EventBus` / `EventBusObserversHolder` from the graph). **`EventBusHandler`** is protocol-only; **`EventBus.swift`** and **`EventBusMock`** are removed. Adds **`RegistrationToken`** (deinit hook for future per-observer teardown; auto-remove is commented out). Tests move to **`CioEventBusTests`** / **`CombinedCacheEventBusHandlerTests`**, including concurrent post+register stress cases; **`UnitTestBase`** no longer clears NotificationCenter observers on teardown.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0968b8a2e72bccfbbcd43d0747e51a3b7fbd8df9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Sources/Common/Communication/EventBusHandler.swift
```diff
@@ -10,12 +10,6 @@ public protocol EventBusHandler {
     func removeFromStorage<E: EventRepresentable>(_ event: E) async
 }
 
-// swiftlint:disable orphaned_doc_comment
-/// `EventBusHandler` acts as a central hub for managing events in the application.
-/// It interfaces with both an event bus for real-time event handling and an event storage system for persisting events.
-// sourcery: InjectRegisterShared = "EventBusHandler"
-// sourcery: InjectSingleton
-// swiftlint:enable orphaned_doc_comment
 public class CioEventBusHandler: EventBusHandler {
```
mahmoud-elmorabea: Is it intentional for this copy of `EventBusHandler` to remain in the source but not wired up in the DI graph?

Sources/Common/Communication/CioEventBus.swift
```diff
@@ -0,0 +1,81 @@
...
+    /// Registers an observer and atomically snapshots the current event cache for replay.
+    ///
+    /// Because registration and snapshot happen inside a single actor turn, any event
+    /// posted concurrently after this call returns is delivered directly to the observer
+    /// and will not appear in `eventsToReplay`.
+    ///
+    /// The cache is intentionally **not** cleared on registration: every observer, past
+    /// and future, receives the full history for its event type.
+    func addObserver(key: String, action: @escaping Action) -> ObserverRegistration {
+        let uuid = UUID()
```
mahmoud-elmorabea, line 40: I think the second identifier per instance of observer for the same type might be complicating this unnecessarily. I understand that today we cannot remove a specific observer but since all these observers are cleared as the singletons are deinit'd. The `removeAll` should be sufficient till we have a case that needs to remove an observer mid-session.

Sources/Common/Communication/CombinedCacheEventBusHandler.swift
```diff
@@ -0,0 +1,120 @@
...
+/// Replaces the `OperationQueue`-based `CioEventBusHandler` with a single `CioEventBus`
+/// actor that serializes observer registration and event enqueuing atomically, eliminating
+/// the race condition where an event could be both directly delivered and replayed.
+///
+/// `CioEventBusHandler` and `CombinedCacheEventBusHandler` both conform to `EventBusHandler`,
+/// allowing them to be swapped for testing or gradual rollout.
+/// `CioEventBusHandler` remains the default registered in the DI graph.
+// sourcery: InjectRegisterShared = "EventBusHandler"
+// sourcery: InjectSingleton
+// swiftlint:enable orphaned_doc_comment
```
mahmoud-elmorabea: The annotations here mean we can remove the old implementation of `EventBusHandler`, no?

Sources/Common/Communication/CombinedCacheEventBusHandler.swift
```diff
@@ -0,0 +1,120 @@
...
+    private let bus: CioEventBus
+    let eventStorage: EventStorage
+    let logger: Logger
+
+    public init(eventStorage: EventStorage, logger: Logger) {
+        self.bus = CioEventBus()
+        self.eventStorage = eventStorage
+        self.logger = logger
+        Task { [weak self] in
+            await self?.loadEventsFromStorage()
```
mahmoud-elmorabea, line 20: Is there a guarantee that this could happen after an observer is added, causing that observer to miss previously persisted events?

mahmoud-elmorabea: The previous implementation's queue was FIFO I think and it might have avoided that?

Sources/Common/Communication/CombinedCacheEventBusHandler.swift
```diff
@@ -0,0 +1,120 @@
...
+                if let typed = event as? E {
+                    action(typed)
+                } else {
+                    self?.logger.debug(
+                        "CombinedCacheEventBusHandler: Event type mismatch for key \(E.key)"
+                    )
+                }
+            }
+            // Replay is performed outside the actor to keep delivery concurrent.
+            // removeFromStorage is safe to call for events not on disk (no-op).
```
mahmoud-elmorabea, line 62: Can this cause a new posted event to be consumed by the observer before the older history has been fully replayed?

Sources/Common/Communication/CombinedCacheEventBusHandler.swift
```diff
@@ -0,0 +1,120 @@
...
+    /// Registration and the cache snapshot are taken atomically inside the actor.
+    /// Events posted after registration returns are delivered directly to the observer;
+    /// events already in the cache at registration time are replayed exactly once.
+    public func addObserver<E: EventRepresentable>(
+        _ eventType: E.Type, action: @escaping (E) -> Void
+    ) {
+        logger.debug("CombinedCacheEventBusHandler: Adding observer for \(eventType)")
+        Task { [weak self] in
+            guard let self else { return }
+            let registration = await bus.addObserver(key: E.key) { [weak self] event in
```
mahmoud-elmorabea, line 51: This `registration` is never kept

Tests/Common/Communication/CombinedCacheEventBusHandlerTests.swift
```diff
@@ -0,0 +1,184 @@
...
+    func test_loadEventsFromStorage_expectEventsSeededAndReplayed() async {
+        let event = ProfileIdentifiedEvent(identifier: "persisted")
+        mockEventStorage.loadEventsClosure = { key in
+            key == ProfileIdentifiedEvent.key ? [event] : []
+        }
+
+        let handler = makeHandler()
+        // Explicitly await loading so the cache is populated before the observer registers.
+        // This avoids a race with the background Task launched in init.
+        await handler.loadEventsFromStorage()
```
mahmoud-elmorabea, line 142: I don't believe this `await loadEventsFromStorage()` happens in production code, where `addObserver` can be called concurrently while loading events.

mahmoud-elmorabea (commented):
Two main points before we merge, the remaining are more of questions:
- Are we intentionally shipping the new implementation while leaving the old one in source unreachable?
- The scope of regression smoke of this spans all modules I believe, do you need help testing any features or is that all covered?

Sources/Common/Communication/CombinedCacheEventBusHandler.swift
hollyschilling-cio: Correct. I left the old implementation in so we could easily swap back and forth for testing, but it's now dead code. I can amend this PR to include the removal of the old implementation if we feel comfortable about this and don't want to preserve the testing path.

Sources/Common/Communication/EventBusHandler.swift
hollyschilling-cio: Duplicate of my response on CombinedCacheEventBusHandler. Yes, the old `CioEventBusHandler` is dead code with the removal of this comment. I preserved it for testing purposes to make it easy to swap the DI back and forth to validate. We could amend the PR to include the removal if the team is confident.

Sources/Common/Communication/CioEventBus.swift
hollyschilling-cio: There is one correction I want to make in your comment. The EventBusHandler (any implementation) is owned as a singleton by the DIGraphShared, which is itself a static global. One of the issues we have in some of our tests is that the DIGraphShared instance persists the registrations between tests. Eventually, we should fix that. 

The RegistrationToken class is absolutely unneeded in this PR. Forward looking, it's a pattern already being used in Geofencing and something we can easily adopt elsewhere. Plus, this lays the groundwork to make it trivial to handle deregistration in a future PR when we are comfortable modifying the EventBusHandler signature. 

If it's a big issue, I could remove it here, but if not I feel it deserves its place here.

Sources/Common/Communication/CombinedCacheEventBusHandler.swift
hollyschilling-cio: The previous queue was a concurrent queue and made no guarantees. We still have no guarantees. It's not an improvement, but also not a regression. 

Image

Sources/Common/Communication/CombinedCacheEventBusHandler.swift
hollyschilling-cio: Correct. That is not a bug. The registration variable here is an `ObserverRegistration`, which contains the `RegistrationToken` (which isn't in use) and the cached events that will be replayed (see line 69).

Sources/Common/Communication/CombinedCacheEventBusHandler.swift
hollyschilling-cio: This doesn't cause it, but it does allow it. This was the behavior before as well. 

Once an observer is registered, it will receive new events as they come in immediately. The change from the old version is that it only get the event once. The old EventBusHandler would deliver it to the new observer directly and add it to the queue of events that would be replayed to it, causing duplicate delivery.

Tests/Common/Communication/CombinedCacheEventBusHandlerTests.swift
hollyschilling-cio: Generally, `loadEventsFromStorage` only gets called from the background of the `init` function. For testing purposes, we call it here explicitly and await it to ensure the EventCache is populated before moving on to the real testing. This is exclusively for testing and would probably be handled by just adding a 100ms delay, but this is cleaner.

Sources/Common/Communication/CombinedCacheEventBusHandler.swift
hollyschilling-cio: When/if we reach a point where we are using the `RegistrationToken` to remove old registrations, we'll return `registration.token` from this method.

Sources/Common/Communication/EventBusHandler.swift
mahmoud-elmorabea: Yes, I think we should combine both PRs for the same release, probably merge the pruning PR into this one, so we avoid multiple releases for this change

Sources/Common/Communication/CioEventBus.swift
mahmoud-elmorabea: Makes sense! Good to have it and push for cleanups in module that could use it.

Sources/Common/Communication/CombinedCacheEventBusHandler.swift
mahmoud-elmorabea: As mentioned above, let's make sure to get the removal in the same release as this one

Sources/Common/Communication/CombinedCacheEventBusHandler.swift
mahmoud-elmorabea: I doubled checked and you are correct! Maybe it's a good candidate for a follow up

mahmoud-elmorabea (approved):
Looks good 💯 

Thanks for replying to comments. One note, let's mark this as `fix` instead of `refactor` so this change is released

PR #1103, opened 2026-06-16
chore(geofence): make location fix event transient and gate lifecycle types with @_spi

mrehan27:
Follow-up to the geofence module split stack (#1094, #1095, #1096) addressing two review items.

## Event persistence
`EventRepresentable` gains `isPersistent` (defaults to `true`); `LocationAcquiredEvent` overrides it to `false`. The EventBus persists an event to disk only when it has no observers, for later replay. A location fix is transient — in a location-only app (no geofence module observing it) there is no consumer, so persisting would accumulate undrained files on disk.

## `@_spi(Internal)` lifecycle types
`AppLifecycleNotifying`, `AppLifecycleObserverToken`, and `RealAppLifecycleNotifying` move from `public` to `@_spi(Internal) public` so the geofence module (which depends on Location) can reuse them without exposing SDK-internal infrastructure on the public API.

## Tests
- Added `test_postEventAndWait_givenNoObserversAndTransientEvent_expectEventNotStored` guarding the transient path. The persistent path is already covered by `test_postEvent_givenNoObservers_expectEventStoredInFile`.
- Test consumers of the lifecycle types now import via `@_spi(Internal)`.

## Notes
- `CioLocation.api` shows no diff: sourcekitten still lists `@_spi` symbols, so the baseline does not reflect the visibility change (the symbols are still hidden from a plain `import CioLocation`).
- `CioInternalCommon.api` updated for the new `isPersistent` requirement.

---

> [!NOTE]
> **Low Risk**
> Targeted EventBus persistence behavior and SPI visibility tightening; existing persistent events keep default behavior and the change is covered by a new unit test.
> 
> **Overview**
> Adds **`isPersistent`** on `EventRepresentable` (default `true`) and gates **disk persistence** in `EventBusHandler` so unobserved events are stored only when `isPersistent` is true. **`LocationAcquiredEvent`** sets `isPersistent` to `false`, avoiding undrained on-disk event files in location-only apps with no geofence observer.
> 
> **`AppLifecycleNotifying`**, **`AppLifecycleObserverToken`**, and **`RealAppLifecycleNotifying`** are now **`@_spi(Internal) public`** instead of plain `public`; geofence and tests import them via `@_spi(Internal) import CioLocation`. **`CioInternalCommon.api`** documents the new protocol property; a test asserts transient events are not stored when there are no observers.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1581cabe6a43f04471d2597aa6eec06165a50429. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Sources/Common/Communication/Event+Location.swift
+++ b/Sources/Common/Communication/Event+Location.swift
@@ -25,6 +25,10 @@ public struct LocationAcquiredEvent: EventRepresentable {
     public let location: LocationData
     public let timestamp: Date
 
+    /// Not persisted: a location fix is only useful in the moment, so it is dropped when unobserved
+    /// rather than written to disk for replay.
+    public var isPersistent: Bool { false }
+
     public init(storageId: String = UUID().uuidString, location: LocationData, timestamp: Date = Date(), params: [String: String] = [:]) {
         self.storageId = storageId
         self.location = location
--- a/Sources/Common/Communication/Event.swift
+++ b/Sources/Common/Communication/Event.swift
@@ -16,6 +16,10 @@ public protocol EventRepresentable: Equatable, Codable {
     var params: [String: String] { get }
     /// Timestamp indicating when the event was created.
     var timestamp: Date { get }
+    /// Whether the event should be persisted to disk when posted with no observers, so it can be
+    /// replayed once an observer registers. Defaults to `true`; transient events override to `false`
+    /// and are dropped when unobserved. The in-memory cache applies regardless.
+    var isPersistent: Bool { get }
 }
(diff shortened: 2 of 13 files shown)


mahmoud-elmorabea (approved):
Looks good! 💯 

One big note to callout is that there is a PR that rewrites event bus going to main -> https://github.com/customerio/customerio-ios/pull/1099

So when we go this feature branch to main, we need to re-do this persistent logic in the new implementation

PR #1107, opened 2026-06-17
feat: Uses Keychain to store DB password

hollyschilling-cio:
This PR introduces the `DbKeyProvider` protocol, it's mock implementation, and the concrete implementation `KeychainDbKeyProvider`. This serves as an intermediary to map the CDP API Key to the actual encryption key, which is stored in the Keychain.

---

> [!NOTE]
> **High Risk**
> Changes how local encrypted data is keyed; existing on-disk databases encrypted with the CDP API key will not decrypt with the new Keychain-derived keys without a migration path.
> 
> **Overview**
> **Local SQLCipher databases no longer use the CDP API key as the encryption password.** `registerStorageManager(cdpApiKey:)` now resolves a separate key via injectable `DbKeyProvider` before opening `Database(path:key:)`.
> 
> A new **`KeychainDbKeyProvider`** implements that contract: for each account (the CDP API key), it loads an existing key from the Keychain or creates a 32-byte random hex key, stored as a generic password with `kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly`. `DbKeyProvider` is registered as a singleton on `DIGraphShared`, with Sourcery-generated `DbKeyProviderMock` and tests that verify storage registration uses the provider and fails closed when key lookup throws.
> 
> **Review note:** Upgrading apps that already have `cio.db` files encrypted with the old API-key passphrase may be unable to open those databases until a migration strategy exists (not in this diff).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 06f04981f07592c2e4ad9bcc0d2f56d844474665. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Sources/Common/DIGraphShared+StorageManager.swift
```diff
@@ -16,7 +16,8 @@ public extension DIGraphShared {
     func registerStorageManager(cdpApiKey: String) -> StorageManager? {
         do {
             let path = Self.databasePath(for: cdpApiKey)
-            let db = try Database(path: path, key: cdpApiKey)
+            let dbKey = try dbKeyProvider.getOrCreateDbKey(account: cdpApiKey)
+            let db = try Database(path: path, key: dbKey)
```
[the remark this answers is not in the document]
hollyschilling-cio: This is in a feature branch and was never released, so this is not a concern.

Sources/Common/Util/KeychainDbKeyProvider.swift
```diff
@@ -0,0 +1,107 @@
...
+            kSecClass: kSecClassGenericPassword,
+            kSecAttrService: Self.service,
+            kSecAttrAccount: account,
+            kSecValueData: data,
+            kSecAttrAccessible: kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly
+        ]
+        let status = SecItemAdd(attributes as CFDictionary, nil)
+        guard status == errSecSuccess else {
+            throw KeychainDbKeyError.writeFailed(status: status)
+        }
```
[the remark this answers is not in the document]
hollyschilling-cio: This is handled by the mutex in `KeychainDbKeyProvider. getOrCreateDbKey` and singleton registration of the object. This could be an issue if rogue code in the app bypassed this class to modify the keychain directly, but that is not the case and should never become the case.

Sources/Common/Util/KeychainDbKeyProvider.swift
```diff
@@ -0,0 +1,107 @@
...
+// sourcery: InjectRegisterShared = "DbKeyProvider"
+// sourcery: InjectSingleton
+/// Keychain-backed implementation of ``DbKeyProvider``.
+///
+/// On first call for a given account, generates a 32-byte random key (hex-encoded)
+/// and stores it as a `kSecClassGenericPassword` item under
+/// `kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly`. This allows the key to
+/// survive app restarts and be accessible during background tasks, while never
+/// leaving the device via iCloud or backups.
+public struct KeychainDbKeyProvider: DbKeyProvider, @unchecked Sendable {
```
mrehan27, line 20: Could this be simply `Sendable` instead of `@unchecked Sendable`?

Sources/Common/Util/KeychainDbKeyProvider.swift
```diff
@@ -0,0 +1,107 @@
...
+    private func storeKey(_ key: String, account: String) throws {
+        guard let data = key.data(using: .utf8) else {
+            throw KeychainDbKeyError.invalidData
+        }
+        let attributes: [CFString: Any] = [
+            kSecClass: kSecClassGenericPassword,
+            kSecAttrService: Self.service,
+            kSecAttrAccount: account,
+            kSecValueData: data,
+            kSecAttrAccessible: kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly
```
mrehan27, line 82: `kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly` looks like the right call given how we want to use the database today. One question though, is the database itself backed up, or is it excluded from backups as well?

Tests/Common/Util/KeychainDbKeyProviderTests.swift
```diff
@@ -0,0 +1,42 @@
+@testable import CioInternalCommon
+@testable import CioInternalCommonMocks
+import XCTest
+
+class KeychainDbKeyProviderTests: XCTestCase {
```
mrehan27, line 5: These tests inject `DbKeyProviderMock`, so they seem to cover only the `registerStorageManager` wiring. The real `KeychainDbKeyProvider` may never be tested, and the key loading/persistence path (`loadKey`) appears to have little or no coverage. Should we add tests for that as well?

mrehan27 (approved):

Sources/Common/Util/KeychainDbKeyProvider.swift
hollyschilling-cio: Good question. There was originally, but but it may not be true anymore. Let me review and I'll update if I can't find a reason.

Sources/Common/Util/KeychainDbKeyProvider.swift
hollyschilling-cio: Let's verify it before merging into main, but the DB should **not** be backed up. Keychain entries can survive even app reinstalls, which potentially means that they db encryption key could be the same even after the DB is recreated, but that shouldn't be a concern.

Tests/Common/Util/KeychainDbKeyProviderTests.swift
hollyschilling-cio: The real Keychain DbKeyProvider cannot be tested because the keychain is not available in test contexts. That's why it's isolated and minimized as much as possible.

Sources/Common/Util/KeychainDbKeyProvider.swift
hollyschilling-cio: `@unchecked` is required because `NSLock` is not `Sendable`. This irony is perfect.

PR #1112, opened 2026-06-18
feat: Added Filtering engine, plugin, and tests

hollyschilling-cio:
This Adds the Engine, Plugins, Data Structures, and Tests for filtering events. I know this is nearly 1,000 LOC, but it's mostly tests.

---

> [!NOTE]
> **High Risk**
> Replaces core event-bus delivery semantics (risk of missed/duplicate internal events) and adds server rules that can silently drop or throttle customer analytics events, including on identify.
> 
> **Overview**
> **4.5.3** ships a **duplicate event delivery fix** and a new **server-driven analytics policy** layer, plus routine release/version/CI updates.
> 
> The internal event path is reworked: **`SharedEventBus` / `CioEventBusHandler` / `NotificationCenter`** are removed in favor of a single **`CioEventBus` actor** plus **`CombinedCacheEventBusHandler`**, so observer registration and cache snapshots happen atomically and late subscribers no longer get the same event twice. DI now wires **`CombinedCacheEventBusHandler`** as the singleton **`EventBusHandler`**.
> 
> Data Pipelines gains **`EventPolicyEngine`**, **`EventPolicyPlugin`**, and **`AggregationRuleset`** decoding. Track/screen/identify (and related) events can be **dropped by filter rules** or **rate limits** (profile vs device scope in SQLite); rules load from Segment integration settings, persist in **`sdk_meta`**, and **`aggregation_state`** is cleared on profile switch / **`clearIdentify`**. The policy plugin is registered **before** published-events plugins so blocked events do not fan out on the event bus.
> 
> Also: **`aggregation_state`** migration, Danger CI (**Node 18**, **danger@12**), **`dangerfile.js`** CommonJS require, changelog/podspecs/**`SdkVersion`**, and API doc updates reflecting the new handler.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5d6a07513dc2faeaf589840443247d155199b10e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Sources/DataPipeline/Plugins/EventPolicyPlugin.swift
```diff
@@ -0,0 +1,56 @@
...
+        self.storage = storage
+    }
+
+    func update(settings: Settings, type: UpdateType) {
+        struct IntegrationSettings: Codable {
+            let aggregationRules: AggregationRuleset?
+        }
+        let config: IntegrationSettings? = settings.integrationSettings(forKey: "Customer.io Data Pipelines")
+        let ruleset = config?.aggregationRules
+        engine.load(ruleset: ruleset)
```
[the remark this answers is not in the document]
hollyschilling-cio: This is by design. If the rules come through empty, we revert to not filtering in any way. This "fail open" behavior is by design.

Sources/Common/Storage/StorageManager.swift
```diff
@@ -55,6 +55,26 @@ private struct CreateSdkMetaSchema: Migration {
...
+
+    func up(_ ctx: MigrationContext) throws {
+        // Per-rule accumulator state (counters for rate limiting).
+        try ctx.execute(
+            """
+            CREATE TABLE IF NOT EXISTS aggregation_state (
+                rule_id         TEXT    NOT NULL PRIMARY KEY,
+                state_json      TEXT    NOT NULL,
+                last_flushed_at INTEGER NOT NULL DEFAULT 0,
+                scope           TEXT    NOT NULL DEFAULT 'profile'
```
mahmoud-elmorabea: I don't think `scope` should have a default, if for some reason, server returns a malformed response, we should ignore that rule instead of applying the wrong scope.

Sources/DataPipeline/EventPolicy/AggregationRuleset.swift
```diff
@@ -0,0 +1,49 @@
+import Foundation
+
+enum RuleScope: String, Codable, Sendable {
+    case profile
+    case device
+
+    init(from decoder: Decoder) throws {
+        let raw = try decoder.singleValueContainer().decode(String.self)
+        self = RuleScope(rawValue: raw) ?? .profile
```
mahmoud-elmorabea: I don't think this should have a default

Sources/DataPipeline/EventPolicy/EventPolicyEngine.swift
```diff
@@ -0,0 +1,67 @@
...
+    /// Returns `true` if the event is allowed through, `false` if it should be dropped.
+    func shouldAllow(eventType: String, name: String) -> Bool {
+        shouldAllow(eventType: eventType, name: name, now: Int64(Date().timeIntervalSince1970))
+    }
+
+    /// Time-injectable overload used by tests.
+    func shouldAllow(eventType: String, name: String, now: Int64) -> Bool {
+        guard let rs = processed.wrappedValue else { return true }
+
+        let exactKey = "\(eventType):\(name)"
```
mahmoud-elmorabea: That string formatting happens in at least 3 places, maybe we can centralize it

Sources/DataPipeline/EventPolicy/EventPolicyEngine.swift
```diff
@@ -0,0 +1,67 @@
...
+        let wildcardKey = "\(eventType):*"
+
+        if rs.filterKeys.contains(exactKey) || rs.filterKeys.contains(wildcardKey) {
+            return false
+        }
+
+        if let rl = rs.rateLimitsByKey[exactKey] ?? rs.rateLimitsByKey[wildcardKey] {
```
mahmoud-elmorabea: This is a bit difficult to read, can we spell out what `rl` and `rs` stand for?

Sources/DataPipeline/Plugins/EventPolicyPlugin.swift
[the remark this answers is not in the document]
mahmoud-elmorabea: Agree 👍

Sources/DataPipeline/Plugins/EventPolicyPlugin.swift
```diff
@@ -0,0 +1,56 @@
...
+
+        if let ruleset,
+           let json = try? JSONEncoder().encode(ruleset),
+           let payload = String(data: json, encoding: .utf8) {
+            try? storage?.setAggregationConfig(payload: payload)
```
mahmoud-elmorabea, line 27: Any parse failure here would keep the stored rules intact, should we treat it as empty rules? Maybe the rules we failed to parse are less restrictive than the ones on disk

Sources/DataPipeline/Plugins/EventPolicyPlugin.swift
```diff
@@ -0,0 +1,56 @@
...
+        } else if config != nil {
+            // Integration key was present but carried no rules — server explicitly removed them.
+            // Delete the cached config so stale rules aren't reloaded on next launch.
+            try? storage?.deleteAggregationConfig()
+        }
+    }
+
+    func track(event: TrackEvent) -> TrackEvent? {
+        engine.shouldAllow(eventType: "track", name: event.event) ? event : nil
+    }
```
mahmoud-elmorabea, line 37: I think we should very clearly document that a "*" filter will block all events including customer's not just internal SDK lifecycle events

Sources/DataPipeline/Plugins/EventPolicyPlugin.swift
```diff
@@ -0,0 +1,56 @@
...
+
+    func screen(event: ScreenEvent) -> ScreenEvent? {
+        engine.shouldAllow(eventType: "screen", name: event.name ?? "") ? event : nil
+    }
+
+    func identify(event: IdentifyEvent) -> IdentifyEvent? {
+        engine.shouldAllow(eventType: "identify", name: "*") ? event : nil
+    }
```
mahmoud-elmorabea, line 45: 🔨 I don't think this will work, have we tested how this works in practice? For `identify` specifically this point is too late to filter it. There is a pipeline of logic on our SDK side that has already operated on the fact that this identify is going through.

(ProfileStore, currentlyIdentifiedProfile, token logic) is already mutated upstream in `DataPipelineImplementation.identify`, so this will leave the SDK in an inconsistent state with the what's sent to the server

Sources/DataPipeline/Plugins/EventPolicyPlugin.swift
mahmoud-elmorabea, line 27: 🔨 To avoid this DB growing unconditionally with orphaned records. When we get new set or rules, we should clean up any filter/rate limits for any rules that no longer exists, right?

Sources/DataPipeline/DataPipelineImplementation.swift
```diff
@@ -70,6 +83,13 @@ class DataPipelineImplementation: DataPipelineInstance, DataPipelineTracking {
         // plugin that adds provider attributes (e.g. location) to identify context
         analytics.add(plugin: IdentifyContextPlugin(registry: diGraph.profileEnrichmentRegistry, logger: logger))
 
+        // Add plugin to enforce server-driven filter and rate-limit rules.
+        // Must be registered before DataPipelinePublishedEvents so that a
+        // blocked event never triggers EventBus notifications downstream.
+        if let engine = eventPolicyEngine {
+            analytics.add(plugin: EventPolicyPlugin(engine: engine, storage: storageManager))
```
mahmoud-elmorabea, line 90: 🔨 Can we make the plugin adding conditional to actually having rules to apply?

Sources/DataPipeline/Storage/StorageManager+EventPolicy.swift
```diff
@@ -0,0 +1,31 @@
...
+        try db.withConnection { conn in
+            if let existing = try conn.fetchOne(AggregationStateRecord.self, id: key),
+               now - existing.lastFlushedAt < windowSeconds {
+                return false
+            }
+
+            try conn.save(AggregationStateRecord(
+                ruleId: key,
+                stateJson: "{}",
+                lastFlushedAt: now,
```
mahmoud-elmorabea, line 25: 🔨 This is a very special case of a rate limiter. I expected the ability to allow N events in a period of time. But this seems to allow only one event if the `windowSeconds` has elapsed from the last sent event

Sources/DataPipeline/DataPipelineImplementation.swift
mahmoud-elmorabea: The majority of workspaces don't suffer volume issues. I'd expect those not to have rules and it would be great if we can make this change as much of a `no-op` for them as possible.

Sources/DataPipeline/EventPolicy/AggregationRuleset.swift
```diff
@@ -0,0 +1,49 @@
...
+    }
+}
+
+struct FilterEntry: Codable, Sendable {
+    let eventType: String
+    let name: String
+    let scope: RuleScope
```
mahmoud-elmorabea: Is filter `scope` used? Doesn't seem to be in this change

mahmoud-elmorabea (approved):
Looks good overall. 

Since this is going into a feature branch, I approved but I added a tiny 🔨 emoji in front of comments that I think we should address either in this PR or a follow up one.

Sources/DataPipeline/Plugins/EventPolicyPlugin.swift
hollyschilling-cio: The * is a dangerous operator and I generally don't like it. I added it only for "emergency" cases where we want to turn off all screen view reporting for an app. It still worries me, but while we control the configuration the risk seems minimal.

Sources/DataPipeline/Plugins/EventPolicyPlugin.swift
hollyschilling-cio: You might be misunderstanding what the code is doing here. 

The parsing happens in lines 20-21 and are loaded into the engine in line 22. Lines 24-27 are serializing and persisting the new rules to replace the old ones on disk. 

In the event that serialization (24-26) fails, we delete the old config (31). In the event that updating the storage (27) fails, we just shrug and move on because if we can't delete if we can't write. 

Or is there something else you're pointing out that I'm not understanding?

hollyschilling-cio: Filter operations do not leave any persistent record. Rate Limit operations have one row per event type. Rows are only deleted at `clearIdentify`. While it is clutter, it is minimal, doesn't grow unbounded, and there is minimal gain from removing it.

Sources/DataPipeline/Plugins/EventPolicyPlugin.swift
hollyschilling-cio: Generally speaking, I don't think the current "dumb" mechanisms filtering/rate-limiting handles in its current state is **EVER** appropriate for filtering identify events. I added the capability because it's one line, but maybe the right call is to pull it out. This is a good callout and it walks the line between future enhancement and limiting current scope.

Sources/DataPipeline/DataPipelineImplementation.swift
hollyschilling-cio: Filter rules are all handled in memory already so if no rules exist, it's effectively a no-op. Rate-limit rules hit a memory check first as well and provided that no rule exists for the event type (or wildcard), it is pretty much a no-op as well. It only goes to disk to check if it needs to check the timestamp of the last occurrence of a rate-limited event. 

All that is to say that even with it remaining in place, it's a memory-resident O(1) operation when not in use.

Sources/DataPipeline/Storage/StorageManager+EventPolicy.swift
hollyschilling-cio: Yes, this is the simplest case. As you've pointed out in the data, rate limiting is unlikely to be a large gain. A complex implementation to allow `x` events in `y` minutes adds a fair bit more complexity and record keeping for diminishing returns. Let's launch this as an MVP and if it proves more effective than anticipated, we can expand the scope of it with fine grain controls.

Sources/DataPipeline/EventPolicy/AggregationRuleset.swift
hollyschilling-cio: Filter scope is used to indicate if the rule is scoped to the profile or device. A profile scoped filter is cleared on `clearIdentify` or change of profile. Device scoped are not. That's why the value defaults to profile everywhere, meaning it will be cleared on identify change.

Sources/DataPipeline/EventPolicy/AggregationRuleset.swift
hollyschilling-cio: See comment on scope use.

Sources/Common/Storage/StorageManager.swift
hollyschilling-cio: See comment on scope use.

mrehan27 (approved):

PR #1117, opened 2026-06-23
chore(geofence): adopt OS-persisted regions, harden API decoding, and event payload

mrehan27:
## What

Fixes found during on-device geofence validation, layered on top of the `origin/main` merge into the feature branch. One squashed commit.

### Adopt OS-persisted regions on launch
iOS keeps `CLLocationManager.monitoredRegions` across process launch and device reboot. On launch the SDK now re-claims the regions it already owns instead of re-registering from a cached anchor. The old path anchored off the Location cache — which movement never updates, so it goes stale on relaunch — and could clobber a good registration, leaving **zero business geofences** monitored. Refresh now anchors off `registrationCenter ?? cache`, and the redundant foreground re-arm trigger was removed.

### Harden geofences API decoding
- Accept region `id` as a JSON number or string. The backend sends a number; the DTO required `String`, so the whole response failed to decode and no geofences loaded.
- Use the singular wire key `ios.max_business_geofence` (the DTO used the plural key, so the server cap was silently ignored).
- Treat region `last_updated` as epoch **milliseconds**.

### Event payload
- Don't persist transient events (`isPersistent == false`, e.g. `LocationAcquiredEvent`) to disk when unobserved — they stay in the in-memory cache only.
- Transition event names are `geofence_entered` / `geofence_exited`.
- `geofence_name` is included in the `/track` payload only when a name is available. Both delivery paths (direct-HTTP and EventBus → DataPipeline) build the payload in one place via the `GeofenceMetric.trackEventProperties` extension.

## Testing
Full SDK test suite green. New coverage: region adoption (relaunch re-claim + host-app isolation), numeric/string `id`, the singular config key, millis units, the transient-event persistence guard, and `geofence_name` present/absent/empty + name resolution. On-device journey validated (fetch → register → movement re-rank → enter/exit + cooldown dedup).

## Stack
Base: `feature/geofence-on-device` (already contains the `origin/main` merge). Squash-merges into the feature branch; no other open PRs in the chain.

MBL-1761

---

> [!NOTE]
> **Medium Risk**
> Changes geofence registration on relaunch, nearby API decoding, and CDP event names/properties—any mismatch with backend dashboards or partial OS region state could affect monitoring or analytics attribution.
> 
> **Overview**
> On-device geofence fixes: **relaunch monitoring**, **API decoding**, and **analytics/event-bus behavior**.
> 
> **Launch / registration:** On cold start, if the OS still monitors the **full** set from last session (business geofences + movement trigger), the SDK **adopts** those regions instead of re-registering from a stale anchor—which could wipe monitoring. Partial OS drops still trigger a fresh register from cache. Identify-time refresh now prefers **last registration center** over the Location cache; the **foreground** lifecycle refresh hook was removed.
> 
> **Geofences API:** Region `id` decodes as **number or string**; config uses **`ios.max_business_geofence`**; `last_updated` is treated as **epoch milliseconds**.
> 
> **Events & analytics:** Unobserved posts only hit disk when `isPersistent` is true (e.g. `LocationAcquiredEvent` stays in-memory). Transition tracks use **`geofence_entered` / `geofence_exited`**; optional **`geofence_name`** is resolved at capture and built once via **`GeofenceMetric.trackEventProperties`** for HTTP and DataPipeline paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f934fec2a5b0ff8ec8950dc9f3e4b9b4ced856c5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Sources/Common/Communication/CombinedCacheEventBusHandler.swift
+++ b/Sources/Common/Communication/CombinedCacheEventBusHandler.swift
@@ -88,8 +88,9 @@ public class CombinedCacheEventBusHandler: EventBusHandler {
     /// Posts `event` and returns only after all current observers have been invoked.
     ///
     /// The event is always added to the in-memory cache. If no observers exist at
-    /// post time, the event is also written to persistent storage so that a future
-    /// observer can receive it via replay.
+    /// post time, a persistent event is also written to storage so a future observer
+    /// can receive it via replay. Transient events (`isPersistent == false`) are never
+    /// written to disk — they are only useful in the moment.
     public func postEventAndWait<E: EventRepresentable>(_ event: E) async {
         logger.debug("CombinedCacheEventBusHandler: Posting event \(event)")
         let observers = await bus.post(event)
@@ -97,7 +98,7 @@ public class CombinedCacheEventBusHandler: EventBusHandler {
         for observer in observers {
             observer(event)
         }
-        if observers.isEmpty {
+        if observers.isEmpty, event.isPersistent {
             logger.debug("CombinedCacheEventBusHandler: No observers, persisting event \(event)")
             do {
                 try await eventStorage.store(event: event)
--- a/Sources/Common/Communication/Event.swift
+++ b/Sources/Common/Communication/Event.swift
@@ -163,18 +163,20 @@ public struct TrackInAppMetricEvent: EventRepresentable {
     }
 }
 
-public struct TrackGeofenceMetricEvent: EventRepresentable {
+public struct TrackGeofenceMetricEvent: EventRepresentable, GeofenceMetric {
     public let storageId: String
     public let params: [String: String]
     public let geofenceId: String
     public let transition: GeofenceTransition
     public let timestamp: Date
+    public let name: String?
 
     public init(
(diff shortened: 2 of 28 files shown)


Sources/Common/Communication/CombinedCacheEventBusHandler.swift
```diff
@@ -88,16 +88,17 @@ public class CombinedCacheEventBusHandler: EventBusHandler {
...
+    /// written to disk — they are only useful in the moment.
     public func postEventAndWait<E: EventRepresentable>(_ event: E) async {
         logger.debug("CombinedCacheEventBusHandler: Posting event \(event)")
         let observers = await bus.post(event)
         // Deliver outside the actor so callbacks run concurrently and cannot deadlock.
         for observer in observers {
             observer(event)
         }
-        if observers.isEmpty {
+        if observers.isEmpty, event.isPersistent {
```
mahmoud-elmorabea, line 101: 💯

mahmoud-elmorabea (approved):

PR #1127, opened 2026-06-26
chore(geofence): movement-trigger-first registration + fail-fast background-delivery timeout

mrehan27:
Two geofence-reliability changes under MBL-1761.

## Movement trigger registered before business regions

iOS caps region monitoring at **20 per app**, shared across all `CLLocationManager`s. `registerWithOsSync` registered business regions first and the movement trigger **last**, so when the shared budget is full (e.g. a host app that also monitors regions) the **trigger was the first casualty** — and losing it freezes the monitored set, since exiting the trigger is what re-ranks toward now-closer geofences as the device moves. Register the trigger **first** so it claims a slot before the business set can exhaust the budget. Pure reorder.

> Not addressed here: if our SDK registers its regions *before* the host app's, the host's `startMonitoring` can fail with no callback. iOS offers no arbitration; the only lever is the existing `maxBusinessGeofence` config. Captured as a config/docs follow-up.

## Fail-fast background-delivery timeout

The background-delivery HTTP client now takes a configurable request/resource timeout (default **8s**, down from 30s). It runs inside the short cold-wake window (e.g. a region-monitoring relaunch), so a slow send should fail fast and be retried on the next flush rather than be suspended mid-flight. Delivery stays **at-least-once** (`append → send → remove-on-success`): a failed or timed-out send keeps the row for the next flush — it is never dropped.

## Also

Drops the unused `location` `UIBackgroundMode` from the APN-UIKit sample. Region monitoring relaunches the app on enter/exit on its own and the SDK only requests one-shot fixes, so the mode was unused — and declaring it without a persistent-location feature risks App Review rejection. Push's `remote-notification` mode stays.

## Tests

- `refresh_givenSuccess_expectRegistrationOrderStopAllThenTriggerThenBusiness`: registration order is now `stopAll → trigger → business`.
- Existing delivery suites (success / failure-retained / anonymous / cooldown / in-process concurrent) pass unchanged.

## Notes

- A follow-up PR adds a unique id to each transition event.
- api-docs regenerated for the additive (default-valued, source-compatible) init signature.

https://linear.app/customerio/issue/MBL-1761

---

> [!NOTE]
> **Medium Risk**
> Touches geofence monitoring order and background event delivery timing; wrong ordering or timeouts could stall re-ranking or delay retries, but changes are localized and test-covered for registration order.
> 
> **Overview**
> Geofence OS registration now **starts the movement trigger before business regions** so it keeps a slot under iOS’s shared 20-region cap when the host app also monitors regions; the coordinator test expects `stopAll → trigger → business`.
> 
> **Background delivery HTTP** drops fixed 30s `URLSession` timeouts in favor of **configurable `requestTimeout` / `resourceTimeout` (default 8s)** for cold-wake sends, with updated internal API docs.
> 
> The **APN UIKit sample** removes **`location` from `UIBackgroundModes`**, leaving `remote-notification` only.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4974c124063c9288975a6f92fd38e389df100e6b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Sources/Common/Service/BackgroundDeliveryHttpClient.swift
+++ b/Sources/Common/Service/BackgroundDeliveryHttpClient.swift
@@ -58,15 +58,23 @@ public final class BackgroundDeliveryHttpClientImpl: BackgroundDeliveryHttpClien
     private let session: URLSession
     private let logger: Logger
 
+    /// - Parameters:
+    ///   - requestTimeout: Per-request timeout. Defaults to 5s to fit a cold-wake background window
+    ///     (e.g. region monitoring), where a slow send should fail fast and be retried by the caller
+    ///     rather than be suspended mid-flight. Callers with a roomier window (e.g. a Notification
+    ///     Service Extension) can pass a larger value.
+    ///   - resourceTimeout: Whole-resource timeout, including waiting for connectivity. Defaults to 5s.
     public convenience init(
         contextStore: BackgroundDeliveryContextStore,
         requestRunner: HttpRequestRunner,
-        logger: Logger
+        logger: Logger,
+        requestTimeout: TimeInterval = 5,
+        resourceTimeout: TimeInterval = 5
     ) {
         let configuration = URLSessionConfiguration.ephemeral
         configuration.allowsCellularAccess = true
-        configuration.timeoutIntervalForRequest = 30
-        configuration.timeoutIntervalForResource = 30
+        configuration.timeoutIntervalForRequest = requestTimeout
+        configuration.timeoutIntervalForResource = resourceTimeout
         self.init(
             contextStore: contextStore,
             requestRunner: requestRunner,
--- a/Sources/Common/Service/BackgroundDeliveryHttpClient.swift
+++ b/Sources/Common/Service/BackgroundDeliveryHttpClient.swift
@@ -59,17 +59,17 @@ public final class BackgroundDeliveryHttpClientImpl: BackgroundDeliveryHttpClien
     private let logger: Logger
 
     /// - Parameters:
-    ///   - requestTimeout: Per-request timeout. Defaults to 5s to fit a cold-wake background window
+    ///   - requestTimeout: Per-request timeout. Defaults to 8s to fit a cold-wake background window
     ///     (e.g. region monitoring), where a slow send should fail fast and be retried by the caller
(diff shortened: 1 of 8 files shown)


Sources/Common/Service/BackgroundDeliveryHttpClient.swift
```diff
@@ -58,15 +58,23 @@ public final class BackgroundDeliveryHttpClientImpl: BackgroundDeliveryHttpClien
...
+    ///     rather than be suspended mid-flight. Callers with a roomier window (e.g. a Notification
+    ///     Service Extension) can pass a larger value.
+    ///   - resourceTimeout: Whole-resource timeout, including waiting for connectivity. Defaults to 5s.
     public convenience init(
         contextStore: BackgroundDeliveryContextStore,
         requestRunner: HttpRequestRunner,
-        logger: Logger
+        logger: Logger,
+        requestTimeout: TimeInterval = 5,
+        resourceTimeout: TimeInterval = 5
```
mahmoud-elmorabea: Do we want to limit it to 5 seconds even if the OS gives a bit more? 5 seems too harsh to me

mahmoud-elmorabea (approved):

mrehan27:
Update: dropped the claim-send-restore (durable remove-before-send) change from this PR and reverted transition delivery back to **at-least-once** — a transition that fails to send is retried on the next flush rather than removed before sending.

Rationale: at-least-once never drops an event; the duplicate a retry can produce is the acceptable trade, and will be reconciled by a follow-up that adds a stable per-transition id. So this PR now lands just the **movement-trigger-first registration** + the **fail-fast background-delivery timeout** (now 8s); the delivery-dedup work moves to a separate PR.

Net diff is now: trigger-first registration, the configurable timeout, and dropping the unused `location` background mode from the sample app.

Sources/Common/Service/BackgroundDeliveryHttpClient.swift
mrehan27: Good point — bumped the default to 8s. With at-least-once delivery a timed-out row isn't lost, it's retried on the next flush, so the timeout only needs to stop us blocking the short background-wake window rather than complete every request. 5s was over-conservative; 8s gives slow connections room while staying inside the wake budget.

PR #1105, opened 2026-06-17
chore(geofence): remove precise location from geofence events

mrehan27:
[MBL-1831](https://linear.app/customerio/issue/MBL-1831)

## Summary
Geofence `Entered`/`Exited` events no longer carry the device's latitude/longitude. A consumer only needs the geofence id + transition type to act, so emitting the user's exact coordinates added needless privacy/legal/compliance burden. The triggering location is still read on-device to drive the movement-trigger re-sync — it is just no longer attached to the event.

Ports the Android change (`customerio-android` `b2c260f5`, PR #739).

## Changes
- Drop `latitude`/`longitude` from `TrackGeofenceMetricEvent` (public — api-docs baseline regenerated), `PendingGeofenceMetric` (incl. `CodingKeys`), and both delivery channels:
  - direct HTTP — `GeofenceDeliveryTracker`
  - EventBus → analytics — `DataPipelineImplementation+GeofenceObserver`
- Drop the now-unused `location` parameter from `GeofenceEventTracker.trackTransition` and the `GeofenceMonitorBinder` call site (the movement-trigger branch still reads the location).

Both delivery channels now emit `geofence_id + transition_type + timestamp` only.

## Scope
Coordinate removal from the **events** only. Android also bundled a top-level `timestamp` tweak ("related to MBL-1780") — handled separately in the follow-up #1106.

## Test plan
- `LocationGeofenceTests` — 153 tests / 13 suites pass.
- `DataPipelineEventBustTests` — 5/5 pass; asserts both channels exclude `latitude`/`longitude`.
- Updated tests to assert coordinates are absent; removed the now-moot null-coordinate tests.
- `make format` / `make lint` clean; api-docs baseline regenerated for the `TrackGeofenceMetricEvent` change.

## Notes
- Backward-compatible with queued data: a `PendingGeofenceMetric` persisted by an older SDK still has `latitude`/`longitude` on disk; the decoder ignores unknown keys, so those rows still decode (coordinates dropped, as intended).
- The privacy manifest update is tracked separately (not in this PR).

## Stack
Both target `feature/geofence-on-device`, merge in order:
1. #1105 — remove precise location from geofence events (this PR)
2. #1106 — attribute direct-HTTP track events to transition time

---

> [!NOTE]
> **Low Risk**
> Privacy-focused payload reduction with no auth or delivery-flow changes; movement-trigger still uses coordinates locally.
> 
> **Overview**
[...]

PR #1091, opened 2026-06-11
chore: APN-UIKit kicks SDK location update on auth grant

mrehan27:
## Summary

Surfaced during MBL-1761 validation testing. In the APN-UIKit sample, granting location permission in the in-app flow alone wasn't kicking off geofence registration — the SDK had no anchor for the initial `/geofences/nearby` fetch until a separate user action (e.g., "Set location" or "SDK Request Location Once") pushed a fix into the pipeline.

The fix: in `locationManagerDidChangeAuthorization` (and its iOS 13 sibling), call `CustomerIO.location.requestLocationUpdate()` whenever auth lands in `.authorizedWhenInUse` / `.authorizedAlways`. The same pattern customers should follow in their own apps to bootstrap geofence sync on first-run permission grant.

Mirrors the Android fix shipped in https://github.com/customerio/customerio-android/pull/728.

## Why the SDK's existing auto-fetch doesn't catch this

`LocationLifecycleObserver` already calls `requestLocationUpdate()` on `didBecomeActive`. But an in-app permission prompt is modal — granting it doesn't trigger a foreground transition. By the time the user grants, the lifecycle hook has already fired once at app start (and no-op'd because permission wasn't granted yet).

## Why this is safe to call on every delegate firing

`requestLocationUpdate()` is idempotent — `guard currentTask == nil else { return }` at [LocationServices.swift:166](https://github.com/customerio/customerio-ios/blob/feature/geofence-on-device/Sources/Location/LocationServices.swift#L166). Multiple grants (WhenInUse → Always upgrade, Settings round-trips) all flow through the same gated entry point, so no first-grant flag is needed in the customer's code.

Ticket: https://linear.app/customerio/issue/MBL-1761/ios-execute-geofencing-validation-and-lifecycle-testing

## Stack

Base: `feature/geofence-on-device` (geofence stack PRs #1081, #1082, #1083, #1084, #1087, #1090 already merged into base).

## Test plan

- [x] `make format && make lint` — 0 violations
- [x] APN-UIKit sample app build — SUCCEEDED
- [x] Manual: fresh install → identify → tap "Grant location" → grant WhenInUse → expect geofence sync to fire without further user action
- [x] Manual: continue rationale → grant Always → no spurious second fetch (SDK in-flight gating)
- [x] Manual: deny in Settings, then re-grant → fetch fires on next delegate callback

---

> [!NOTE]
> **Low Risk**
> Sample-app delegate wiring and documentation only; no SDK or production app behavior changes.
> 
[...]

PR #1090, opened 2026-06-11
chore: rename geofence wire events

mrehan27:
## Summary

Renames the CDP track event names emitted by the geofence transition path:

- `"GeoFence Entered"` → `"CIO Geofence Entered"`
- `"GeoFence Exited"` → `"CIO Geofence Exited"`

Ports the same rename Android shipped in [customerio-android@006132fc](https://github.com/customerio/customerio-android/commit/006132fc4f1171255ace2fe69fb559a8bbac830b) so the two SDKs stay in lockstep on the wire format the backend sees.

Both delivery paths (direct-HTTP via `GeofenceDeliveryTracker`, EventBus via `DataPipelineImplementation+GeofenceObserver`) read the name from `GeofenceTransition.trackEventName`, so a single source-of-truth change covers both.

Ticket: https://linear.app/customerio/issue/MBL-1780/rename-geofence-events-in-mobile-sdks

## Test plan

- [x] `make generate && make format && make lint` — 0 violations
- [x] `CI=1 ./scripts/generate-api-docs.sh` — no baseline diff (string values aren't part of public signatures)
- [x] `xcodebuild build-for-testing` — clean
- [x] Touched test classes (`GeofenceDeliveryTrackerTests`, `BackgroundDeliveryHttpClientTests`, `DataPipelineEventBusTests`) — green

---

> [!NOTE]
> **Medium Risk**
> Changing analytics event names can break dashboards, journeys, or filters that still key on the old "GeoFence" strings until backends and customers migrate.
> 
> **Overview**
> Renames the CDP **track** event names for geofence enter/exit from `"GeoFence Entered"` / `"GeoFence Exited"` to **`"CIO Geofence Entered"`** / **`"CIO Geofence Exited"`**, aligning iOS with the Android SDK wire format.
> 
> The strings come from **`GeofenceTransition.trackEventName`**, so both delivery paths (direct HTTP via **`GeofenceDeliveryTracker`** and EventBus/DataPipeline) pick up the new names without separate call-site changes. Related doc comments and tests are updated to match.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 281b9f83a158ca3e237336d26a8c05beb590fe65. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #1130, opened 2026-06-30
chore(geofence): add support for geofencing feature

mrehan27:
### Summary

This PR adds support for the upcoming geofence feature.

### Highlights

- **On-device geofence monitoring** — registers nearby geofences and tracks enter/exit transitions through Core Location region monitoring.
- **Movement-based tiered refresh** — re-ranks and re-fetches nearby geofences as the user moves (exit-triggered via a movement-trigger region), within the OS 20-region budget.
- **Reliable delivery** — transitions delivered at-least-once: persisted to a file-backed pending store before sending and retried on next launch/identify (survives process death and background cold-wake), with an EventBus → DataPipeline fallback for anonymous capture; concurrent sends arbitrated by a shared pending-delivery store, duplicate transitions suppressed by a cooldown filter, and each transition tagged with a stable id for reconciliation.
- **Lifecycle resilience** — adopts OS-persisted regions after reboot/relaunch, re-fetches on login, and snapshots `userId` per transition to prevent cross-user attribution.
- **New `CioLocationGeofence` module** with the on-device implementation split out of `Location`.

### Changes

Includes previously approved/merged PRs (oldest first):

- #1036
- #1047
- #1048
- #1060
- #1061
- #1062
- #1064
- #1065
- #1066
- #1067
- #1068
- #1069
- #1070
- #1071
- #1072
- #1074
- #1075
- #1076
- #1078
- #1081
- #1082
- #1083
- #1084
- #1087
- #1090
- #1091
- #1094
- #1095
- #1096
- #1103
- #1105
- #1106
- #1114
- #1117
- #1120
- #1121
- #1122
- #1125
- #1127
- #1129
- #1133
- #1140
- #1146
- #1148
- #1149
- #1154
- #1156
- #1158
- #1161
- #1160
- #1162
- #1168
- #1169
- #1172
- #1173
- #1177
- #1179
- #1180

---

> [!NOTE]
> **High Risk**
> Large new location/geofencing surface with cold-wake HTTP delivery, persisted API keys when opted in, and identity-sensitive transition attribution—bugs could mis-deliver events, leak credentials on disk, or attribute transitions to the wrong user.
> 
> **Overview**
> Introduces an opt-in **`CioLocationGeofence`** / **`LocationGeofence`** product (SwiftPM, CocoaPods, umbrella `CustomerIO` subspec) that monitors nearby geofences, emits **Geofence Transition** analytics, and keeps working across cold-wake and partial SDK init.
> 
> **SDK surface:** Register **`GeofenceModule`** alongside **`LocationModule`**. Public API includes **`CustomerIO.geofence.refreshFromCurrentLocation()`**, **`GeofenceModule.bootstrapForBackgroundDelivery`** for wrapper/cold-wake hosts, and **`SDKConfigBuilder.allowBackgroundDelivery(_:)`** to optionally persist the CDP API key for authenticated direct HTTP when DataPipeline is not up.
> 
> **Location integration:** Location gains silent **`requestLocationUpdateSilently()`**, **`getLastKnownLocation()`**, a **`track`** flag on sync so geofence fixes do not emit **CIO Location Update** or enrich identify, and **`LocationAcquiredEvent`** on the event bus for geofence refresh re-arming.
> 
> **Common / DataPipeline:** New background-delivery stack (**`BackgroundDeliveryContextStore`**, **`BackgroundDeliveryHttpClient`**, **`BackgroundTaskRunner`**), **`TrackGeofenceMetricEvent`**, transient-event persistence rules, and DataPipeline mirroring **userId** / **apiHost** / optional **cdpApiKey** for identify, reset, and geofence track handling.
> 
> **Geofence implementation (high level):** CDP **/geofences/nearest** fetch, distance-filtered OS registration within the 20-region budget (movement-trigger region for re-rank/refetch), sync coordinator with remote/local refresh and sign-out reset, transition tracking with cooldown, per-geoset fan-out, file-backed pending queue, direct HTTP vs EventBus routing, and bootstrap that adopts OS-persisted regions when possible.
> 
> Sample apps and CI workflows wire the new pod, deploy tier, and APN-UIKit settings/location UI for testing **Always** authorization and geofence refresh.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit eebdce56948ba0d2f833ccc141575dff9414d4fc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Sources/LocationGeofence/Storage/PendingGeofenceMetricStore.swift
```diff
@@ -0,0 +1,155 @@
...
+/// File-backed queue of geofence transition events awaiting direct-HTTP delivery.
+///
+/// Same persistence shape as `PendingPushDeliveryStore` but in the app's container
+/// (geofence callbacks run in the main process — no app group needed). File uses
+/// `completeUntilFirstUserAuthentication` Data Protection and is excluded from backups.
+///
+/// Actor-isolated: every method's load → modify → save runs without `await`, so
+/// concurrent callers can't observe a half-applied state.
+///
+/// Leftover rows are flushed on next module init. RN/Flutter wrappers may defer that
```
Shahroz16, line 13: Could we add a bounded retry when the app becomes active or connectivity returns, alongside the init and identify flushes? A long-lived RN or Flutter process can otherwise retain a failed HTTP row until the next identify or a restart. This is a follow-up, not a merge blocker, but it would make the at-least-once delivery behavior more responsive in practice.

Sources/LocationGeofence/Storage/PendingGeofenceMetricStore.swift
```diff
@@ -0,0 +1,155 @@
...
+    /// Removes one pending entry by key. Returns `true` when the entry was found and removed.
+    func remove(key: String) -> Bool {
+        var items = loadFromDisk()
+        let originalCount = items.count
+        items.removeAll { $0.key == key }
+        guard items.count != originalCount else { return false }
+        return saveToDisk(items)
+    }
+
+    /// Removes entries whose keys are in `keys` in one read-modify-write.
```
Shahroz16: Could we either remove these APIs or add their production callers in this PR? `removeAll(keys:)` and `clearAll()` are only used by their own tests. In particular, `clearAll` says it is used on sign-out, while reset currently retains queued rows for replay. Keeping both paths makes the intended delivery policy harder to follow.

Shahroz16 (approved):

Sources/Location/LocationServices.swift
```diff
@@ -42,6 +42,22 @@ public protocol LocationServices: AnyObject {
     /// The SDK does not request location permission. The host app must prompt for authorization
     /// (e.g. via `CLLocationManager.requestWhenInUseAuthorization()`) and only call this when permission is granted.
     func requestLocationUpdate()
+
+    /// Acquires a one-shot location fix for internal consumers (e.g. geofencing) with no analytics
+    /// side effects: no `CIO Location Update` track event, and the fix is **not** cached or persisted,
+    /// so it never reaches identify context enrichment and is not observable via `getLastKnownLocation`.
```
Shahroz16: The bot observation is valid: this SPI says a silent fix is not observable from getLastKnownLocation, but processLocationUpdate stores it in the in-memory last-known slot. Could we settle one contract and align docs and code? Keeping the in-memory location is reasonable for geofencing, in which case the docs should say it is not persisted or sent as analytics rather than unobservable.

Sources/LocationGeofence/Event/GeofenceEventTracker.swift
```diff
@@ -0,0 +1,191 @@
...
+        guard activeDeliveryKeys.mutating({ $0.insert(metric.key).inserted }) else { return }
+        defer { activeDeliveryKeys.mutating { _ = $0.remove(metric.key) } }
+
+        // Rows without a stamped userId (anonymous capture or legacy pre-stamping)
+        // route through EventBus instead — DataPipeline tracks anonymously under
+        // whatever identity is current at consume time.
+        guard let stampedUserId = metric.userId, !stampedUserId.isEmpty else {
+            postEventBus(metric: metric)
+            _ = await pendingStore.remove(key: metric.key)
+            return
```
Shahroz16: The bot concern makes sense here. postEvent starts a separate task, so process termination can occur between scheduling it and the EventBus accepting it. Could we use postEventAndWait before removing the row? That closes the window when an observer is present and gives persistent EventBus storage a chance to run when it is not.

Sources/LocationGeofence/Event/GeofenceEventTracker.swift
```diff
@@ -0,0 +1,191 @@
...
+        // ProfileIdentifiedEvent-triggered flush, or two flushes) can't both send the
+        // same row. The claim set is in-memory only — on app kill the row stays on
+        // disk and is retried by flushPending in the next process.
+        guard activeDeliveryKeys.mutating({ $0.insert(metric.key).inserted }) else { return }
+        defer { activeDeliveryKeys.mutating { _ = $0.remove(metric.key) } }
+
+        // Rows without a stamped userId (anonymous capture or legacy pre-stamping)
+        // route through EventBus instead — DataPipeline tracks anonymously under
+        // whatever identity is current at consume time.
+        guard let stampedUserId = metric.userId, !stampedUserId.isEmpty else {
```
Shahroz16: While tracing this through the current backend, I found a contract mismatch here. A row without userId becomes a Geofence Transition with anonymous_id. The Customer.io destination maps that to Track v2 action: geofence, but Edge deliberately rejects anonymous geofence requests with HTTP 400. Can we align which side owns this? If geofencing is identified-only, this fallback should not publish and remove the row as a valid anonymous event. If anonymous geofencing is intended, Edge and its integration tests need to support it.

Sources/LocationGeofence/Resources/PrivacyInfo.xcprivacy
```diff
@@ -0,0 +1,46 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<!-- Geofence transitions are sent via /track with the user id; the device's
+	     coordinates are never sent, so no location data type is declared. -->
```
Shahroz16: Could we recheck this manifest against the actual request path? fetchNearbyGeofences posts exact latitude and longitude to /geofences/nearest, and I traced the current backend forwarding both unchanged to geo_query, so the statement that coordinates are never sent is not accurate. If the service-request exception applies because coordinates are not retained, it would help to document that basis; otherwise Apple classifies this precision as precise location. The linked flags also seem worth revisiting because the primary transition path includes userId and stores the event against that person.

Sources/LocationGeofence/GeofenceMonitorBinder.swift
```diff
@@ -0,0 +1,33 @@
...
+    ) {
+        monitor.setOnTransition { [weak tracker, weak coordinator] identifier, transition, location in
+            // CLLocationManager delivers on main; both handlers below are async with their
+            // own serialization (tracker active-delivery dedup, coordinator refresh gate),
+            // so fire-and-forget Tasks are safe.
+            if identifier == GeofenceConstants.movementTriggerIdentifier {
+                // EXIT is the only registered transition for the movement trigger; the
+                // guard defends against an unexpected ENTER reaching this dispatch.
+                guard transition == .exit, let location else { return }
+                Task { _ = await coordinator?.handleMovement(latitude: location.latitude, longitude: location.longitude) }
```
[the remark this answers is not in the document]
mrehan27: Accepted as-is: `manager.location` is rarely nil at the moment the OS delivers a region exit (it needed a fix to detect the crossing), and a missed re-rank self-heals at the next launch/identify refresh or the 24h staleness refetch. A fallback would need a background GPS acquisition to compute a re-rank we get anyway.

Sources/LocationGeofence/GeofenceModuleState.swift
```diff
@@ -0,0 +1,147 @@
...
+                // only here — not before the async reads — avoids a false arm when an anchor exists.
+                self.lastSkippedForNoLocation.wrappedValue = true
+                self.autoAcquireIfNeeded()
+                return
+            }
+            self.lastSkippedForNoLocation.wrappedValue = false
+            _ = await di.geofenceSyncCoordinator.refresh(
+                latitude: anchor.latitude,
+                longitude: anchor.longitude
+            )
```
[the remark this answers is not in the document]
mrehan27: By design, at parity with Android: launch/identify refreshes anchor at the last registration center, and movement re-ranking is the OS movement-trigger's job (delivered even to force-quit apps). The stale-anchor case requires the OS to have dropped our regions AND relocation while the app was dead; damage is bounded by the 24h staleness refetch (server-tunable expiry), and an iOS 18+ re-registration without a current fix already self-heals (the trigger's assumed-inside seeding makes CLMonitor emit the healing exit). A targeted improvement — always seeding the movement trigger assumed-inside so a stale-center registration emits that exit immediately — is tracked as a follow-up ticket.

Sources/LocationGeofence/Storage/PendingGeofenceMetricStore.swift
mrehan27: Agreed — tracking as a follow-up ticket. Since this comment, #1161 also flushes the backlog on every new crossing, which covers the long-lived RN/Flutter process whenever the user moves; an app-active/connectivity trigger remains the follow-up.

Sources/LocationGeofence/Storage/PendingGeofenceMetricStore.swift
mrehan27: Done — `removeAll(keys:)` and `clearAll()` are removed; the store surface is now `append`/`loadAll`/`remove(key:)`. Rows deliberately survive sign-out: delivery policy is governed by per-row userId stamping and identified-only capture.

Sources/Location/LocationServices.swift
mrehan27: Settled on the contract you suggested: a silent fix updates the in-memory last-known slot (observable via `getLastKnownLocation` — geofencing anchors on it) but is not persisted, not sent as analytics, and does not enrich identify. The `requestLocationUpdateSilently` doc now states exactly that.

Sources/LocationGeofence/Event/GeofenceEventTracker.swift
mrehan27: Done in #1156 — the EventBus path uses `postEventAndWait` and the row is drained only after the handoff resolves.

Sources/LocationGeofence/Event/GeofenceEventTracker.swift
mrehan27: Aligned with the backend in #1156: geofencing is identified-only. Anonymous transitions are dropped at capture, before cooldown or persistence, so the EventBus fallback never publishes an anonymous Geofence Transition.

Sources/LocationGeofence/Resources/PrivacyInfo.xcprivacy
mrehan27: Valid — `fetchNearbyGeofences` does send exact coordinates, so the manifest statement is not accurate as written. Correcting it (declaration and basis, with Legal input) is tracked as a separate ticket to land before the feature ships publicly.

Sources/LocationGeofence/Event/GeofenceEventTracker.swift
```diff
@@ -0,0 +1,309 @@
...
+    /// exists yet). That is not a durable-persistence ack — the strongest signal EventBus exposes
+    /// today — so a crash before DataPipeline persists re-delivers on the next flush (deduped by
+    /// transitionId).
+    private func deliverViaEventBus(metric: PendingGeofenceMetric) async {
+        guard activeDeliveryKeys.mutating({ $0.insert(metric.key).inserted }) else { return }
+        defer { activeDeliveryKeys.mutating { _ = $0.remove(metric.key) } }
+
+        let effective = await resolvingLiveValues(metric)
+        await postEventBus(metric: effective)
+        _ = await pendingStore.remove(key: metric.key)
```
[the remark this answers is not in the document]
mrehan27: Real but narrow: it takes no live observer AND EventBus disk persistence failing in the same window, and the fix needs a Common API change (`postEventAndWait` surfacing persistence success). Tracked as a follow-up ticket rather than expanding Common's API pre-release.

Sources/LocationGeofence/GeofenceModuleState.swift
```diff
@@ -0,0 +1,147 @@
...
+        }
+        di.eventBusHandler.addObserver(ResetEvent.self) { [weak self] _ in
+            // Drop any pending refresh intent so a previous user's request can't drive a sync for
+            // the next user — matches the coordinator's user-scoped reset. Cleared synchronously
+            // here so it lands before a re-login's `ProfileIdentifiedEvent` re-arms.
+            self?.explicitRefreshRequested.wrappedValue = false
+            self?.lastSkippedForNoLocation.wrappedValue = false
+            Task { @MainActor in
+                _ = await di.geofenceSyncCoordinator.reset()
+            }
```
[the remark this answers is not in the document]
mrehan27: Already handled at the other end of the gate: every gated operation compensates at its single exit while still holding the gate — `cleanupIfUserChanged` (GeofenceSyncCoordinator+Refresh.swift) re-reads the user and, on a mid-operation change, runs the same teardown `reset()` would (`stopMonitoringAll` + `clearUserScopedState`), with `retryForCurrentUser` converging a sign-in. The one holder without the hook, `applyCachedRegistration`, is synchronous on the main actor (no network) behind the bootstrap's identity re-check, so the uncovered window is a few milliseconds — accepted.

Sources/Location/LocationServices.swift
```diff
@@ -162,13 +198,13 @@ actor LocationServicesImplementation: LocationServices {
         await locationSyncCoordinator.processLocationUpdate(locationData)
     }
 
-    private func startRequestIfNeeded() async {
+    private func startRequestIfNeeded(track: Bool, respectMode: Bool) async {
         guard currentTask == nil else { return }
```
Shahroz16: Could we preserve the tracked intent when these two request paths overlap? On a first identified session, the `.onAppStart` lifecycle request (`track: true`) and the automatic geofence bootstrap (`track: false`) can race into this guard. If the silent request wins, this drops the tracked request, and the lifecycle gate will not retry it again in this process. The silent result only updates last-known location, so it does not emit `CIO Location Update` or populate the cached location used by the identify retry. This is a normal startup/onboarding path, not just a theoretical interleaving. Could a tracked request upgrade the in-flight silent request, or set a pending `wantsTracking` flag for the result?

Sources/LocationGeofence/Resources/PrivacyInfo.xcprivacy
Shahroz16: Could we update this explanation to match the actual flow?. Or leaving location out of this manifest looks correct.?

Shahroz16 (approved):

Sources/LocationGeofence/Resources/PrivacyInfo.xcprivacy
mrehan27: Addressed in #1168 (open): it declares PreciseLocation (Linked=false — the nearest request carries no user identity) and rewrites this comment to match the actual flow; the missing UserDefaults required-reason entry is added there too. Leaving location out would only be correct under Apple's ephemeral real-time-servicing exemption, and unredacted server logs on the nearest endpoint currently rule that out — that basis, plus the future slim-down path if backend redacts the logs, is recorded in the PR.

Sources/Location/LocationServices.swift
mrehan27: Fixed in #1173 (merged into this branch) — implemented as the pending-`wantsTracking` flag you suggested, named `pendingTrackUpgrade` on the `LocationServicesImplementation` actor.

Shape: a tracked request arriving while another is in flight latches the intent instead of vanishing; the in-flight request consumes it at result time and posts a **tracked** result, so one GPS fix serves both consumers. The `.off` gate is re-applied at consumption so the upgrade cannot smuggle a track the public request would itself have refused, and a request that ends without a fix drops the latch — matching what the tracked request would have done on its own.

A follow-up commit also clears the latch when the owning task clears `currentTask`, one actor turn later; without it a tracked request arriving in that gap latched an intent with no request left to consume it, and the next acquisition (often a geofence fix much later) would have posted it as a tracked update for an unrelated place and time.

Documented residual: a tracked request landing during the single actor hop into `LocationSyncCoordinator` while the result is being posted is still dropped. Same as the previous behaviour, with the window down from a multi-second GPS acquisition to one hop, and the launch race this targets lands during the provider call, which the latch does cover.

Four tests pin the interleavings (upgrade delivers one tracked result from one provider call; reverse order unchanged; `.off` refuses the upgrade; a failed request drops the latch).

Sources/Location/LocationServices.swift
```diff
@@ -162,31 +202,54 @@ actor LocationServicesImplementation: LocationServices {
...
+            }
         }
         if let result = await locationProvider.requestLocationOnce() {
             switch result {
             case .success(let snapshot):
-                await postLocation(snapshot)
+                // The upgrade must not emit a track the public request itself would have
+                // refused under `.off` (silent requests run regardless of mode).
+                let upgraded = pendingTrackUpgrade && config.mode != .off
+                await postLocation(snapshot, track: track || upgraded)
```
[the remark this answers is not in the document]
mrehan27: Duplicate of the residual already assessed on #1173 ([thread](https://github.com/customerio/customerio-ios/pull/1173#discussion_r3650011355)) — it surfaces here only because that PR is now merged into this branch. Accepted deliberately, not an open item.

Correct on mechanism: an intent arriving during the actor hop into `LocationSyncCoordinator` is cleared unconsumed. Two things bound it. It is **not a regression** — that intent is dropped on the base branch too, and the PR shrinks the window from a multi-second GPS acquisition to a single hop. And the race actually targeted, the `.onAppStart` request against geofence's silent acquire at launch, lands during the provider call, which the latch does cover.

Every way to close the last hop costs more than the hop: posting twice emits a duplicate `LocationAcquiredEvent`, which geofencing observes as a refresh re-arm; letting the late intent start its own request adds a GPS acquisition on the launch path; a track-only coordinator entry point adds API surface for a microsecond window.

Documented as a known residual in the #1173 description.

Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator+Refresh.swift
```diff
@@ -0,0 +1,174 @@
...
+        parsedConfig: GeofenceConfig?,
+        effectiveConfig: GeofenceConfig,
+        anchor: LocationData
+    ) async -> Bool {
+        guard regionsUnchanged(regions, await storage.getCachedGeofences()),
+              await canSkipReregistration(
+                  nearestIds: nearestIds,
+                  previouslyRegisteredIds: storage.getRegisteredBusinessIds(),
+                  registerMovementTrigger: effectiveConfig.maxBusinessGeofences > 0
+              )
```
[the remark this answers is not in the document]
mrehan27: Not reproducible — the premise does not hold. A single `await` covers every async call in the enclosing expression, so the outer `await` on `canSkipReregistration(...)` also covers the nested `storage.getRegisteredBusinessIds()`:

```swift
guard regionsUnchanged(regions, await storage.getCachedGeofences()),
      await canSkipReregistration(
          nearestIds: nearestIds,
          previouslyRegisteredIds: storage.getRegisteredBusinessIds(),
          registerMovementTrigger: effectiveConfig.maxBusinessGeofences > 0
      )
```

Verified by compiling the same shape (`await canSkip(previously: store.ids())`, actor-isolated `ids()`, no inner `await`) — builds and runs. Decisively: a genuinely missing `await` on an async call is a compile error, and this target builds.

Actor isolation is therefore intact and the stored ids are read under it, so the fast path cannot act on a stale set.

PR #1133, opened 2026-07-01
chore(geofence): sync nearby geofences with device location

mrehan27:
Adds a location-bound geofence sync mode alongside fetch-all, and makes it the default.

## What changed
- **`GeofenceSyncMode.fetchNearby`** (new, now the shipped `active` mode): sends the device location to `/geofences/nearby` so the backend returns the nearby set, and re-fetches once the device moves beyond `remoteFetchRefreshTriggerRadius` from the last fetch anchor.
- **`fetchAll`** is unchanged and still sends no location. Both modes are fully implemented; `active` is a compile-time choice (never a runtime/server toggle).
- The nearby request carries **no user identifier** (only the workspace API key), so the coordinate can't be attributed to a person — the device location is sent as-is, without coarsening.
- The movement/refresh decision re-fetches (rather than only re-ranking) once the device leaves the area the nearby set was fetched for; `fetchAll` still only re-ranks.

## Tests
- API service: the device coordinate is sent on the wire as-is for nearby; fetch-all sends no query params.
- Coordinator: nearby re-fetches beyond the radius / re-ranks within it / bootstraps with no anchor via nearby; fetch-all behavior is unchanged.

## Notes
- Parallel to Android MBL-1945 — same wire contract (device location sent as-is; the request carries no user identity).

https://linear.app/customerio/issue/MBL-1945

---

> [!NOTE]
> **High Risk**
> Default SDK behavior now transmits precise device location to Customer.io on geofence sync (privacy-sensitive, compile-time only but a breaking posture change for integrators expecting no off-device location).
> 
> **Overview**
> Introduces **`GeofenceSyncMode.fetchNearby`** and sets it as the compile-time **`active`** mode (replacing **`fetchAll`** as the default shipped behavior). **`fetchAll`** is unchanged: no location on the wire and movement only re-ranks the full cached workspace.
> 
> **`fetchNearby`** adds **`GeofenceApiService.fetchNearbyGeofences`**, which GETs `/geofences/nearby` with **`latitude`** / **`longitude`** query params (exact coordinates; auth is workspace API key only, no user id). Remote refresh goes through **`awaitApiFetch(latitude:longitude:)`**, which switches on **`syncMode`**.
> 
> Refresh and movement logic now **re-fetch from the server** when the device moves beyond **`remoteFetchRefreshTriggerRadius`** from the last fetch anchor (**`movedBeyondRefetchRadius`** in **`refreshAction`**; same check in **`handleMovement`**). Within that radius, behavior stays local re-rank. Docs/constants are updated so **`remoteFetchRefreshTriggerRadius`** is described as used by **`fetchNearby`**, not reserved for a future mode.
> 
> Tests cover nearby query shaping and coordinator paths (bootstrap, beyond/within refetch radius, time-fresh + far move); **`GeofenceApiServiceMock`** gains **`fetchNearbyGeofences`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b66ccf71522db8a1339fcbcd8c2bf113ebe7a425. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Tests/LocationGeofence/Geofence/Api/CoordinateCoarsenerTests.swift
```diff
@@ -0,0 +1,23 @@
...
+        #expect(coarse.longitude == -122.417457)
+    }
+
+    @Test
+    func coarsen_givenHighLatitude_expectLongitudeGridWidensToStayUniform() {
+        // At 60° (cos ≈ 0.5) a degree of longitude is ~half the ground distance, so the longitude
+        // grid roughly doubles in degrees to hold the ~500 m floor — where fixed-degree rounding
+        // would over-refine longitude instead.
+        let coarse = CoordinateCoarsener.coarsen(latitude: 60.0, longitude: 10.123456)
+        #expect(abs(coarse.longitude - 10.123456) < 0.0090) // ~the ~0.009° cell at this latitude
```
mahmoud-elmorabea: Same comment as android regarding more preceise bounds for the test expectation

mahmoud-elmorabea (approved):

Tests/LocationGeofence/Geofence/Api/CoordinateCoarsenerTests.swift
mrehan27: Same as Android, we decided not to use coarse location since the API doesn't send any user specific information. So we reverted the coarse location for now as this is no longer needed.

PR #1139, opened 2026-07-01
fix: stabilize flaky tests that keep breaking CI on main

jcpsimmons:
## Problem

The **Tests** workflow on `main` has been failing intermittently for weeks (latest: [run 28548337547](https://github.com/customerio/customerio-ios/actions/runs/28548337405), [run 28243876792](https://github.com/customerio/customerio-ios/actions/runs/28243876765)). Each red run fails on a *different* single test, so this is a cluster of race conditions rather than one bad commit:

| Run date | Failing test | Root cause |
|---|---|---|
| 07-01 | `CombinedCacheEventBusHandlerTest.test_removeObserver_...` | `addObserver`/`removeObserver` each spawn an independent `Task`, so the remove can be applied to the bus actor *before* the add, leaving the observer registered |
| 06-26 | `SynchronizedEquatableTests.equalsConcurrentlyDoesNotDeadlock` | 2000 blocks dispatched to `DispatchQueue.global()` cause GCD thread explosion on loaded runners; the 30s `group.wait` times out even though there is no deadlock |
| 06-05 | `LocationServicesImplementationTests.setLastKnownLocation_givenMultipleCalls_...` | back-to-back `setLastKnownLocation` calls spawn independent tasks; when the second wins the race, the filter tracks NYC instead of SF |
| 06-04, 05-21, 04-22 | `GistQueueNetworkTest` (two tests) | real `URLSession.shared` round-trip with no assertions in the handler; the wait exceeds its timeout on CI |
| 05-08 | `MessagingPushPendingPushFlushTests.test_initialize_whenNoPendingMetrics_...` | the detached flush task from the *previous* test outlives it, resolves `DataPipelineTracking` at flush time, and tracks into the next test's mock |

## Fixes

- **`CombinedCacheEventBusHandler`**: operations from synchronous entry points are now chained FIFO (each enqueued `Task` awaits the previous one), so `addObserver` / `removeObserver` / `postEvent` are applied to the bus actor in call order. `postEventAndWait` joins the chain, which also let the test drop its sleeps. This fixes a real production-visible ordering race, not just the test.
- **`SynchronizedTests`**: the deadlock test now uses exactly two threads looping in opposite lock order (stronger ABBA pressure, no thread explosion) with a generous timeout.
- **`MessagingPush`**: the pending-metrics flush task is stored in an internal `@Atomic` static so tests can await its completion instead of racing wall-clock expectations; the flush can no longer leak across tests.
- **`LocationServicesImplementationTests`**: the multiple-calls test waits for the first location to be tracked before posting the second.
- **`GistQueueNetworkTest`**: dropped the two remaining real-network waits, same rationale as the existing flake-fix comments in that file.

## Verification

No Xcode available locally, so the two `CioInternalCommon` fixes were verified with a standalone stress harness against the real implementation:

- 2,000 iterations of add-remove-post: removed observer never received an event (previously flaked)
- 2,000 iterations of concurrent post + register: exactly-once delivery every time
- 200 x 2,000 opposite-order `==` comparisons: no deadlock, completes in milliseconds

The simulator-dependent suites (Location, MessagingInApp, MessagingPush) are covered by this PR's CI run. Suggest re-running the Tests workflow a few times on this branch before merging to confirm stability.

---

> [!NOTE]
> **Medium Risk**
> The event bus FIFO change alters observer registration and delivery ordering for all SDK event subscribers; incorrect chaining could affect profile/identity and other cached events, though the change targets a known race.
> 
> **Overview**
> Addresses intermittent **Tests** workflow failures by fixing real ordering races and making async tests deterministic instead of timing-dependent.
> 
> **`CombinedCacheEventBusHandler`** now serializes `addObserver`, `removeObserver`, and `postEvent` through a FIFO `enqueue` chain (each operation awaits the previous `Task`), so synchronous call order is preserved on the actor. **`postEventAndWait`** joins that chain and shares **`postAndDeliver`** with async posts. This fixes a production case where immediate add-then-remove could leave an observer registered.
> 
> **`MessagingPush`** records the detached pending-metrics flush in **`pendingMetricsFlushTask`** so tests can **`await`** completion and prevent flush work from leaking into the next test’s DI mocks.
> 
> Test-only hardening: event-bus remove test drops sleeps; **`Synchronized`** deadlock test uses two tight-loop threads (avoids GCD thread explosion); location multi-call test waits for the first track before the second; **Gist** queue tests stop waiting on real **URLSession** round-trips; push flush tests use async **`await`** on the flush task instead of wall-clock expectations.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 514818d8c443b36bd4259690250da43e50d80664. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Sources/Common/Communication/CombinedCacheEventBusHandler.swift
+++ b/Sources/Common/Communication/CombinedCacheEventBusHandler.swift
@@ -12,12 +12,34 @@ public class CombinedCacheEventBusHandler: EventBusHandler {
     let eventStorage: EventStorage
     let logger: Logger
 
+    /// Tail of the FIFO operation chain. Every bus operation started from a synchronous
+    /// entry point (`addObserver`, `removeObserver`, `postEvent`) is appended here so that
+    /// operations reach the actor in call order. Without this, each entry point spawning
+    /// an independent `Task` allowed e.g. `addObserver` followed by `removeObserver` to be
+    /// applied in reverse, leaving the observer registered.
+    private let lastOperation = Synchronized<Task<Void, Never>?>(nil)
+
     public init(eventStorage: EventStorage, logger: Logger) {
         self.bus = CioEventBus()
         self.eventStorage = eventStorage
         self.logger = logger
-        Task { [weak self] in
+        enqueue { [weak self] in
             await self?.loadEventsFromStorage()
+        }
+    }
+
+    /// Appends `operation` to the FIFO chain. Each operation awaits the previous one,
+    /// guaranteeing call-order execution while callers remain synchronous.
+    @discardableResult
+    private func enqueue(_ operation: @escaping () async -> Void) -> Task<Void, Never> {
+        lastOperation.mutating { last in
+            let previous = last
+            let next = Task {
+                await previous?.value
+                await operation()
+            }
+            last = next
+            return next
         }
     }
 
@@ -46,7 +68,7 @@ public class CombinedCacheEventBusHandler: EventBusHandler {
         _ eventType: E.Type, action: @escaping (E) -> Void
     ) {
         logger.debug("CombinedCacheEventBusHandler: Adding observer for \(eventType)")
-        Task { [weak self] in
+        enqueue { [weak self] in
             guard let self else { return }
             let registration = await bus.addObserver(key: E.key) { [weak self] event in
                 guard self != nil else { return }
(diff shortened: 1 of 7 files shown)


jcpsimmons:
Closing. Logged as MBL-2202. main Tests have been green since Jul 1 so this is less urgent than when I opened it, but the underlying flakiness was never actually fixed.

PR #1141, opened 2026-07-03
feat: Visual Notification Inbox overlay (integration / build distribution)

mahmoud-elmorabea:
Draft integration PR for the Visual Notification Inbox overlay. Opened primarily to publish sample-app builds to Firebase App Distribution so another team can test the feature end-to-end.

Aggregates the already-reviewed sub-PRs: module skeleton, data layer, overlay UI, interaction/web-parity, message actions + branding colors, dismiss-resurrection guard, and APN-UIKit sample demo.

Firebase App Distribution builds (APN-UIKit + CocoaPods-FCM) are posted as a comment on this PR once CI completes.

---

> [!NOTE]
> **Medium Risk**
> New customer-facing messaging UI and concurrent HTTP/SSE inbox sync paths (known overwrite race documented); moderate scope but touches real-time delivery and profile/session state.
> 
> **Overview**
> Adds **`CioMessagingInbox`** as a distributable SwiftPM/CocoaPods module (Jist-backed SwiftUI overlay) and wires it through deploy workflows, the umbrella `CustomerIO` pod, and sample apps.
> 
> The **visual inbox data layer** lives in MessagingInApp: queue polls now honor **`x-cio-inbox-enabled`**, handle **204** empty queues, fetch **templates/branding** (once per session, serve-stale), and drive **`VisualInboxLoadState`** visibility. A **`@_spi(VisualInbox)`** provider exposes snapshots to the overlay without expanding the public headless API. **SSE lifecycle** changes move HTTP backfill to the server **`connected`** event (and on user-id changes) to reduce races with live inbox updates.
> 
> **Headless inbox** gains **`InboxEventListener`**, per-session dedupe for shown/opened with profile-change resets, and richer click tracking (`actionValue`). Sample apps mount **`NotificationInboxOverlay`** / bell views, bundle DM Sans/Abril fonts for Jist, and APN-UIKit bumps minimum iOS to **15.0**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b85ecd5044202179194f70d0444fabba51bf5608. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Sources/MessagingInApp/Inbox/Data/VisualInboxSPI.swift
```diff
@@ -0,0 +1,346 @@
...
+    /// task's child tasks; serialized by the `lastSnapshotLock` since those run concurrently.
+    private func emitSnapshot(into continuation: AsyncStream<VisualInboxSnapshot>.Continuation) async {
+        let snapshot = await currentSnapshot()
+        lastSnapshotLock.lock()
+        let changed = lastEmittedSnapshot == nil || lastEmittedSnapshot != snapshot
+        if changed { lastEmittedSnapshot = snapshot }
+        lastSnapshotLock.unlock()
+        guard changed else { return }
+        continuation.yield(snapshot)
+    }
```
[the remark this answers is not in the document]
mahmoud-elmorabea: Valid, low severity. There are two independent subscribers to `inboxMessages`: `VisualInboxSPI.observe()` emits a snapshot on each change, while `VisualInboxRepository` recomputes `loadState` in a separate `Task`. So `observe()` can emit a snapshot with the new message list but a not-yet-recomputed `loadState` (e.g. `.visible` + empty list right after dismissing the last message), then correct it on the next emission once `loadStateChanges()` fires. It self-corrects within one cycle, so worst case is a brief flicker — but the window is real. A tighter fix would derive `loadState` and the message list from a single coalesced source so they can't disagree. Tracking as a follow-up.

Sources/MessagingInApp/Inbox/DefaultNotificationInbox.swift
```diff
@@ -15,6 +15,22 @@ class DefaultNotificationInbox: NotificationInbox, @unchecked Sendable {
...
+    /// Message ids already reported to the host via `inboxMessageShown`, so "shown" fires at most once
+    /// per message per app session even though the UI may call `notifyMessageShown` on every render.
+    /// Guarded by `shownMessageIdsLock` because `notifyMessageShown` can be called off the main thread.
+    private var shownMessageIds: Set<String> = []
+    private let shownMessageIdsLock = NSLock()
+
+    /// Ids already reported "opened" so the host `inboxMessageOpened` callback fires at most once per
+    /// message per session (mirrors `shownMessageIds`). Guarded by `openedMessageIdsLock`.
+    private var openedMessageIds: Set<String> = []
+    private let openedMessageIdsLock = NSLock()
```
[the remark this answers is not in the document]
mahmoud-elmorabea: Low risk / borderline. Accurate that `shownMessageIds`/`openedMessageIds` are process-scoped and not cleared on `ResetEvent`/`resetState()`. But the dedupe is keyed by `queueId`, and queueIds are treated as globally unique / never reused (the Android `snapshot()` note documents the same invariant). So after logout + a new identify, the new profile's messages carry new queueIds → `inboxMessageShown`/`inboxMessageOpened` still fire; the 'at most once per app session' behavior is also intentional. This only bites if the exact same queueIds are re-delivered after reset, which the queueId-uniqueness invariant rules out — so no missed callbacks in practice. Fine to leave; we could clear these sets on reset defensively if we ever stop guaranteeing queueId uniqueness.

Sources/MessagingInbox/NotificationInboxView.swift
```diff
@@ -0,0 +1,243 @@
...
+    /// it). nil for standalone embedding, where the host owns presentation.
+    private let onNavigate: (() -> Void)?
+
+    /// Performs the SDK default navigation for an un-intercepted action. Injected (constructor DI)
+    /// rather than reaching into `DIGraphShared.shared` inline, so navigation is testable.
+    private let navigator: InboxActionNavigating
+
+    /// Creates a standalone inbox list backed by the SDK's shared Visual Inbox data layer.
+    public init() {
+        _model = ObservedObject(wrappedValue: VisualInboxModel())
```
Shahroz16: Could we preserve the internally created model across SwiftUI reconstruction? `@ObservedObject` does not own it, so a parent update can replace the model after the original lifecycle has started. Since `@StateObject` is iOS 14+, the iOS 13 path will need an equivalent owner. The same applies to `NotificationInboxBell`.

Sources/MessagingInApp/Gist/Gist.swift
```diff
@@ -126,6 +126,11 @@ class Gist: GistProvider {
             // SSE is now active - stop polling
             logger.logWithModuleTag("Gist: SSE enabled for identified user - stopping polling timer", level: .info)
             invalidateTimer()
+            // SSE does not deliver pending messages on connect. Backfill any messages that already
+            // exist for the user (mirrors SseLifecycleManager's foreground fetch); otherwise a fresh
+            // login shows an empty inbox until the next foreground. Calls fetchUserQueue directly
+            // because fetchUserMessages() no-ops when shouldUseSse is true.
+            fetchUserQueue(state: state)
```
Shahroz16: Could we make this backfill finish before SSE starts, or version the two full-state updates? `fetchUserQueue` starts the HTTP request in the background and returns immediately. If SSE publishes a newer inbox snapshot first, the older HTTP response can replace it afterward.

Sources/MessagingInApp/Inbox/Data/VisualInboxSelector.swift
```diff
@@ -0,0 +1,54 @@
...
+
+    /// Selects, filters, and sorts messages for the visual inbox.
+    /// - Parameters:
+    ///   - messages: All inbox messages from state.
+    ///   - now: Reference time used for expiry evaluation (injectable for tests).
+    /// - Returns: visual-inbox messages, expired entries removed, sorted priority asc → sentAt desc.
+    static func select(messages: [InboxMessage], now: Date = Date()) -> [InboxMessage] {
+        messages
+            .filter { hasVisualInboxTopic($0) }
+            .filter { !isExpired($0, now: now) }
```
Shahroz16, line 24: Could we schedule a refresh for the next message expiry? Expiry is checked only when `select()` runs, and the UI only recomputes on store or load-state changes. With a quiet SSE connection, an expired row and its badge count can remain visible until an unrelated update.

Sources/MessagingInApp/Inbox/NotificationInbox.swift
```diff
@@ -64,6 +64,35 @@ public protocol NotificationInbox: Sendable {
...
+    ///
+    /// - Parameters:
+    ///   - message: The inbox message that was clicked
+    ///   - actionName: Optional name of the action clicked (e.g., "view_details", "dismiss")
+    ///   - actionValue: Optional value of the action clicked (e.g., the deep link / URL)
+    func trackMessageClicked(message: InboxMessage, actionName: String?, actionValue: String?)
+
+    /// Registers (or clears, with `nil`) the host listener notified of inbox message actions.
+    /// Used by `MessagingInApp.setInboxEventListener(_:)`; not intended for direct host use.
+    func setInboxEventListener(_ listener: InboxEventListener?)
```
Shahroz16, line 82: Could we keep these listener plumbing methods off the public `NotificationInbox` requirement list, or provide defaults? They are mandatory for existing external conformers and test doubles, so this otherwise additive feature becomes source-breaking. The concrete `MessagingInApp.setInboxEventListener` API can remain public while the SDK-only forwarding lives on an internal protocol; its new requirement on `MessagingInAppInstance` needs the same compatibility treatment.

.github/workflows/deploy-sdk.yml
```diff
@@ -222,10 +222,12 @@ jobs:
           # Pods within a tier do not depend on each other, so they push in parallel.
           # Tiers push in order because `pod trunk push` lints each podspec and the
           # dependency versions pushed by the previous tier must already exist.
+          # CustomerIOMessagingInbox depends on CustomerIOMessagingInApp (tier 2), so
+          # it pushes in tier 3.
           tiers=(
             "CustomerIOCommon.podspec"
             "CustomerIOTrackingMigration.podspec CustomerIOMessagingPush.podspec CustomerIOMessagingInApp.podspec CustomerIOLocation.podspec"
-            "CustomerIODataPipelines.podspec CustomerIOMessagingPushAPN.podspec CustomerIOMessagingPushFCM.podspec CustomerIOLocationGeofence.podspec"
+            "CustomerIODataPipelines.podspec CustomerIOMessagingPushAPN.podspec CustomerIOMessagingPushFCM.podspec CustomerIOMessagingInbox.podspec CustomerIOLocationGeofence.podspec"
```
Shahroz16, line 230: Could we also add `CustomerIOMessagingInbox` to the CDN-propagation `pods` list below? The release smoke test immediately rewrites the sample app to this published pod, so `pod install` can run before the new pod is visible on the CDN.

Package.swift
```diff
@@ -19,6 +19,7 @@ var products: [PackageDescription.Product] = [
     .library(name: "MessagingPushAPN", targets: ["CioMessagingPushAPN"]),
     .library(name: "MessagingPushFCM", targets: ["CioMessagingPushFCM"]),
     .library(name: "MessagingInApp", targets: ["CioMessagingInApp"]),
+    .library(name: "MessagingInbox", targets: ["CioMessagingInbox"]),
```
Shahroz16, line 22: Could we add `MessagingInbox:CioMessagingInbox` to `scripts/generate-api-docs.sh` as part of this product addition? The current API compatibility job does not inspect the new public SwiftUI surface.

Sources/MessagingInbox/VisualInboxModel.swift
```diff
@@ -0,0 +1,326 @@
...
+        chromeLoadTask = Task { @MainActor [weak self] in
+            guard let self else { return }
+            let resolved = await self.provider.brandingChrome()
+            if Task.isCancelled { return }
+            if let resolved {
+                self.chrome = resolved
+                // Build the bell glyph once here (not per render). nil svg / unparseable → nil → bundled bell.
+                self.bellGlyph = resolved.bellIconSvg.flatMap(InboxBellGlyph.build(fromSVG:))
+            }
+        }
```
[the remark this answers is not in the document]
mahmoud-elmorabea: Not acting on this one — it is unreachable in practice.

The code shape is real: `reloadChrome()` only assigns inside `if let resolved`, so `chrome` / `bellGlyph` do survive a nil branding load. But nothing renders them in that state. Bell and panel visibility is driven by the published `state` (`.hidden` renders nothing), not by `chrome`, and `computeLoadState` requires non-nil branding to resolve to a visible state. So whenever branding is unavailable the overlay draws nothing regardless of what `chrome` still holds.

The only way in is a branding payload that resolves once and then returns *without* `patterns.inbox`, leaving the inbox visible while `brandingChrome()` goes nil — at which point the consequence is a stale colour set, not the cross-workspace leak the finding describes.

Closing. If it ever shows up in practice the fix is a two-line `else` that nils both.

Sources/MessagingInApp/Inbox/Data/Cache/InboxRenderAssetsCache.swift
```diff
@@ -0,0 +1,46 @@
...
+///
+/// There is **no wall-clock TTL/expiry**: freshness is decided by once-per-session server
+/// revalidation in `VisualInboxRepository`, matching Android's model. This store's sole job is to
+/// hold the **last-known payload** so it can be served on a subsequent (same-session) read without a
+/// network call, and served stale when a revalidation fails.
+struct InboxRenderAssetsCache {
+    private let keyValueStore: SharedKeyValueStorage
+
+    init(keyValueStore: SharedKeyValueStorage) {
+        self.keyValueStore = keyValueStore
```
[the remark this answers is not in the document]
mahmoud-elmorabea: Addressed as far as this stack goes — closing with a pointer.

The false claim is gone: #1183 (merged) corrects the comment to state that `InboxRenderAssetsCache` resolves to the app-wide `SharedKeyValueStorage` suite rather than per-`siteId` storage, and documents what logout does and does not cover.

The actual re-scoping is deliberately not done here. It changes the storage key layout and needs a migration for already-persisted assets, which is outside the overlay work. The residual exposure is narrow: logout clears the cache, so reaching it takes re-initialising the SDK with a different site id inside a single process.

Closing this thread rather than keeping it as the tracker, since it disappears when `feat/overlay-inbox` merges — the re-scoping needs its own issue to outlive that.

mrehan27 (approved):

Sources/MessagingInApp/Gist/Utilities/SseLifecycleManager.swift
```diff
@@ -104,6 +110,30 @@ actor CioSseLifecycleManager: SseLifecycleManager {
...
+    /// `connected` event rather than the transition that decides to connect, so the fetch starts
+    /// against a stream that is already live and is skipped when a connection never establishes.
+    ///
+    /// This narrows the overwrite window but does not close it: the fetch is fire-and-forget, so an
+    /// `inbox_messages` event arriving while it is in flight can still be overwritten by the older
+    /// HTTP snapshot, since both publish a full-state write. Ordering them properly needs the two
+    /// writes versioned or merged; tracked separately.
+    ///
+    /// Not the only backfill trigger: `Gist.handleUserIdentificationChange` also fetches, because a
+    /// `userId` change while SSE is already connected produces no `connected` event to hook.
```
Shahroz16, line 127: Could we reconnect SSE and clear the previous inbox state for a direct A → B user change? This comment correctly notes that there is no new `connected` event, but the existing stream remains authenticated as A because `startConnection()` is a no-op while connected. The SDK has already switched its state to B, so a later event from A can replace B’s inbox. Since this can expose one user’s inbox to another, I think it should be fixed before. A focused A → B test would help cover the fix.

Sources/MessagingInApp/Gist/Utilities/SseLifecycleManager.swift
```diff
@@ -104,6 +110,30 @@ actor CioSseLifecycleManager: SseLifecycleManager {
...
+    ///
+    /// SSE only delivers events that arrive AFTER the connection is established, so a fresh login
+    /// would otherwise show an empty inbox until the next foreground. Keyed off the server's
+    /// `connected` event rather than the transition that decides to connect, so the fetch starts
+    /// against a stream that is already live and is skipped when a connection never establishes.
+    ///
+    /// This narrows the overwrite window but does not close it: the fetch is fire-and-forget, so an
+    /// `inbox_messages` event arriving while it is in flight can still be overwritten by the older
+    /// HTTP snapshot, since both publish a full-state write. Ordering them properly needs the two
+    /// writes versioned or merged; tracked separately.
```
Shahroz16, line 124: Could we track a follow-up to prevent the HTTP backfill from overwriting newer inbox state received through SSE? If an SSE update arrives while the backfill request is still running, the older HTTP response can replace it, and the inbox may stay stale until another SSE event or reconnect. I don’t think this needs to block this PR, but linking the follow-up here would make the remaining limitation clear.

Sources/MessagingInApp/Gist/Utilities/SseLifecycleManager.swift
mahmoud-elmorabea: I think this is the whole of in-app, not just inbox. Maybe if we want to formally support that login while a user already logged in path we could. I suggest handling that as a follow up as well

Sources/MessagingInApp/Gist/Utilities/SseLifecycleManager.swift
mahmoud-elmorabea: The window is too narrow and it exceptionally unlikely that these two responses have different messages. I suggest we take as a follow up to sequence the two calls or have some kind of logic to merge them

PR #1146, opened 2026-07-07
chore(geofence): emit one transition event per (fence, geoset), fence as metadata

jcpsimmons:
Starting point for MBL-2008, for Rehan to take over when he's online. Ticket deliberately left unclaimed. Targets `feature/geofence-on-device`.

## What this does

The SDK learns geoset membership from the nearby response and stamps it onto transition events, so every event stands alone and campaign/segment matching needs no joins at ingest:

- `GeofenceApiRegion` decodes optional `geoset_ids` per fence (the endpoint's snake_case wire convention via `convertFromSnakeCase`); absent or empty means no membership, so backend can roll the field out gradually
- Cached `Geofence` carries `geosetIds`; geofences cached by pre-geoset SDK versions still decode (missing key defaults to empty)
- `trackTransition` fans one physical transition out to one `PendingGeofenceMetric` per geoset. Each event carries a scalar `geosetId` plus `geofenceId` / `geofenceName` metadata; a fence in sets Y and Z produces two standalone events, a fence in no geoset produces today's single fence-only event
- The cooldown/dedup gate still runs once per fence, before fan-out, so all rows of one transition are suppressed or emitted together
- The pending-store composite key gains a `geosetId` suffix so fan-out rows of one transition persist and retry independently (rows written by older versions keep their key shape)
- Both delivery paths carry `geosetId`: direct HTTP (`GeofenceMetric.trackEventProperties`) and EventBus (`TrackGeofenceMetricEvent`, new optional field, defaulted so persisted events replay)
- Event name unchanged (`Geofence Transition`), so ingest, journey-start gating, and analytics are untouched

Scalar `geosetId` rather than an array: campaign triggers and segment conditions match on scalar attribute equality (proposal section 5.4).

## Testing

New coverage in `GeofenceEventTrackerTests` (fan-out per geoset on both delivery paths, no-geoset fallback, failure retention of all fan-out rows, cooldown suppressing the whole transition), `GeofenceApiResponseTests` (wire decode present/absent), `GeofenceStorageTests` and `PendingGeofenceMetricStoreTests` (round-trips plus legacy decode compatibility for pre-geoset cached data). Existing tests compile unchanged (new params are defaulted).

Note: verified by CI only; local `swift build` on this machine can't target iOS.

## Open questions for Rehan

- Each fanned-out event gets its own `transitionId` (each is a standalone event per the ticket). If downstream dedup should instead treat the fan-out as one physical transition, they could share one id; flagging since #1129 introduced the field
- The pending-store key doc says it matches Android's `PendingGeofenceDelivery.key`; Android will need the same geoset suffix when it picks up its half
- `geosetId` (camelCase) follows the payload convention from #1121; confirm that's the agreed event property name

Linear: https://linear.app/customerio/issue/MBL-2008/sdk-emit-one-transition-event-per-fence-geoset-fence-as-metadata

---

> [!NOTE]
> **Medium Risk**
> Changes geofence analytics volume (N events per crossing), alters the nearby-fetch API contract and radius semantics, and touches persist-first delivery edge cases; backward-compatible decoding limits upgrade risk.
> 
> **Overview**
> Adds **geoset-aware geofence transitions**: the API response can include `geoset_ids` (decoded as strings, including int64 wire values), cached `Geofence` models carry `geosetIds`, and **`trackTransition` fans one OS crossing into one metric/event per distinct geoset** (or a single event with no `geosetId` when the fence has none). Cooldown still applies once per fence before fan-out; fan-out rows share one **`transitionId`** and get distinct pending keys via a **`geosetId` suffix**. **`geosetId`** is added to **`GeofenceMetric` / track properties** and **`TrackGeofenceMetricEvent`**, with optional decode for older persisted events.
> 
> **Delivery and durability** changes: the pending queue **`append` accepts batches** so all fan-out rows persist atomically; delivery runs **concurrently** per row; if persist fails after claiming cooldown, **`releaseCooldown`** runs and send is skipped.
> 
> **Sync/API**: nearby fetch moves from **GET `/geofences/nearby` with query params** to **POST `/geofences/nearest`** with a JSON body (`latitude`, `longitude`, **`radius`**). **`GeofenceConfig.searchRadius`** supplies that radius from cached config (max of refetch distance and monitoring cap). Fetch-all stays **GET `/geofences/nearby`**.
> 
> Tests and mocks are updated for the new API signature, geoset decode/fan-out, persist-failure cooldown behavior, and legacy on-disk compatibility.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d8e8a0eeb6bda2f3d9064fd54cfb6d02acce57d7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Sources/Common/Type/GeofenceMetric.swift
+++ b/Sources/Common/Type/GeofenceMetric.swift
@@ -12,6 +12,10 @@ public protocol GeofenceMetric {
     var name: String? { get }
     /// Uniquely identifies this transition; stable across delivery retries.
     var transitionId: String { get }
+    /// The geoset this event was emitted for, or `nil` when the geofence is in
+    /// no geoset. A geofence in N geosets produces N metrics per transition,
+    /// each carrying one geoset ID, so every event stands alone.
+    var geosetId: String? { get }
 }
 
 public extension GeofenceMetric {
@@ -19,7 +23,7 @@ public extension GeofenceMetric {
     /// `transition` property.
     var trackEventName: String { "Geofence Transition" }
 
-    /// `/track` event properties. `geofenceName` is included only when a name is available.
+    /// `/track` event properties. `geofenceName` and `geosetId` are included only when available.
     /// `transitionId` uniquely identifies the transition and stays stable across delivery retries.
     /// `timestamp` is not a property — it is set on the event envelope by each path.
     var trackEventProperties: [String: Any] {
@@ -30,6 +34,9 @@ public extension GeofenceMetric {
         ]
         if let name {
             properties["geofenceName"] = name
+        }
+        if let geosetId {
+            properties["geosetId"] = geosetId
         }
         return properties
     }
--- a/Sources/Common/Type/GeofenceMetric.swift
+++ b/Sources/Common/Type/GeofenceMetric.swift
@@ -36,8 +36,7 @@ public extension GeofenceMetric {
             properties["geofenceName"] = name
         }
         if let geosetId {
-            // Numeric ids go out as a number (type-preserving) so scalar campaign matching is consistent.
-            properties["geosetId"] = Int64(geosetId) ?? geosetId
+            properties["geosetId"] = geosetId
         }
         return properties
     }
--- a/Sources/Common/Type/GeofenceMetric.swift
+++ b/Sources/Common/Type/GeofenceMetric.swift
@@ -36,7 +36,8 @@ public extension GeofenceMetric {
             properties["geofenceName"] = name
         }
         if let geosetId {
-            properties["geosetId"] = geosetId
+            // Numeric ids go out as a number (type-preserving) so scalar campaign matching is consistent.
+            properties["geosetId"] = Int64(geosetId) ?? geosetId
         }
         return properties
     }
--- a/Sources/LocationGeofence/Event/GeofenceEventTracker.swift
+++ b/Sources/LocationGeofence/Event/GeofenceEventTracker.swift
@@ -83,12 +83,15 @@ final class GeofenceEventTracker: @unchecked Sendable {
         let cachedGeofence = await storage.getCachedGeofences().first { $0.id == geofenceId }
         let cachedName = cachedGeofence?.name
         let geofenceName = (cachedName?.isEmpty == false) ? cachedName : nil
-        // One standalone event per geoset, each carrying a scalar geosetId plus the
-        // geofence id/name as metadata, so segment and campaign matching needs no joins
-        // at ingest. No geoset membership (or the geofence has left the cache) emits a
-        // single event without a geosetId, preserving pre-geoset behavior.
-        let memberGeosetIds = cachedGeofence?.geosetIds ?? []
+        // One event per geoset (scalar geosetId + geofence id/name as metadata) so matching needs no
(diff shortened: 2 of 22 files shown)


mrehan27:
Picked this up — all three open questions are resolved in the commit on top:

**1. transitionId (own vs shared):** Went with **shared** — one `transitionId` across a transition's geoset fan-out, with `geosetId` differentiating the rows. Matches the "one physical crossing = one transition" intent (and Android).

**2. Pending-store key / Android parity:** Android already adds the same `geosetId` suffix to `PendingGeofenceDelivery.key`, so the two stay in sync — no cross-team gap.

**3. `geosetId` name/type:** Kept `geosetId` (camelCase, consistent with `geofenceId` / `transitionId`), and made it **type-preserving** — emitted as a number when numeric, string otherwise — to match Android so campaign scalar matching is consistent across platforms.

Also on top of the fan-out: `POST /geofences/nearest` with `radius` in the body (`max(reFetchRadius, monitoringCap)`, matching Android), atomic batch-persist plus concurrent fan-out delivery so N geosets don't serialize in the OS background wake window, and a duplicate-geoset guard. Removing `fetchAll` is a separate follow-up (both platforms).

Sources/Common/Type/GeofenceMetric.swift
```diff
@@ -31,6 +35,9 @@ public extension GeofenceMetric {
         if let name {
             properties["geofenceName"] = name
         }
```
Shahroz16, line 37: Could we add a cross-SDK contract test that exercises numeric IDs and pins the emitted representation? CDP accepts strings and numbers today, so this is not blocking. It would help keep iOS and Android payloads intentionally aligned as the ingestion contract evolves.

Sources/LocationGeofence/Event/GeofenceEventTracker.swift
```diff
@@ -74,21 +77,45 @@ final class GeofenceEventTracker: @unchecked Sendable {
...
+                transition: transition,
+                timestamp: now,
+                userId: stampedUserId,
+                name: geofenceName,
+                transitionId: transitionId,
+                geosetId: geosetId
+            )
+        }
+        // Persist all rows in one atomic write: a per-row loop could save some and lose the rest on
+        // an app kill, and the cooldown is already spent so the lost ones would never retry.
```
Shahroz16, line 108: Could we handle the `false` result from `append` here? The cooldown has already been recorded. If queue persistence fails and the HTTP send then fails, the metric is neither queued for `flushPending()` nor eligible for a prompt retry. A small recovery path, or at least a diagnostic log, would make the persist-before-send guarantee easier to trust.

Shahroz16 (approved):

Sources/Common/Type/GeofenceMetric.swift
mrehan27: Added a new test

Sources/LocationGeofence/Event/GeofenceEventTracker.swift
mrehan27: Updated to remove cooldown if write fails

PR #1149, opened 2026-07-09
chore(geofence): hold a background-task assertion across transition delivery

mrehan27:
When the OS wakes the app briefly for a geofence transition, delivery ran as an async task with no background-execution assertion — so the app could be suspended mid-send, leaving delivery to the next `flushPending`. This wraps the fan-out delivery (and the `flushPending` replay) in a `UIApplication` background-task assertion so the parallel sends can finish in the same wake.

## What changed
- New `Common/Service/BackgroundTaskRunner.swift`:
  - `BackgroundTaskRunner` protocol (`withBackgroundTime`).
  - `NoBackgroundTaskRunner` — runs inline, identical to today's behavior; extension-safe; the constructor default.
  - `UIKitBackgroundTaskRunner(name:)` — real assertion, `@available(iOSApplicationExtension, unavailable)` (mirrors the existing `UIKitWrapper` pattern so Common still compiles for the NSE). `name` is required so each subsystem labels its task in Instruments/traces.
- `GeofenceEventTracker` wraps both `trackTransition` delivery and `flushPending` replay. **Persist happens before the assertion** — durability never depends on it.
- `BackgroundDeliveryHttpClient` request/resource timeouts raised 8s→15s/20s — the assertion now guarantees the runtime to afford them, kept under the assertion floor to leave headroom for flush/cleanup. `flushPending`'s sequential replay is the one backlog caveat; delivery is persist-first + at-least-once, so a timeout only defers, never loses.

## Why it's safe
Best-effort timeliness aid, not a correctness dependency:
- Rows are persisted to disk before delivery; `deliver()` drains a row only on success.
- Anything unfinished replays via `flushPending` (module init / `ProfileIdentifiedEvent` / next wake), deduped by `transitionId` (at-least-once).
- `endBackgroundTask` is idempotent (guards `.invalid`) — the completion path and the OS expiration handler can't double-end. If the OS denies the grant, work runs unprotected = today's behavior.

## Tests
- Spy runner verifies delivery runs strictly inside the assertion for both `trackTransition` and `flushPending`, and that an empty queue requests no background time.
- `NoBackgroundTaskRunner` (the default) is exercised by every existing tracker test.
- api-docs baseline regenerated for the new public Common symbols.

## Stack
Independent of the fetchAll removal (touches different files), based on the fan-out it extends:
1. **This PR** → `mbl-2008-geoset-fanout`
2. #1148 (`chore/geofence-remove-fetchall`) → `mbl-2008-geoset-fanout` — sibling
3. #1146 (`mbl-2008-geoset-fanout`) → `feature/geofence-on-device`
4. #1130 (`feature/geofence-on-device`) → `main`

Merge order: #1146 first, then this and #1148 (either order), then #1130.

---

> [!NOTE]
> **Medium Risk**
> Touches cold-wake geofence delivery and shared background HTTP defaults; correctness still depends on persist-first + retry, but longer timeouts and OS background limits can change timing of first-attempt sends.
> 
> **Overview**
> Geofence transition delivery and pending replay now run inside a **UIApplication background-task assertion** so brief OS wakes are less likely to suspend the app mid-HTTP send. A new **`BackgroundTaskRunner`** abstraction in Common provides **`UIKitBackgroundTaskRunner`** (main app, labeled `io.customer.geofence.delivery`) and **`NoBackgroundTaskRunner`** for extensions and tests.
> 
> **`GeofenceEventTracker`** still **persists metrics before** requesting background time; it wraps concurrent fan-out delivery in `trackTransition` and sequential replay in **`flushPending`** (skipping the assertion when the queue is empty). **`BackgroundDeliveryHttpClient`** default timeouts increase from **8s to 15s / 20s** to align with the longer assertion window. Tests use a spy runner; internal API docs list the new public symbols.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e66266564eeabeaaeaee5e361fa1668d7a8a253d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Sources/Common/Service/BackgroundTaskRunner.swift
+++ b/Sources/Common/Service/BackgroundTaskRunner.swift
@@ -1,0 +1,69 @@
+import Foundation
+
+/// Requests extra background execution time so a short async unit of work can finish after the OS
+/// wakes the app briefly, instead of the app being suspended mid-flight. The assertion is released
+/// when the work returns or the OS expires it, whichever comes first.
+///
+/// Best-effort timeliness aid only — callers must not depend on it for correctness. No-op where
+/// `UIApplication` is unavailable (non-UIKit platforms, app extensions).
+public protocol BackgroundTaskRunner: Sendable {
+    func withBackgroundTime(_ work: @Sendable () async -> Void) async
+}
+
+/// No-op runner: runs the work inline, identical to not requesting background time. Default where
+/// `UIApplication` is unavailable, and in tests.
+public struct NoBackgroundTaskRunner: BackgroundTaskRunner {
+    public init() {}
+
+    public func withBackgroundTime(_ work: @Sendable () async -> Void) async {
+        await work()
+    }
+}
+
+#if canImport(UIKit)
+import UIKit
+
+/// Wraps `work` in a `UIApplication` background-task assertion. Unavailable in app extensions —
+/// `UIApplication.shared` is a main-app-only API.
+@available(iOSApplicationExtension, unavailable)
+public struct UIKitBackgroundTaskRunner: BackgroundTaskRunner {
+    /// Debug label for the assertion, shown in the debugger/Instruments. Each subsystem passes its
+    /// own (e.g. geofence vs push delivery) so tasks are distinguishable in traces.
+    private let name: String
+
+    public init(name: String) {
+        self.name = name
+    }
+
+    public func withBackgroundTime(_ work: @Sendable () async -> Void) async {
+        let assertion = BackgroundTaskAssertion()
+        await assertion.begin(name: name)
+        await work()
+        await assertion.end()
+    }
+}
+
+/// Holds the assertion id on the main actor so the normal completion path and the OS expiration
+/// handler both end it exactly once.
+@available(iOSApplicationExtension, unavailable)
+@MainActor
+private final class BackgroundTaskAssertion {
+    private var id: UIBackgroundTaskIdentifier = .invalid
+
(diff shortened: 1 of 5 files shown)


Sources/Common/Service/BackgroundTaskRunner.swift
```diff
@@ -0,0 +1,69 @@
...
+    private var id: UIBackgroundTaskIdentifier = .invalid
+
+    nonisolated init() {}
+
+    func begin(name: String) {
+        id = UIApplication.shared.beginBackgroundTask(withName: name) { [weak self] in
+            // OS is out of patience — end synchronously (this runs on the main thread) so the app
+            // isn't killed by the watchdog. In-flight work is abandoned; the caller owns any retry.
+            self?.end()
+        }
```
Shahroz16, line 61: Could we adjust this wording? Ending an assertion does not itself cancel work, so “In-flight work is abandoned” is stronger than the implementation promises. Saying the work may be suspended and persisted rows are retried would make the guarantee precise.

Shahroz16 (approved):
I found one documentation mismatch. The expiration handler ends the assertion but does not cancel the underlying work. The inline comment asks for wording that reflects the actual suspension and retry behavior.

Sources/Common/Service/BackgroundTaskRunner.swift
mrehan27: Updated

PR #1154, opened 2026-07-10
chore(geofence): enrich transition events with geofence metadata

mrehan27:
Enriches the "Geofence Transition" event with workspace-defined `metadata` so campaigns/segments can match on fence attributes (brand, category, promo, etc.) without a join at ingest. Values arrive on the nearby response and are stamped onto the event.

## What this does
- **`GeofenceMetadataValue`** (Common): a `Codable`/`Sendable` scalar enum — `string` / `int` (Int64) / `double` / `bool` — so a JSON primitive keeps its type all the way through decode → cache → persisted row → event payload (a number matches numerically, a bool matches as a bool).
- **Decode** (`GeofenceApiRegion`): tolerant at the API boundary — null/array/object values are dropped rather than failing the whole region; scalars keep their type. Disk decode stays strict (our own encoder).
- **Freshness — prefer-live-at-send**: each pending row snapshots the fence's `name` + `metadata` at crossing time; at send, `deliver` re-resolves both from the current cache and falls back to the snapshot only if the fence has left the cache. Name and metadata move together so they never mix points in time.
- **Always emitted**: `metadata` is a nested object on every event, empty `{}` when the fence carries none (never omitted/null). Flows through both delivery paths (direct-HTTP and EventBus).
- **Defensive caps** (`GeofenceConstants`): count (100) + total key+value byte budget (100 KB), applied sorted-by-key for determinism. The server fully validates metadata; these only stop a runaway payload from bloating a background request.

## Testing
New `GeofenceMetadataValueTests` (typed decode/encode, number-beyond-Int64 → double fallback). Added metadata coverage to `GeofenceApiResponseTests` (type-preservation, missing/empty → `[:]`, null/non-scalar dropped, count + byte caps), `GeofenceEventTrackerTests` (snapshot, prefer-live, evicted-fence fallback, EventBus carries metadata), `GeofenceDeliveryTrackerTests` (nested object, empty present), `PendingGeofenceMetricStoreTests` (round-trip + legacy rows without metadata decode to nil). api-docs baseline regenerated for the new public Common symbols.

## Android parity
Matches Android (`d3344112`): typed/type-preserved values, nested always-present `metadata`, prefer-live-at-send (`withFreshestEventData`), tolerant sanitize, identical caps (100 / 100 KB). iOS re-registers wholesale (no diff), so Android's `equalsForRegistration` has no iOS counterpart. (Android bundled the request-radius removal into this commit; on iOS that lives in #1148 — request-path cleanup — so this PR is metadata-only.)

## Stack
1. **This PR** (`mbl-2045-geofence-metadata`) → `chore/geofence-delivery-background-task`
2. #1149 (`chore/geofence-delivery-background-task`) → `mbl-2008-geoset-fanout`
3. #1148 (`chore/geofence-remove-fetchall`) → `mbl-2008-geoset-fanout` — sibling
4. #1146 (`mbl-2008-geoset-fanout`) → `feature/geofence-on-device`
5. #1130 (`feature/geofence-on-device`) → `main`

Merge order: #1146, then #1148 and #1149 (either order), then this, then #1130.

Linear: https://linear.app/customerio/issue/MBL-2045

---

> [!NOTE]
> **Medium Risk**
> Changes geofence transition event shape and delivery paths (HTTP + EventBus) with backward-compatible persistence, but new always-present metadata and prefer-live resolution could affect downstream analytics expectations.
> 
> **Overview**
> **Geofence Transition** `/track` events now carry workspace **`metadata`** from the nearby API so ingest can match on fence attributes without joins.
> 
> Adds public **`GeofenceMetadataValue`** (string / int / double / bool) so primitives keep their JSON types through cache, pending rows, HTTP, and EventBus. **`GeofenceMetric.trackEventProperties`** always includes a nested **`metadata`** object (`{}` when none); empty **`geofenceName`** is omitted like legacy `""` names.
> 
> Metadata is decoded on **`GeofenceApiRegion`** with tolerant handling (bad types / non-scalars dropped, region still parses) and **count + byte caps** before it lands on **`Geofence`**. At delivery, **`GeofenceEventTracker`** snapshots metadata at crossing and **re-reads name + metadata from cache at send**, falling back to the persisted row if the fence was evicted. **`Geofence.name`** is now optional (`nil` when absent) instead of defaulting to `""`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e230acca2fa7c60644dd10ba57d51b3fa9c01982. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Sources/LocationGeofence/Event/GeofenceEventTracker.swift
+++ b/Sources/LocationGeofence/Event/GeofenceEventTracker.swift
@@ -104,7 +104,9 @@ final class GeofenceEventTracker: @unchecked Sendable {
                 userId: stampedUserId,
                 name: geofenceName,
                 transitionId: transitionId,
-                geosetId: geosetId
+                geosetId: geosetId,
+                // Snapshot as the fallback for an evicted geofence; `deliver` prefers the live cache.
+                metadata: cachedGeofence?.metadata
             )
         }
         // Persist all rows in one atomic write: a per-row loop could save some and lose the rest on
@@ -157,17 +159,19 @@ final class GeofenceEventTracker: @unchecked Sendable {
         guard activeDeliveryKeys.mutating({ $0.insert(metric.key).inserted }) else { return }
         defer { activeDeliveryKeys.mutating { _ = $0.remove(metric.key) } }
 
+        let effective = await resolvingLiveValues(metric)
+
         // Rows without a stamped userId (anonymous capture or legacy pre-stamping)
         // route through EventBus instead — DataPipeline tracks anonymously under
         // whatever identity is current at consume time.
-        guard let stampedUserId = metric.userId, !stampedUserId.isEmpty else {
-            postEventBus(metric: metric)
+        guard let stampedUserId = effective.userId, !stampedUserId.isEmpty else {
+            postEventBus(metric: effective)
             _ = await pendingStore.remove(key: metric.key)
             return
         }
 
         let success = await withCheckedContinuation { (continuation: CheckedContinuation<Bool, Never>) in
-            deliveryTracker.trackMetric(metric: metric, userId: stampedUserId) { result in
+            deliveryTracker.trackMetric(metric: effective, userId: stampedUserId) { result in
                 switch result {
                 case .success:
                     continuation.resume(returning: true)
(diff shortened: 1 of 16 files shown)


Shahroz16 (approved):
I did not find an actionable issue. I verified that metadata remains available after decoding, caching, persistence, and either delivery path.

Sources/LocationGeofence/Event/GeofenceEventTracker.swift
```diff
@@ -170,11 +174,23 @@ final class GeofenceEventTracker: @unchecked Sendable {
...
+    /// Prefers the freshest cached `name`/`metadata` at send, falling back to the row snapshot when
+    /// the geofence has left the cache. All other fields — including the dedup key inputs — are
+    /// unchanged, so the returned copy addresses the same persisted row.
+    private func resolvingLiveValues(_ metric: PendingGeofenceMetric) async -> PendingGeofenceMetric {
+        guard let live = await storage.getCachedGeofences().first(where: { $0.id == metric.geofenceId }) else {
+            return metric
+        }
+        // Keep the snapshot name when the cached one is empty — don't regress a name we once had to blank.
+        let liveName = live.name.isEmpty ? metric.name : live.name
+        return metric.withResolved(name: liveName, metadata: live.metadata)
```
Shahroz16: The bot raises a useful contract question. The decoder makes a missing name and an intentional empty name indistinguishable, while this fallback retains the old snapshot. Could we document that empty means missing, or preserve the distinction in decoding? That would make a server-side name clear behave predictably.

mrehan27: Updated

PR #1155, opened 2026-07-13
chore(inbox): native-sheet overlay UI (bell glyph + branding position + web-parity actions)

mahmoud-elmorabea:
Reworks the drop-in `NotificationInboxOverlay` to present the inbox in a native system sheet (medium/large detents, iOS 16+) instead of a custom panel, and brings the overlay to behavioral parity with the web SDK:

- **Presentation:** bell → system `.sheet` with `.presentationDetents([.medium, .large])` + grabber; removes the custom panel/scrim/animation. iOS 16+.
- **Chrome from branding:** bell glyph bundled from `patterns.inbox.floatingIcon.svg` (vector, tinted); bell anchor from `patterns.inbox.position`; unread badge honors `unreadIndicator.showAlert`.
- **Reactive lifecycle fix:** the data subscription no longer stops when the bell hides, so the bell reappears when messages arrive to an empty inbox.
- **Message actions:** corrected to the real (web) schema — `openUrl`/`openDeeplink`/`performAction`/`dismiss` reading `action`/`name`; deep links open via `UIApplication.open` and dismiss the sheet.
- **Sample:** overlay hosted on a SwiftUI screen; UIKit path uses the lower-level `NotificationInboxBell` + `NotificationInboxView`.

Unit tests updated (`MessagingInboxTests`); device-verified on the APN-UIKit sample.

---

> [!NOTE]
> **Medium Risk**
> Touches public overlay behavior (iOS 16 floor, removed panel callback), deep-link/URL navigation paths, and SPI surface changes; large vendored SVG dependency but isolated to bell rendering.
> 
> **Overview**
> **`NotificationInboxOverlay`** now opens the message list in a native SwiftUI **sheet** (medium/large detents, drag indicator) instead of a custom slide-out panel, scrim, and `onPanelPresentationChange` callback. Minimum availability moves to **iOS 16+**; a stable `Color.clear` anchor keeps the data subscription alive when the bell hides so new messages can bring the bell back.
> 
> **Branding** flows through the SPI: `floatingIcon.svg`, `position`, and `unreadIndicator.showAlert` are parsed and mapped into `VisualInboxChrome`. The bell renders workspace SVG (vendored **SVGPath** + `InboxBellGlyph`) or a new **`CioInboxAssets`** fallback, with **`Bundle.cioInboxResources`** for SPM and CocoaPods.
> 
> **Message actions** align with the web schema (`action` / `name`, `openUrl`, `openDeeplink`, `performAction`); default navigation uses injectable **`InboxActionNavigating`** (`DeepLinkUtil` for deep links). The sheet dismisses before navigation when a valid URL is involved. **`NotificationInboxView`** shows nothing when hidden, uses a UIKit spinner on iOS 13, and supports `onNavigate`.
> 
> **Samples:** dashboard drops the UIKit passthrough overlay; inbox demo pushes a SwiftUI screen with `NotificationInboxOverlay` in a `ZStack`. Lint/format exclude vendored `Vendor` code.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 144e99208a70e1d0c7146297aba2ed50292b007e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Sources/MessagingInbox/NotificationInboxBell.swift
```diff
@@ -85,6 +86,22 @@ public struct NotificationInboxBell: View {
...
+
+    /// The bell glyph: the workspace's branding SVG (`floatingIcon.svg`) when present and parseable,
+    /// otherwise the bundled default bell asset. Both are tinted with the branding glyph color.
+    @ViewBuilder
+    private func bellGlyph(colors: ResolvedInboxColors) -> some View {
+        if let svg = model.chrome?.bellIconSvg, InboxBellIcon.canRender(svg) {
+            // Even-odd fill matches CIO bell SVGs' `fill-rule="evenodd"` (outlined bell).
+            InboxBellIcon(svg: svg).fill(colors.bellIcon, style: FillStyle(eoFill: true))
+        } else {
+            Image("cio-inbox-bell", bundle: .module)
```
matt-frizzell: 👀 Non-blocking, validate.

**CocoaPods build break:** `Bundle.module` is synthesized only by SwiftPM — CocoaPods does not generate a resource-bundle accessor, and there's no shim in `Sources/` — so this line won't compile when the pod is built via `CustomerIOMessagingInbox.podspec` (`type 'Bundle' has no member 'module'`). This is the first `Bundle.module` usage in the SDK, so nothing else has hit it yet.

Even with an accessor, the podspec ships **two** resource bundles (`CioMessagingInbox_Privacy`, `CioMessagingInbox_InboxAssets`), so the asset lookup needs to target the right one explicitly. Suggest a small resolver:

```swift
enum InboxAssets {
    static var bundle: Bundle {
        #if SWIFT_PACKAGE
        return .module
        #else
        // CocoaPods: assets are compiled into the CioMessagingInbox_InboxAssets resource bundle.
        let container = Bundle(for: BundleFinder.self)
        return container.url(forResource: "CioMessagingInbox_InboxAssets", withExtension: "bundle")
            .flatMap(Bundle.init(url:)) ?? container
        #endif
    }
    private final class BundleFinder {}
}
```

…and `Image("cio-inbox-bell", bundle: InboxAssets.bundle)` here. Worth adding a `pod lib lint` (or CocoaPods-FCM sample build) to the checklist for this branch since SPM builds won't catch this class of issue.

Sources/MessagingInbox/InboxBellIcon.swift
```diff
@@ -0,0 +1,84 @@
...
+/// markup and fits the combined path into the target rect (centered, aspect-preserving).
+///
+/// Fill with `FillStyle(eoFill: true)`: CIO bell glyphs are authored with the even-odd rule (the bell
+/// body is an outlined ring), matching the source SVG's `fill-rule="evenodd"`.
+@available(iOS 13.0, *)
+struct InboxBellIcon: Shape {
+    /// Raw SVG markup, e.g. `<svg viewBox="0 0 24 25"><path d="…"/>…</svg>`.
+    let svg: String
+
+    func path(in rect: CGRect) -> Path {
```
matt-frizzell: 👀 Non-blocking, validate.

**Perf: the SVG is fully re-parsed on every render pass.** `Shape.path(in:)` runs on every layout, and the call site (`NotificationInboxBell.bellGlyph`, line 94) also calls `canRender(svg)` — another full `parse` — on every body evaluation. Body re-evaluates on each model publish, and `observe()` emissions + `reloadChrome()` make that frequent, so we're running regex extraction + the SVGPath character parser per frame for a static glyph.

The parse is deterministic per SVG string, so caching makes it one-time:

```swift
private static let cache = NSCache<NSString, ParsedSVG>() // ParsedSVG: class wrapping (Path, CGRect)

private static func parse(_ svg: String) -> (path: Path, viewBox: CGRect)? {
    if let hit = cache.object(forKey: svg as NSString) { return (hit.path, hit.viewBox) }
    // ...existing parse...
    cache.setObject(ParsedSVG(path: combined, viewBox: box), forKey: svg as NSString)
    return (combined, box)
}
```

This also makes the `canRender` + `path(in:)` double-parse free.

Sources/MessagingInbox/NotificationInboxView.swift
```diff
@@ -147,6 +162,16 @@ public struct NotificationInboxView: View {
...
             }
+            // Close the inbox after a DEEP-LINK action so the opened in-app screen isn't left sitting
+            // behind the inbox sheet. Only `openDeeplink` dismisses: it routes within/into the app, so
+            // the sheet would otherwise cover the destination. `openUrl` opens externally (browser) and
+            // leaves nothing behind the sheet; `performAction`/`dismiss` keep the inbox open. Fires
+            // whether the SDK opened it or the host handled it; skipped for a missing message or an
+            // empty destination.
+            let isDeeplinkNavigation = resolution.behavior == .openDeeplink
+            if isDeeplinkNavigation, outcome != .messageMissing, resolution.actionValue?.isEmpty == false {
+                await MainActor.run { onNavigate?() }
```
matt-frizzell: 👀 Non-blocking, validate.

Related to Bugbot's "openUrl in-app leaves sheet" comment below, but distinct: that one is about dismissal not happening for openUrl; this is about dismissal happening after navigation for openDeeplink.

**Ordering: the deep link is handled *before* the sheet starts dismissing.** `performDefaultNavigation` (line 158) enqueues `deepLinkUtil.handleDeepLink(url)` via `DispatchQueue.main.async` (line 199) before this `MainActor.run { onNavigate?() }` is enqueued — so the host's `deepLinkCallback` runs while the inbox sheet is still presented. If the host routes by *pushing* on a nav stack underneath, that's fine (and matches the intent here). But if the host routes by **presenting modally**, it will present on top of the sheet — and the sheet dismissal that follows tears the host's screen down with it.

Since the stated goal is "don't leave the opened screen behind the inbox," dismissing first is strictly safer:

- flip `onNavigate?()` before `performDefaultNavigation(resolution)`, or
- run the navigation from the sheet's `onDismiss` in `NotificationInboxOverlay`.

If the current order is intentional (e.g., verified fine on device with push-based routing), a comment documenting that hosts must not present modally from `deepLinkCallback` would help.

Sources/MessagingInbox/Vendor/SVGPath/SVGPath.swift
```diff
@@ -0,0 +1,699 @@
+//
+// Vendored from nicklockwood/SVGPath v1.3.0 (https://github.com/nicklockwood/SVGPath) — MIT License.
+// Copied into the SDK (not a package dependency) so it ships via both SwiftPM and CocoaPods with no
+// external dependency. Access levels were lowered from `public` to internal to keep these types off
+// the SDK's public API surface; otherwise unmodified. See LICENSE.md in this directory.
```
matt-frizzell, line 5: 👀 Non-blocking, validate.

**Missing file:** all three vendored file headers (and the `.swiftlint.yml` exclusion comment) say "See LICENSE.md in this directory," but `Vendor/SVGPath/` contains only the three `.swift` files — there's no `LICENSE.md` on this branch. The MIT text embedded in each header keeps us legally covered, but the reference is dangling; please add the upstream `LICENSE.md` from https://github.com/nicklockwood/SVGPath.

Sources/MessagingInApp/Inbox/Data/VisualInboxSPITypes.swift
```diff
@@ -85,35 +88,51 @@ public struct VisualInboxChrome: Equatable {
...
 
     public init(
         bellBackground: String?,
         bellIconColor: String?,
+        bellIconSvg: String?,
         panelBackground: String?,
         dividerColor: String?,
         badgeBackground: String?,
         cornerRadius: Double?,
+        position: String?,
```
matt-frizzell, line 110: 👀 Non-blocking, validate.

**Test gap:** `InboxBrandingTest.swift` wasn't updated for this — its `floatingIcon` fixture has no `svg` key, so the new field parse (and its absent-key → `nil` path) is untested, as is the `VisualInboxProviderImpl` mapping of `svg` / `position` / `showAlert` into `VisualInboxChrome`. These are one-line assertions on the existing fixture; worth adding since the rest of the branding parse is covered there.

matt-frizzell (approved):
None of these are architectural blockers and could be follow-ups but seemed relevant enough to raise.

Sources/MessagingInbox/NotificationInboxOverlay.swift
```diff
@@ -7,147 +7,83 @@ import SwiftUI
...
-        if #available(iOS 14.0, *) {
-            ignoresSafeArea()
-        } else {
-            edgesIgnoringSafeArea(.all)
+        // System sheet with medium/large detents + grabber — no header (matches web parity). The list
+        // shares this overlay's model, so bell and sheet observe the same state.
+        .sheet(isPresented: $isInboxPresented) {
+            // Dismiss the sheet after a navigation action so the opened screen (deep link / URL) isn't
+            // left behind the inbox.
+            NotificationInboxView(model: model, onNavigate: { isInboxPresented = false })
```
Shahroz16, line 90: Are we intentionally moving away from the workspace’s panel background and corner radius when using the native sheet? The previous panel applied both values, and web and Android still do, but neither resolved value has a consumer after this change. If system sheet styling is the product decision, could we make that explicit and remove the unused resolution? Otherwise, could we apply the branded surface and corner treatment here?

Sources/MessagingInbox/NotificationInboxView.swift
```diff
@@ -155,23 +180,27 @@ public struct NotificationInboxView: View {
...
+            // links, so inbox deep links behave consistently with the rest of the SDK.
+            guard let value = resolution.actionValue, let url = URL(string: value) else {
+                logger.debug("[CIO-Inbox] openDeeplink action has no openable value (name=\(resolution.actionName))")
                 return
             }
-            // `performDefaultNavigation` runs inside an unstructured Task (after awaiting the
-            // main-actor model), so hop back to the main actor: UIApplication.shared.open is a
-            // UIKit/main-thread API.
-            DispatchQueue.main.async { UIApplication.shared.open(url) }
+            DispatchQueue.main.async { DIGraphShared.shared.deepLinkUtil.handleDeepLink(url) }
```
Shahroz16: Would it make sense to inject the deep link navigator here rather than resolving `DIGraphShared.shared.deepLinkUtil` inside the view? Our usual pattern keeps the shared graph at module initialization and uses constructor injection below that boundary. It would also make host interception and the ordering of navigation and dismissal easier to test.

Sources/MessagingInbox/VisualInboxMessageRow.swift
```diff
@@ -78,31 +79,38 @@ struct VisualInboxMessageRow: View {
...
+        let actionValue = data?["action"]?.stringValue ?? data?["url"]?.stringValue
+        // Tracking name is `name`; fall back to the Jist node name.
+        let actionName = data?["name"]?.stringValue ?? event.name
         let behavior: InboxActionResolution.Behavior
         switch data?["behavior"]?.stringValue {
-        case "openUrl": behavior = .openUrl
-        case "newTab": behavior = .newTab
-        case "deeplink": behavior = .deeplink
-        default: behavior = .none
+        // `newTab` is the web "open in new tab" variant; on mobile it is a plain external browser open.
```
Shahroz16: Could we align this with the shape gist-web actually emits? `newTab` is a Boolean next to `behavior: "openUrl"`, rather than a behavior of its own. The new test currently locks in a payload the web parser does not produce. I also noticed the dismiss fixtures use the string `"true"` but do not exercise the real `.bool(true)` path. If these string forms are intentional legacy support, could we label them that way and add coverage for the current schema as well?

Sources/MessagingInbox/InboxActionResolution.swift
```diff
@@ -3,24 +3,36 @@ import Foundation
...
-    /// `newTab`). `none` when the message carried no behavior — we then infer from the url scheme.
+    /// The action `behavior` from the message's `properties[actionName]`. Matches the web enum
+    /// `InboxActionBehavior` (minus `dismiss`, which is resolved before this type is built).
     enum Behavior: Equatable {
+        /// Open `actionValue` as a URL.
         case openUrl
-        case newTab
-        case deeplink
-        case none
+        /// Open `actionValue` as a deep link. Web treats this identically to `openUrl`; on iOS both
```
Shahroz16: Could we update this description to match the new routing path? `openDeeplink` now uses `deepLinkUtil.handleDeepLink`, including the host callback and universal link handling, rather than going directly through `UIApplication.open`. As written, the type describes different behavior from the implementation.

Sources/MessagingInbox/NotificationInboxOverlay.swift
```diff
@@ -7,147 +7,83 @@ import SwiftUI
...
-            // The slide-out panel sits above the scrim but behind the floating button.
-            if isPanelOpen, model.showsChrome {
-                panel
-                    .transition(.move(edge: .trailing).combined(with: .opacity))
+        // Bell anchor is branding-driven (`patterns.inbox.position`); defaults to bottom-right.
+        let bellPosition = InboxBellPosition.resolve(model.chrome?.position)
+        // Bell is shown unless the inbox is hidden (the bell view hides itself based on chrome state);
+        // tapping it presents the inbox list in a native sheet with system detents.
+        //
+        // The bell + lifecycle live in a ZStack anchored by a stable, non-interactive `Color.clear`.
```
Shahroz16, line 52: I may be missing a SwiftUI detail here, but do we still need `Color.clear` for lifecycle reasons? The lifecycle modifiers are attached to the outer overlay, so replacing the bell’s conditional child should not remove that outer view. If the clear view is only making the alignment container expand, could the expanding frame carry `bellPosition.alignment` instead? That might let us remove both the extra view and this fairly involved explanation.

Tests/MessagingInbox/InboxBellIconTests.swift
```diff
@@ -0,0 +1,70 @@
...
+
+        XCTAssertFalse(path.isEmpty)
+        let bounds = path.boundingRect
+        // Scaled-to-fit + centered: stays within the target rect.
+        XCTAssertGreaterThanOrEqual(bounds.minX, rect.minX - 0.5)
+        XCTAssertGreaterThanOrEqual(bounds.minY, rect.minY - 0.5)
+        XCTAssertLessThanOrEqual(bounds.maxX, rect.maxX + 0.5)
+        XCTAssertLessThanOrEqual(bounds.maxY, rect.maxY + 0.5)
+        // Aspect preserved: viewBox is taller than wide, so height >= width in a square rect.
+        XCTAssertGreaterThanOrEqual(bounds.height, bounds.width - 0.5)
```
Shahroz16: Could we make this assertion more specific to aspect preservation? A path stretched incorrectly from 24×25 to 100×100 would still satisfy these bounds and the height versus width check. For this fixture, asserting the expected fitted bounds, roughly x=2, y=0, width=96, height=100, or directly comparing the aspect ratios would catch that regression.

Shahroz16 (commented):
Thanks for pulling the native sheet and action behavior together., I tried to test it out as well.

Before the feature branch is merged to main, could we make sure the iOS 13 and CocoaPods release paths are completed? With Jist #39 applied, a disposable iOS 13 build still failed on `ProgressView` until a compatible spinner was added, and the current branch still keeps the package and Jist rendering gates at iOS 15. CocoaPods also needs a published and declared Jist dependency, along with the existing `Bundle.module` issue resolved. The green sample does not currently include this pod. The `VisualInboxProviderTest` crash also occurs on the base commit, so it does not look introduced here, though it would be good to get a green rerun before merging. I left a few inline questions about branding, schema fidelity, dependency injection, and test coverage.

The [binary size report](https://github.com/customerio/customerio-ios/pull/1155#issuecomment-4962573230) compares the whole feature branch against main and shows an increase of about 202 KiB, its because of the SVGs added, I am okay with it, but if we want even more compressed assets that could reduce the size

CustomerIOMessagingInbox.podspec
```diff
@@ -20,7 +20,8 @@ Pod::Spec.new do |spec|
   path_to_source_for_module = "Sources/MessagingInbox"
   spec.source_files = "#{path_to_source_for_module}/**/*{.swift}"
   spec.resource_bundle = {
-    "#{spec.module_name}_Privacy" => "#{path_to_source_for_module}/Resources/PrivacyInfo.xcprivacy"
+    "#{spec.module_name}_Privacy" => "#{path_to_source_for_module}/Resources/PrivacyInfo.xcprivacy",
+    "#{spec.module_name}_InboxAssets" => "#{path_to_source_for_module}/Resources/CioInboxAssets.xcassets"
```
Shahroz16, line 24: The resource bundle fix looks right. Before this stack merges, could we also close the CocoaPods dependency gap? The podspec compiles everything under `Sources/MessagingInbox`, including files that `import Jist`, but still declares only `CustomerIOMessagingInApp` and `CustomerIOCommon`. Would a CocoaPods consumer hit `no such module 'Jist'` here? It seems like the inbox pod cannot be released until Jist is published and declared, and the pod is linted against that dependency.

Sources/MessagingInbox/NotificationInboxView.swift
```diff
@@ -58,11 +66,16 @@ public struct NotificationInboxView: View {
             .onAppear { if marksOpenedOnAppear { model.markVisibleMessagesOpened() } }
     }
 
-    /// Panel body driven by load state: spinner while loading, empty placeholder when visible-but-empty,
-    /// otherwise the Jist-rendered list.
+    /// Panel body driven by load state: nothing when the inbox is hidden, spinner while loading,
+    /// empty placeholder when visible-but-empty, otherwise the Jist-rendered list.
```
Shahroz16, line 76: Could we finish the iOS 13 path here? `NotificationInboxView` is public on iOS 13, but this loading branch still uses `ProgressView` (iOS 14+), while Jist rendering remains gated to iOS 15 and `Package.swift` still sets iOS 15. Once Jist #39 lands, would it make sense to lower the deployment target, drop the iOS 15 gates, and use an iOS 13-compatible spinner so this is actually compiled against an iOS 13 destination?

Shahroz16 (changes requested):
The CocoaPods pod cannot currently resolve Jist I belive, that would be a blocker to fix

Sources/MessagingInbox/NotificationInboxOverlay.swift
mahmoud-elmorabea: We're keeping `Color.clear` deliberately — it's a lifecycle anchor, not just an alignment spacer, and removing it reintroduced a real bug on device.

The problem it solves: when the inbox goes hidden, `NotificationInboxBell` renders nothing (it hides itself based on chrome state). The lifecycle modifiers (`onAppear { model.start() }` / `onDisappear { model.stop() }`) are on the outer overlay — but if the *only* child in the `ZStack` is the bell, SwiftUI can tear down/rebuild the alignment container when its content flips to empty, and in practice we saw `onDisappear` fire → `model.stop()` cancels the data subscription **permanently**, so a later message can never re-show the bell. The always-present, non-interactive `Color.clear` keeps a stable child in the stack so the overlay (and its subscription) stays mounted while the bell is hidden; `allowsHitTesting(false)` preserves touch passthrough to the content behind it.

On the "could the expanding frame carry `bellPosition.alignment` instead" suggestion — the `.frame(maxWidth/Height: .infinity)` already carries alignment, so placement doesn't need the clear view. But the frame alone doesn't guarantee a stable non-empty child when the bell renders nothing, which is the part that matters for the subscription lifecycle. I confirmed the regression (bell never returns after hidden→message) on device when the anchor is dropped, so I'd rather keep it. The inline comment at that block already spells out this lifecycle reason so it doesn't read as dead scaffolding.

Sources/MessagingInbox/NotificationInboxBell.swift
mahmoud-elmorabea: Fixed — added a `Bundle.cioInboxResources` shim (`.module` on SwiftPM; resolves the resource bundle on CocoaPods).

Sources/MessagingInbox/InboxBellIcon.swift
mahmoud-elmorabea: Fixed — the glyph is now parsed once in `VisualInboxModel` when branding resolves and held; `InboxBellIcon` only applies the fit transform per layout. Dropped the static cache.

Sources/MessagingInbox/NotificationInboxView.swift
mahmoud-elmorabea: Fixed — dismiss now runs before navigation is enqueued.

Sources/MessagingInApp/Inbox/Data/VisualInboxSPITypes.swift
mahmoud-elmorabea: Done — added `svg` assertions to `InboxBrandingTest` plus a `brandingChrome` → `VisualInboxChrome` mapping test (svg/position/showAlert) in `VisualInboxProviderTest`.

Sources/MessagingInbox/NotificationInboxView.swift
mahmoud-elmorabea: Done — injected `InboxActionNavigating` (`DefaultInboxActionNavigator`) instead of resolving `DIGraphShared.shared` inside the view.

Sources/MessagingInbox/VisualInboxMessageRow.swift
mahmoud-elmorabea: Fixed — dropped `newTab` as a behavior; realigned to the web boolean shape, and added a real `.bool(true)` dismiss test.

Sources/MessagingInbox/InboxActionResolution.swift
mahmoud-elmorabea: Fixed — doc now describes the `deepLinkUtil.handleDeepLink` routing.

Tests/MessagingInbox/InboxBellIconTests.swift
mahmoud-elmorabea: Done — asserts exact fitted bounds (x=2, y=0, 96×100).

Sources/MessagingInbox/NotificationInboxOverlay.swift
mahmoud-elmorabea: Done — the sheet now applies the branded background + corner radius (`BrandedInboxSheetStyle`: presentation APIs on iOS 16.4+, content background on 16.0–16.3).

Sources/MessagingInbox/Vendor/SVGPath/SVGPath.swift
mahmoud-elmorabea: Fixed — added `Vendor/SVGPath/LICENSE.md`.

Sources/MessagingInbox/VisualInboxMessageRow.swift
```diff
@@ -78,31 +79,38 @@ struct VisualInboxMessageRow: View {
...
+        case "openUrl", "newTab": behavior = .openUrl
+        case "openDeeplink": behavior = .openDeeplink
+        case "performAction": behavior = .performAction
+        default: behavior = .unknown
```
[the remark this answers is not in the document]
mahmoud-elmorabea: Intentional — we align to gist-web's action schema, which emits `openDeeplink` (and `openUrl`), never a bare `deeplink` behavior, so mapping it would re-introduce a shape the web parser doesn't produce. It resolves to `unknown`, and the host is still offered the action via the listener.

Sources/MessagingInbox/NotificationInboxView.swift
mahmoud-elmorabea: Partially addressed — the loading branch now uses an iOS-13-compatible spinner (`InboxLoadingSpinner`) instead of `ProgressView`. Lowering the deployment target + dropping the iOS-15 Jist gates is a post-Jist-#39 follow-up.

CustomerIOMessagingInbox.podspec
mahmoud-elmorabea: We won't publish the Jist version until all testing on this feature is done, so declaring/linting the `Jist` dependency here is deferred to later on this feature branch — it doesn't block this PR.

PR #1156, opened 2026-07-14
chore(geofence): address geoset PRs review feedback

mrehan27:
Addresses review feedback from the geoset delivery PRs.

Ticket: [MBL-2087](https://linear.app/customerio/issue/MBL-2087/mobile-address-geosets-prs-review-feedback)

## Fixes
- **Silent-location doc**: corrected — a silent fix *is* observable via `getLastKnownLocation` (unpersisted, no analytics/enrichment); geofencing anchors its refresh on it.
- **Removed unused pending-store APIs** (`removeAll(keys:)`, `clearAll`) whose only reference was their own tests; `clearAll`'s doc also implied a sign-out wipe that contradicts the retain-for-replay policy.
- **Drop anonymous transitions at capture** — geofencing is identified-only (the backend rejects anonymous geofence tracks), so an anonymous crossing is dropped before cooldown/persist instead of being sent anonymously and wrongly treated as delivered.
- **Split delivery by attempt** (Android parity): a fresh transition goes out over direct HTTP (works before DataPipeline is up on a cold-wake); the `flushPending` replay hands the backlog to EventBus → DataPipeline (publish-then-remove).
- **Pinned attribution on replay**: `TrackGeofenceMetricEvent` now carries the snapshot `userId`, which the DataPipeline observer pins onto the track (with timestamp) so a delayed replay attributes to the crossing's user and time, not the current identity/send time.

## Notes
- Verified against Android review-fix commit `f6747557` — full parity on everything this touches.

---

> [!NOTE]
> **Medium Risk**
> Changes geofence delivery channels and identity attribution on replay; behavior is covered by extensive test updates but affects analytics correctness and cold-wake delivery paths.
> 
> **Overview**
> Geofence transitions are now **identified-only**: crossings with no user at capture are dropped before cooldown or persistence (backend rejects anonymous tracks), with a new debug log for that case.
> 
> **Delivery is split by attempt** (Android parity): a fresh transition still goes out over **direct HTTP** under a background task; **`flushPending` replays** the on-disk queue via **EventBus → DataPipeline** only (no HTTP retry on flush, no background assertion on replay).
> 
> **Attribution is pinned on replay**: `TrackGeofenceMetricEvent` and pending rows require a snapshot **`userId`**; DataPipeline sets `trackEvent.userId` from the event (with the existing timestamp pin) so delayed replay does not follow the current identity. EventBus posts use **`postEventAndWait`** so rows are removed only after handoff resolves.
> 
> Removes unused **`PendingGeofenceMetricStore.clearAll`** / **`removeAll(keys:)`**. Updates **`requestLocationUpdateSilently`** docs to note the in-memory last-known fix is visible via **`getLastKnownLocation`** for geofencing refresh.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4b561091a079dfdd220f2917301104ebd7662e8f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Sources/LocationGeofence/Event/GeofenceEventTracker.swift
+++ b/Sources/LocationGeofence/Event/GeofenceEventTracker.swift
@@ -167,9 +167,10 @@ final class GeofenceEventTracker: @unchecked Sendable {
         // HTTP failure: row stays for next flush.
     }
 
-    /// Replay via EventBus → DataPipeline. Awaits the durable handoff (observer invoked, or the
-    /// event persisted to the EventBus cache when none exists yet) before removing our copy, so a
-    /// crash in between re-delivers on the next flush (deduped by transitionId) rather than dropping.
+    /// Replay via EventBus → DataPipeline, which then owns delivery and retry. We drop our copy once
+    /// the handoff resolves (observer invoked, or written to the EventBus cache when none exists yet).
+    /// That is not a durable-persistence ack — the strongest signal EventBus exposes today — so a
+    /// crash before DataPipeline persists re-delivers on the next flush (deduped by transitionId).
     private func deliverViaEventBus(metric: PendingGeofenceMetric) async {
         guard activeDeliveryKeys.mutating({ $0.insert(metric.key).inserted }) else { return }
         defer { activeDeliveryKeys.mutating { _ = $0.remove(metric.key) } }
@@ -192,8 +193,8 @@ final class GeofenceEventTracker: @unchecked Sendable {
 
     /// Hands a row to DataPipeline via EventBus, carrying the snapshot userId + timestamp so a
     /// delayed replay attributes to the right person and time, not the current identity/send time.
-    /// `postEventAndWait` (not `postEvent`) so the caller can drain the row only once the handoff is
-    /// durable — `postEvent` returns before its detached delivery/persist task runs.
+    /// `postEventAndWait` (not `postEvent`) so the caller drains the row only once the handoff has
+    /// resolved — `postEvent` returns before its detached delivery/persist task even runs.
     private func postEventBus(metric: PendingGeofenceMetric) async {
         await eventBusHandler.postEventAndWait(TrackGeofenceMetricEvent(
             geofenceId: metric.geofenceId,
--- a/Sources/LocationGeofence/Event/GeofenceEventTracker.swift
+++ b/Sources/LocationGeofence/Event/GeofenceEventTracker.swift
@@ -3,24 +3,18 @@ import Foundation
 
 // sourcery: InjectRegisterShared = "GeofenceEventTracker"
 // sourcery: InjectCustomShared
-/// Sends geofence transition events through a three-layer delivery flow:
-/// 1. Cooldown-based deduplication (keyed by "geofenceId:transitionType")
-/// 2. Persist to `PendingGeofenceMetricStore` before any send attempt, stamping
-///    the currently-identified userId so A's transitions always attribute to A
-/// 3. Dispatch in `deliver`:
-///    - Stamped userId → direct-HTTP path with that userId; success drains the
-///      row, failure retains it for the next flush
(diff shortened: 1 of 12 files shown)


Sources/LocationGeofence/Storage/PendingGeofenceMetric.swift
```diff
@@ -1,14 +1,15 @@
...
+/// EventBus → DataPipeline. Always carries the userId identified at capture; geofencing is
+/// identified-only, so anonymous crossings are dropped before a row is ever created.
 struct PendingGeofenceMetric: Codable, Equatable, Sendable, GeofenceMetric {
     let geofenceId: String
     let transition: GeofenceTransition
     let timestamp: Date
-    /// The userId identified at capture time, or `nil` if none was identified.
-    let userId: String?
+    /// The userId identified at capture time.
+    let userId: String
```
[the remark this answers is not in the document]
mrehan27: Valid mechanism, but not reachable here. Geofencing has never shipped, so no device holds a persisted pending-metric file from an earlier SDK — there are no legacy rows with a missing/`null` `user_id` to decode. And the drop-at-capture guard means every row persisted from now on carries a `user_id`, so the on-disk schema is self-consistent from the first release. No change needed.

Sources/LocationGeofence/Event/GeofenceEventTracker.swift
```diff
@@ -186,6 +167,18 @@ final class GeofenceEventTracker: @unchecked Sendable {
...
+    /// Replay via EventBus → DataPipeline. Awaits the durable handoff (observer invoked, or the
+    /// event persisted to the EventBus cache when none exists yet) before removing our copy, so a
+    /// crash in between re-delivers on the next flush (deduped by transitionId) rather than dropping.
+    private func deliverViaEventBus(metric: PendingGeofenceMetric) async {
+        guard activeDeliveryKeys.mutating({ $0.insert(metric.key).inserted }) else { return }
+        defer { activeDeliveryKeys.mutating { _ = $0.remove(metric.key) } }
+
+        let effective = await resolvingLiveValues(metric)
+        await postEventBus(metric: effective)
+        _ = await pendingStore.remove(key: metric.key)
```
Shahroz16, line 180: Could we keep the geofence row until the normal observer path has acknowledged durable persistence? `postEventAndWait` waits for callback invocation, but the DataPipeline observer calls `analytics.process`, and the default CioAnalytics `TransientDB.append` schedules its disk write asynchronously. This line can therefore remove the only durable copy before that write completes, so process death in this window still loses the transition. The no-observer path is durable when `eventStorage.store` succeeds, but `CombinedCacheEventBusHandler` catches storage errors and still returns normally, which creates the same removal-on-failure gap. Could this await a success or completion from the destination persistence, or retain another durable copy until that succeeds?

Shahroz16 (approved):
Thanks for catering feedback.

Sources/LocationGeofence/Event/GeofenceEventTracker.swift
mrehan27: Good catch — right on both paths. `postEventAndWait` resolves on observer invocation (or the EventBus cache write when no observer exists yet), not on a downstream persistence ack: the DataPipeline observer's `analytics.process` enqueues and returns before disk flush, and the no-observer path swallows `eventStorage.store` errors. So the copy can be dropped before durable commit.

The window is narrow and crash-only (no duplication), bounded by at-least-once + `transitionId` dedup, so it stays within the best-effort contract. I've reworded the comments here to stop claiming a "durable handoff" and call out this limitation instead.

A real fix means surfacing a persistence ack across the EventBus boundary (`postEventAndWait` returns `Void` and swallows storage errors today, and the handler is shared with push delivery) — that's EventBus scope, so we can track it separately rather than handle it in this project.

PR #1157, opened 2026-07-14
chore(live-activities): Segments + Countdown Timer bundled templates

mahmoud-elmorabea:
## What

Reintroduces a `CioLiveActivities_Templates` module with **two ready-to-use Live Activity templates** apps can render without writing their own widget — they only supply local branding:

- **Segments** (`io.customer.liveactivities.segments`) — status headline + a discrete step progress bar.
- **Countdown Timer** (`io.customer.liveactivities.countdowntimer`) — status + a live `MM:SS` countdown to a target time.

## Design

- **Branding is local**, supplied at the widget call site in the app's `WidgetBundle` (logo, background, text/progress styles) — the widget renders in a separate process, so nothing rides the push payload and the SDK delivers no assets.
- **`background`** accepts a solid `Color` *or* any gradient/`ShapeStyle`. A shared internal `TemplateBackground` routes a `Color` through the system `activityBackgroundTint` (StandBy/AOD-adaptive) and anything else to a directly-drawn background. `textColor`/progress styles accept any `ShapeStyle`.
- Both attributes types conform to `CIOActivityAttribute` (push-to-start capable), carry `cioMetadata` (delivered/opened + deep link), and model date fields with `EpochSecondsDate` (epoch **seconds** on the wire).
- Accessibility: lock-screen cards read as one combined VoiceOver element; the countdown clock rests at `0:00` (finished state is push-driven — documented on `endTime`).

## Wire contract

- **Local** `start`/`update`/`end` → `Live Notification Event` (start includes `attributes` + `contentState`; update/end send `contentState`).
- **Server** push start/update/end are decoded by ActivityKit and are **not** echoed as lifecycle events; the SDK emits token events + `delivered`/`opened` metrics from `cioMetadata`.

## Sample

Wires both templates into **APN-UIKit**: the Chica food-delivery Segments preset (orange gradient + logo) and a sale Countdown preset, each with SwiftUI previews. (An alternate HIIT Segments branding was used during development and is not included.)

## Notes

- Adds the `api-docs/CioLiveActivities_Templates.api` baseline + `generate-api-docs.sh` entry (required by Danger for a new product).
- Targets `feat/live-activities` (the LA integration branch; not yet in `main`).

---

> [!NOTE]
> **Medium Risk**
> New public SDK surface and notification-type identifier changes can break backends still on `io.customer.liveactivities.*`; otherwise mostly additive sample and widget code.
> 
> **Overview**
> Adds a new public **`LiveActivities_Templates`** SPM product with two ready-made Live Activity UIs (**Segments** step progress and **Countdown Timer**), plus shared styling helpers (`TemplateBackground`, branding types, Lock Screen + Dynamic Island layouts).
> 
> Introduces **`CIOActivityTemplate`** and a **`register(_:)`** overload on `LiveActivityConfigBuilder` so built-in types register with their own `io.customer.livenotifications.*` identifiers; docs and the sample **`DeliveryActivityAttributes`** identifier move from `liveactivities` to **`livenotifications`**.
> 
> **APN-UIKit** wires the module into the app and widget extension, registers both templates in `AppDelegate`, adds them to the widget bundle with Chica/sale branding demos, and extends the Live Activities demo screen with auto-run drivers (5s steps, cancel prior auto-task).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0de5ae6ecffa8e0c28d18d0424e896ad88957d2f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Sources/LiveActivities_Templates/CountdownTimer/CIOCountdownTimerLiveActivity.swift
```diff
@@ -0,0 +1,136 @@
...
+/// its own — the view isn't re-evaluated per second). The finished state appears when a push
+/// delivers a new content-state (no `endTime` + a "done" `title`). Rendering nothing for an absent /
+/// past `endTime` also avoids constructing an invalid `timerInterval` range.
+@available(iOS 16.2, *)
+private struct CountdownText: View {
+    let endTime: EpochSecondsDate?
+
+    var body: some View {
+        if let date = endTime?.date, date > Date() {
+            Text(timerInterval: Date() ... date, countsDown: true, showsHours: false)
```
Shahroz16: This is a very narrow boundary case, so I do not think it needs to block this PR.

Could we capture Date once and use the same value for the comparison and the timer range? If endTime passes between these two calls, the condition succeeds but the second Date is already later than endTime, which makes the range invalid.

Using one now value removes that edge case and makes it straightforward to test.

Sources/LiveActivities_Templates/Segments/CIOSegmentsLiveActivity.swift
```diff
@@ -0,0 +1,125 @@
...
+// MARK: - Segmented progress bar
+
+@available(iOS 16.2, *)
+private struct SegmentedBar: View {
+    let state: CIOSegmentsAttributes.ContentState
+    let branding: CIOSegmentsBranding
+
+    var body: some View {
+        HStack(spacing: 6) {
+            ForEach(0 ..< max(state.segmentsTotal, 0), id: \.self) { index in
```
Shahroz16: This is defensive hardening for an extreme payload, not a blocker for this PR.

Could we track a reasonable maximum for segmentsTotal before release? Services currently allows any integer, and a very large value would make the widget extension create a huge number of Capsule views.

Ideally iOS, Android, and services would share the same limit.

Shahroz16 (approved):
From a runtime perspective, this looks good to approve. The templates line up with the services contract, and the LiveActivities_Templates scheme builds successfully.

I left two non-blocking hardening suggestions for malformed or boundary-case payloads. They do not need to hold up this PR. The only current merge issue I see is the SwiftLint failure: LiveActivitiesViewController is 256 lines and the limit is 250.

PR #1158, opened 2026-07-15
chore(geofence): harden background delivery on modern iOS

mrehan27:
Improves the reliability of geofence transition monitoring on modern iOS by adopting the current-generation Core Location monitoring API (`CLMonitor` + `CLServiceSession`) on iOS 18+. Contained to the `LocationGeofence` monitoring layer behind the existing `GeofenceRegionMonitoring` seam — delivery, attribution, cooldown, and storage are unchanged. It also closes a cross-platform gap with a diff-based **initial enter-when-inside** (Android `INITIAL_TRIGGER_ENTER` parity) that runs identically on iOS ≤17 and 18+.

Ticket: [MBL-2105](https://linear.app/customerio/issue/MBL-2105)

## Why
Background geofence delivery has reliability gaps on modern iOS that surface as missed or duplicate transitions:
- **Cold-wake session lifetime** — on iOS 18+, background location delivery is scoped to an explicit service session; without one, transitions delivered after the app is suspended are unreliable.
- **Registration baseline** — the initial monitored state (already-inside a region, or geometry that changed since last registration) needs to be established consistently, or the first crossing after a re-register can be missed or double-fired.
- **Boundary/geometry edge cases** on the newer monitoring stack.

The classic `CLLocationManager` region-monitoring APIs are superseded on newer iOS; `CLMonitor` is Apple's current API for this, and its `CLServiceSession` is the documented mechanism for keeping background delivery alive.

## Approach
Adopt a `CLMonitor`-backed `GeofenceRegionMonitoring` implementation, selected by the DI accessor **on iOS 18+**; iOS 13–17 keep the classic monitor. iOS 18 is the floor because only there does `CLServiceSession` provide a dependable background story — the session is held for the monitor's lifetime and created **only when `Always` is already granted**, so the SDK never triggers a permission prompt (the host owns permission).

`CLMonitor` re-emits a condition's current state (not just crossings) and has no per-transition-type filter, so a per-condition **dedup baseline** and **delivery filter** are persisted in `GeofenceStorage`; the monitor stays a thin adapter. Baselines are seeded from the device's actual state at registration, so registration is silent yet the first real crossing still delivers.

Cold-wake delivery depends on the events consumer being attached when the OS hands over the transition, so the bootstrap constructs the monitor **before** its async storage reads, and `CLMonitor` creation is **not gated on protected-data availability**: that flag is false for the whole wake window while the device is locked (the normal crossing scenario) and can read false on prewarmed launches with no notification ever following — either way the process would end up with no consumer. The reboot corner gating aimed at is unreachable for location launches, and an empty conditions read self-heals via reconcile. Events arriving before the bootstrap binds the transition handler are buffered and drained on bind (processing them earlier would advance the dedup baseline and drop the delivery), and bootstrap runs are serialized so re-triggers (authorization change, reconciliation) can't interleave. A sign-out during a gated refresh (whose `reset()` is dropped on the held dedup gate) is undone at the refresh's single exit point, whichever path the body took — freshness-skip and failed-fetch exits included — and when a new user is already signed in by then, the refresh re-runs for them so a user switch converges to a registered state instead of an unmonitored gap.

Adopted conditions are also **re-armed in place** (removed and re-added) at launch rather than trusting the OS-side monitoring state: per `CLMonitor.h`, CoreLocation stops monitoring a condition whose pending event no monitor was configured to receive — after a reboot the app is not relaunched, so the first missed crossing silently kills monitoring while the persisted store still lists the condition, and every later crossing is lost even once the app is opened. The re-add seeds `assuming:` from the stored baseline, so unchanged conditions are event-silent and transitions missed while unmonitored deliver as genuine catch-up at the next launch. What remains — crossings between the reboot and the first open — is OS policy (no relaunch after reboot; Android covers the same gap with `BOOT_COMPLETED`), and nets out as catch-up at that first open.

## Initial enter-when-inside
Neither iOS OS layer reports an ENTER for a geofence the device is **already inside** when that region is first registered (first launch, sign-in while inside, or the server adding a fence around you). Matching Android's `INITIAL_TRIGGER_ENTER`, the coordinator diffs each registration against the previously-registered set and emits one ENTER for a genuinely-new region containing the anchor — but only for fences the OS actually registered (a blocked-permission or invalid-coordinate drop emits nothing, so there's no enter that can't be balanced by an exit), and only for the user the diff was computed for (re-checked before the fire-and-forget delivery, so a mid-flight sign-in can't misattribute it). A wholesale re-register of unchanged regions stays silent (the diff), a sign-out clears the set so the next sign-in re-fires, and delivery runs through the tracker so it is cooldown-deduped. It sits at the coordinator — above the monitor — so **iOS ≤17 and 18+ behave identically**, and is delivered off the refresh gate so a slow send can't stall a refresh.

## Scope
- **In:** the Core Location monitoring implementation within `LocationGeofence`, plus the coordinator-level initial enter-when-inside.
- **Out:** transition delivery, attribution, cooldown, and storage plumbing — unchanged (the initial-enter reuses them).
- Single monitor per process, version-routed — no double or missing path; the classic path's monitoring is untouched below iOS 18, and the initial-enter path is shared by both.

## Testing
- The monitor's decision logic (dedup baseline, delivery filter, cold-start persistence, changed-geometry reseed) lives in `GeofenceStorage` and is unit-tested; the initial enter-when-inside diff is unit-tested at the coordinator (containment, transition-type filter, new-vs-already-registered, registration-rejected, local + remote paths). Full `LocationGeofenceTests` suite green at 251.
- The CoreLocation adapter itself isn't unit-testable (real `CLMonitor` actor keyed by a process-global name, real `CLLocationManager`, async event stream) — same as the classic monitor. It is validated on-device: background delivery, force-quit cold-wake, reboot region persistence, and offline-fail → replay.
- Terminated-app field test (iOS 26.6, real walk): a crossing relaunched the swiped-up app in the background, the consumer attached ~15ms after bootstrap, the transition delivered ~1.5s after the wake, and four further crossings delivered over the following 25 minutes with nothing left in the pending store.
- Reboot field test (iOS 26.6, real drive): crossings after a reboot with the app never opened were lost as expected (no relaunch), the first open re-armed all 7 adopted conditions and delivered the net transitions missed while unmonitored as catch-up within ~1s of launch, and subsequent crossings delivered in real time from background wakes — force-quit included — with the pending store empty afterwards. A relaunch with a current baseline produced zero condition emissions (re-arm is event-silent when nothing changed).

## Notes
- All pre-merge validations are complete: the reboot cold-wake field test (see Testing) and the initial enter-when-inside on both paths — 18+ in the field (sign-in inside a fence delivered its synthetic ENTER), and the classic ≤17 path on-device with the monitor selection forced to classic (fresh sign-in inside a fence delivered exactly one synthetic ENTER for the containing fence, none for the others).
- Fast transits of small (≈500m) geofences can be missed in the background. This is an OS-level detection limitation of low-power background location (shared by the classic path), not specific to this change; larger radii / dwell are the mitigation. Tracked separately.

---

> [!NOTE]
> **High Risk**
> Touches background location delivery, user attribution during sign-out/switch, and synthetic enter events—areas where bugs cause missed or misattributed geofence transitions.
> 
> **Overview**
> Hardens **geofence monitoring on iOS 18+** by routing `GeofenceRegionMonitoring` through a new **`CLMonitor` + `CLServiceSession`** implementation (classic `CLLocationManager` unchanged on iOS 13–17). The adapter adds persisted **per-condition dedup baselines** and **transition-type filtering** in `GeofenceStorage`, a **FIFO mutation pipeline**, **event buffering** until bootstrap binds handlers, **mirror/reconcile** for adopt-vs-re-register, **re-arm on adopt** after reboot, and **serialized** `GeofenceBootstrap.wireMonitor` runs with **authorization + reconciliation** re-wires.
> 
> Adds coordinator-level **initial enter-when-inside** (Android `INITIAL_TRIGGER_ENTER` parity): after remote/local refresh, **diff new OS-registered fences** against prior IDs and emit synthetic **ENTER** via `GeofenceTransitionEmitting`, respecting OS clamp radius and user identity. **Gated refresh/movement** now runs **exit cleanup** when identity changes mid-operation (including dropped `reset()`), then **self-heal retry** for the current user.
> 
> Sample app gains **`location` background mode**. Extensive unit tests cover coordinator initial-enter, identity races, and CLMonitor storage logic.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit da92c9ed2827c76bad6b998f33a8ad762ef793d2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor.swift
```diff
@@ -0,0 +1,363 @@
...
+        let mirrored = Set(userDefaults.stringArray(forKey: Self.conditionMirrorKey) ?? [])
+        self.knownConditionIdentifiers = mirrored
+        // Everything under our private monitor name was registered by the SDK, so persisted
+        // conditions are owned as soon as the process starts — a cold-wake event must find its
+        // identifier in the filter before any async work has run.
+        self.ownedRegionIdentifiers = mirrored
+        // Auth-status changes only; no region callbacks arrive here.
+        authManager.delegate = self
+        updateServiceSession()
+        enqueueMonitorOperation { [weak self] monitor in
```
Shahroz16, line 88: Could we make reconciliation with live identifiers a readiness step before bootstrap decides whether to adopt or re-register, and before event consumption starts? Right now init seeds from UserDefaults, queues live reconciliation, then starts the consumer, while bootstrap immediately reads osMonitoredRegionIdentifiers. This is why both Bugbot findings hold: a stale extra mirror entry can adopt a condition CLMonitor dropped, and a missing mirror entry can make stop-all skip a live condition. It also lets a pending event be filtered before reconciliation adds its identifier. A small async preparation or live-monitor seam would let us cover live identifiers differing from the mirror and keep the mirror as a warm cache only.

Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor.swift
```diff
@@ -0,0 +1,363 @@
...
+        }
+        startConsuming()
+    }
+
+    // MARK: - CLMonitor lifecycle
+
+    private func monitorInstance() -> Task<CLMonitor, Never> {
+        if let monitorTask { return monitorTask }
+        let name = Self.monitorName
+        let task = Task { await CLMonitor(name) }
```
Shahroz16, line 104: Can we defer creating CLMonitor until protected data is available and resume on UIApplicationProtectedDataDidBecomeAvailable? Apple's CLMonitor header explicitly requires that ordering because its persisted conditions live in app-container storage. We create it immediately from init, including the reboot cold-wake path; before first unlock the persisted state may be unavailable, which undermines the scenario this PR is intended to harden. Could we add a deferral test and include before-first-unlock in the device validation?

Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor.swift
```diff
@@ -0,0 +1,363 @@
...
+
+    private func startConsuming() {
+        consumeTask = Task { [weak self] in
+            guard let self else { return }
+            let monitor = await self.monitorInstance().value
+            do {
+                for try await event in await monitor.events {
+                    await self.handle(event: event)
+                }
+            } catch {
```
Shahroz16: Could we restart the event consumer with bounded backoff when the sequence throws or completes? This catch permanently ends the only subscription, and the log already acknowledges that transitions may stop until relaunch. A sequential subscription after the old loop has ended does not create the simultaneous-consumer problem described above. A small events abstraction would also make this failure path testable.

Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor.swift
```diff
@@ -0,0 +1,363 @@
...
+        // lives on the (deprecated) CLLocationManager, but it's a harmless property read and there is no
+        // CLMonitor equivalent, so use it to keep both paths registering identical geometry.
+        let clampedRadius = min(radius, authManager.maximumRegionMonitoringDistance)
+
+        // The device's ACTUAL state relative to the circle seeds both CLMonitor's `assuming:` hint and
+        // the stored baseline (see `recordMonitorRegistration` for why that keeps registration silent
+        // yet the first real crossing delivers). Fall back to the geometric expectation when there's no
+        // fix: the movement trigger is centered on the device (inside); business geofences are outside.
+        let isMovementTrigger = identifier == GeofenceConstants.movementTriggerIdentifier
+        let isInside = isDeviceInside(center: coordinate, radius: clampedRadius) ?? isMovementTrigger
```
Shahroz16, line 273: Could we avoid treating this as the device's actual state when location can be nil or stale? CLLocationManager.location is only the last location received, and this manager never requests a fresh fix; nil also assumes every business geofence is outside. Apple documents that a wrong assumed state is corrected with an event. Because the storage baseline uses the same guess and recordMonitorEvent delivers any difference, registering while already inside can generate a synthetic enter rather than a real crossing. Starting as unknown and suppressing the first resolved state, or using a freshness-bounded fix, would preserve classic silent registration. Could we cover the nil and wrong-guess paths?

Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor.swift
```diff
@@ -0,0 +1,363 @@
...
+
+/// `CLMonitor`-backed implementation of `GeofenceRegionMonitoring`, where the classic
+/// `CLLocationManager` region-monitoring APIs are deprecated. Available at iOS 17, but the DI
+/// accessor only routes to it on iOS 18+ (13–17 keep the classic monitor) — iOS 18 is the floor
+/// because only there does `CLServiceSession` keep background delivery alive (see the accessor).
+///
+/// Behavioral contract: match the classic monitor — deliver only genuine boundary crossings, only
+/// for the registered transition types. `CLMonitor` differs from classic in ways this type has to
+/// compensate for:
+/// - It re-emits a condition's CURRENT state on process start and system re-evaluation
```
Shahroz16: Have we validated on device that CLMonitor actually re-emits an unchanged state on a new events iteration? Apple's documented model says it emits when the observed state differs from Record.lastEvent and recommends using the monitor's records instead of a parallel event-state table. If that behavior holds, we could persist the transition filter but source the last state from the CLMonitor record, avoiding two state stores that can drift. If iOS 18 behaves differently in practice, documenting that reproduction would make the reason for this second baseline clear.

Shahroz16 (commented):
I left some comments on lifecycle paths that the storage tests cannot exercise. I agree with both `Bugbot` findings: bootstrap reads a fallible mirror before live CLMonitor reconciliation.

Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor.swift
mrehan27: Done in `5f958ff9`: `CLMonitor` is now created behind a `ProtectedDataAvailability` provider that awaits `isProtectedDataAvailable` / `protectedDataDidBecomeAvailableNotification` before constructing it, so the reboot-before-first-unlock path does not build it against unavailable container storage. Ill fold before-first-unlock into the on-device reboot cold-wake pass thats still open before merge.

Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor.swift
mrehan27: Done in `5f958ff9`: `startConsuming` now re-subscribes with bounded backoff (1s→30s) on throw or completion instead of ending permanently, logging each failure via `geofenceMonitorEventStreamFailed`.

Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor.swift
mrehan27: Addressed in `5f958ff9`: once `CLMonitor` reconciles the UserDefaults mirror against its live identifiers it fires `setOnReconciled`, which re-runs `wireMonitor` so adopt/re-register is decided against live truth, not a stale mirror. A new `GeofenceBootstrapTests` case covers the re-run. The two Bugbot mirror findings key off this same seam.

Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor.swift
mrehan27: We have leaned into this rather than away from it: `7fa048d2` makes enter-when-inside a deliberate feature — a genuinely-new registration where the device is already inside now emits one ENTER (Android `INITIAL_TRIGGER_ENTER` parity). It is decided at the coordinator by diffing against the previously-registered set, so it fires only for new fences (not on every re-evaluation), and it is delivered through the tracker so it is cooldown-deduped. The CLMonitor baseline still seeds from actual state so plain registration stays silent; a wrong nil/stale guess costs at most one corrective event, which the baseline absorbs on the next crossing.

Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor.swift
mrehan27: Yes — validated on device: iOS 18 `CLMonitor` re-emits every conditions current state on relaunch (and ~14x on an unlock), not just differences from `Record.lastEvent`; documented in the class doc (`303f2a64`). We keep the separate baseline rather than reading `Record.lastEvent` because it also has to carry the per-region transition-type filter CLMonitor does not honor and survive the stop-all + start-all re-registration that runs on every sync — neither of which `Record.lastEvent` gives us.

Sources/LocationGeofence/GeofenceBootstrap.swift
```diff
@@ -37,6 +37,21 @@ enum GeofenceBootstrap {
...
+        // - authorization changes → re-attempt registration when permission improves mid-process.
+        // - reconciliation → re-decide adopt-vs-re-register once live OS truth is known, correcting a
+        //   choice made off the pre-reconcile mirror (CLMonitor path only; no-op on classic, whose
+        //   osMonitoredRegionIdentifiers is already live).
+        let rewire: @MainActor () -> Void = {
+            Task { @MainActor in await GeofenceBootstrap.wireMonitor(di: di) }
+        }
+        monitor.setOnAuthorizationChanged(rewire)
+        monitor.setOnReconciled(rewire)
+
```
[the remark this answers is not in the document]
mrehan27: Mitigated rather than fully eliminated, deliberately. The first adopt/re-register decision does read the UserDefaults mirror, kept for the cold-wake no-await window — we cannot `await` live `CLMonitor.identifiers` before registering without risking a queued cold-wake transition landing in an empty filter. After `CLMonitor` reconciles the mirror against its live identifiers, `setOnReconciled` re-runs `wireMonitor` against live truth, correcting a stale mirror that adopted a dropped condition on that second pass. Making reconciliation a synchronous pre-step would reintroduce the await we specifically avoid.

Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor.swift
```diff
@@ -0,0 +1,400 @@
...
+        let clampedRadius = min(radius, authManager.maximumRegionMonitoringDistance)
+
+        // The device's ACTUAL state relative to the circle seeds both CLMonitor's `assuming:` hint and
+        // the stored baseline (see `recordMonitorRegistration` for why that keeps registration silent
+        // yet the first real crossing delivers). Fall back to the geometric expectation when there's no
+        // fix: the movement trigger is centered on the device (inside); business geofences are outside.
+        let isMovementTrigger = identifier == GeofenceConstants.movementTriggerIdentifier
+        let isInside = isDeviceInside(center: coordinate, radius: clampedRadius) ?? isMovementTrigger
+        let initialTransition: GeofenceTransition = isInside ? .enter : .exit
+        let assumedState: CLMonitor.Event.State = isInside ? .satisfied : .unsatisfied
```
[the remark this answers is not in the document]
mrehan27: Acknowledged — the coordinators initial-enter uses the refresh anchor while the `CLMonitor` baseline uses `authManager.location`, so they can disagree for a genuinely-new fence. If they disagree toward "inside", the extra enter is cooldown-deduped; toward "outside", it is the same stale-anchor residual that already bounds registration accuracy (the OS delivers the real crossing on the next move). We keep the anchor as the coordinators source because it is monitor-agnostic — the classic <=17 path has no `CLMonitor` record to read — and it matches Android, which also diffs/enters at its last known location.

Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator+InitialEnter.swift
```diff
@@ -0,0 +1,27 @@
...
+                && region.distanceTo(anchor) <= region.radius
+        }
+        guard !newInside.isEmpty else { return }
+        // Deliver off the refresh gate (like the binder does for real crossings) so a slow send can't
+        // stall the next refresh; `trackTransition` persists first, so an interrupted send is retried.
+        Task { [transitionEmitter] in
+            for region in newInside {
+                await transitionEmitter.trackTransition(geofenceId: region.id, transition: .enter)
+            }
+        }
```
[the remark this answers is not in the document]
mrehan27: Fixed in 6e8edff4. `registerWithOsSync` now returns the OS-accepted identifier set and `emitInitialEnters` gates on it, so a candidate the monitor dropped (blocked permission / invalid coordinates) no longer produces a synthetic enter it could never balance with an exit — parity with Android firing the initial trigger only on a successful add. Covered by `refresh_givenNewGeofenceInsideButRegistrationRejected_expectNoInitialEnter`.

Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator+InitialEnter.swift
```diff
@@ -0,0 +1,27 @@
...
+                && region.transitionTypes.contains(.enter)
+                && region.distanceTo(anchor) <= region.radius
+        }
+        guard !newInside.isEmpty else { return }
+        // Deliver off the refresh gate (like the binder does for real crossings) so a slow send can't
+        // stall the next refresh; `trackTransition` persists first, so an interrupted send is retried.
+        Task { [transitionEmitter] in
+            for region in newInside {
+                await transitionEmitter.trackTransition(geofenceId: region.id, transition: .enter)
+            }
```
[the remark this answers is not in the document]
mrehan27: Fixed in 6e8edff4. The emit task now re-checks `contextStore.currentUserId == expectedUserId` before delivering — the same guard `performRemoteRefresh` applies before registering — so a sign-out/sign-in between sync completion and the detached task cannot reattribute the enter to whoever is current.

Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator+InitialEnter.swift
```diff
@@ -0,0 +1,40 @@
...
+        registeredIds: Set<String>,
+        previouslyRegisteredIds: Set<String>,
+        expectedUserId: String,
+        anchor: LocationData
+    ) {
+        let newInside = candidates.filter { region in
+            registeredIds.contains(region.id)
+                && !previouslyRegisteredIds.contains(region.id)
+                && region.transitionTypes.contains(.enter)
+                && region.distanceTo(anchor) <= region.radius
```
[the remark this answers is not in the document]
mrehan27: Not addressing this — it's a non-issue for realistic geofences. Region radius is capped at 10 km server-side, well below `maximumRegionMonitoringDistance`, so `distanceTo <= region.radius` and the clamped `min(radius, cap)` are always equal — the clamp would never change behavior. And even if a radius did exceed the OS cap, emitting on the defined radius is the correct semantic: the device genuinely is inside the geofence's defined area, the OS monitoring a smaller circle is a platform limitation, and any later duplicate real-ENTER is caught by the cooldown dedup. Not worth adding the OS-cap plumbing for that case.

Shahroz16: I think the open bot comment is valid here. Services `origin/main` currently validates fence radii up to 10,000 m, but that does not establish that every supported device has a `maximumRegionMonitoringDistance` of at least 10 km. Apple defines that value as the platform capability, and both monitor paths register `min(radius, maximumRegionMonitoringDistance)`. Here we compare against the unclamped radius, so a device outside the circle actually registered with the OS can still get a synthetic enter. Could we use the same effective radius for this check, ideally from the accepted registration, and cover a cap below the fence radius in a test?

Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator+InitialEnter.swift
```diff
@@ -0,0 +1,40 @@
...
+        }
+        guard !newInside.isEmpty else { return }
+        // Deliver off the refresh gate (like the binder does for real crossings) so a slow send can't
+        // stall the next refresh; `trackTransition` persists first, so an interrupted send is retried.
+        Task { [transitionEmitter, contextStore] in
+            // The diff was computed for `expectedUserId`; a sign-out/sign-in before this detached task
+            // runs must not reattribute the enter to whoever is current now (as `performRemoteRefresh`).
+            guard contextStore.currentUserId == expectedUserId else { return }
```
Shahroz16: The pre-loop guard covers an identity change before this task starts, but every `trackTransition` is awaited. If the first delivery is slow and the user signs out or changes before the next iteration, the tracker reads the new current user and stamps the remaining initial enters to that profile. Could we recheck `expectedUserId` for every iteration, or pass the captured user ID into the emitter so the whole batch has fixed attribution? A test that suspends the first emit and changes the user would pin this down.

Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator.swift
```diff
@@ -263,7 +267,9 @@ final class GeofenceSyncCoordinatorImpl: GeofenceSyncCoordinator, @unchecked Sen
         let effectiveConfig = parsedConfig ?? cachedConfig ?? .fallback
         let anchor = LocationData(latitude: latitude, longitude: longitude)
         let nearest = distanceFilter.nearest(regions, to: anchor, limit: effectiveConfig.maxBusinessGeofences, maxDistance: effectiveConfig.maxMonitoringDistance)
-        await MainActor.run {
+        // Read before `recordRegistration` overwrites it — the diff decides which registrations are new.
+        let previouslyRegisteredIds = await storage.getRegisteredBusinessIds()
```
Shahroz16: Could we make sign-out win over an in-flight refresh around this await? `performRemoteRefresh` only checks the user at line 260, and the local path has no pre-registration recheck. If sign-out lands during this storage read, its `reset()` can lose the same gate and return `.alreadyInProgress`; there is no retry, while the refresh can resume, register regions, and persist state for the signed-out user. I think the safer invariant is for reset to cancel or queue behind the refresh, or for the refresh to detect the changed user and perform cleanup before releasing the gate. Could we add a race test where reset fires while this read is suspended?

Shahroz16 (commented):
I found three remaining lifecycle edges in the new initial enter and identity paths.

Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator+InitialEnter.swift
mrehan27: Added the clamp in `2b187d0a`. In practice `maximumRegionMonitoringDistance` sits well above the 10 km server radius cap, so this isn't a case we expect to hit — but since Apple documents no floor for it and it's a device value we can't raise, the inside check now clamps to the effective registered radius (`distanceTo(anchor) <= min(region.radius, maxMonitoringRadius)`, sourced from the accepted registration) so the decision can't disagree with the circle the OS actually monitors. No-op when `radius <= cap`, i.e. the normal case. Covered by `refresh_givenDeviceInsideConfiguredRadiusButOutsideOsCap` (100 km fence, 200 m cap, device ~5 km out).

Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator+InitialEnter.swift
mrehan27: Fixed in `2b187d0a`. Moved the `currentUserId == expectedUserId` recheck inside the delivery loop, so a slow send that spans a sign-out/switch stops the remaining enters instead of letting the tracker stamp them to the new profile. Test `refresh_givenUserChangesMidInitialEnterBatch` changes the user after the first send and asserts exactly one enter is delivered — both would fire without the per-iteration guard.

Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator.swift
mrehan27: Fixed in `2b187d0a`. Rather than have `reset()` win the gate, the refresh self-corrects: after register + persist (still holding the gate) it re-checks the user via `cleanupIfUserChanged` and, if it changed, undoes the stale-user state (`stopMonitoringAll` + `clearUserScopedState`) before returning — so a sign-out that raced the register window isn't left half-applied after its `reset()` was dropped as `.alreadyInProgress`. Applied to both the remote and local paths (local had no post-fetch recheck). Race test `refresh_givenSignOutDuringRegisterPersist` signs out during the first post-check storage write and asserts registration + last-sync are cleared and no enters fire. Kept a single race test on the remote path since the local path calls the same shared helper — happy to add a local-specific one if you'd prefer it pinned separately.

Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator.swift
```diff
@@ -280,6 +286,16 @@ final class GeofenceSyncCoordinatorImpl: GeofenceSyncCoordinator, @unchecked Sen
...
+        // The post-fetch check above only covers the fetch window; a sign-out/switch can still land
+        // during the register + persist awaits, and reset() can't run while we hold the gate. Undo.
+        if await cleanupIfUserChanged(expectedUserId: expectedUserId) { return .success(()) }
+        emitInitialEnters(
+            candidates: nearest,
+            osRegistration: osRegistration,
+            previouslyRegisteredIds: previouslyRegisteredIds,
+            expectedUserId: expectedUserId,
+            anchor: anchor
+        )
```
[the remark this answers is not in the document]
mrehan27: Real edge, but leaving it as-is. It needs a sign-out AND a same-user sign-in to both land inside a single refresh's gate window — two auth events for the same profile within ~1-2s. When hit, the impact is bounded: one missed synthetic "already-inside" enter for that session (real crossings are unaffected) and cooldowns that expire on their own within the dedup window; there's no cross-user attribution issue since the id is unchanged.

A correct fix is more than it looks: a coordinator-local "reset pending" flag still has an irreducible set/read race between `reset()` and the refresh releasing the gate, so it only shrinks the window. Doing it airtight means an identity generation counter on `BackgroundDeliveryContextStore`, captured per refresh — shared-infra surface I don't think this low-probability, bounded, mostly self-healing case justifies. Happy to revisit if it shows up in the field.

Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor.swift
mrehan27: Update: the deferral shipped in `5f958ff9`, but field testing showed it disabling delivery, so it's removed in `2faf01ce`. Two failure modes from device container logs:

1. On a prewarmed/background-state launch, `isProtectedDataAvailable` read `false` while the device was unlocked, and `protectedDataDidBecomeAvailableNotification` never fired — it only posts on a real lock→unlock transition. The wait never resolved, so that process ran with no events consumer at all; even foreground catch-up was dead.
2. When a crossing wakes the app while the device is locked — the primary delivery scenario — protected data is unavailable by definition, so the deferral blocked monitor creation for the entire wake window and the event was lost until the next manual open.

The header note is real, but following it unconditionally conflicts with CLMonitor's own background contract: the system wakes the app for events while the device is locked, which is exactly when the flag is false. The note guards the persistence read before first unlock — a launch that can't occur for location events (iOS doesn't background-launch apps for location before first unlock, and a manual launch implies unlocked). If a monitor were ever created against unreadable storage, it comes up with zero conditions and the reconcile path (`setOnReconciled` → re-run `wireMonitor`) re-registers from cache. At-rest protection is unchanged — the storage files keep `completeUntilFirstUserAuthentication`; the deferral only gated read timing, and the rationale now lives in a comment at the creation site.

After removal, the terminated-app walk passes end-to-end: a crossing relaunched the swiped-up app in the background, the consumer attached ~15ms after bootstrap, the transition delivered ~1.5s after the wake, and four later background crossings all delivered.

Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator.swift
```diff
@@ -288,15 +327,18 @@ final class GeofenceSyncCoordinatorImpl: GeofenceSyncCoordinator, @unchecked Sen
...
+    func performLocalRefresh(
+        expectedUserId: String,
         latitude: Double,
         longitude: Double,
         config: GeofenceConfig,
         cachedRegions: [Geofence]
     ) async -> Result<Void, GeofenceSyncError> {
         let anchor = LocationData(latitude: latitude, longitude: longitude)
         let nearest = distanceFilter.nearest(cachedRegions, to: anchor, limit: config.maxBusinessGeofences, maxDistance: config.maxMonitoringDistance)
-        await MainActor.run {
```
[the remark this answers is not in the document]
mrehan27: Same family as the re-sign finding above (see the reply there): any check-then-act identity guard leaves a residual window, and each additional re-check just relocates it. In this one, harm requires a genuine boundary crossing to be delivered between `registerWithOsSync` and the exit cleanup a few storage writes later (registration itself emits nothing — classic is silent, CLMonitor's initial evaluation is baseline-suppressed), with a sign-in landing in the same milliseconds. Events in that window are stamped with the user current at capture time, which is the intended attribution; the stale piece is only the fence ranking, undone by the exit cleanup and the new user's own refresh. The airtight fix is an identity generation counter on the shared context store — tracked as the revisit-if-seen-in-field follow-up rather than another narrowing re-check here.

Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator+Refresh.swift
```diff
@@ -0,0 +1,117 @@
...
+    /// can't run while the gate is held — it returns `.alreadyInProgress` and is dropped. Runs at
+    /// each gated method's single exit (gate still held) to undo the stale user's state, whichever
+    /// path the body took. Initial-enter delivery re-checks the user per iteration on its own.
+    /// Returns true when it cleaned, so the caller can trigger the current-user retry.
+    func cleanupIfUserChanged(expectedUserId: String?) async -> Bool {
+        guard let expectedUserId, contextStore.currentUserId != expectedUserId else { return false }
+        await MainActor.run { monitor.stopMonitoringAll() }
+        await storage.clearUserScopedState()
+        logger.geofenceSyncSupersededByUserChange()
+        return true
```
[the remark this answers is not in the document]
mrehan27: The interleaving is real, but it's the same-id re-sign case documented as won't-fix above (https://github.com/customerio/customerio-ios/pull/1158#discussion_r3601883045): by the time the exit check runs, `currentUserId` has returned to the captured id, so the intermediate sign-out is unobservable via id comparison — distinguishing "still A" from "A → nil → A" needs an identity generation counter in Common, which is out of scope for this PR.

Impact is narrower than the user-switch case the exit cleanup fixes: the surviving state belongs to the same user who is current again, so monitoring, attribution, and storage are all correct. The only loss is the synthetic enter-when-inside event a sign-out/sign-in cycle would re-emit for regions the device is already inside — real crossings and subsequent refreshes behave normally. Given the window (sign-out + re-sign-in as the same id inside one in-flight refresh) and the self-recovery, accepting this until the generation counter exists.

Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor.swift
```diff
@@ -0,0 +1,390 @@
...
+        }
+    }
+
+    private func handle(event: CLMonitor.Event) async {
+        // Hold events until the bootstrap binds `onTransition`: processing earlier would advance the
+        // persisted dedup baseline and then drop the delivery on the nil handler, suppressing the
+        // later re-emission of the same state. New arrivals queue behind any backlog so
+        // per-condition order holds. Capped against a process that never binds — dropping oldest is
+        // safe because CLMonitor re-emits current state.
+        if onTransition == nil || !pendingEvents.isEmpty {
```
Shahroz16: Could we include `isDrainingPendingEvents` in this condition? The drain removes its last queued event before awaiting `process`, so an event arriving during that await sees an empty queue and processes alongside the backlog. Two rapid events for the same condition can then reach `onTransition` out of order, even though the code intends new arrivals to stay behind the backlog. Queueing while the drain is active should preserve the FIFO invariant.

Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor.swift
```diff
@@ -0,0 +1,390 @@
...
+
+    var osMonitoredRegionIdentifiers: Set<String> {
+        knownConditionIdentifiers
+    }
+
+    func adoptExistingRegions(matching identifiers: Set<String>) {
+        let adopted = identifiers.intersection(knownConditionIdentifiers)
+        guard !adopted.isEmpty else { return }
+        ownedRegionIdentifiers.formUnion(adopted)
+        rearmConditions(adopted)
```
Shahroz16, line 234: Since the CLMonitor path now removes and re-adds adopted conditions, could we update the `adoptExistingRegions` protocol description and the bootstrap comments/tests that still say adoption happens without re-registering? The classic implementation still adopts in place, but this implementation deliberately does not, so the current contract is misleading when reasoning about launch behavior.

Shahroz16 (approved):
I found one remaining sign-out cleanup path. The radius clamp and per-iteration identity handling from the previous review are addressed, but reset contention can still leave the signed-out user's state behind.

Approving regardless because, It is practical, but probably not frequent in normal usage, would still encourage to harden it.

Sources/LocationGeofence/Storage/GeofenceStorage.swift
```diff
@@ -93,18 +93,85 @@ actor GeofenceStorage {
...
+    func recordMonitorEvent(_ transition: GeofenceTransition, forIdentifier identifier: String) -> GeofenceMonitorEventOutcome {
+        var state = loadFromDisk() ?? GeofenceState()
+        var records = state.monitorRegionRecords ?? [:]
+        guard var record = records[identifier] else {
+            // No registration record (condition predates this bookkeeping). Establish the baseline
+            // without delivering — mirrors classic registration, which is silent about the initial state.
+            records[identifier] = MonitorRegionRecord(lastState: transition, transitionTypes: [.enter, .exit])
+            state.monitorRegionRecords = records
+            saveToDisk(state)
+            return .suppressedNoBaseline
```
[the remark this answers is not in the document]
mrehan27: The permissive seed is a deliberate recovery posture, not a filter bypass. The sign-out path can't complete: events only reach `recordMonitorEvent` for owned conditions, reset drops ownership together with the records, and anything squeezing through the same-actor-hop window is then dropped at the tracker's identified-only gate (the user is signed out). A genuinely lost record self-heals at the next sync, which rewrites `transitionTypes`; until then the exposure is one transition type of a real crossing, still identity- and cooldown-gated. Seeding restrictively instead would fail closed — suppressing wanted crossings indefinitely when sync can't run (e.g. offline) — a worse trade than briefly over-delivering. Keeping as-is.

PR #1161, opened 2026-07-20
chore(geofence): flush queued transition events on each new crossing

mrehan27:
Retries queued geofence transition events on every new crossing instead of waiting for the next launch or identify. Contained to `GeofenceEventTracker.trackTransition`/`flushPending` plus one read-only accessor on `BackgroundDeliveryContextStore`.

Ticket: [MBL-2105](https://linear.app/customerio/issue/MBL-2105)

## Why
A row that fails its fresh HTTP send (offline, timeout) stays in the pending store, but the store was only flushed on module init, cold-wake bootstrap, and identify. A process that stays alive in the background and keeps receiving crossings never retried its backlog — each crossing delivered only its own rows. Field-observed on a real drive: an offline background wake queued a crossing at 18:39, and later crossings at 19:00 in the same process delivered themselves while the 18:39 rows sat untouched until the next relaunch. Android already converges per crossing (WorkManager network-aware retry); this closes that gap.

## Approach
`trackTransition` persists and sends the current crossing **first**: the monitor's dedup baseline has already advanced by the time the tracker sees a crossing, so a process suspended during a slow backlog replay must not be able to lose an un-persisted crossing (it would never re-emit); attribution (user, timestamp) is captured at crossing time for the same reason. The backlog flush then runs on every crossing — anonymous and cooldown-suppressed ones included, since backlog rows are self-contained (stamped userId) — **excluding the rows just written**, so a failed fresh send stays on disk for the next trigger instead of re-attempting on the network that just failed. The flush routes by who owns delivery best:

- **DataPipeline live in-process** (`hasLiveCdpApiKeyProvider`) → EventBus: durable queue, its own retry cadence, batches with other traffic.
- **Not live but persisted API key** (wrapper cold-wake) → direct HTTP inside the background-time window, so the backlog ships during the wake instead of deferring to the next JS/Dart launch.
- **Neither** → EventBus persistent event, replayed by DataPipeline at the next initialized launch.

No fallback-on-failure between channels: a failed row stays queued for the next flush. Duplicate handoffs are absorbed by the active-delivery claims and server-side `transitionId` dedup (at-least-once contract, unchanged).

## Testing
- Tracker suite extended to 38 tests: routing per channel (live provider / persisted-key cold-wake / neither), backlog replay on gated crossings, stamped-userId over HTTP, failed-row retention, concurrent flush dedup, and fresh-row durability while a blocked backlog replay is mid-flight (attempted exactly once — the flush excludes it). Full `LocationGeofenceTests` + `CommonTests` green.
- `api-docs/internal/CioInternalCommon.api` regenerated for the new public accessor.

## Notes
- Base is `feature/geofence-on-device` (independent of #1158; umbrella feature→main PR is #1130).
- Known accepted trade: an initialized-but-backgrounded backlog rides DataPipeline's flush cadence rather than sending immediately — the backlog is already late by definition, and the fresh path stays direct HTTP for real-time delivery.

---

> [!NOTE]
> **Medium Risk**
> Changes background geofence delivery ordering and flush routing; behavior is localized to `GeofenceEventTracker` with broad new tests, but mis-routing could affect metric delivery in cold-wake vs foreground SDK states.
> 
> **Overview**
> Fixes **stuck pending geofence metrics** in long-lived background processes by **replaying the backlog after every crossing**, not only on init, identify, or cold-wake bootstrap.
> 
> **`trackTransition`** now **persists and sends the current crossing first** (`deliverCurrentCrossing`), then calls **`flushPending(excluding:)`** so a slow backlog replay cannot block durability of the new crossing and a just-failed fresh send is not immediately retried on the same flush.
> 
> **`flushPending`** chooses the replay channel: **EventBus when DataPipeline is live** (`hasLiveCdpApiKeyProvider` on `BackgroundDeliveryContextStore`); **direct HTTP with a background task** when only a **persisted CDP key** exists (wrapper cold-wake); otherwise **EventBus** for disk persistence until the next init. HTTP failures keep rows queued without falling back to EventBus.
> 
> Tests cover routing, stamped user attribution, anonymous crossings still flushing backlog, and ordering under slow replay.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b292e736b1ee3fad892937e09cd54acd1c2780fd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Sources/LocationGeofence/Event/GeofenceEventTracker.swift
+++ b/Sources/LocationGeofence/Event/GeofenceEventTracker.swift
@@ -9,9 +9,11 @@ import Foundation
 /// transitionId:
 /// - Fresh transition → direct HTTP (`deliverFresh`): works before DataPipeline is up (cold-wake);
 ///   failure retains the row for the next flush.
-/// - Replay (`flushPending` → `deliverViaEventBus`) → EventBus → DataPipeline's durable queue.
+/// - Replay (`flushPending`) → EventBus when DataPipeline is live, else direct HTTP off the
+///   persisted context; routing rationale on the method.
 ///
-/// `flushPending()` runs on module init, cold-wake bootstrap, and `ProfileIdentifiedEvent`.
+/// `flushPending()` runs on module init, cold-wake bootstrap, `ProfileIdentifiedEvent`, and at the
+/// start of every tracked transition — so a backlog retries whenever a crossing wakes the device.
 /// Concurrent deliveries of the same row are deduped via the in-memory active-delivery set.
 ///
 /// `@unchecked Sendable`: all stored properties are `let`; mutable state is wrapped in `Synchronized`.
@@ -58,6 +60,10 @@ final class GeofenceEventTracker: @unchecked Sendable {
         geofenceId: String,
         transition: GeofenceTransition
     ) async {
+        // Retry any backlog first: a crossing proves the device is awake, and queued rows are
+        // self-contained (stamped userId) — this crossing's identity/cooldown gates don't apply.
+        await flushPending()
+
         // Identified-only: the backend rejects anonymous geofence tracks, so drop before cooldown or
         // persist. Snapshot the userId so a later sign-out/sign-in can't reattribute the row.
         guard let stampedUserId = contextStore.currentUserId, !stampedUserId.isEmpty else {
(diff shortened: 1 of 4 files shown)


Sources/LocationGeofence/Event/GeofenceEventTracker.swift
```diff
@@ -58,6 +60,10 @@ final class GeofenceEventTracker: @unchecked Sendable {
         geofenceId: String,
         transition: GeofenceTransition
     ) async {
+        // Retry any backlog first: a crossing proves the device is awake, and queued rows are
+        // self-contained (stamped userId) — this crossing's identity/cooldown gates don't apply.
+        await flushPending()
```
Shahroz16: Could we make the current crossing durable before waiting for the backlog here? This requires an existing backlog plus a slow cold-wake replay, so it will not affect every transition. In that scenario, though, replay can spend up to the 20-second resource timeout on older rows while the background assertion is best effort. If iOS suspends or terminates the process before this await returns, the new crossing has not reached pendingStore.append and is permanently lost. Capturing the current user and transition time, then persisting the fresh rows first, would also keep capture-time attribution accurate. Could we reorder this while retaining direct delivery, and add a test that blocks backlog delivery and confirms the fresh row is already queued?

Sources/LocationGeofence/Event/GeofenceEventTracker.swift
```diff
@@ -130,12 +136,28 @@ final class GeofenceEventTracker: @unchecked Sendable {
...
-            await deliverViaEventBus(metric: metric)
+        let metrics = await pendingStore.loadAll()
+        guard !metrics.isEmpty else { return }
+        let persistedKey = contextStore.currentCdpApiKey
+        if !contextStore.hasLiveCdpApiKeyProvider, let persistedKey, !persistedKey.isEmpty {
+            // Same assertion + concurrent fan-out rationale as the fresh send in `trackTransition`.
+            await backgroundTaskRunner.withBackgroundTime { [self] in
+                await withTaskGroup(of: Void.self) { group in
+                    for metric in metrics {
+                        group.addTask { await self.deliverFresh(metric: metric) }
```
Shahroz16, line 169: Nonblocking: could we bound the replay pressure here? This matters mainly when a device stays offline and continues crossing fences. The pending store can return 100 rows, and each later crossing schedules the full set again; activeDeliveryKeys only deduplicates overlapping calls. A small concurrency limit or retry gate could reduce repeated work. I do not see this as an approval blocker once the current crossing is durable.

Shahroz16 (approved):

Sources/LocationGeofence/Event/GeofenceEventTracker.swift
mrehan27: Good catch — the baseline has already advanced by the time the tracker sees the crossing, so a suspension mid-replay would have lost it permanently. Addressed in b292e736: the crossing's rows are persisted (with user and timestamp captured at crossing time) and sent first, and the backlog flush runs after, excluding the rows just written — so a failed fresh send stays on disk for the next trigger instead of re-attempting on the network that just failed. Added the test you suggested: backlog delivery blocked mid-flight, fresh row asserted durable on disk, and attempted exactly once.

Sources/LocationGeofence/Event/GeofenceEventTracker.swift
mrehan27: Agreed it's real for a device that stays offline across many crossings, and agreed it isn't this PR's scope — bounding replay (concurrency cap and/or a retry gate) belongs with the background-delivery tuning follow-up (HTTP timeout adjustments, tracked alongside the #1149 work), where the same knobs apply to both the fresh fan-out and the replay. Deferring to that.

PR #1162, opened 2026-07-20
chore(geofence): fix event-loss races and cross-user cooldown carryover

mrehan27:
Fixes four event-loss and state-carryover defects surfaced by the #1130 review pass and a field-diagnosed missed transition. All are contained to `LocationGeofence`; one commit per fix.

Ticket: [MBL-2105](https://linear.app/customerio/issue/MBL-2105/ios-harden-geofence-background-delivery-on-ios-17)

## Fixes

**1. Per-user cooldown key** (`userId:geofenceId:transition`) — a fast re-login can skip the async sign-out cleanup, leaving the previous user's cooldown timestamps in shared storage and suppressing the new session's first crossings. Registration/freshness state deliberately stays user-agnostic (geofences are workspace-scoped); only the cooldown needed scoping.

**2. Bootstrap adopt/re-register skipped when the identified user changed mid-read** — bootstrap snapshots the userId, then awaits storage reads; a sign-out landing in that window ran the decision against a torn-down state: the restore re-registered regions the reset had just removed, and the adopt branch's re-arm re-added conditions after the reset's queued removes — with no later reset to stop either (the guard fronts both branches per review). Sign-in and user-switch paths are unaffected: `applyCachedRegistration` already skips a nil user, and a switch converges via reset-clears-state → `hasUnregisteredCache` → local re-register on the new user's identify refresh.

**3. Classic (≤17) cold-wake events buffered until the bootstrap binds the handler** — on a cold wake, any DI path constructing the monitor (e.g. the identify-replay refresh resolving the coordinator) puts the `CLLocationManager` delegate live before bind/adopt run; a queued crossing delivered in that window was dropped with no re-emission — losing the very crossing the OS woke the app for. The buffer is the classic twin of the CLMonitor path's pre-bind queue (drain-order guard included); ownership is checked at drain time, after adopt has populated it, so buffered host-app-region events are still filtered.

**4. CLMonitor events delivered when owned at observation time** — the post-await ownership re-check in `process()` dropped a transition whose dedup baseline had already advanced whenever a sync's stop-all + re-add swap raced the storage hop. The re-registration then preserved the advanced baseline and re-armed CLMonitor as agreeing, so the crossing was lost permanently. Field-diagnosed from a real missed enter: the state file showed the baseline traveling exit→enter→exit with no enter ever reaching the tracker — only this path can do that. A region genuinely removed in that window now delivers one last identity/cooldown-gated event instead, which is the right side of the trade.

## Testing
- Unit: cooldown scoping has a behavioral test (new user's crossing delivers mid-window, same user stays suppressed; fails against the old key); a latent vacuous assertion in the persist-fail test was fixed alongside. Full `LocationGeofenceTests` green at 260. Fixes 2–4 are structurally untestable at unit level (identity mutation mid-actor-read; simulator auth can't seed monitor ownership) — the monitors keep their established device-validated posture.
- Field (fix 3): forced-classic build, force-quit app, real drive — the first cold wake delivered two crossings **9ms before bind/adopt finished**; both were buffered, drained in FIFO order after adopt, and delivered within ~1s. iOS's duplicate re-callbacks a second later were cooldown-suppressed. Pre-fix, both crossings were silently lost.
- Field (fix 4): the failing case is reproduced from a real drive's state file (above); a confirmation drive on the fixed build is in progress and results will be posted here.

## Notes
- Deliberately out of scope, tracked separately: the privacy-manifest correction, and a sweep for classic regions orphaned by an iOS 17→18 upgrade (battery-only impact; id-collision question deserves its own look).

---

> [!NOTE]
> **Medium Risk**
> Changes race-sensitive background geofence delivery and identity timing; fixes reduce lost events but CLMonitor may emit an extra transition when ownership races sync teardown.
> 
> **Overview**
> Hardens **LocationGeofence** background delivery by fixing four defects that dropped or mis-attributed transitions.
> 
> **Duplicate suppression** now keys cooldowns as `userId:geofenceId:transition` so a fast re-login is not blocked by the prior user’s recent crossing when sign-out cleanup did not run.
> 
> **Bootstrap** skips adopt/re-register if `currentUserId` changes during the async storage reads after the snapshot, avoiding resurrecting regions torn down by sign-out; the next identify refresh registers instead.
> 
> **Classic `CLLocationManager` monitor** buffers region callbacks until `onTransition` is bound (and drains with ownership checks), matching the CLMonitor pre-bind queue so cold-wake crossings before bind/adopt are not lost.
> 
> **CLMonitor `process`** no longer re-checks ownership after `recordMonitorEvent`’s await, so a genuine crossing is not permanently dropped when sync’s stop-all/re-add races the storage hop (may deliver one last gated event if a region was removed in that window).
> 
> Adds a unit test for per-user cooldown behavior and updates cooldown key expectations in existing tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3f1d575c0d0eb585c7088de78bb947ce779b0117. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Sources/LocationGeofence/Event/GeofenceEventTracker.swift
+++ b/Sources/LocationGeofence/Event/GeofenceEventTracker.swift
@@ -84,7 +84,9 @@ final class GeofenceEventTracker: @unchecked Sendable {
             return []
         }
 
-        let cooldownKey = "\(geofenceId):\(transition.rawValue)"
+        // Cooldown is scoped per user: a re-login must not be suppressed by the previous user's
+        // recent transition, even when a fast re-login skips the async sign-out cleanup.
+        let cooldownKey = "\(stampedUserId):\(geofenceId):\(transition.rawValue)"
         let now = dateUtil.now
         // Cached config wins when present so a workspace can tune the dedup window without
         // an SDK release; constructor default applies otherwise.
--- a/Sources/LocationGeofence/GeofenceBootstrap.swift
+++ b/Sources/LocationGeofence/GeofenceBootstrap.swift
@@ -81,13 +81,13 @@ enum GeofenceBootstrap {
         // regions dropped, e.g. a monitoring failure or only the trigger surviving) falls through to
         // re-register so the missing business geofences come back rather than staying unmonitored
         // until the next refresh.
-        if !expectedOwnedRegions.isEmpty, expectedOwnedRegions.isSubset(of: monitor.osMonitoredRegionIdentifiers) {
-            monitor.adoptExistingRegions(matching: expectedOwnedRegions)
-        } else if di.backgroundDeliveryContextStore.currentUserId != userId {
-            // Identity changed during the reads above: registering now could resurrect regions a
-            // sign-out reset just tore down, with no later reset to stop them. The next
-            // identify-driven refresh registers instead.
+        if di.backgroundDeliveryContextStore.currentUserId != userId {
+            // Identity changed during the reads above: adopting or registering now could resurrect
+            // regions a sign-out reset just tore down (adopt's FIFO'd re-adds land after the
+            // reset's queued removes). The next identify-driven refresh registers instead.
             di.logger.geofenceSyncSkipped(reason: "identified user changed during bootstrap")
+        } else if !expectedOwnedRegions.isEmpty, expectedOwnedRegions.isSubset(of: monitor.osMonitoredRegionIdentifiers) {
+            monitor.adoptExistingRegions(matching: expectedOwnedRegions)
         } else {
             // First launch after install, the OS dropped our regions (e.g. permission revoked then
             // re-granted, which clears `monitoredRegions`), or a partial drop. Register fresh from cache.
--- a/Sources/LocationGeofence/GeofenceBootstrap.swift
+++ b/Sources/LocationGeofence/GeofenceBootstrap.swift
@@ -83,6 +83,11 @@ enum GeofenceBootstrap {
         // until the next refresh.
         if !expectedOwnedRegions.isEmpty, expectedOwnedRegions.isSubset(of: monitor.osMonitoredRegionIdentifiers) {
             monitor.adoptExistingRegions(matching: expectedOwnedRegions)
+        } else if di.backgroundDeliveryContextStore.currentUserId != userId {
+            // Identity changed during the reads above: registering now could resurrect regions a
+            // sign-out reset just tore down, with no later reset to stop them. The next
+            // identify-driven refresh registers instead.
+            di.logger.geofenceSyncSkipped(reason: "identified user changed during bootstrap")
         } else {
             // First launch after install, the OS dropped our regions (e.g. permission revoked then
             // re-granted, which clears `monitoredRegions`), or a partial drop. Register fresh from cache.
--- a/Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor.swift
+++ b/Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor.swift
@@ -207,9 +207,9 @@ final class CLMonitorGeofenceMonitor: NSObject, GeofenceRegionMonitoring, @preco
             return
         }
         guard case .deliver = await storage.recordMonitorEvent(transition, forIdentifier: identifier) else { return }
-        // The await hopped off the main actor; re-check ownership so a region removed during that
-        // window isn't delivered (parity with the classic delegate's synchronous ownership check).
-        guard ownedRegionIdentifiers.contains(identifier) else { return }
+        // No ownership re-check after the await: the baseline already advanced, so dropping here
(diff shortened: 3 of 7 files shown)


Sources/LocationGeofence/GeofenceBootstrap.swift
```diff
@@ -83,6 +83,11 @@ enum GeofenceBootstrap {
         // until the next refresh.
         if !expectedOwnedRegions.isEmpty, expectedOwnedRegions.isSubset(of: monitor.osMonitoredRegionIdentifiers) {
             monitor.adoptExistingRegions(matching: expectedOwnedRegions)
+        } else if di.backgroundDeliveryContextStore.currentUserId != userId {
```
Shahroz16: Non-blocking: could we move this identity check ahead of the adopt branch as well? We snapshot `userId` before awaiting `getRegisteredBusinessIds()`, but if sign-out lands during that await and the OS still reports the complete set, the adopt branch runs without reaching this check. On `CLMonitor`, `stopMonitoringAll()` clears ownership immediately but queues the live condition and mirror removal, so bootstrap can briefly reclaim the previous set while reset is in flight. Hoisting the check would make the guard cover both outcomes and keep the ownership state consistent.

Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor.swift
```diff
@@ -207,9 +207,9 @@ final class CLMonitorGeofenceMonitor: NSObject, GeofenceRegionMonitoring, @preco
             return
         }
         guard case .deliver = await storage.recordMonitorEvent(transition, forIdentifier: identifier) else { return }
-        // The await hopped off the main actor; re-check ownership so a region removed during that
-        // window isn't delivered (parity with the classic delegate's synchronous ownership check).
-        guard ownedRegionIdentifiers.contains(identifier) else { return }
+        // No ownership re-check after the await: the baseline already advanced, so dropping here
```
Shahroz16, line 210: One follow-up I think is worth tracking: now that we intentionally deliver after this storage await even if ownership was removed, the transition still gets its user identity later in `GeofenceEventTracker`. If A signs out and B is identified while `recordMonitorEvent` is doing its file read and write, the handler can run afterward and stamp the crossing to B. The window is very small and mostly pre-existing, so I would not block this PR on it, but carrying the observed identity through, or dropping when it changed, would close the account-attribution gap for both monitor paths.

Sources/LocationGeofence/Event/GeofenceEventTracker.swift
```diff
@@ -84,7 +84,9 @@ final class GeofenceEventTracker: @unchecked Sendable {
             return []
         }
 
-        let cooldownKey = "\(geofenceId):\(transition.rawValue)"
+        // Cooldown is scoped per user: a re-login must not be suppressed by the previous user's
+        // recent transition, even when a fast re-login skips the async sign-out cleanup.
+        let cooldownKey = "\(stampedUserId):\(geofenceId):\(transition.rawValue)"
```
Shahroz16, line 89: Small cleanup: could we update the cooldown-key comments in `GeofenceConfig` and `GeofenceState` too? Both still describe `geofenceId:transitionType`, while the persisted key now includes the user ID.

Shahroz16 (approved):

Sources/LocationGeofence/GeofenceBootstrap.swift
mrehan27: Good catch — and it invalidated my earlier reasoning that adopt was reclaim-only: since the re-arm, adopt re-adds conditions through the FIFO pipeline, and those re-adds land after a reset's queued removes, so it can genuinely resurrect the torn-down set. Hoisted the guard ahead of the whole adopt/re-register decision in 3f1d575c.

Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor.swift
mrehan27: Agreed on both the gap and the framing — deferring to a follow-up ticket. The window predates this PR on both monitor paths and is an uncorrelated ms-scale coincidence (crossings process in the background mid-route; account switches happen foregrounded, usually stationary — unlike the swap race, nothing ties the two together), with impact bounded to one mislabeled event and the sign-out-only case already dropped at the tracker's identity gate. The clean fix — capturing observed identity at the monitor and carrying it through the handler chain (or dropping on mismatch) — is a cross-cutting signature change with an Android-parity dimension, so it deserves its own review, paired with the direct-identify-switch staleness work that shares the attribution question.

Sources/LocationGeofence/Event/GeofenceEventTracker.swift
mrehan27: Done in 3f1d575c — both GeofenceConfig and GeofenceState now document the userId-scoped key.

PR #1165, opened 2026-07-21
chore: enable Live Activities

mahmoud-elmorabea:
Live Notification feature

---

> [!NOTE]
> **High Risk**
> Large new surface area touching ActivityKit lifecycle, push tokens, identity-gated CDP events, and logout/reset behavior; incorrect observation or metric attribution would affect production messaging analytics.
> 
> **Overview**
> Adds **Live Activities** (Live Notifications) as a first-class Customer.io iOS SDK capability: three distributable pieces (**`LiveActivities`**, **`LiveActivities_Attributes`**, **`LiveActivities_Templates`**) via SPM and CocoaPods, wired into the umbrella **`CustomerIO`** pod and deploy workflows.
> 
> The **`LiveActivities`** module registers through **`SDKConfigBuilder.addModule(LiveActivitiesModule(...))`** and exposes **`CustomerIO.liveActivities`** for local **`start` / `update` / `end`**, **`adopt`**, ActivityKit token/lifecycle observation, delivery **`opened`** / **`delivered`** metrics, and **`handleWidgetUrl`** for widget taps. Persistence keys and reporting distinguish user dismissals from SDK-initiated ends and buffer opens before init.
> 
> **APN-UIKit** gains a widget extension (custom delivery UI plus branded SDK templates), dashboard demo screen, and deep-link routing. **CocoaPods-FCM** adds a template-only widget extension and dashboard start/end controls. SwiftLint disables **`multiple_closures_with_trailing_closure`** for SwiftUI-style APIs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f86b7775db5467f19a790c612935e06a577ed20d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Package.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version:5.5
+// swift-tools-version:5.10
```
Shahroz16: Was this Swift tools jump to 5.10 intentional? It changes whether the entire package can be resolved before a customer selects any Live Activities product. The SDK still advertises Swift 5.3, and the minimum-version checklist calls for coordinated manifest, CI, README and podspec changes; I couldn't find anything in these new targets that requires a 5.10 manifest. Could we keep the previous tools version, or make the support bump explicit across those surfaces?

CustomerIOLiveActivities.podspec
```diff
@@ -0,0 +1,26 @@
+Pod::Spec.new do |spec|
+  spec.name         = "CustomerIOLiveActivities"
+  spec.version      = "4.6.0" # Don't modify this line - it's automatically updated
```
Shahroz16: Could we align all three new Live Activities podspecs with the repository's current 4.6.1 before this lands? They are 4.6.0 and pin CustomerIOCommon 4.6.0, while DataPipelines and the rest of this exact branch pin Common 4.6.1. A fresh wrapper pod install now fails with those two exact constraints, so the current integration branch cannot be consumed even before release. The same change is needed in Attributes and Templates.

Sources/LiveActivities/LiveActivitiesInstance.swift
```diff
@@ -0,0 +1,185 @@
...
+    #endif
+
+    @discardableResult
+    func handleWidgetUrl(_ url: URL) -> URL? {
+        // Hosts are told to route *every* opened URL through here, so the overwhelmingly common case
+        // is an ordinary app link. Hand those straight back without a diagnostic — there is nothing
+        // Customer.io could have reported, so a log would only be noise on an app's normal deep links.
+        guard let parsed = CioLiveActivityWidgetUrl.parse(url) else { return url }
+
+        // A Customer.io URL did arrive but no module is live, so the `opened` metric is lost. Say why
```
Shahroz16: Could we preserve the opened metric when this URL arrives before module initialization? This is the normal cold-launch order for the React Native and Flutter wrappers in this stack: AppDelegate receives the Live Activity URL before JS/Dart calls CustomerIO.initialize. We return the customer deep link correctly here, but every terminated-app tap loses attribution. Could we buffer the parsed delivery and flush it once the Live Activities module is registered, while still returning the redirect immediately?

Shahroz16 (commented):
I went through the aggregate again at the current head and re-ran the Live Activities package tests. The recent identity/start/end fixes look good, and I don't think the two remaining Cursor claims around adoption or instance ids are reachable in the supported flow. I did find three current integration problems below. These are normal paths in the wrapper stack, so I would fix them before shipping this aggregate.

Sources/LiveActivities/LiveActivitiesModuleState.swift
```diff
@@ -0,0 +1,78 @@
...
+        )
+        implementation.performInitialization()
+
+        lock.lock()
+        instance = implementation
+        lock.unlock()
+
+        for metadata in pendingOpens.drain() {
+            implementation.reportBufferedOpen(metadata: metadata)
+        }
```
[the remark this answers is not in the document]
mahmoud-elmorabea: A single tap will only ever be reported once, if there are multiple report opens, those are multiple taps and that is valid

Shahroz16 (approved):

PR #1166, opened 2026-07-21
fix(live-activities): report start when issued right after identify()

mahmoud-elmorabea:
Fixes the identify()→startLiveActivity() race.

The reporter gates lifecycle events on the SDK's synchronous identified flag (backed by the existing `DataPipelineTracking` reverse-dependency, updated on the caller thread during identify()/clearIdentify()) instead of the async event-bus-fed identity cache. A startLiveActivity() issued immediately after identify() is now reported rather than dropped by a stale flag. Token registration is unchanged (still event-driven via the registrar).

All 96 `LiveActivitiesTests` pass locally.

---

> [!NOTE]
> **Medium Risk**
> Changes Live Activities analytics gating and logout teardown paths where mis-reported `end` events or dropped `start` events would affect customer data; behavior is covered by new reset tests but touches identity timing and terminal classification.
> 
> **Overview**
> Fixes the **identify() → startLiveActivity()** race by having `LiveActivityReporter` gate lifecycle events on **`sdk.isUserIdentified`** (synchronous `DataPipelineTracking`) instead of the async event-bus `userId` cache, so a start right after `identify()` is no longer dropped.
> 
> **Logout / reset** no longer relies on clearing identity alone to hide force-ended activities: `endAllActivities` now takes a **`prepareLocalEnd`** callback so each activity is marked in `LiveActivityLocalEndTracker` before `.immediate` end, preventing `.dismissed` from being reported as a user swipe. Reset **stops calling `localEndTracker.clearAll()`** so in-flight app-initiated ends still suppress late terminals after account switch. `handleReset()` is exposed for tests; new **`LiveActivitiesResetTests`** cover account-switch and in-flight local-end cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 894c822bf42371af8df0bc5d41d3ebeb38f3725c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Sources/LiveActivities/LiveActivitiesModule.swift
+++ b/Sources/LiveActivities/LiveActivitiesModule.swift
@@ -37,7 +37,7 @@ public final class LiveActivitiesModule {
     private let tokenStorage: LiveActivityTokenStorage
 
     private let identity = LiveActivityIdentity()
-    private let localEndTracker = LiveActivityLocalEndTracker()
+    private let localEndTracker: LiveActivityLocalEndTracker
     private let reporter: LiveActivityReporter
     private let registrar: LiveActivityRegistrar
     private let deliveryTracker: LiveActivityDeliveryTracker
@@ -68,11 +68,13 @@ public final class LiveActivitiesModule {
     init(
         config: LiveActivityConfig,
         sdk: CIOLiveActivitiesSDKProviding,
-        tokenStorage: LiveActivityTokenStorage
+        tokenStorage: LiveActivityTokenStorage,
+        localEndTracker: LiveActivityLocalEndTracker = LiveActivityLocalEndTracker()
     ) {
         self.config = config
         self.sdk = sdk
         self.tokenStorage = tokenStorage
+        self.localEndTracker = localEndTracker
 
         let identity = self.identity
         let reporter = LiveActivityReporter(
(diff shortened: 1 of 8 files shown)


Sources/LiveActivities/LiveActivitiesModule.swift
```diff
@@ -77,7 +77,10 @@ public final class LiveActivitiesModule {
         let identity = self.identity
         let reporter = LiveActivityReporter(
             track: { name, properties in sdk.track(name: name, properties: properties) },
-            currentUserId: { identity.userId },
+            // Gate on the SDK's synchronous identified flag (backed by DataPipelineTracking),
+            // not the async event-bus-fed identity cache, so a start() right after identify()
+            // isn't dropped. Token registration still uses `identity` via the registrar.
+            isUserIdentified: { sdk.isUserIdentified },
```
Shahroz16, line 85: This breaks the reset suppression for a back-to-back logout/login. `handleReset()` sets `identity.userId = nil` before force-ending user A's activities so their terminal callbacks can't report an `end`. After this closure reads `sdk.isUserIdentified` instead, `clearIdentify(); identify("B")` can put B into the synchronous analytics store before the async `ResetEvent` / `handleReset()` runs. The forced `.immediate` ends can then be classified as user dismissals and tracked under B. This could be a normal account-switch sequence?. Could we keep an explicit reporter suppression during reset, or mark these activities as local ends before ending them, and add a regression test for clearIdentify → identify(B) with an active activity?

Sources/LiveActivities/LiveActivitiesModule.swift
```diff
@@ -307,19 +312,33 @@ public final class LiveActivitiesModule {
...
-        localEndTracker.clearAll()
+        // Do NOT clear existing local-end markers here: an activity the app ended via
+        // CIOLiveActivity.end() just before logout is already gone from Activity.activities, so the
+        // force-end loop below won't re-mark it. Dropping its marker would let its late `.dismissed`
+        // terminal be misreported as a user swipe (under the next user). Markers are in-memory,
+        // one-shot, and ULID-keyed, so keeping them is inert for any other activity.
         latestMetadata.wrappedValue = [:]
         for registration in config.registrations {
-            await registration.endAllActivities()
+            await registration.endAllActivities { [tokenStorage, localEndTracker] activityId, attributesInstanceId in
```
Shahroz16, line 335: One broader case: `clearIdentify(); identify(B); start(...)` can create B's activity before the async reset runs, and this loop can then end B's activity too. I checked the exact base and that was already possible before this PR, so this is not blocking here. Like you mentioned earlier, it fits better with a cleaner user-change notification path instead of relying on the async bus. I'm good with tracking it separately and handling it after the Live Activities release.

Shahroz16 (approved):

PR #1171, opened 2026-07-22
feat(live-activities): stop reporting locally-triggered updates

mahmoud-elmorabea:
Local Live Activity updates driven by the host app now render/apply locally but no longer emit a `Live Notification Event` (`eventType=update`) to CDP — only `start` and `end` are reported. Removes the dead `reportUpdate` path and the `update` eventType from the wire contract. Public `update()` API is unchanged.

---

> [!NOTE]
> **Medium Risk**
> Changes the shared CDP wire contract (`eventType` values) and analytics behavior for apps that relied on local update events; public `update()` API is unchanged so integration risk is mainly downstream data pipelines.
> 
> **Overview**
> **Local Live Activity updates no longer emit CDP lifecycle events.** `CIOLiveActivity.update()` still applies content via ActivityKit, but the SDK no longer calls `LiveActivityReporter` or sends `Live Notification Event` with `eventType=update`.
> 
> **Wire contract and reporter are trimmed to `start`/`end` only.** `LiveActivityContract.EventType.update` and `reportUpdate` are removed; docs on `LiveActivitiesModule`, `CIOLiveActivity`, and `LiveActivityObserver` state that only local `start`/`end` are reported (push-driven changes remain unreported). The `reportUpdate_sendsContentStateOnly` test is dropped.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e786a7f1c830f4382fb9cf4843e1ec3dec0bbf06. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Sources/LiveActivities/LiveActivityHandle.swift
+++ b/Sources/LiveActivities/LiveActivityHandle.swift
@@ -7,12 +7,14 @@ import ActivityKit
 
 /// A handle to a Live Activity the SDK is managing locally.
 ///
-/// Returned by `LiveActivitiesModule.start(...)` and `adopt(_:)`. Driving the activity through
-/// this handle's `update`/`end` is what emits the local `Live Notification Event`s — a backend
-/// push that changes or ends the activity is applied by the OS and is never reported.
+/// Returned by `LiveActivitiesModule.start(...)` and `adopt(_:)`. Starting and ending the activity
+/// through this handle is what emits the local `start`/`end` `Live Notification Event`s. A local
+/// `update` is applied to the activity but is intentionally not reported — only `start`/`end` emit a
+/// CDP event. A backend push that changes or ends the activity is applied by the OS and is never
+/// reported either.
 ///
-/// > Important: Ending or updating the underlying `activity` directly (bypassing this handle)
-/// > performs the ActivityKit operation but emits no Customer.io event.
+/// > Important: Ending the underlying `activity` directly (bypassing this handle) performs the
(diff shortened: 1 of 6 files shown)


Sources/LiveActivities/LiveActivityHandle.swift
```diff
@@ -7,12 +7,14 @@ import ActivityKit
...
 /// A handle to a Live Activity the SDK is managing locally.
 ///
-/// Returned by `LiveActivitiesModule.start(...)` and `adopt(_:)`. Driving the activity through
-/// this handle's `update`/`end` is what emits the local `Live Notification Event`s — a backend
-/// push that changes or ends the activity is applied by the OS and is never reported.
+/// Returned by `LiveActivitiesModule.start(...)` and `adopt(_:)`. Starting and ending the activity
+/// through this handle is what emits the local `start`/`end` `Live Notification Event`s. A local
+/// `update` is applied to the activity but is intentionally not reported — only `start`/`end` emit a
+/// CDP event. A backend push that changes or ends the activity is applied by the OS and is never
+/// reported either.
```
Shahroz16, line 14: Could we carry this wording through the remaining API docs too? `LiveActivitiesModule.start` still says the returned handle reports both update and end events, and `adopt` says it can report update/end. The reporter and observer comments also still describe start/update/end reporting. The code itself looks right, so this is not a runtime blocker.

Shahroz16 (approved):

PR #1174, opened 2026-07-23
fix(in-app): stop corrupting escaped regex in page route rules

mahmoud-elmorabea:
[MBL-2192: iOS SDK: dotted in-app page rules never match (cleanPageRule corrupts escaped regex)](https://linear.app/customerio/issue/MBL-2192/ios-sdk-dotted-in-app-page-rules-never-match-cleanpagerule-corrupts)

In-app page route rules arrive as a ready-to-use regex with metacharacters already escaped (e.g. `\.` for a literal dot). `cleanPageRule` was replacing every `\` with `/`, corrupting those escapes so any route containing a `.` could never match and the message was silently dropped. This returns the route rule unmodified and adds regression tests for single- and double-escaped dots (contains + equals forms).

Linear: MBL-2192

---

> [!NOTE]
> **Low Risk**
> Small behavioral fix in page-rule string handling with targeted tests; restores intended regex matching without touching auth or data paths.
> 
> **Overview**
> Fixes in-app messages **silently failing to show** when page rules target routes that include dots (e.g. `dashboard.account.settings`).
> 
> **`cleanPageRule`** no longer replaces backslashes with slashes. Backend-supplied `routeRuleApple` regex already escapes metacharacters (`\.` for literal dots); the old transform turned `\.` into `/.`, breaking matching and dropping those messages.
> 
> Regression tests cover contains/equals patterns with escaped dots, JSON decode from `properties`, and identity checks that `cleanPageRule` preserves backslashes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9e60f67af082c7171f475daba68f70ab7057c9d6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Tests/MessagingInApp/Gist/Managers/Models/MessageTest.swift
+++ b/Tests/MessagingInApp/Gist/Managers/Models/MessageTest.swift
@@ -1,4 +1,5 @@
 @testable import CioMessagingInApp
+import Foundation
 import XCTest
 
 class MessageTest: XCTestCase {
@@ -70,8 +71,9 @@ class MessageTest: XCTestCase {
     // untouched so these escapes survive; corrupting "\" would make any dotted
     // route impossible to match.
     func test_doesPageRuleMatch_givenContainsRuleWithEscapedDots_expectMatchDottedRoute() {
-        // `#"\."#` in a raw string is a backslash followed by a dot, exactly as the
-        // route rule arrives over the wire for a "contains dashboard.account.settings" rule.
+        // `#"\."#` in a raw string is a single backslash followed by a dot: the pattern
+        // as it exists in memory after JSON decoding, for a "contains
+        // dashboard.account.settings" rule.
         let message = Message(pageRule: #"^(.*dashboard\.account\.settings.*)$"#)
         XCTAssertTrue(message.doesPageRuleMatch(route: "dashboard.account.settings"))
     }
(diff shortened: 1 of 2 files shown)


Tests/MessagingInApp/Gist/Managers/Models/MessageTest.swift
```diff
@@ -62,4 +62,36 @@ class MessageTest: XCTestCase {
...
+    }
+
+    // MARK: - cleanPageRule preserves escaped metacharacters
+
+    func test_cleanPageRule_givenEscapedDots_expectDotsNotCorruptedToSlash() {
+        let message = Message(pageRule: #"^(.*dashboard\.account\.settings.*)$"#)
+        XCTAssertEqual(message.cleanPageRule, #"^(.*dashboard\.account\.settings.*)$"#)
+    }
+
+    // Some rules arrive with the backslash itself escaped (two backslashes before
```
Shahroz16: Non-blocking: this comment seems to conflate JSON wire escaping with the string `Message` receives. A wire value containing `\\.` decodes to runtime `\.`, while the raw Swift value used in this test contains two backslashes; ICU interprets that differently and it would not match a literal dot. Could we either reword this as a generic “preserves every backslash” case or, preferably, decode a small JSON payload and assert `doesPageRuleMatch` so the regression covers the actual transport boundary?

Shahroz16 (approved):

PR #1175, opened 2026-07-24
chore(live-activities): open-attribution via cioWidgetUrl

mahmoud-elmorabea:
## What
Adds Live Activity **open attribution** for iOS via a Customer.io tracking `widgetURL`.

- **`CioLiveActivityWidgetUrl`** (in the `LiveActivities_Attributes` package): builds a Customer.io tracking URL from the content-state's `cioMetadata` (`deliveryId` + `deliveryToken` + optional deep-link `redirect`) and parses it back. Exposes a **`.cioWidgetUrl(_:)`** SwiftUI modifier for both `View` (Lock Screen) and `DynamicIsland`.
- **`LiveActivitiesModule.handleWidgetUrl(_:)`**: reports an `opened` metric for the exact delivery the user tapped and returns the deep link to navigate to (`nil` if none); passes non-CIO URLs through unchanged.
- Built-in template widgets (**Segments**, **Countdown Timer**) and the sample **custom Delivery** widget now use `.cioWidgetUrl` on Lock Screen + Dynamic Island; the sample `SceneDelegate` forwards taps to `handleWidgetUrl`.

## Why
A tap on a Live Activity is only observable if its `widgetURL` is set. Rather than matching a customer deep link (which several activities can share — ambiguous attribution), the identity is derived from the **same `ContentState` that renders the activity**, so the open attributes to the exact update shown, with no observer lookup and no deep-link dependency. Works from a cold launch (terminated app).

## Behavior
- **Attribution:** per-delivery; the tap reports `opened` against whichever start/update was displayed.
- **Deep link:** optional. Sent as the **top-level `deep_link`** field on the `/live_notifications/start` request (the backend stamps it into `cioMetadata.deepLink`); omit it and the open still reports (navigation just no-ops).
- **No Info.plist URL scheme and no entitlement** required — the tracking URL is an intra-app `widgetURL` delivered to the containing app.
- iOS **16.1+**.

## Testing
- `CioLiveActivityWidgetUrlTests` — URL build/parse, token-only → no URL, round-trip, non-CIO rejection.
- `LiveActivityHandleWidgetUrlTests` — reports `opened` + returns redirect for a CIO URL; deep-link-only navigates without an open; non-CIO passthrough.
- Verified end-to-end on a physical device (prod workspace): warm tap, tap with deep-link navigation, and cold-launch taps on a fully terminated app (activity started while killed, and killed mid-activity) — all report `opened` for the correct delivery.

---

> [!NOTE]
> **Medium Risk**
> New public SDK surface and changed recommended tap flow affect engagement metrics and deep-link routing; CocoaPods tier changes must stay ordered for releases.
> 
> **Overview**
> Adds **per-delivery Live Activity open attribution** by replacing raw customer deep links in `widgetURL` with Customer.io **`cio-live-activity://`** tracking URLs built from the on-screen `cioMetadata`.
> 
> **Widget / app API:** `CioLiveActivityWidgetUrl` builds and parses those URLs; **`.cioWidgetUrl(_:)`** on Lock Screen and Dynamic Island sets the tap target. **`LiveActivitiesModule.handleWidgetUrl(_:)`** reports `opened` from URL-encoded delivery id/token (no in-memory deep-link matching, including cold launch) and returns the optional `redirect` deep link; other URLs pass through unchanged. Built-in templates and the sample Delivery widget switch from `.widgetURL(deepLink)` to `.cioWidgetUrl`; the sample `SceneDelegate` calls `handleWidgetUrl` instead of `handleDeepLinkOpen` for taps.
> 
> **Distribution:** New CocoaPods (`CustomerIOLiveActivitiesAttributes`, `CustomerIOLiveActivities`, `CustomerIOLiveActivitiesTemplates`), umbrella `CustomerIO` subspecs, and deploy workflow tier/CDN wait lists updated for them.
> 
> **Tests:** URL build/parse edge cases and `handleWidgetUrl` metric/navigation behavior; test helpers gain `CapturingEventBusHandler`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cb8af5b0dab10a4243ec534536ef1375a99dd109. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


CustomerIO.podspec
```diff
@@ -54,4 +54,16 @@ Pod::Spec.new do |spec|
   spec.subspec "Location" do |ss|
     ss.dependency "CustomerIOLocation", "= #{spec.version.to_s}"
   end
+
+  spec.subspec "LiveActivities" do |ss|
+    ss.dependency "CustomerIOLiveActivities", "= #{spec.version.to_s}"
```
Shahroz16, line 59: Could we wire these three new specs into the release workflow as part of this PR? `deploy-sdk.yml` publishes a hard-coded tier list, so right now it skips `CustomerIOLiveActivitiesAttributes`, `CustomerIOLiveActivities`, and `CustomerIOLiveActivitiesTemplates`, then tries to publish this umbrella spec with exact-version dependencies that do not exist. I reproduced that path with `pod lib lint CustomerIO.podspec`: all three aliases fail to resolve. The specs need to be added in dependency order, and the same three names should go into the CDN propagation list. Otherwise the first CocoaPods release containing this will fail after partially publishing the other pods.

Shahroz16 (approved):
One unrelated CI test is red: `MessagingPushPendingPushFlushTests` so that appears flaky

PR #1181, opened 2026-07-27
fix(inbox): correctness and cross-platform parity fixes

mahmoud-elmorabea:
Addresses the correctness findings from the cross-platform review of the inbox overlay work. Paired with the equivalent Android PR so both SDKs behave identically.

- A 204 poll is now handled as an authoritative empty queue, matching gist-web (which treats 204 exactly like a 200 with no messages). It previously fell into the JSON-parse path, which threw — leaving stale inbox rows on screen and skipping the visual-inbox pipeline entirely. Note the reducer's `clearMessageQueue` is deliberately untouched: it is also dispatched on network failure, where the inbox must be retained.
- Logout clears the persisted render assets, the enablement flag, and the revalidation gate. These are workspace-scoped, so one user's inbox could previously be served to the next for the life of the process. The repository doc claimed this already happened; it didn't.
- `x-cio-inbox-enabled` is read via `value(forHTTPHeaderField:)`, which matches case-insensitively unlike an `allHeaderFields` subscript.
- `messageOpened`/`messageDismissed` now fire from the visual inbox rather than the headless `markMessageOpened`/`markMessageDeleted`. Both platforms' docs scope this listener to the visual inbox, and firing from the headless API meant hosts with no overlay mounted received the callbacks. Both platforms now also report the post-action opened state.
- Hidden-reason composition matches Android's early-return form, so the byte-for-byte parity claim in the code is now true.
- Jist is pinned exactly in both `Package.swift` and the podspec, matching the Android Gradle pin.
- `pod install` no longer runs behind `|| true`, which was masking a real resolution failure as a confusing build error several steps later.
- Deleted the stale `MessagingInboxTargetTests` placeholder.

`VisualInboxRepository.swift` hit the 400-line lint limit, so store observation and session lifecycle moved into `VisualInboxRepositoryObservation.swift`, following the existing `VisualInboxSPITypes` split.

Verified locally: 458 tests across `MessagingInAppTests` + `MessagingInboxTests` pass, `swiftlint --strict` is clean, and the SDK builds.

---

> [!NOTE]
> **Medium Risk**
> Changes affect inbox polling, persisted cache on logout, load-state visibility, and when host listeners fire—important for multi-user and overlay correctness, but covered by new/updated tests.
> 
> **Overview**
> Aligns the iOS visual inbox with Android and gist-web on queue handling, session boundaries, visibility rules, and host callbacks.
> 
> **Queue polling:** HTTP **204** is treated as an authoritative empty inbox (clears cache, publishes `[]`, still runs the visual-inbox pipeline). **`x-cio-inbox-enabled`** is read with case-insensitive header lookup.
> 
> **Session / cache:** Logout clears workspace-scoped render assets and reopens the once-per-session revalidation gate, with guards for in-flight fetches and cold-start `nil` userId. Visibility is driven only by **templates + branding**—zero messages stays **visible-and-empty**, not hidden. Coalesced **`snapshot()`** / **`loadStateAndJistMessages()`** reduce inconsistent overlay reads.
> 
> **Host API:** **`messageOpened`** / **`messageDismissed`** fire from the visual inbox SPI path only (not headless `markMessageOpened` / `markMessageDeleted`).
> 
> **Tooling:** Jist pinned to **0.1.0** across SPM/CocoaPods; sample-app workflow includes **CustomerIOMessagingInbox**; **`pod install`** no longer masks failures.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b360a0e7d087a7aae9deee158541b8f3a74c27c3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Sources/MessagingInApp/Inbox/NotificationInbox.swift
```diff
@@ -93,6 +93,17 @@ public protocol NotificationInbox: Sendable {
     /// to call on every render. No-op if no listener is set.
     func notifyMessageShown(message: InboxMessage)
 
+    /// Notifies the registered ``InboxEventListener`` that a message was opened in the visual inbox.
+    /// Deduped by the SDK so it fires at most once per message per app session. Separate from
+    /// ``markMessageOpened(message:)`` so the callback stays scoped to the visual inbox rather than
+    /// firing for every headless caller. No-op if no listener is set.
+    func notifyMessageOpened(message: InboxMessage)
```
Shahroz16: Could these listener-dispatch methods stay inside the native visual inbox instead of becoming requirements on the public `NotificationInbox` protocol? Flutter and React Native use `MessagingInApp.setInboxEventListener` and the native views; neither calls these methods directly. If they must stay public, could we add default implementations so existing conformers keep compiling?

Sources/MessagingInApp/Inbox/Data/VisualInboxSPI.swift
```diff
@@ -182,7 +184,7 @@ final class VisualInboxProviderImpl: VisualInboxProvider, @unchecked Sendable {
     /// snapshot messages and `unopenedCount` are derived from that same array. Reading the list and
     /// the count via separate reads could observe a store change in between and produce a badge that
     /// disagrees with the list it's shown next to; deriving both from one read keeps them consistent.
-    private func currentSnapshot() async -> VisualInboxSnapshot {
+    func snapshot() async -> VisualInboxSnapshot {
         async let state = self.state()
```
Shahroz16: Could we build this snapshot from a single store read? A queue or SSE update can land between these async reads, leaving the state and messages from different versions.

Sources/MessagingInApp/Inbox/Data/VisualInboxRepository.swift
```diff
@@ -93,14 +94,23 @@ actor VisualInboxRepositoryImpl: VisualInboxRepository {
     private var isFetchInFlight = false
 
     /// Session-scoped revalidation gate: false until this session's first revalidation, then
-    /// same-session loads serve the stored payload with no network call. Resets ONLY on process
-    /// restart (DI singleton) — i.e. a new session revalidates exactly once.
-    private var didRevalidateThisSession = false
+    /// same-session loads serve the stored payload with no network call. Reopens on process restart
+    /// (DI singleton) and on logout — i.e. a new session revalidates exactly once.
+    var didRevalidateThisSession = false
```
Shahroz16, line 104: Could we leave this gate open when neither fetched nor cached templates or branding are available? A transient first-load failure currently closes it, so later updates keep reading an empty cache and the inbox stays hidden until restart.

Sources/MessagingInApp/Inbox/Data/VisualInboxRepository.swift
```diff
@@ -298,15 +270,18 @@ actor VisualInboxRepositoryImpl: VisualInboxRepository {
         templates: InboxTemplatesRegistry?,
         branding: InboxBranding?
     ) -> VisualInboxLoadState {
-        if !messages.isEmpty, templates != nil, branding != nil {
-            return .visible(messageCount: messages.count)
+        // Reason composition mirrors Android exactly: an empty selection short-circuits on its own,
+        // and only the templates/branding pair ever combines into one reason.
+        if messages.isEmpty {
+            return .hidden(reason: "no selected messages")
```
Shahroz16: Not sure what the best UI/UX would be here but maybe an enabled inbox with no messages show the documented “You’re all caught up” state? Returning `.hidden` here makes that branch unreachable, so the standalone inbox becomes blank after its last message is dismissed.?

Sources/MessagingInApp/Gist/Managers/QueueManager.swift
```diff
@@ -67,6 +67,11 @@ class QueueManager {
                         } catch {
                             completionHandler(.failure(error))
                         }
+                    case 204:
+                        // An authoritative empty queue, not a failure: parsing the empty body would
+                        // throw, stranding stale rows on screen and skipping the inbox pipeline.
+                        self.publishInboxMessages([], inboxEnabledHeader: inboxEnabledHeader)
+                        completionHandler(.success([]))
```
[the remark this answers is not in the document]
mahmoud-elmorabea: Confirmed and fixed in b3f1fb0 — the 204 path now calls `clearCachedUserQueue()` alongside clearing the inbox state, so a later 304 falls through to `.success(nil)` (keep the local store) instead of re-parsing the stale body.

Reproduced before fixing: with the 204 leaving the cache in place, a 200 → 204 → 304 sequence dispatches `processInboxMessages` with counts `[1, 0, 1]` — the cleared row comes back on the 304, exactly as described. With the fix it stays `[1, 0]`. That is now `test_fetchUserQueue_whenNoContentThenNotModified_expectNoStaleRepublish`.

Worth flagging for whoever picks up the queue layer: **Android has this same latent issue on its base branch**, and it predates this PR. `Queue.kt` routes 204 to `handleNoContent` → `ClearMessageQueue(isContentEmpty = true)` but never clears `inAppPreferenceStore.saveNetworkResponse(...)`. A later 304 gets its body swapped for that stale payload by the OkHttp interceptor, rewritten to `200 + x-from-cache`, and dispatched as `ProcessInboxMessages(stale)`. Left out of scope here since these paired PRs are limited to the cross-platform divergences, but it should get its own fix.

One honest caveat on severity: reaching this needs the server to emit a 304 whose ETag maps to a pre-204 non-empty body. I have not confirmed the queue endpoint does that, so treat the fix as hardening rather than a demonstrated production bug.

Sources/MessagingInApp/Inbox/Data/VisualInboxRepository.swift
```diff
@@ -261,8 +239,12 @@ actor VisualInboxRepositoryImpl: VisualInboxRepository {
...
         // Mark the session gate regardless of outcome: this session has now revalidated once.
         // A failed revalidation served stale; we do not re-hit the network again this session
-        // (until a process restart reopens the gate).
-        didRevalidateThisSession = true
+        // (until a process restart reopens the gate). Skipped when a logout reset the session
+        // mid-flight: that reopened the gate, and closing it here would let the next session serve
+        // assets it never revalidated.
+        if sessionGeneration == generationAtStart {
+            didRevalidateThisSession = true
+        }
```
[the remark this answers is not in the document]
mahmoud-elmorabea: Confirmed and fixed in a074af1. Good catch — this is the sharper of the two rounds.

`fetchTemplatesAndBranding` writes `assetsCache.setData(...)` inside each per-endpoint helper as the request succeeds, well before any guard runs, so the logout's `assetsCache.clear()` was simply undone when the parked fetch landed. The generation check now early-returns and re-drops those payloads.

Reproduced rather than assumed: with the re-clear removed, the post-logout assertions come back with a fully populated `InboxTemplatesRegistry` and `InboxBranding` still in the cache — so the write-back is real, not theoretical, and the new assertions are not passing vacuously.

One deliberate narrowing: the guard calls a new `clearRenderAssets()` (templates + branding only) rather than `clear()`. A later poll for the next user may already have re-established the enablement flag, and wiping it here would hide the inbox until the following poll re-sent the header.

Also worth recording against my own earlier reply on this PR: I argued there that the workspace-scoping of these assets made a stale-carryover benign. That holds for Android, which keys its store by URL, but **not** for iOS — `InboxRenderAssetsCache` uses fixed keys (`.inboxTemplatesCache`, `.inboxBrandingCache`) with no workspace component. The mitigation I claimed was weaker here than I said.

Sources/MessagingInApp/Inbox/NotificationInbox.swift
mahmoud-elmorabea: Done in e0d9d63 — went with your first option rather than the default-implementation fallback.

`notifyMessageOpened`/`notifyMessageDismissed` now live on an internal `VisualInboxEventDispatching` that only `DefaultNotificationInbox` conforms to, so they are off the public surface entirely and host conformers have nothing extra to satisfy. The SPI dispatches through a conditional cast, which has a nice side effect: `NoOpNotificationInbox` no longer needs the two empty overrides at all — not conforming *is* the no-op.

The public API baseline drops the two entries, so `check-api-breaking-changes` confirms they left the surface.

Worth noting `notifyMessageActionTaken` and `notifyMessageShown` were already protocol requirements before this PR, so I left them where they are — moving them is a separate, larger call about the pre-existing surface, and I did not want to widen a review-fix PR into that. Happy to follow up if you want the whole notify* group behind the internal protocol.

Sources/MessagingInApp/Inbox/Data/VisualInboxSPI.swift
mahmoud-elmorabea: Fixed in e0d9d63, and you were right that my earlier pass only got half of this.

Round 1 made the messages and `unopenedCount` derive from one `jistMessages()` read, so the badge could not disagree with the list beside it. But `state()` was still its own `async let`, so exactly what you describe remained possible: an update landing between the two reads could pair a load state with a message list from a different version — including a `.visible(messageCount:)` whose count disagreed with the array.

`snapshot()` now calls a single `repository.loadStateAndJistMessages()`. It reads the messages first and the state second inside one actor call, so whatever store change the message read observed is already reflected in the state it returns; the pair cannot straddle versions.

Sources/MessagingInApp/Inbox/Data/VisualInboxRepository.swift
mahmoud-elmorabea: Fixed in e0d9d63 — the most consequential of your six, I think.

The gate closed unconditionally, and the rationale in the comment (do not re-hit a failing server every poll within a session) only actually holds when there is a stale value to fall back on. With an empty cache there is nothing to serve, so closing it stranded the inbox hidden for the whole process exactly as you describe: later store changes hit `recomputeLoadStateFromCurrentMessages`, which re-reads the same empty cache.

It now closes only when both templates and branding resolved (fresh OR stale). A cycle that served something still closes it, so the anti-retry-loop property is intact; a cycle that produced nothing leaves it open for the next poll to retry.

Android had the identical behavior — its test `(c)` asserted it — so it is fixed there too (customerio-android#799, 1b0df36), with `loadTemplatesAndBranding_givenFailureAndNoCache_expectGateStaysOpen` and `test_enableAndLoad_whenRevalidationFailsWithNoCache_expectGateStaysOpen` covering the two sides. The existing serve-stale tests still pass unchanged, since those have a cache and so still close the gate.

Sources/MessagingInApp/Inbox/Data/VisualInboxRepository.swift
mahmoud-elmorabea: Fixed in e0d9d63. You were right that the branch was unreachable, and it was unreachable on both platforms.

`computeLoadState` returned `.hidden(reason: "no selected messages")` on an empty selection, while `NotificationInboxView` had a fully written "You're all caught up" state keyed off *visible with no messages* — a state the data layer could never produce. Android was the same: `computeVisibilityInternal` short-circuited to Hidden, and `NotificationInboxOverlay` had an `EmptyState` branch behind `state.messages.isEmpty()` that never ran.

The message count no longer gates visibility on either platform. Renderable (enabled + templates + branding) means visible; the count only decides what the list draws.

On your UI/UX uncertainty — the reason this is safe is that overlay chrome is gated separately from visibility. iOS `showsChrome` requires `!renderableMessages.isEmpty` and Android uses `state.isVisible && hasRenderableMessages`, so the floating bell still disappears when the inbox empties. No bell over a blank panel; the change only makes the standalone/embedded inbox able to say "caught up" instead of going blank.

Flagging one thing for the record: this reverses the message-driven-visibility decision taken during the data-layer PR, so if that was load-bearing for something I have not seen, this is the commit to point at.

Sources/MessagingInApp/Inbox/Data/VisualInboxRepositoryReads.swift
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    func jistMessages() async -> [JistInboxMessage] {
+        let selected = await selectedMessages()
+        return InboxMessageJistAdapter.toJist(selected)
+    }
+
+    func loadStateAndJistMessages() async -> (state: VisualInboxLoadState, messages: [JistInboxMessage]) {
+        let messages = await jistMessages()
+        // Read the state AFTER the messages, in the same actor call: whatever store change the
```
Shahroz16: Could we soften “can never straddle versions” here? `jistMessages()` awaits the store, so `currentLoadState` can still advance before this returns. The UI handles that safely today, but the comment promises stronger consistency than the method provides.

Tests/MessagingInApp/Inbox/Data/VisualInboxProviderTest.swift
```diff
@@ -174,6 +174,8 @@ private final class NoOpNotificationInboxFake: NotificationInbox, @unchecked Sen
     }
 
     func notifyMessageShown(message: InboxMessage) {}
+    func notifyMessageOpened(message: InboxMessage) {}
```
Shahroz16: Could we remove these two methods from the fake? It no longer conforms to `VisualInboxEventDispatching`, so the conditional cast never calls them and they are now dead test code.

Shahroz16 (approved):

Sources/MessagingInApp/Inbox/Data/VisualInboxSPI.swift
mahmoud-elmorabea: Softened in b360a0e — you were right, and Bugbot independently flagged the same thing on `VisualInboxRepositoryReads.swift` a few minutes after you did.

The comment promised more than the method delivers. `jistMessages()` awaits the store, and the store subscriber can recompute `currentLoadState` from a newer update while the actor is suspended there, so the pair can still straddle versions — reading the state after the messages just changes which side is stale. Corrected in all three places I had overstated it: the implementation comment, the protocol doc, and the `snapshot()` doc.

They now state what actually holds: the messages and `unopenedCount` are consistent with each other (one array, one selection), and the load state is a best-effort visibility hint that resolves on the next emission.

On the code itself — we discussed it and decided to leave the residual window rather than restructure, which matches your read that the UI handles it safely today. Making it airtight would mean deriving the state from the same selection instead of reading the separately-maintained field, and that is a bigger change than this PR should carry. Happy to pick it up separately if you would rather not leave it.

Sources/MessagingInApp/Inbox/NotificationInbox.swift
mahmoud-elmorabea: Good catch — removed in b360a0e. Confirmed dead exactly as you describe: once the fake stopped conforming to `VisualInboxEventDispatching`, the conditional cast in the SPI could never reach either method, so they were unreachable stubs rather than no-op behavior.

461 tests, 0 failures; swiftlint clean.

PR #1185, opened 2026-07-29
fix(inbox): backfill on SSE connect confirmation and own the inbox view models

mahmoud-elmorabea:
The two live findings from the #1141 review, plus the api-docs baseline that is currently failing `check-api-breaking-changes` on the base branch. Paired with customerio-android#802 for the SSE half.

## 1. SSE backfill ordering (Shahroz16 + Bugbot, independently)

We introduced this in #1143 while fixing the fresh-login empty inbox. `fetchUserQueue` is fire-and-forget (`threadUtil.runBackground`), so issuing it alongside `startConnection()` left the two unordered: the HTTP response could land after an SSE `inbox_messages` update and replace the newer list with an older snapshot, since both publish a full-state write.

**Reviewing whether that fix was the right shape turned out to be the useful part.** gist-web has the same need and solves it differently — `src/managers/queue-manager.ts` fetches from `sseSource.addEventListener('connected', ...)`, *after* the server confirms the connection, not before opening the `EventSource`. Our shape was the divergence.

The fetch now runs from the server's `connected` event — the only transition into `.connected` — via a hook `SseLifecycleManager` registers on `SseConnectionManager`. This orders the two by construction instead of reconciling them afterwards, so it needs no versioning tokens, no merge action, and no reducer changes. It also stops fetching when a connection never establishes, and collapses three trigger sites into one.

Considered and rejected: versioning the full-state writes (several times the size, and depends on allocating tokens at request rather than publish time — easy to get wrong later); awaiting the fetch before connecting (opposite of web, delays the connection, a hung fetch stalls it).

**Scope note:** this also removes the *pre-existing* pre-fetch on the foreground path (`SseLifecycleManager`, present on `main`). Once connection confirmation drives the fetch, keeping it would double-fetch — and it carried the same race. Android had no equivalent.

## 2. Inbox view model ownership (Shahroz16)

`NotificationInboxView` and `NotificationInboxBell` built their model with `ObservedObject(wrappedValue: VisualInboxModel())`. `@ObservedObject` does not own its value, so a parent update could hand the view a brand-new model after the original had already started its lifecycle — losing load state, dedupe guards, and in-flight work.

Both are `@available(iOS 13.0, *)` against an iOS 13 module floor, so `@StateObject` is not an option (only `NotificationInboxOverlay` can use it — it is iOS 16). Each public view now owns the model with `@State` and renders an internal `InboxListView` / `InboxBellView` that takes it as `@ObservedObject`. `VisualInboxModel.init` has no side effects — work begins in `start()` — so the instances SwiftUI discards are inert.

**Public API is unchanged**: same type names, same public initializers, and no existing api-docs baseline moved.

## 3. Missing api-docs baseline

`CioMessagingInbox` was added to `generate-api-docs.sh` without committing the generated baseline, so `check-api-breaking-changes` fails on `feat/overlay-inbox` with `+ CioMessagingInbox.api`. Committed here.

Worth flagging: the generated baseline contains only `class CioMessagingInbox` — the generator does not pick up the public SwiftUI views. So adding the module to the script does not actually give the compatibility job visibility into the new public view surface, which was the point of that request. Probably needs its own look.

## Tests

`SseLifecycleManagerTest`: the two tests asserting a fetch on foreground now assert the connection opens without fetching, then confirm the backfill fires once the hook runs. The hook is captured at `start()` because several tests call `resetMock()`, which clears the mock's recorded arguments.

469 tests, 0 failures; swiftlint clean.

---

> [!NOTE]
> **Medium Risk**
> Touches real-time message/inbox sync and user-switch backfill; a residual race between in-flight HTTP backfill and SSE full-state writes is documented but not fixed in this PR.
> 
> **Overview**
> **SSE inbox backfill** now runs after the server sends its `connected` event (via `setOnConnectionConfirmed` on `SseConnectionManager`), aligned with gist-web, instead of alongside `startConnection()` or on foreground. That removes the race where a fire-and-forget HTTP snapshot could overwrite a newer SSE `inbox_messages` update. Foreground and SSE-enable paths no longer prefetch; `Gist` still calls `fetchUserQueue` when `userId` changes while SSE is already up (user swap with no new `connected`).
> 
> **Visual inbox SwiftUI** public `NotificationInboxBell` and `NotificationInboxView` own `VisualInboxModel` with `@State` (iOS 13–safe) and delegate to internal `InboxBellView` / `InboxListView`; the overlay composes those internals with a shared model. Public API surface is unchanged.
> 
> Adds **`CioMessagingInbox.api`** api-docs baseline and tests for connection-confirmed hook firing once (not on transport open only).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b9ee6af6a72874521f40785681a970866eec1d91. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Sources/MessagingInApp/Gist/Gist.swift
+++ b/Sources/MessagingInApp/Gist/Gist.swift
@@ -152,9 +152,16 @@ class Gist: GistProvider {
             // User became identified and SSE is enabled - stop polling (SSE will take over)
             logger.logWithModuleTag("Gist: User identified with SSE enabled - stopping polling (SSE will handle messages)", level: .info)
             invalidateTimer()
-            // The backfill for messages that predate the stream is driven by the server's `connected`
-            // event in SseLifecycleManager, not from here: issued at this point it would race the
-            // connection and could overwrite a newer SSE update with an older snapshot.
+            // Fetch here as well as from the `connected` hook, because this transition fires on ANY
+            // `userId` change — including one identified user replacing another. In that case SSE is
+            // already connected, so `startConnection()` no-ops, no `connected` event follows, and the
+            // new user's existing inbox would never load until an unrelated reconnect.
+            //
+            // Accepted cost: on anonymous → identified, where SSE was not yet connected, this fetches
+            // once here and again from `connected`. And being fire-and-forget it is not ordered
+            // against SSE, so this path keeps the overwrite window the `connected` hook removes for
+            // the others. Guaranteeing the new user sees their inbox is worth both.
+            fetchUserQueue(state: state)
         } else if !state.isUserIdentified, state.useSse {
             // User became anonymous but SSE flag is still enabled - start polling
             // (SSE won't be used for anonymous users)
--- a/Sources/MessagingInApp/Gist/Gist.swift
+++ b/Sources/MessagingInApp/Gist/Gist.swift
@@ -51,9 +51,9 @@ class Gist: GistProvider {
 
         // Start the SSE lifecycle manager to observe app foreground/background events
         Task {
-            await sseLifecycleManager.start { [weak self] state in
+            await sseLifecycleManager.start(queueFetchHandler: { [weak self] state in
                 self?.fetchUserQueue(state: state)
-            }
+            })
         }
     }
 
@@ -126,11 +126,9 @@ class Gist: GistProvider {
             // SSE is now active - stop polling
             logger.logWithModuleTag("Gist: SSE enabled for identified user - stopping polling timer", level: .info)
             invalidateTimer()
-            // SSE does not deliver pending messages on connect. Backfill any messages that already
-            // exist for the user (mirrors SseLifecycleManager's foreground fetch); otherwise a fresh
-            // login shows an empty inbox until the next foreground. Calls fetchUserQueue directly
-            // because fetchUserMessages() no-ops when shouldUseSse is true.
-            fetchUserQueue(state: state)
+            // The backfill for messages that predate the stream is driven by the server's `connected`
+            // event in SseLifecycleManager, not from here: issued at this point it would race the
+            // connection and could overwrite a newer SSE update with an older snapshot.
         } else if !state.useSse {
             // SSE disabled - start polling
             logger.logWithModuleTag("Gist: SSE disabled - starting polling with interval: \(state.pollInterval)s", level: .info)
@@ -154,11 +152,9 @@ class Gist: GistProvider {
             // User became identified and SSE is enabled - stop polling (SSE will take over)
             logger.logWithModuleTag("Gist: User identified with SSE enabled - stopping polling (SSE will handle messages)", level: .info)
             invalidateTimer()
-            // SSE does not deliver pending messages on connect. Backfill the messages that already
-            // exist for the now-identified user (mirrors SseLifecycleManager's foreground fetch);
-            // otherwise a fresh login shows an empty inbox until the next foreground. Calls
-            // fetchUserQueue directly because fetchUserMessages() no-ops when shouldUseSse is true.
-            fetchUserQueue(state: state)
+            // The backfill for messages that predate the stream is driven by the server's `connected`
+            // event in SseLifecycleManager, not from here: issued at this point it would race the
+            // connection and could overwrite a newer SSE update with an older snapshot.
         } else if !state.isUserIdentified, state.useSse {
             // User became anonymous but SSE flag is still enabled - start polling
             // (SSE won't be used for anonymous users)
--- a/Sources/MessagingInApp/Gist/Network/SSE/SseConnectionManager.swift
+++ b/Sources/MessagingInApp/Gist/Network/SSE/SseConnectionManager.swift
@@ -9,6 +9,16 @@ protocol SseConnectionManagerProtocol: AutoMockable {
 
     /// Stops the current SSE connection.
     func stopConnection() async
+
+    /// Registers the hook invoked when the server confirms the connection (its `connected` event),
+    /// which is the only transition into `.connected`.
+    ///
+    /// Exists so the owner can sync state SSE will not deliver: the stream only carries messages
+    /// that arrive AFTER the connection is established, so whatever already exists for the user has
+    /// to be fetched over HTTP. Driving that from here rather than alongside `startConnection()`
+    /// means the fetch cannot race an SSE update that lands first (mirrors gist-web, which fetches
+    /// from its `connected` listener).
+    func setOnConnectionConfirmed(_ handler: @escaping @Sendable () -> Void) async
 }
 
 // sourcery: InjectRegisterShared = "SseConnectionManagerProtocol"
@@ -63,6 +73,9 @@ actor SseConnectionManager: SseConnectionManagerProtocol {
     private var activeConnectionGeneration: UInt64 = 0
 
     private var connectionState: SseConnectionState = .disconnected
+
+    /// Hook invoked from `setupSuccessfulConnection`. See `setOnConnectionConfirmed`.
+    private var onConnectionConfirmed: (@Sendable () -> Void)?
     private var streamTask: Task<Void, Never>?
     private var retryTask: Task<Void, Never>?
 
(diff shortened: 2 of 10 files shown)


Sources/MessagingInApp/Gist/Utilities/SseLifecycleManager.swift
```diff
@@ -104,6 +110,24 @@ actor CioSseLifecycleManager: SseLifecycleManager {
...
     }
 
+    /// Fetches the messages that already existed before the stream opened, once the server has
+    /// confirmed the connection.
+    ///
+    /// SSE only delivers events that arrive AFTER the connection is established, so a fresh login
+    /// would otherwise show an empty inbox until the next foreground. Keyed off the server's
+    /// `connected` event rather than the transition that decides to connect, because the fetch is
+    /// fire-and-forget: issued alongside `startConnection()` its response could land after an SSE
+    /// update and overwrite the newer list with an older snapshot. Waiting for confirmation also
```
Shahroz16: Could we also protect this backfill from a newer SSE snapshot, or merge the result instead of replacing the inbox? Waiting for `connected` ensures the subscription exists, but it does not order the `/queue` response against later SSE events. Both paths replace the full inbox, so an update received while this request is in flight can still be overwritten by an older HTTP snapshot. If that is intentionally out of scope, could we soften this guarantee and track the ordering fix separately?

Sources/MessagingInApp/Gist/Network/SSE/SseConnectionManager.swift
```diff
@@ -157,6 +170,10 @@ actor SseConnectionManager: SseConnectionManagerProtocol {
     /// The cleanup operations include the connection generation, so they only affect
     /// the connection that was active when stopConnection() was called. If a new connection
     /// starts during the await points, it won't be affected by this cleanup.
+    func setOnConnectionConfirmed(_ handler: @escaping @Sendable () -> Void) {
```
Shahroz16: Could we move `setOnConnectionConfirmed` above this comment, or move the comment back above `stopConnection()`? Right now “Stops the active SSE connection…” documents the setter, while `stopConnection()` has no doc.

Sources/MessagingInApp/Gist/Network/SSE/SseConnectionManager.swift
```diff
@@ -544,6 +561,8 @@ actor SseConnectionManager: SseConnectionManagerProtocol {
 
         // Start heartbeat timer with initial timeout (default + buffer)
         await heartbeatTimer.startTimer(timeoutSeconds: HeartbeatTimer.initialTimeoutSeconds, generation: generation)
+
+        onConnectionConfirmed?()
```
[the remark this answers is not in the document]
Shahroz16: I confirmed this on the current head: `setupSuccessfulConnection` is reached from both transport `.connectionOpen` and the server `connected` event, and the callback is unconditional. A normal connection therefore starts two backfills, with the first still happening before server confirmation. Could we invoke the hook only from the `connected` branch and add a connection-manager test for `open → connected` that expects one callback?

Shahroz16 (commented):
Same as android:

iOS receives transport `.connectionOpen` and later the server `.connected` event.
Both call `setupSuccessfulConnection()`.
That method invokes the backfill callback unconditionally, so iOS also performs two backfills per connection, with the first occurring too early.

Sources/MessagingInApp/Gist/Network/SSE/SseConnectionManager.swift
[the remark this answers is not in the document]
mahmoud-elmorabea: Confirmed and fixed in 464d4d2 — same defect as the Android side, same origin.

`setupSuccessfulConnection` is reached from both `handleConnectionOpen` (line 290) and the `.connected` branch (line 359), so the hook fired twice and the first fire was on transport open, before the server confirmed anything. I had asserted the opposite — that the server event was the only transition into `.connected` — on the strength of the class doc line "ConnectionOpenEvent/CONNECTED event", which actually says the reverse of what I took from it.

The invoke now lives in the `.connected` branch only, with the two tests you asked for:

- `test_openThenConnected_expectConnectionConfirmedExactlyOnce`
- `test_transportOpenOnly_expectNoConnectionConfirmed`

Both verified against the regression: with the old placement restored, the first reports 3 invocations and the second reports 1.

471 tests, 0 failures; swiftlint clean.

Sources/MessagingInApp/Gist/Utilities/SseLifecycleManager.swift
mahmoud-elmorabea: Out of scope here, and you are right that I overstated it — the guarantee is softened rather than claimed.

Waiting for `connected` gets two things: the fetch starts against a stream that is already live, and it is skipped entirely when a connection never establishes. It does NOT order the `/queue` response against later SSE events, so an `inbox_messages` event arriving while the request is in flight can still be overwritten by the older HTTP snapshot, because both publish a full-state write. The code comments and the PR body now say exactly that instead of claiming the race is closed.

Closing it properly needs the two full-state writes versioned or merged. The options I looked at:

- **Additive backfill** — a union-instead-of-replace action for this path only. Small, and it matches what the backfill is actually for, since pruning was never its job.
- **Versioned writes** — a token on the action, reducer drops anything older. Also fixes overlapping polls, but needs the token allocated at REQUEST time rather than publish time, or a late response simply wins with a higher number. Several times the size and easy to get subtly wrong later.

Happy to take either as a follow-up on both platforms — worth deciding which, since the additive one is cheap but only covers this path.

Sources/MessagingInApp/Gist/Network/SSE/SseConnectionManager.swift
mahmoud-elmorabea: Fixed in 464d4d2 — good catch, that was a straightforward insertion error on my part. `setOnConnectionConfirmed` now sits above the doc block, so "Stops the active SSE connection…" documents `stopConnection()` again.

Sources/MessagingInApp/Gist/Gist.swift
```diff
@@ -154,11 +152,9 @@ class Gist: GistProvider {
...
             logger.logWithModuleTag("Gist: User identified with SSE enabled - stopping polling (SSE will handle messages)", level: .info)
             invalidateTimer()
-            // SSE does not deliver pending messages on connect. Backfill the messages that already
-            // exist for the now-identified user (mirrors SseLifecycleManager's foreground fetch);
-            // otherwise a fresh login shows an empty inbox until the next foreground. Calls
-            // fetchUserQueue directly because fetchUserMessages() no-ops when shouldUseSse is true.
-            fetchUserQueue(state: state)
+            // The backfill for messages that predate the stream is driven by the server's `connected`
+            // event in SseLifecycleManager, not from here: issued at this point it would race the
+            // connection and could overwrite a newer SSE update with an older snapshot.
```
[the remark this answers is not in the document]
mahmoud-elmorabea: Valid — a regression I introduced by removing this call. Fixed in b9ee6af.

Confirmed the mechanism: `Gist` subscribes to `keyPath: \.userId`, so this handler fires when one identified user replaces another, and `startConnection()` no-ops in that case because a connection is already live. No `connected` event follows, so the new user's inbox would not load until an unrelated reconnect.

Restored the fetch on this transition. Tradeoff recorded at the call site: on anonymous → identified it now fetches both here and from the `connected` hook, and being fire-and-forget this path is not ordered against SSE, so it keeps the overwrite window the hook removes elsewhere. Guaranteeing the new user sees their inbox is worth both.

Two notes for whoever reviews this hunk:

**Android is not affected.** It subscribes to `isUserIdentified` — a `Bool` — so one identified user replacing another produces no emission there at all. Its backfill lived in that same handler, so Android never covered this transition either before or after the change. That is a pre-existing gap rather than a regression, and a real divergence in what "the user changed" means between the two SDKs.

**This branch is not covered by a test, deliberately.** There is no `Gist` test file, no `GistDelegateMock`, and `GistDelegate` has six members, so testing it means a hand-rolled double plus a real `QueueManager` and comparator probing to reach the subscriber — disproportionate scaffolding for one assertion. The connection-hook behaviour this depends on IS covered (`test_openThenConnected_expectConnectionConfirmedExactlyOnce`, `test_transportOpenOnly_expectNoConnectionConfirmed`). Flagging it rather than implying coverage.

471 tests, 0 failures; swiftlint clean.

Shahroz16 (approved):

PR #1190, opened 2026-07-31
chore(geofence): reconcile OS regions by diff instead of wholesale re-registration

mrehan27:
### Summary

Region registration rebuilt the entire monitored set on every sync: stop everything, then re-add the new set. This replaces that with a diff — stop what left the set, start what is new or reshaped, leave carry-over regions registered as they are. An unchanged set now issues no OS calls at all.

This came out of continued real-world drive testing after 4.7.0, on routes long enough to exercise many sync cycles in the background. It brings iOS in line with how Android already registers regions.

### Why

Two behaviours trace back to the same root: the stop and the add are separate operations on an async pipeline, and nothing persists the intended set between them.

**Interrupted re-registration.** If the app is suspended part-way through the sequence, the OS is left holding fewer regions than intended. The movement trigger is re-registered in that same sequence, so the SDK cannot schedule its own wake to correct it; the set is restored on the next app run. On-device measurement: 3 of 17 re-registration cycles applied the removes without the adds.

**Transition loss on set change.** A boundary crossing the OS had detected but not yet delivered was discarded by the teardown. Since carry-over regions are no longer torn down, this window closes too.

Neither is reachable while the app is in the foreground, and neither leaves persistent state behind — the next sync reconciles.

### The change

`GeofenceRegionMonitoring` gains `setMonitoredRegions(_:)`, which takes the desired set and returns what it changed. The diff lives in the monitors rather than the coordinator so each compares against the geometry it actually registered, after the OS radius clamp — otherwise a region larger than the cap reads as "changed" on every pass and churns forever.

A reshaped region is removed before being re-added. On the `CLMonitor` path (iOS 18+) that ordering is required: `add` over an identifier that is already live keeps the original condition and reports no error, verified with a probe. The removal is keyed on what the OS holds rather than on what this process has adopted, so a condition adopted at launch is still repaired if its geometry changes.

What counts as unchanged is decided from this process's ownership plus the circle it recorded — deliberately not from the mirror of drained OS operations. That mirror only updates when a queued add finishes, so consulting it re-registers any region whose add is still in flight — staged either by a sync that landed before an earlier one's operations drained, or by the launch re-arm. That is the churn this PR exists to remove. Ownership and the recorded circle are sufficient because every path that records geometry also queues the matching OS add, and every path that invalidates the OS side clears one of the two synchronously.

Removed along the way:

- `canSkipReregistration` and the two unchanged-set fast paths — the degenerate case of the diff.
- `recenterMovementTrigger` — the trigger is now part of the same reconciliation.

The registration log line counts only the regions the monitor actually took, so one dropped for blocked permission or invalid coordinates is not reported as registered.

The launch re-arm, which re-adds adopted conditions at the OS, now records them in the condition mirror the way a normal registration does. It never did, so the mirror under-reported anything it revived; because the next process seeds ownership from that mirror, a cold-wake event for a revived condition was dropped. Until the unchanged check stopped consulting the mirror this was masked — the next sync re-registered those conditions and rewrote the mirror on the way through, which is precisely the churn being removed here.

### Validation

- Paired A/B drive on device, three unmodified control apps against one carrying this change. The three controls each lost their full region set once during the run and stayed empty until reopened; the app with the fix held all 7 regions through 10 sync cycles.
- Cold-wake path exercised on the same drive. On a cold process the registered-geometry map starts empty, so every region reads as new and the diff degenerates to today's behaviour — adds only, no removes.
- Upgrade path (new binary over an existing container) confirmed: regions persisted by the OS are adopted at launch and then reconciled.
- Full suite green at the head commit — 1604 passed, 0 failures, 17 skipped. SwiftLint `--strict` clean, no public API changes.

New unit coverage for the diff itself: `GeofenceRegionRequestTests` (tolerance and clamp behaviour of `matchesRegistered`) plus coordinator cases for add/remove/unchanged, permission-dropped regions, and the reshape-before-re-add ordering.

Two things here resist unit testing: `CLMonitor` cannot be instantiated from a test bundle, and what the ordering guarantees turn on is pipeline-drain latency, which the synchronous mock cannot represent. Those parts were checked with an executable model of the registration functions against a fake `CLMonitor` carrying the measured semantics (`add` over a live identifier silently ignored, `remove` of an absent one a no-op). It enumerates sync / adopt / `.unmonitored` / relaunch sequences with every drain point after each — 2.56M paths — and asserts three invariants: the OS converges on the desired set and geometry, a region registered with unchanged geometry is never re-registered, and the persisted mirror equals what the OS holds. No violations, under both possible `.unmonitored` semantics. The result is only worth as much as the harness, so each superseded version of this code was run back through it and each reproduces the defect it had — that is how the mirror gap above was found rather than reasoned about. Making `CLMonitor` injectable so the adapter is directly testable is worth a separate ticket.

### Reviewing

`CLMonitorGeofenceMonitor+Registration.swift` is the file to read closely — it holds the registered-geometry mirror and the ordering guarantees. `GeofenceRegionMonitoring.swift` states the contract that unchanged regions must not be re-registered, which is the part that must not regress later.

### Merge order

Targets **`release-july-launch-fixes`**, not `main` — the release branch gathers this and #1191 (plus anything else going into the same patch) so one release is cut for all of it rather than one per merge.

Order into the release branch no longer affects what ships, but merging #1191 first keeps `GeofenceStorage.swift`'s history clean. The two merge cleanly either way — they share only that file, in different places.

---

> [!NOTE]
> **High Risk**
> Large change to background geofence OS registration (CLMonitor ordering, adoption geometry, transition preservation); mistakes can drop events or leave stale regions, though coverage and on-device validation are strong.
> 
> **Overview**
> Geofence sync no longer tears down the full OS region set on every refresh. **`registerWithOsSync`** builds a desired list (movement trigger first, then business fences) and calls **`setMonitoredRegions`**, which adds, removes, or skips each region based on whether its circle actually changed.
> 
> **`GeofenceRegionMonitoring`** gains **`GeofenceRegionRequest`**, **`GeofenceRegionDiff`**, and **`matchesRegistered`** (tolerance + OS radius clamp). The contract is explicit: unchanged circles must stay registered so pending boundary crossings are not dropped.
> 
> The coordinator drops **`canSkipReregistration`**, **`recenterMovementTrigger`**, and the remote/local “unchanged set” fast paths—those cases are handled inside the diff. Logging switches from rerank-unchanged to **`geofenceRegistrationDiff`**.
> 
> On **CLMonitor**, registration moves to **`CLMonitorGeofenceMonitor+Registration`**: an in-memory **`registeredConditions`** map drives **`isRegisteredUnchanged`**; adds **`remove` then `add`** because silent overwrites are ignored; **`setMonitoredRegions`** can sweep stray OS identifiers. **Bootstrap** passes persisted **`monitorRecords`** into **`adoptExistingRegions`** so geometry is seeded before the queued re-arm. **Re-arm** updates the condition mirror per revived id; **`.unmonitored`** clears storage via **`clearMonitorRegionRecord`**.
> 
> **Classic `CLLocationManager`** gets the same **`setMonitoredRegions`** diff, comparing against live **`CLCircularRegion`** geometry.
> 
> Tests and the mock monitor were expanded for diff behavior, adoption records, and **`matchesRegistered`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1810fe007ee9eb0b877b996ffc7ac3ee6226ebc5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor+Registration.swift
+++ b/Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor+Registration.swift
@@ -176,9 +176,16 @@ extension CLMonitorGeofenceMonitor {
 
     /// True when this monitor owns the condition and registered it with the same circle, so
     /// re-adding would only risk absorbing an undelivered crossing.
+    ///
+    /// Ownership plus the recorded circle is sufficient: every path that records geometry also
+    /// queues the matching OS add on the FIFO, and every path that invalidates the OS side clears
+    /// ownership or the record synchronously. `knownConditionIdentifiers` must NOT be consulted —
+    /// it is only updated when queued operations drain, so requiring it re-registers regions whose
+    /// add is still in flight (a sync landing before an earlier one's ops drained) and regions
+    /// revived by `rearmConditions` (which never inserts into it) — each an absorbing remove + add
+    /// for a circle the OS already holds or is about to.
     private func isRegisteredUnchanged(_ region: GeofenceRegionRequest) -> Bool {
         guard ownedRegionIdentifiers.contains(region.identifier),
-              knownConditionIdentifiers.contains(region.identifier),
               let existing = registeredConditions[region.identifier]
         else { return false }
         return region.matchesRegistered(
--- a/Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor+Registration.swift
+++ b/Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor+Registration.swift
@@ -1,0 +1,174 @@
+import CioInternalCommon
+import CoreLocation
+import Foundation
+
+/// Region registration for the CLMonitor path, split out to keep the monitor's event and lifecycle
+/// plumbing readable. Members are `internal` (not `private`) only because they live in a separate
+/// file from their state; they remain monitor implementation detail.
+@available(iOS 17.0, *)
+extension CLMonitorGeofenceMonitor {
+    func adoptExistingRegions(matching identifiers: Set<String>) {
+        let adopted = identifiers.intersection(knownConditionIdentifiers)
+        guard !adopted.isEmpty else { return }
+        ownedRegionIdentifiers.formUnion(adopted)
+        rearmConditions(adopted)
+        logger.geofenceRegionsAdopted(count: adopted.count)
+    }
+
+    func startMonitoring(identifier: String, center: LocationData, radius: Double, transitionTypes: Set<GeofenceTransition>) {
+        reportPermissionTier()
+        guard CoreLocationGeofenceMonitor.permissionTier(for: authManager.authorizationStatus) != .blocked else { return }
+
+        let coordinate = CLLocationCoordinate2D(latitude: center.latitude, longitude: center.longitude)
+        guard CLLocationCoordinate2DIsValid(coordinate) else {
+            logger.geofenceInvalidCoordinatesForRegion(identifier)
+            return
+        }
+
+        // Populate the ownership filter synchronously so a fast-arriving event isn't dropped.
+        ownedRegionIdentifiers.insert(identifier)
+
+        // Parity with the classic monitor's clamp; `maximumRegionMonitoringDistance` is a deprecated
+        // but harmless read with no CLMonitor equivalent — both paths register identical geometry.
+        let clampedRadius = min(radius, authManager.maximumRegionMonitoringDistance)
+
+        // The device's ACTUAL state seeds both CLMonitor's `assuming:` hint and the stored baseline
+        // (see `recordMonitorRegistration`: registration stays silent, the first real crossing
+        // delivers). No fix → geometric expectation: trigger is device-centered (inside),
+        // business geofences outside.
+        noteRegisteredCondition(
+            identifier: identifier,
+            center: LocationData(latitude: coordinate.latitude, longitude: coordinate.longitude),
+            radius: clampedRadius,
+            transitionTypes: transitionTypes
+        )
+
+        let isMovementTrigger = identifier == GeofenceConstants.movementTriggerIdentifier
+        let isInside = isDeviceInside(center: coordinate, radius: clampedRadius) ?? isMovementTrigger
+        let initialTransition: GeofenceTransition = isInside ? .enter : .exit
+        let assumedState: CLMonitor.Event.State = isInside ? .satisfied : .unsatisfied
+
+        enqueueMonitorOperation { [weak self] monitor in
+            guard let self else { return }
+            // Persist before the OS add: storage keys off recorded geometry to preserve the baseline
+            // on an unchanged re-register and reseed on a new/changed circle. The decision lives in
+            // storage because this runs after stop-all, when CLMonitor's own record is already gone.
+            await self.storage.recordMonitorRegistration(
+                identifier: identifier,
+                transitionTypes: transitionTypes,
+                initialState: initialTransition,
+                center: LocationData(latitude: coordinate.latitude, longitude: coordinate.longitude),
+                radius: clampedRadius
+            )
+            let condition = CLMonitor.CircularGeographicCondition(center: coordinate, radius: clampedRadius)
+            await monitor.add(condition, identifier: identifier, assuming: assumedState)
+            self.knownConditionIdentifiers.insert(identifier)
+            self.persistConditionMirror()
+        }
+    }
+
+    func stopMonitoring(identifier: String) {
+        guard ownedRegionIdentifiers.remove(identifier) != nil else { return }
+        registeredConditions.removeValue(forKey: identifier)
+        enqueueConditionRemoval(identifier)
+    }
+
+    /// Drops the condition at the OS. Split from `stopMonitoring` because re-registration must be
+    /// able to clear a condition this process has not adopted — see `setMonitoredRegions`.
+    ///
+    /// The storage record intentionally survives removal: a remove + re-register cycle relies on
+    /// the persisted baseline to suppress CLMonitor's re-evaluation of an unchanged state.
+    private func enqueueConditionRemoval(_ identifier: String) {
+        enqueueMonitorOperation { [weak self] monitor in
+            guard let self else { return }
+            await monitor.remove(identifier)
+            self.knownConditionIdentifiers.remove(identifier)
+            self.persistConditionMirror()
+        }
+    }
+
+    func stopMonitoringAll() {
+        ownedRegionIdentifiers.removeAll()
+        registeredConditions.removeAll()
+        // Clear against CLMonitor's LIVE identifiers, not the owned/mirror snapshot: an empty owned
+        // set or a lossy mirror must not leave a stale SDK condition holding an OS slot.
+        enqueueMonitorOperation { [weak self] monitor in
+            guard let self else { return }
+            for identifier in await monitor.identifiers {
+                await monitor.remove(identifier)
+            }
+            self.knownConditionIdentifiers.removeAll()
+            self.persistConditionMirror()
+        }
+    }
+
+    @discardableResult
+    func setMonitoredRegions(_ regions: [GeofenceRegionRequest]) -> GeofenceRegionDiff {
+        let desiredIdentifiers = Set(regions.map(\.identifier))
+        var removed: Set<String> = []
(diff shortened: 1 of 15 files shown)


Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor+Registration.swift
```diff
@@ -0,0 +1,174 @@
...
+        guard ownedRegionIdentifiers.contains(region.identifier),
+              knownConditionIdentifiers.contains(region.identifier),
+              let existing = registeredConditions[region.identifier]
+        else { return false }
+        return region.matchesRegistered(
+            center: existing.center,
+            radius: existing.radius,
+            transitionTypes: existing.transitionTypes,
+            clampedTo: authManager.maximumRegionMonitoringDistance
+        )
```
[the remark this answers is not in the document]
mrehan27: Looked into this — I don't think the skip is reachable the way it's described, and a live-identifier check wouldn't detect the failure mode named.

**`knownConditionIdentifiers` is seeded from live OS truth.** `reconcileKnownConditions` is the pipeline's first operation on every process and assigns `knownConditionIdentifiers = Set(await monitor.identifiers)`; it also drops `registeredConditions` entries for anything the OS no longer lists. It's maintained on every add/remove after that, not an independent guess.

**The silent-stop case leaves the condition listed.** Per the `CLMonitor.h` note quoted in `rearmConditions`, when CoreLocation stops monitoring a condition whose pending event no monitor was configured to receive, the persisted store still lists it. So the zombie condition is still in `monitor.identifiers` — checking that set would report it healthy. That case is handled by unconditionally re-adding on adoption in `rearmConditions`, which is the only thing that revives it.

**A skip requires this process to have registered the condition itself.** `registeredConditions` is deliberately not seeded from the persisted mirror, so it starts empty on every process. `isRegisteredUnchanged` needs an entry there, which means the first pass in a new process re-registers everything unless `rearmConditions` filled it in by actually re-adding at the OS. Recovery is the next app run rather than a full teardown.

On the classic path, `CoreLocationGeofenceMonitor.isRegisteredUnchanged` already reads live state via `manager.monitoredRegions`. CLMonitor exposes no way to read a condition back, which is why the mirror exists on that path only.

There's a narrow residual: if CoreLocation removed an identifier from `monitor.identifiers` mid-session, the set would stay stale until the process restarts. Reusing the `await monitor.identifiers` read in the sweep above to prune it would misfire — that operation is queued and runs after this pass's synchronous bookkeeping but before the enqueued `add` reaches the OS, so a just-registered region isn't listed yet and would be pruned, re-registering on every subsequent sync. Re-registration is what discards undelivered crossings, which is the behaviour this PR is fixing, so leaving it as-is.

Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor.swift
```diff
@@ -127,6 +145,7 @@ final class CLMonitorGeofenceMonitor: NSObject, GeofenceRegionMonitoring, @preco
         let drifted = persisted != knownConditionIdentifiers
         knownConditionIdentifiers = persisted
         ownedRegionIdentifiers.formUnion(persisted)
+        registeredConditions = registeredConditions.filter { persisted.contains($0.key) }
```
Shahroz16: Could this preserve registrations already staged behind the reconciliation task? `startMonitoring` writes `registeredConditions` synchronously, while its OS add is queued after this initial identifier read. If cached registration starts before CLMonitor finishes loading, this filter removes the new geometry because it is absent from the pre-add snapshot; the later add updates only `knownConditionIdentifiers`. The next sync then removes and re-adds those unchanged regions, reopening the transition-loss window. Recording geometry in the queued add, or keeping pending registrations here, would keep the mirrors aligned.

Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor+Registration.swift
```diff
@@ -0,0 +1,174 @@
...
+        for region in regions where !isRegisteredUnchanged(region) {
+            // CLMonitor SILENTLY IGNORES an add over a live identifier — it keeps the original
+            // circle and reports no error — so a reshaped region must be removed first; the
+            // pipeline is FIFO, so the remove lands before the add. `stopMonitoring` only reaches
+            // regions this process owns, and a condition the OS holds that adoption missed would
+            // otherwise keep its old circle forever: the add is dropped, but the new geometry is
+            // still recorded here, so every later pass reads "unchanged" and never repairs it.
+            if ownedRegionIdentifiers.contains(region.identifier) {
+                stopMonitoring(identifier: region.identifier)
+            } else if knownConditionIdentifiers.contains(region.identifier) {
```
Shahroz16: Could this changed-region path wait for the first live reconciliation before deciding whether removal is needed? If the persisted mirror omits an identifier that CLMonitor still holds, bootstrap can reach this branch while the ID is in neither `ownedRegionIdentifiers` nor `knownConditionIdentifiers`, so no removal is queued. CLMonitor then silently ignores the add over the live ID, while `registeredConditions` records the new geometry; the reconciliation rerun sees that staged geometry as unchanged and leaves the OS monitoring the old circle. Deferring the first diff until reconciliation completes, or removing against live identifiers in the FIFO, would close this.

Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor.swift
mrehan27: Good catch — fixed in `1560a3a7` by deleting the filter.

It turned out that line could only ever do harm. `registeredConditions` is deliberately not seeded from the persisted mirror and starts empty each process, and `reconcileKnownConditions` is the pipeline's first operation — so at the moment it ran, the map's only entries were ones staged synchronously while `CLMonitor` was still loading, whose adds are queued behind the reconcile. Those are exactly the identifiers `persisted` does not contain yet. There was no stale prior-process geometry for it to prune, so filtering could only drop live geometry, and the next sync would re-register regions that hadn't changed — reopening the transition-loss window, as you say.

Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor+Registration.swift
mrehan27: Agreed, and taken the second option — fixed in `1560a3a7` by removing against live identifiers from inside the FIFO rather than deferring the first diff.

`startMonitoring`'s queued operation now does `await monitor.remove(identifier)` before the add, so the removal is keyed on the OS instead of on either mirror and a lost identifier can't defeat it. Removing one the OS doesn't hold is a no-op, and only new or reshaped regions reach `startMonitoring`, so nothing unchanged is torn down. The `ownedRegionIdentifiers` / `knownConditionIdentifiers` branch you commented on is gone entirely — the diff is smaller than before.

I preferred this over deferring the first diff until reconciliation because the cold-wake path must not add an `await` between constructing the monitor and registering; a suspension there is what the parent commit exists to avoid.

Three related things came out of reviewing that change, all in the same commit:

- Ownership is now released before each re-registration. Dropping the old conditional removal also dropped the ownership release it was doing, which left a region that `startMonitoring` rejects (blocked permission, invalid coordinates) still counted in `monitoredRegionIdentifiers` — so `emitInitialEnters` could fire an enter for a region the OS isn't monitoring.
- Those rejected paths now clear the identifier at the OS too. With ownership released and nothing owning it, no later pass repaired it, so an old circle could hold one of the 20 slots for as long as the region stayed in the desired set.
- `.unmonitored` now clears the stored baseline. Nothing else prunes it while the region stays desired, so the device could cross while unmonitored and an unchanged-geometry re-register would carry the stale state and suppress the next genuine arrival. The clear runs on the pipeline and skips when `knownConditionIdentifiers` shows a registration has already re-added the identifier.

Coverage caveat worth stating: `CLMonitorGeofenceMonitor` can't be instantiated from the test bundle, so the queued removal and the `.unmonitored` call site are covered by reading plus device runs, not by tests. The new coordinator tests exercise the same decisions through the mock. Making `CLMonitor` injectable so the adapter becomes testable is worth a separate ticket.

Shahroz16 (approved):
Rechecked the latest head. The earlier findings are addressed, and I don't see anything remaining that should block this.

PR #1191, opened 2026-07-31
chore(geofence): select nearest regions by boundary distance and prune stale baselines

mrehan27:
### Summary

Two related correctness improvements in how the nearest region set is chosen and how per-region state is retained. Both surfaced from the same question: what happens to a region the device is *inside* when the set changes.

No urgency — nothing here is tied to a reported failure. Kept separate from the registration fix so that one stays minimal.

### 1. Rank by boundary distance, not centre

Nearest-set selection ranked and capped on distance to each region's centre, ignoring radius. A region the device sits inside therefore ranks at its own radius rather than at zero, so a large region is evicted as soon as `maxBusinessGeofences` regions have nearer centres. An evicted region is unmonitored, so its exit is never reported.

The `maxMonitoringDistance` cutoff had the same shape: a region whose centre fell outside the cap was filtered out even when its area contained the device.

Ranking and filtering now use `edgeDistanceTo`, which clamps to zero inside the region, so an occupied region sorts first and survives both the cap and the limit.

Two things worth knowing:

- When all regions share a radius the ranking order is unchanged — subtracting a constant preserves order. Accounts with mixed radii are the ones affected.
- The distance cap does change: measuring to the boundary widens the effective cutoff by one radius, so regions just outside the old limit now register. The server's own selection is ordered by centre distance, so in a dense workspace a large region containing the device can still be cut before the SDK sees it; aligning that is a backend follow-up.

Demonstrated on a simulator: a 2 km region the device sat inside was dropped once 20 nearer-centred regions appeared, and its exit never arrived.

### 2. Prune monitor baselines to the registered set

A per-region baseline deliberately survives `stopMonitoring`, so an unchanged re-register keeps it and `CLMonitor` re-evaluating the same state is not delivered as a duplicate. That is correct for a region being replaced and incorrect for one being evicted: an evicted region goes unmonitored, so no EXIT ever balances an `.enter` baseline. Re-registering the same circle later preserves that stale value instead of the state the device is actually in, and the next genuine arrival reads as no change.

Cost when it happens is one missed arrival plus an EXIT with no matching ENTER; the EXIT clears the baseline, so it does not persist beyond that.

`recordRegistration` now retains records only for the registered set plus the movement trigger. Regions still in the set keep their baseline, so the duplicate suppression these records exist for is unaffected. This also bounds the records the same way `monitoredGeofenceIds` is already bounded — previously they were only ever added to, or cleared wholesale on sign-out.

**Existing installs**: no migration step. The prune runs on every sync, so the first sync after updating clears anything a previous build stranded. Clearing all records on upgrade was considered and rejected — with no baseline the next crossing is suppressed, which would cost every user one event.

`CLMonitor` path only (iOS 18+); the classic path keeps no baseline.

Ports Android `21914158`, whose commit message names this gap on iOS.

### Why these two together

They share a precondition. Centre-distance ranking is what makes an occupied region get evicted in the first place, which is what strands the baseline. Fixing the ranking removes most of the trigger; pruning handles the rest.

### Validation

- Both changes revert-verified: each new test fails with only its own fix backed out.
- Simulator A/B for the ranking change — the containing region now survives eviction and its exit is reported.
- Storage-level reproduction of the stale baseline before and after the prune.
- Full suite green — 1040 XCTest, 575 Swift Testing, 0 failures. SwiftLint `--strict` clean, no public API changes.

### Merge order

Targets **`release-july-launch-fixes`**, not `main` — the release branch gathers this and #1190 (plus anything else going into the same patch) so one release is cut for all of it rather than one per merge.

Order into the release branch no longer affects what ships. Merging this one first keeps `GeofenceStorage.swift`'s history clean; the two merge cleanly either way, sharing only that file in different places.

---

> [!NOTE]
> **Medium Risk**
> Changes core geofence selection and persisted CLMonitor dedup state; behavior shifts for mixed radii and distance caps, but scope is localized with strong test coverage.
> 
> **Overview**
> Fixes two geofence correctness issues: which regions stay monitored when the OS cap bites, and whether **CLMonitor** enter events still fire after a region drops out of the set and comes back.
> 
> **Boundary-based nearest selection** — `GeofenceDistanceFilter` now ranks and applies `maxDistance` using new `Geofence.edgeDistanceTo` (0 inside the circle) instead of center distance. Occupied large regions are no longer evicted in favor of many small nearby centers, and regions whose center is outside the cap but whose area contains the device still qualify.
> 
> **Baseline pruning on registration** — `GeofenceStorage.recordRegistration` filters `monitorRegionRecords` to the current business IDs plus the movement trigger. Evicted regions lose stale `.enter` baselines that could suppress a real enter on re-registration; retained regions keep dedup behavior unchanged.
> 
> Tests cover boundary ranking edge cases and the eviction/revisit regression.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0949d28311b009498f12f6b9b92f8cb1a29ebd75. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Sources/LocationGeofence/Coordinator/GeofenceDistanceFilter.swift
+++ b/Sources/LocationGeofence/Coordinator/GeofenceDistanceFilter.swift
@@ -6,7 +6,13 @@ import Foundation
 /// monitored regions; one slot is reserved for the movement-trigger geofence).
 struct GeofenceDistanceFilter: Sendable {
     /// Ranks and caps by distance to each region's *boundary* (`edgeDistanceTo`), so a region the
-    /// device is inside always ranks first and survives both the limit and the distance cap.
+    /// device is inside ranks first among the candidates and survives both the limit and the
+    /// distance cap.
+    ///
+    /// Only among the candidates: the backend applies its own limit first, ordered by distance to
+    /// each region's center, so in a dense workspace a large region containing the device can be
+    /// cut before it ever reaches this sort. Aligning that server-side ordering with boundary
+    /// distance is a separate change.
     ///
     /// Ties broken by ascending `id` for deterministic ordering. Distances are rounded to whole
     /// meters before comparison: `CLLocation.distance` can return sub-meter-varying values for
--- a/Sources/LocationGeofence/Coordinator/GeofenceDistanceFilter.swift
+++ b/Sources/LocationGeofence/Coordinator/GeofenceDistanceFilter.swift
@@ -5,15 +5,19 @@ import Foundation
 /// to cap business-geofence registrations at the OS-allowed count (iOS allows 20 total
 /// monitored regions; one slot is reserved for the movement-trigger geofence).
 struct GeofenceDistanceFilter: Sendable {
+    /// Ranks and caps by distance to each region's *boundary* (`edgeDistanceTo`), so a region the
+    /// device is inside always ranks first and survives both the limit and the distance cap.
+    ///
     /// Ties broken by ascending `id` for deterministic ordering. Distances are rounded to whole
     /// meters before comparison: `CLLocation.distance` can return sub-meter-varying values for
     /// identical inputs, which would otherwise defeat the id tiebreak and make the order of
-    /// equidistant regions nondeterministic. Regions farther than `maxDistance` are excluded
-    /// (`GeofenceConstants.noMonitoringDistanceCap` for no cap). Returns empty when `limit <= 0`.
+    /// equidistant regions nondeterministic. Regions whose boundary is farther than `maxDistance`
+    /// are excluded (`GeofenceConstants.noMonitoringDistanceCap` for no cap). Returns empty when
+    /// `limit <= 0`.
     func nearest(_ regions: [Geofence], to location: LocationData, limit: Int, maxDistance: Double) -> [Geofence] {
         guard limit > 0, !regions.isEmpty else { return [] }
         return regions
-            .map { ($0, $0.distanceTo(location).rounded()) }
+            .map { ($0, $0.edgeDistanceTo(location).rounded()) }
             .filter { $0.1 <= maxDistance }
             .sorted { lhs, rhs in
                 if lhs.1 != rhs.1 { return lhs.1 < rhs.1 }
--- a/Sources/LocationGeofence/Storage/GeofenceStorage.swift
+++ b/Sources/LocationGeofence/Storage/GeofenceStorage.swift
@@ -247,6 +247,18 @@ actor GeofenceStorage {
         var state = loadFromDisk() ?? GeofenceState()
         state.movementTriggerCenter = center
         state.monitoredGeofenceIds = businessIds
+        // Drop per-condition baselines for regions this registration no longer covers. A record
+        // survives `stopMonitoring` on purpose, so an unchanged re-register keeps its baseline and
+        // CLMonitor's re-evaluation stays silent — but a region *evicted* from the set is a
+        // different case. It goes unmonitored, so no EXIT ever balances a `.enter` baseline, and a
+        // later re-registration with the same circle keeps that stale value instead of the state
+        // the device is actually in. The next genuine arrival then reads as no change and is
+        // dropped. Retaining exactly the registered set bounds the records the same way
+        // `monitoredGeofenceIds` is bounded, and clears anything a previous version stranded.
+        if let records = state.monitorRegionRecords {
+            let retained = businessIds.union([GeofenceConstants.movementTriggerIdentifier])
+            state.monitorRegionRecords = records.filter { retained.contains($0.key) }
+        }
         saveToDisk(state)
     }
 
--- a/Sources/LocationGeofence/Model/Geofence+Distance.swift
+++ b/Sources/LocationGeofence/Model/Geofence+Distance.swift
@@ -14,4 +14,18 @@ extension Geofence {
     func distanceTo(_ location: LocationData) -> CLLocationDistance {
         distanceTo(latitude: location.latitude, longitude: location.longitude)
     }
+
+    /// Distance in meters from this geofence's *boundary* to the given location, `0` when the
+    /// location is inside.
+    ///
+    /// Monitoring relevance is proximity to the boundary, not to the center: ranking on center
+    /// distance evicts a large region the device currently occupies as soon as
+    /// `maxBusinessGeofences` regions have nearer centers, and an unmonitored region can never
+    /// report its exit.
+    ///
+    /// Not a containment test — this is `0` for every point inside. Use `distanceTo` against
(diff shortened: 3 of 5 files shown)


Sources/LocationGeofence/Storage/GeofenceStorage.swift
```diff
@@ -247,6 +247,18 @@ actor GeofenceStorage {
...
+        // Drop per-condition baselines for regions this registration no longer covers. A record
+        // survives `stopMonitoring` on purpose, so an unchanged re-register keeps its baseline and
+        // CLMonitor's re-evaluation stays silent — but a region *evicted* from the set is a
+        // different case. It goes unmonitored, so no EXIT ever balances a `.enter` baseline, and a
+        // later re-registration with the same circle keeps that stale value instead of the state
+        // the device is actually in. The next genuine arrival then reads as no change and is
+        // dropped. Retaining exactly the registered set bounds the records the same way
+        // `monitoredGeofenceIds` is bounded, and clears anything a previous version stranded.
+        if let records = state.monitorRegionRecords {
+            let retained = businessIds.union([GeofenceConstants.movementTriggerIdentifier])
```
Shahroz16, line 259: Could we remove the stored baseline when CLMonitor reports that a condition is no longer held? The call sites pass desired `nearestIds`, not conditions the OS still holds, so `.unmonitored` clears the monitor’s ownership and mirror but this pruning retains the record while the region remains desired. Re-adding the same circle then preserves the stale `lastState`; because `assuming:` uses the current state, CLMonitor emits no corrective event, and the next genuine crossing can be suppressed. Clearing the storage record in the `.unmonitored` path would close that gap.

Sources/LocationGeofence/Coordinator/GeofenceDistanceFilter.swift
```diff
@@ -5,15 +5,19 @@ import Foundation
...
     /// meters before comparison: `CLLocation.distance` can return sub-meter-varying values for
     /// identical inputs, which would otherwise defeat the id tiebreak and make the order of
-    /// equidistant regions nondeterministic. Regions farther than `maxDistance` are excluded
-    /// (`GeofenceConstants.noMonitoringDistanceCap` for no cap). Returns empty when `limit <= 0`.
+    /// equidistant regions nondeterministic. Regions whose boundary is farther than `maxDistance`
+    /// are excluded (`GeofenceConstants.noMonitoringDistanceCap` for no cap). Returns empty when
+    /// `limit <= 0`.
     func nearest(_ regions: [Geofence], to location: LocationData, limit: Int, maxDistance: Double) -> [Geofence] {
         guard limit > 0, !regions.isEmpty else { return [] }
         return regions
```
Shahroz16, line 25: I checked the current CDP and Services path and there is one earlier cap to account for. The SDK omits `limit`, CDP defaults to 100, and Services orders by center distance before applying that limit. In a dense workspace, a large fence containing the device can fall outside those 100 and never reach this boundary-distance sort. Could we align the server-side limit with boundary distance, or track that as a required backend follow-up before describing occupied fences as always retained?

Sources/LocationGeofence/Storage/GeofenceStorage.swift
mrehan27: Good catch, and agreed — the prune keys on the desired set, so a region the OS gave up on while it stays desired keeps its record, and an unchanged-geometry re-register then carries the stale `lastState`.

Fixed, but in #1190 rather than here: the only place that knows the OS gave up is the `.unmonitored` handler in `CLMonitorGeofenceMonitor`, which is a file that PR touches. It now clears the stored record alongside ownership and the mirror entry, using a new `GeofenceStorage.clearMonitorRegionRecord(identifier:)`. The clear runs on the monitor pipeline and skips itself when `knownConditionIdentifiers` shows a registration has already re-added the identifier, so a re-add that wrote a baseline from the device's real position isn't deleted.

Both PRs ship in the same patch, so nothing is left unfixed between them.

Sources/LocationGeofence/Coordinator/GeofenceDistanceFilter.swift
mrehan27: Thanks for checking the CDP and Services path — that answers a question I had open and couldn't settle from the SDK side.

Taken the second option: `0949d283` drops the unconditional claim. The doc now says an occupied region ranks first *among the candidates returned*, and names why — the backend applies its own limit first, ordered by center distance, so in a dense workspace a large region containing the device can be cut before this sort ever sees it.

Aligning that server-side ordering with boundary distance is a backend change, so it is out of scope for this PR. Flagging it rather than committing anyone to it — the call belongs to whoever owns the `/geofences/nearest` endpoint.

One SDK-side lever worth noting for that conversation: `radius` and `limit` are both optional in the nearest request body and we omit both, which is why the default of 100 applies. Requesting a higher limit would narrow the window, since a containing region would have to fall outside the top-N by center distance rather than the top-100. It is a mitigation rather than a fix — a large enough workspace still cuts it — and it inflates the response and cached set in exactly the dense workspaces where it matters, so I would not do it blind.

Shahroz16 (approved):
Rechecked the latest head. The earlier feedback is addressed, and I don't see anything remaining that should block this.

mahmoud-elmorabea (approved):

Sources/LocationGeofence/Storage/GeofenceStorage.swift
mrehan27: One update since my reply above, so the description here stays accurate.

`13725835` in #1190 also makes `rearmConditions` record the identifiers it re-adds in `knownConditionIdentifiers` (it previously re-added at the OS without writing the mirror). That widens the skip condition I described: a condition the launch re-arm has revived now suppresses the record clear too, not only one re-added by `startMonitoring`.

That is the behaviour we want here rather than an exception to it — the re-arm passes `assuming: record.lastState`, so it arms the condition *from* the stored baseline, and deleting that baseline immediately afterwards would leave a live condition whose next event establishes a new baseline without delivering.

Sources/LocationGeofence/Coordinator/GeofenceDistanceFilter.swift
mrehan27: Closing the loop on one thing your comment exposed beyond the code doc: the PR description carried the same unconditional claim, saying the boundary cap "matches the server, which selects its nearest set by boundary distance." That was wrong in exactly the way you describe. Corrected — the body now states that the server orders by center distance and that a large region containing the device can be cut before the SDK sees it.

So within this PR the claim is scoped correctly in both places. The server-side ordering itself stays open and is not something this PR can close; it needs the backend change, and until then the improvement here is bounded to the candidates that do come back.

PR #1199, opened 2026-08-07
fix(in-app): guard modal sizing against viewport-dependent message height

mahmoud-elmorabea:
### Bug

Modal in-app messages are sized to the content height the renderer reports (about once a second). If a message's CSS instead takes its height from the viewport — `height: 100vh`, or `height: 100%` on `html`/`body` — the two are mutually recursive and the height never resolves:

- **Collapsed:** reported == container, so zero is a fixed point. The WebView starts at zero, so the message stays at zero forever behind a full-screen overlay that shows nothing and swallows every touch. On iOS there is no outside-tap dismissal, so it cannot be escaped at all.
- **Runaway:** reported == container + constant, so the height grows by that constant on every report until the container clamps it, rendering full-screen with content pinned to the top.

Reproduced on device: `0.0 x 0.0` then `402.0 x 0.0` repeating; and `864 → 896 → 906` settling at screen height + 32.

### Fix

`ModalSizePolicy` recognises both shapes from the reported heights alone, so it needs no knowledge of the container:

| Reported | Verdict | Behaviour |
| --- | --- | --- |
| collapsed (≤ 20pt) × 4 | degenerate | fail the message → overlay dismissed, host gets `messageError` |
| constant positive delta × 4 | viewport-dependent | keep applying (container clamps, so rendering unchanged) + log the cause once |
| tall but constant | apply | unchanged — a legitimately tall message |
| irregular growth | apply | unchanged — height still settling while images/fonts load |

Requiring a *constant* delta is what keeps a normal message that grows while loading from being flagged.

Reported sizes now flow through an overridable `handleSizeChanged` on `BaseMessageManager`, because `sizeChanged` is declared on the `EngineWebDelegate` extension and methods on a class extension are statically dispatched.

Scoped to modal messages — inline legitimately reports a zero height to collapse itself.

### Notes

- Both paths log a diagnostic naming the likely CSS cause; a blank overlay is otherwise very hard to diagnose.
- The impression logged at `routeLoaded` stays counted — the overlay did display, and correcting it would need new metric semantics.
- Policy and tests mirror the Android counterpart so the two SDKs cannot drift.

### Tests

`MessagingInAppTests`: 488 tests, 0 failures (12 new). swiftformat + swiftlint `--strict` clean.

Ref: MBL-2242

### DIFF
--- a/Sources/MessagingInApp/Gist/EngineWeb/EngineWeb.swift
+++ b/Sources/MessagingInApp/Gist/EngineWeb/EngineWeb.swift
@@ -130,7 +130,7 @@ public class EngineWeb: NSObject, EngineWebInstance {
     @objc
     func forcedTimeout() {
         logger.logWithModuleTag("Timeout triggered, triggering message error.", level: .info)
-        inAppMessageManager.dispatch(action: .engineAction(action: .messageLoadingFailed(message: currentMessage)))
+        inAppMessageManager.dispatch(action: .engineAction(action: .messageLoadingFailed(message: currentMessage, suppressRetry: false)))
         delegate?.error()
     }
 }
--- a/Sources/MessagingInApp/Gist/Managers/MessageManager.swift
+++ b/Sources/MessagingInApp/Gist/Managers/MessageManager.swift
@@ -125,6 +125,18 @@ open class BaseMessageManager {
         engine.delegate = nil
     }
 
+    // MARK: - Size Reporting
+
+    /// Applies a size reported by the message renderer.
+    ///
+    /// Declared on the class body rather than on the `EngineWebDelegate` extension below because
+    /// methods defined in a class extension are statically dispatched and cannot be overridden.
+    /// Subclasses override this when they need to inspect reported sizes.
+    func handleSizeChanged(width: CGFloat, height: CGFloat) {
+        gistView.delegate?.sizeChanged(message: currentMessage, width: width, height: height)
+        logger.logWithModuleTag("Message size changed Width: \(width) - Height: \(height)", level: .debug)
+    }
+
     // MARK: - Helpers
 
     public func showNewMessage(url: URL) {
@@ -262,8 +274,7 @@ extension BaseMessageManager: EngineWebDelegate {
     }
 
     public func sizeChanged(width: CGFloat, height: CGFloat) {
-        gistView.delegate?.sizeChanged(message: currentMessage, width: width, height: height)
-        logger.logWithModuleTag("Message size changed Width: \(width) - Height: \(height)", level: .debug)
+        handleSizeChanged(width: width, height: height)
     }
 
     public func routeError(route: String) {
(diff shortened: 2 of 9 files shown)


mahmoud-elmorabea:
Pushed fixes for the review. Verified each finding against the code first — 12 of 13 were valid, one needed qualifying.

**Correctness**

- **Persistent messages looped.** Confirmed: `shouldMarkMessageAsShown` is `!persistent` for `DisplayMessage`, so persistent messages were never marked shown and the failure left them eligible. Added `suppressRetry` to `MessageLoadingFailed` (defaults to `false`, so transient failures stay retryable) and the reducer now honours it. My original PR body asserted this was safe — that was wrong, and I've corrected the body.
- **Verdict now requires elapsed time, not just report count.** The inline controller documents that `sizeChanged` fires multiple times, and `bridge.js` binds `reportMessageSize` to the window `resize` event, so bursts are real. A collapsed run must now span 2.5s.
- **Latch moved behind explicit `onDegenerateHandled` / `onViewportDependentHandled`**, and the controller returns early when there is no current message — a dropped verdict no longer disables the guard permanently.
- **`arm()` is idempotent** rather than clearing state. That removes iOS's `hasArmedSizePolicy` flag, which only existed to work around the non-idempotence.
- **Index guard** so a sample count below two cannot throw.

**Diagnosability**

- Both diagnostics go through a new `logViewError`, so they carry the module's `[InApp][Modal]` tag, and no longer shadow the base logger.
- They report the policy instance's configured thresholds instead of the static defaults.
- Corrected the runaway comment: Android has no explicit clamp (iOS does), so the height grows until the layout can grow no further.

**Cleanup**

- Dropped the unused height payloads from the verdicts (also removes an allocation per size report).
- `sizePolicy` is injected so tests can drive its clock.

**One qualification — the "container does not clamp" finding.** Correct about the code, and the comment is fixed. But the implied unbounded growth did not occur on device: it converged at `947` (screen 915 + 32) and held for 90 consecutive reports, having applied `2486px` inside a `2400px` parent. Layout/clipping bounds it in practice, so no behaviour change was needed there.

**Not addressed:** iOS still has no manager-level tests (the review's `test-coverage` finding). `sizePolicy` is now injectable, which unblocks it, but constructing `ModalMessageManager` in a test still means standing up `DIGraphShared` and a WebView — worth its own change rather than bolting onto this one.

Tests: Android 544, iOS 492 — 0 failures. ktlint / lint / apiCheck / swiftformat / swiftlint all clean.

mahmoud-elmorabea:
Closing in favour of a renderer-level fix (MBL-2242).

The root cause is a circular dependency in the sizing protocol: the SDK sizes the WebView to the height the page reports, while the page derives its height from that same viewport. Any SDK-side detection is therefore a threshold on an ambiguous number — a reported 0 can mean "no content" or "the container you are measuring against is 0 tall" — so it can only guess, and it adds latency before it guesses.

Fixing it in the renderer (giving the message a layout viewport that is not derived from the reported height) removes the circularity instead of compensating for it, fixes iOS and Android at once, and reaches already-shipped SDK versions. It also makes a reported 0 unambiguous, at which point the only thing left for the SDK is a small sanity check rather than the sampling/timing policy in this PR.

Branch `fix/inapp-modal-size-guard` is left pushed if any of it is needed later.

PR #1204, opened 2026-08-11
chore(geofence): run movement passes on a fresh fix instead of the manager's cached one

mrehan27:
## Problem

`CLLocationManager.location` in a long-suspended process can remain frozen at the fix cached around process start. Movement passes (triggered by movement-trigger EXITs) read that property directly, so every pass in such a process runs against the same stale coordinates:

- the local re-rank evaluates candidates from the wrong point,
- the movement trigger re-centers where the device is not,
- the moved-beyond check never observes real displacement, so the remote refetch can starve for the duration of the trip — geofences beyond the cached window are not discovered until the next app open or cold wake.

The condition requires one process instance to stay suspended-alive across the trip; a cold wake reseeds the cache and is unaffected. Confirmed on device via registration state: the trigger was re-registered at coordinates bit-identical to the trip-start fix across multiple passes.

## Fix

New `MovementFixResolver` (internal): movement-trigger EXITs resolve their fix before dispatch.

- Cached fix ≤ 30 s old → used as-is, no request. (The common case: on an instrumented device drive, all passes hit this branch.)
- Older or missing → one one-shot `requestLocation()` (hundred-meter accuracy) on a resolver-owned manager. A delivered fix is accepted only when it is itself recent with valid accuracy — Core Location can echo a cached location as a new manager's first delivery, and a fix as stale as the one that prompted the request must not complete the pass as fresh. On failure or 10 s timeout the pass falls back to the cached fix.

Both monitors also prefer the newest known fix (manager cache vs. resolver's last delivered fix) for event locations and the inside/outside staging check.

Design constraints:

- **A pass is never dropped or stalled**: dispatch is fire-and-forget, every path completes (fresh, fallback, or the pre-existing nil behavior), and the pending-event drain is never blocked.
- **Worst case is the previous behavior**: request failure/timeout yields exactly what the code did before this change.
- **No posture change**: no continuous location updates, no background-mode additions; at most one one-shot request per movement pass, only when the cache is stale.
- **The wake window can't cut the pass short**: a background-task assertion is held for the life of the fresh-fix request, and the movement pass runs under its own — the trigger EXIT is consumed once dispatched, so a wake window expiring mid-pass would otherwise lose it with no retry.
- The resolver owns its `CLLocationManager` because movement passes must work on wrapper cold wakes, where the Location module is not initialized.
- Fix age is logged on every movement pass, so staleness frequency is measurable in the field.

No public API changes.

## Validation

- 13 unit tests on the resolver covering every branch (fresh/stale, failure fallback, fallback newest-wins, timeout exactly-once with late-delivery retention, coalescing concurrent resolutions into one request, invalid-coordinate guard, stale-echo rejection, invalid-accuracy rejection, background-window held until completion, repeat cycles). Full LocationGeofenceTests: 322/322.
- Simulator end-to-end on both monitor paths: CLMonitor (iOS 26) and classic `CLLocationManager` regions (iOS 17.5) — driven routes with movement cycles, all enter/exit events delivered exactly once at a local mock CDP (10 transitions, 0 duplicates), registration diffs steady at `+1/-0`.
- The stale branch was additionally exercised on-simulator with forced-stale builds: real `requestLocation()` issued and the freshly delivered fix used by the pass; with the recency guard forced to reject every delivery, the pass completed through the timeout fallback instead of stalling.
- Instrumented device drive on a build containing this change: movement cycles, background remote refreshes, and event delivery all healthy; every pass logged its fix age.

---

> [!NOTE]
> **Medium Risk**
> Touches core geofence background delivery and location sourcing on cold wake; behavior is designed to match or improve on the prior fallback path, but incorrect fix freshness logic could still mis-rank or delay sync.
> 
> **Overview**
> Fixes movement geofence passes using **stale** `CLLocationManager.location` after a long-suspended process, which could starve remote refetches for an entire trip.
> 
> Introduces **`MovementFixResolver`**: on movement-trigger EXIT, uses a cache ≤30s old as-is; otherwise issues a one-shot `requestLocation()` (with stale-echo and accuracy guards), coalesces concurrent resolves, and falls back to cached fix on failure or 10s timeout. Both monitors route EXIT through the resolver (fire-and-forget so pending-event drains are not blocked) and **`bestKnownFix()`** so resolver fixes beat a frozen manager cache for other reads.
> 
> **`GeofenceBackgroundTime`** selects UIKit vs no-op background runners; **`GeofenceMonitorBinder`** wraps `handleMovement` in a background-task assertion so wake windows cannot drop the pass mid-flight; the resolver holds a separate assertion for the fresh-fix request. Adds movement-fix logging and 13 unit tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1657d3126b58562482f1c43d5d707a5dd912ce08. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Sources/LocationGeofence/Monitor/MovementFixResolver.swift
+++ b/Sources/LocationGeofence/Monitor/MovementFixResolver.swift
@@ -1,0 +1,132 @@
+import CioInternalCommon
+import CoreLocation
+import Foundation
+
+/// Ensures the fix attached to a movement-trigger EXIT is fresh before it drives a sync pass.
+///
+/// `CLLocationManager.location` on a long-suspended process can stay frozen at the fix cached
+/// around process start, anchoring every movement pass (re-rank point, trigger re-center, the
+/// moved-beyond check) to a stale position for a whole trip. A cached fix older than
+/// `GeofenceConstants.movementFixMaxAge` triggers a one-shot request; on failure or after
+/// `movementFixRequestTimeout` the pass falls back to the cached fix, so it is never worse off
+/// than without the resolver. Concurrent resolutions coalesce onto one in-flight request, and
+/// `latestFix` retains the freshest delivered fix for the monitor's other cached-fix reads.
+///
+/// Owns its manager instead of routing through the Location module's provider: on a wrapper cold
+/// wake `CustomerIO.initialize` has not run, and movement passes must work in exactly that state.
+@MainActor
+final class MovementFixResolver: NSObject, @preconcurrency CLLocationManagerDelegate {
+    private let logger: Logger
+    private let maxAge: TimeInterval
+    private let requestTimeout: TimeInterval
+
+    /// Created lazily so tests using the `requestFreshFix` seam never touch CoreLocation.
+    private lazy var manager: CLLocationManager = {
+        let manager = CLLocationManager()
+        manager.desiredAccuracy = kCLLocationAccuracyHundredMeters
+        manager.delegate = self
+        return manager
+    }()
+
+    /// Freshest fix this resolver has received, retained even when it arrives after a timeout.
+    private(set) var latestFix: CLLocation?
+    private var pendingCompletions: [(LocationData?) -> Void] = []
+    /// Newest cached fix seen while a request is in flight — the fallback on failure/timeout.
+    private var fallbackFix: CLLocation?
+    private var timeoutTask: Task<Void, Never>?
+
+    /// Test seam: replaces the manager's one-shot request. Tests inject this and then feed
+    /// `handleResolvedFix` / `handleRequestFailure` directly.
+    var requestFreshFix: (() -> Void)?
+
+    init(
+        logger: Logger,
+        maxAge: TimeInterval = GeofenceConstants.movementFixMaxAge,
+        requestTimeout: TimeInterval = GeofenceConstants.movementFixRequestTimeout
+    ) {
+        self.logger = logger
+        self.maxAge = maxAge
+        self.requestTimeout = requestTimeout
+    }
+
+    deinit {
+        timeoutTask?.cancel()
+    }
+
+    /// Completes with a fix no older than `maxAge` when one can be obtained, exactly once per call.
+    /// `cached` should be the caller's best currently-known fix.
+    func resolve(cached: CLLocation?, completion: @escaping (LocationData?) -> Void) {
+        let age = cached.map { -$0.timestamp.timeIntervalSinceNow }
+        if let cached, let age, age <= maxAge {
+            logger.geofenceMovementFixResolved(ageSeconds: age, requested: false)
+            completion(locationData(from: cached))
+            return
+        }
+        logger.geofenceMovementFixStale(ageSeconds: age)
+        if let cached, fallbackFix.map({ cached.timestamp > $0.timestamp }) ?? true {
+            fallbackFix = cached
+        }
+        pendingCompletions.append(completion)
+        guard pendingCompletions.count == 1 else { return }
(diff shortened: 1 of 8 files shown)


Sources/LocationGeofence/Monitor/MovementFixResolver.swift
```diff
@@ -0,0 +1,132 @@
...
+            completion(locationData(from: cached))
+            return
+        }
+        logger.geofenceMovementFixStale(ageSeconds: age)
+        if let cached, fallbackFix.map({ cached.timestamp > $0.timestamp }) ?? true {
+            fallbackFix = cached
+        }
+        pendingCompletions.append(completion)
+        guard pendingCompletions.count == 1 else { return }
+        startTimeout()
```
Shahroz16, line 78: Could we hold a background-task assertion across this resolution and the movement re-registration? A region wake gives the app only a short execution window, and this request can use the full 10 seconds before `onTransition` or `handleMovement` starts. On a cold wake with a slow fix, iOS can suspend the app before the timeout or re-center completes; the EXIT has already been consumed, so there may be no retry. Starting the assertion before the request and keeping it through the movement pass would make the no-drop behavior reliable.

Sources/LocationGeofence/Monitor/MovementFixResolver.swift
```diff
@@ -0,0 +1,132 @@
...
+            requestFreshFix()
+        } else {
+            manager.requestLocation()
+        }
+    }
+
+    // MARK: - CLLocationManagerDelegate
+
+    func locationManager(_ manager: CLLocationManager, didUpdateLocations locations: [CLLocation]) {
+        guard let fix = locations.last, CLLocationCoordinate2DIsValid(fix.coordinate) else { return }
```
Shahroz16: Could we validate recency before completing here? Core Location can initially deliver a cached value to a newly started manager and recommends checking timestamps. This accepts the first coordinate-valid result regardless of age, so the same stale fix that triggered the request can be treated and logged as fresh. Waiting for a recent, valid-accuracy result and falling back only at the existing timeout would cover that path; a stale-callback test would lock it in.

Sources/LocationGeofence/Monitor/MovementFixResolver.swift
mrehan27: Good catch — the trigger EXIT is consumed at dispatch, so a window expiring mid-flight would lose the pass with no retry. Done in 1657d312: the fresh-fix request holds a background-task assertion for its whole life (released on fix/failure/timeout), and the movement pass runs under its own via the binder, both through the existing `BackgroundTaskRunner`. One scoping note: on a cold wake the manager cache reseeds from the system fix, so the fresh branch runs with no request — the exposure is the suspended-alive stale path, which is exactly the population this fix serves.

Sources/LocationGeofence/Monitor/MovementFixResolver.swift
mrehan27: Agreed — an echoed cached fix would have completed the pass while logging it as freshly requested. Done in 1657d312: `didUpdateLocations` now accepts a fix only when it is recent (≤ `movementFixMaxAge`) with valid accuracy; anything else is retained for `latestFix` bookkeeping but keeps the request waiting, and the existing timeout falls back. Locked in with the stale-echo and invalid-accuracy tests, plus a sim run with the guard forced to reject every delivery: the pass completed through the timeout fallback instead of stalling.

Shahroz16 (approved):

PR #1212, opened 2026-08-13
ci: move CI onto supported node and ruby runtimes

mahmoud-elmorabea:
Node 18 → 24, Ruby 3.2 → 3.4, and the last actions still on pre-node20 runtimes.

Contributes to MBL-2245

---

> [!NOTE]
> **Medium Risk**
> Touches all primary CI paths (tests, sample builds, deploy, PR checks) via runtime and fastlane dependency bumps; SDK size reporting logic changed but remains CI-only with no product code impact.
> 
> **Overview**
> **CI runtimes:** Node **18 → 24** (Danger in `pr-review`), Ruby **3.2 → 3.4** in sample-app builds, API-breaking checks, and the shared iOS test workflow (required because updated gems such as **excon 1.7** need Ruby ≥ 3.3). **`ruby/setup-ruby`** is pinned to **v1.321.0** everywhere it’s used.
> 
> **SDK size reports:** The three **bloaty** steps no longer use **`mathiasvr/command-output`**; they run as plain **`bash`** steps and write multiline output to **`GITHUB_OUTPUT`** with a **random heredoc delimiter** so report text can’t accidentally terminate the block.
> 
> **Deploy workflow:** **`sd`** installation moves from a forked action to **`kenji-miyake/setup-sd@v2.0.0`**.
> 
> **Ruby deps:** **`Gemfile.lock`** is refreshed (notably **fastlane 2.237.0**, transitive AWS/Google gems, **Bundler 2.6.9**, and **arm64-darwin-25**).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1d0931f9e6a0ed80ccd81a58513abc7b1f6ace49. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


.github/actions/generate-sdk-size-report/action.yml
```diff
@@ -71,25 +71,36 @@ runs:
...
-          -n 0 
-    
+    run: |
+      report=$(bloaty --source-filter ".*(customerio-ios\/Sources).*" \
+        -d compileunits --debug-file=${{ steps.build-sample-app.outputs.app-xcarchive-path }}/dSYMs/APN\ UIKit.app.dSYM/Contents/Resources/DWARF/APN\ UIKit \
+        ${{ steps.build-sample-app.outputs.app-xcarchive-path }}/Products/Applications/APN\ UIKit.app/APN\ UIKit \
+        -n 0)
+      echo "$report"
+      {
+        echo 'stdout<<CIO_REPORT_EOF'
```
Shahroz16: Non-blocking hardening: could we generate a unique delimiter for these multiline outputs instead of using the fixed `CIO_REPORT_EOF` value? That avoids an output line accidentally terminating the GitHub output block. The same pattern appears in all three report steps.

Shahroz16 (approved):

PR #1215, opened 2026-08-17
chore(samples): fix APN universal-link matcher

Shahroz16:
## Summary

- fix the APN UIKit sample's malformed universal-link matcher
- compare URL schemes and hosts case-insensitively, as required by URL semantics
- replace placeholder sample tests with focused positive and negative matcher coverage

## Architectural decision

The previously proposed public scene lifecycle coordinator was removed. This PR now makes no production SDK or public API change. Host applications and wrappers continue routing their own lifecycle callbacks through the already released `CustomerIO.liveActivities.handleWidgetUrl` primitive.

## Release impact

This is a sample-only correction and should be squash-merged with the current `chore(samples):` title. It does not require an SDK release.

## Validation

- APN UIKit sample build-for-testing: passed
- `APN UIKitTests/Tests`: 6/6 passed on iPhone 17 Pro, iOS 26.5 simulator
- SwiftFormat and strict SwiftLint: passed, 0 violations across 735 files
- pre-commit and pre-push hooks: passed
- `git diff --check`: passed
- exact net diff against `main`: 2 sample files

## Adversarial review

Claude Code Opus completed a fresh high-effort source-command review of exact head `c522289c21cf0c841b1a115fb48f1dac0bbb02cb`. It found no production SDK, architecture, compatibility, PII, or merge-blocking issue and classified the change as L1 risk. Its remaining suggestions would broaden the supported path behavior or add abstraction without a demonstrated requirement, so they were intentionally not included in this narrow sample fix.


mahmoud-elmorabea (changes requested):
`handleSceneOpenURLs` (`CioAppLifecycleCoordinator.swift:203`, guard `:210`) rejects any multi-context activation and routes nothing, and `SceneDelegate.swift:37` went from a per-URL loop to a single call — two coalesced opens now lose both the deep link and the `opened` metric for each delivery.

- Restore per-URL handling for the warm callback; keep fail-closed on the single-activation connect path.
- Rebase off #1211 so this ships independently — that PR has no Xcode 27 dependency.
- Needs an owner before Xcode 27 stable: AppDelegate-only stops being viable once apps rebuild against the iOS 27 SDK, and hand-wiring the coordinator is a step back from `CioAppDelegateWrapper`'s auto-wiring.

Shahroz16:
Addressed all three points in `fa2f7bbf`:

- Warm `scene(_:openURLContexts:)` now routes every coalesced URL exactly once and returns handled when any route accepts it. Cold `willConnect` remains fail-closed for multiple candidates.
- The PR is unstacked and targets `main`; it contains no #1211 implementation.
- The coordinator now has one instance per scene. SwiftUI documentation also clarifies that `UIApplicationDelegateAdaptor` retains only application-global initialization/token/notification ownership while UI activation remains topology-exclusive.

Focused coordinator tests (26), package test build, APN sample build, format, and strict lint are green. Claude Opus reviewed the correction; its applicable ownership/documentation findings were incorporated. Please re-review when ready.

api-docs/CioDataPipelines.api
```diff
@@ -6,6 +6,23 @@ public final class CioDataPipelines.AutoTrackingScreenViews {
 	public var autoScreenViewBody: (() -> Map<String, Any>)?;
 	public fun init(filterAutoScreenViewEvents: ((UIViewController) -> Boolean)? = nil, autoScreenViewBody: (() -> List<String: Any>)? = nil);
 }
+public final class CioDataPipelines.CioAppLifecycleCoordinator {
```
mahmoud-elmorabea: I think this might be convenient for us but confusing for its consumers. How about we split `CioAppLifecycleCoordinator` into multiple coordinators -> `CioSceneLifecycleCoordinator`, `CioAppLifecycleCoordinator`...etc?

That way a consumer only needs to chose which correct type to use and auto-complete helps them instead of work against them by providing potentially incorrect calls that will only be rejected at runtime

mahmoud-elmorabea: We also lose the `Topology` concept which might not be digestible widely

Shahroz16: Agreed. The unreleased facade is now split into `CioAppDelegateLifecycleCoordinator`, `CioSceneLifecycleCoordinator`, and `CioSwiftUILifecycleCoordinator`. The topology enum and runtime topology-rejection case are removed, so autocomplete exposes only callbacks that are legal for the chosen host integration. Samples and the public API baseline were updated; the package builds, 24 focused coordinator tests pass, and format/lint are clean. Claude Opus 5 independently agreed with the split and accepted the corrected design after its test and documentation follow-ups were addressed.

Sources/DataPipeline/Integration/CioAppLifecycleCoordinator.swift
```diff
@@ -0,0 +1,134 @@
...
+    func handleConnection<URLActivation>(
+        _ activation: SceneConnectionActivation<URLActivation>,
+        routeURL: (URLActivation) -> Bool,
+        continueUserActivity: (NSUserActivity) -> Bool,
+        performShortcut: (UIApplicationShortcutItem) -> Bool
+    ) -> CioSceneLifecycleHandlingResult {
+        let candidateCount = activation.urlActivations.count
+            + activation.userActivities.count
+            + (activation.shortcutItem == nil ? 0 : 1)
+        guard candidateCount <= 1 else {
```
mahmoud-elmorabea: **Behavior regression vs `main` on the cold Live Activity path.**

Pre-diff, `scene(_:willConnectTo:options:)` called `handle(urlContexts: connectionOptions.urlContexts)`, which looped and routed every context. Now any `candidateCount >= 2` returns early and runs *no* routing closure.

The case that worries me isn't two URLs — it's one URL plus one user activity. This PR newly counts `connectionOptions.userActivities` toward `candidateCount`, so a cold Live Activity tap that iOS co-delivers with a Handoff/Spotlight `NSUserActivity` now goes from "widget URL routed, `opened` reported" to "nothing happens." That's a lost delivery metric and a dead deep link, and it fails closed silently apart from a log line.

It also contradicts the warm path 30 lines down: `handleOpenURLs` routes every URL from the identical `Set<UIOpenURLContext>` shape. Two opposite policies for the same data is going to be hard to explain to a host.

If the intent is "don't guess which of two UI intents the user meant," that reasoning applies to *navigation*, not to the metric report — `handleWidgetUrl` is idempotent per delivery and reporting it is never ambiguous. Could the URL candidates route unconditionally, with the ambiguity rule limited to activity-vs-shortcut-vs-URL *navigation*?

Sources/DataPipeline/Integration/CioAppLifecycleCoordinator.swift
```diff
@@ -0,0 +1,134 @@
...
+            return rejectAmbiguous(callback: "scene connection", candidateCount: candidateCount)
+        }
+        guard candidateCount == 1 else {
+            return activation.hasNotificationResponse ? .notificationOwnedByApplication : .noActivation
```
mahmoud-elmorabea: `.notificationOwnedByApplication` is only reachable when `candidateCount == 0`, so the signal is erased exactly where a host would act on it.

Cold connection with `notificationResponse != nil` **and** one URL context returns `.handled`/`.unhandled`; with two candidates it returns `.rejectedAmbiguousInput` (your own `testSceneHandleConnection_givenNotificationAndMultipleURLs_thenRejectsSceneAmbiguity` asserts this). Either way the host can't tell a notification response was also delivered.

A host switching on the result to decide "do I still need to run my own notification handling?" gets the wrong answer — which is the failure mode a single aggregate enum was meant to prevent. Consider returning a small struct (`uiOutcome` + `hadNotificationResponse`) so the two independent facts stay independent.

Sources/DataPipeline/Integration/CioAppLifecycleCoordinator.swift
```diff
@@ -0,0 +1,134 @@
...
+        }
+        if let urlActivation = activation.urlActivations.first {
+            return routeOnce { routeURL(urlActivation) }
+        }
+        if let userActivity = activation.userActivities.first {
+            return routeOnce { continueUserActivity(userActivity) }
+        }
+        guard let shortcutItem = activation.shortcutItem else { return .noActivation }
```
mahmoud-elmorabea: This guard is unreachable. Getting here requires `candidateCount == 1` with both arrays empty, which means `shortcutItem != nil` by construction. The `else { return .noActivation }` branch can never execute, and it reads as a real fallback — worth collapsing to `activation.shortcutItem!` with a comment, or restructuring the three checks into a `switch` so the exhaustiveness is visible.

Sources/DataPipeline/Integration/CioAppLifecycleCoordinator.swift
```diff
@@ -0,0 +1,134 @@
...
+    /// Handles warm URL activations delivered to a UIKit scene.
+    ///
+    /// UIKit may coalesce multiple independent URL opens into one callback. Each URL is routed once
+    /// in unspecified order; the aggregate result is handled when at least one route accepts it.
+    @discardableResult
+    public func handleOpenURLContexts(
+        _ urlContexts: Set<UIOpenURLContext>,
+        route: (UIOpenURLContext) -> Bool
+    ) -> CioSceneLifecycleHandlingResult {
+        handleOpenURLs(Array(urlContexts), route: route)
```
mahmoud-elmorabea: `Array(urlContexts)` iterates in `Set` hash order, which is per-process seeded. When a warm callback coalesces two URLs that both mutate UI, which screen the user ends up on differs run to run on the same build.

The doc comment normalizes this ("in unspecified order") but a host has no hook to make it deterministic. Sorting by `url.absoluteString` before routing would at least make it reproducible.

Sources/DataPipeline/Integration/CioAppLifecycleCoordinator.swift
```diff
@@ -0,0 +1,134 @@
...
+        let userActivities: [NSUserActivity]
+        let shortcutItem: UIApplicationShortcutItem?
+        let hasNotificationResponse: Bool
+    }
+
+    /// Creates a UIKit scene lifecycle coordinator.
+    public init() {
+        // This is a public host-integration entry point. Resolve the shared logger lazily so the
+        // SDK can finish top-level module initialization before the first lifecycle callback.
+        self.loggerProvider = { DIGraphShared.shared.logger }
```
mahmoud-elmorabea: This reaches into `DIGraphShared` from an object created once per scene delegate. Repo `CLAUDE.md` is explicit on both counts:

- line 24: "Always use a constructor-based dependency injection pattern, and use `DIGraphShared` only for top-level module initialization"
- line 112: "`DIGraphShared` serves as the central dependency registry, which should be used only for top-level module initialization"

A scene coordinator isn't top-level module init. Separately, `loggerProvider: () -> Logger` keeps a closure alive for the object's lifetime just to defer one lookup (and in `init(logger:)` the closure captures a value you already own). `private lazy var logger: Logger = DIGraphShared.shared.logger` with a `Logger?` injection point gets the same laziness with no captured environment.

Sources/DataPipeline/Integration/CioAppLifecycleTypes.swift
```diff
@@ -0,0 +1,22 @@
+/// The outcome of one scene lifecycle activation ingress.
+public enum CioSceneLifecycleHandlingResult: Equatable, Sendable {
```
mahmoud-elmorabea: We ship source (no `-enable-library-evolution`), so this public enum is effectively frozen the moment it's released: any customer can `switch` over it exhaustively without `@unknown default`, and adding a sixth case in a minor version breaks their build.

The five cases plainly anticipate growth, and this surface has already been reshaped once in this PR. Either make it a struct with static members, or document that hosts should read `wasHandled` and include `@unknown default` if they switch.

Apps/APN-UIKit/APN UIKit/SceneDelegate.swift
```diff
@@ -103,10 +111,11 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
...
-        guard let universalLinkUrl = userActivity.webpageURL else {
-            return
-        }
+        _ = route(userActivity: userActivity)
+    }
 
-        _ = deepLinkHandler.handleUniversalLinkDeepLink(universalLinkUrl)
+    private func route(userActivity: NSUserActivity) -> Bool {
+        guard let universalLinkURL = userActivity.webpageURL else { return false }
+        return deepLinkHandler.handleUniversalLinkDeepLink(universalLinkURL)
```
mahmoud-elmorabea: Heads up that this makes the sample demonstrate the new API incorrectly, through no fault of the PR.

`handleUniversalLinkDeepLink` returns `doesMatchUniversalLink(url)`, which compares against `URL(string: "http://applinks:ciosample.page.link/spm")`. That string has a non-numeric port, so `URL(string:)` returns **nil** — I ran it on Swift 6.2 to confirm. `universalLink?.host` is nil, so the comparison is always `false`.

That bug is pre-existing on `main` and this PR doesn't touch `DeepLinksHandlerUtil.swift`. But before this PR the Bool was discarded; now it becomes the value of a *public SDK contract*. On a successful cold universal-link launch the sample posts `showDeepLinkScreenOnDashboard` (navigation did happen) while `handleConnection` reports `.unhandled` / `wasHandled == false`. Any host copying this pattern and using `wasHandled` to gate a fallback will double-handle.

Worth fixing the helper here (`URL(string: "https://ciosample.page.link/spm")`) so the reference integration tells the truth.

Apps/APN-UIKit/APN UIKit/SceneDelegate.swift
```diff
@@ -103,10 +111,11 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
     // To handle Universal Links from the Customer.io SDK, see `AppDelegate` file for implementation.
     // Learn more: https://customer.io/docs/sdk/ios/push/#universal-links-deep-links
     func scene(_ scene: UIScene, continue userActivity: NSUserActivity) {
-        guard let universalLinkUrl = userActivity.webpageURL else {
-            return
-        }
+        _ = route(userActivity: userActivity)
```
mahmoud-elmorabea: Two seats now call `route(userActivity:)`: the coordinator's `continueUserActivity` closure (line 32, from `connectionOptions.userActivities`) and this one. `scene(_:continue:)` never enters the coordinator, so the coordinator *cannot* dedupe — and its own doc comment says repeated identical values in later callbacks are treated as new occurrences.

On any launch path where iOS puts the activity in `connectionOptions.userActivities` and *also* calls `scene(_:continue:)`, `handleUniversalLinkDeepLink` fires twice → two `showDeepLinkScreen*` notifications → duplicate popup. Pre-diff the cold path ignored `userActivities` entirely, so this exposure is new to this PR.

I'm not certain iOS does both on any given path, but the design makes it unverifiable from inside the SDK, which is the deeper issue: the coordinator claims to own scene activation but has seats for only 2 of the 4 activation callbacks (no warm `scene(_:continue:)`, no `windowScene(_:performActionFor:)`). Either give it every seat so dedup is possible, or drop `continueUserActivity` from the cold path and let `scene(_:continue:)` be the single owner.

Tests/DataPipeline/APITest.swift
```diff
@@ -118,4 +119,14 @@ class DataPipelineAPITest: UnitTest {
             autoScreenViewBody: { self.dictionaryData }
         )
     }
+
+    @MainActor
+    func test_sceneLifecycleCoordinatorPublicFunctions() throws {
+        try skipRunningTest()
+
+        let sceneCoordinator = CioSceneLifecycleCoordinator()
+        let _: CioSceneLifecycleHandlingResult = sceneCoordinator.handleOpenURLContexts([]) { _ in true }
```
mahmoud-elmorabea: This omits `handleConnection(options:routeURL:continueUserActivity:performShortcut:)` — the method customers actually integrate against.

Per this file's own header ("If a public function of the SDK is modified, this test class will not successfully compile"), dropping `public` from `handleConnection`, renaming `options:` → `connectionOptions:`, or reordering its four closures all compile green today. The `api-docs` check compares generated text, not call-site compilability, so it doesn't cover this either. One more line here closes it.

Tests/DataPipeline/Integration/CioSceneLifecycleCoordinatorTests.swift
```diff
@@ -0,0 +1,317 @@
...
+
+    private func makeSceneCoordinator() -> CioSceneLifecycleCoordinator {
+        CioSceneLifecycleCoordinator(logger: logger)
+    }
+
+    private func makeShortcut() -> UIApplicationShortcutItem {
+        UIApplicationShortcutItem(type: "settings", localizedTitle: "Settings")
+    }
+
+    private func handleConnection(
```
mahmoud-elmorabea: All 16 tests go through the internal generic overloads via `makeSceneActivation` / `handleOpenURLs([URL])`. The public wrappers are never called, so these four field reads have zero coverage:

```swift
urlActivations: Array(connectionOptions.urlContexts),
userActivities: Array(connectionOptions.userActivities),
shortcutItem: connectionOptions.shortcutItem,
hasNotificationResponse: connectionOptions.notificationResponse != nil
```

Swapping `urlContexts` for `userActivities` there, or inverting the `!= nil`, keeps all 16 green while breaking every real cold launch.

I accept `UIScene.ConnectionOptions` has no public initializer, so this may be genuinely untestable — the generic seam is a reasonable response to that. But then the adapter is our unverified surface and it should be as boring as possible, plus called from `APITest` so at least the signature is pinned.

Apps/APN-UIKit/APN UIKit/SceneDelegate.swift
```diff
@@ -72,24 +80,24 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
...
-            if #available(iOS 17.2, *), url.host == LiveActivitiesViewController.deepLinkHost {
-                routeToLiveActivities()
-                continue
-            }
-            _ = deepLinkHandler.handleAppSchemeDeepLink(url)
+    private func route(url: URL) -> Bool {
+        // A tap on a Customer.io Live Activity arrives as a CIO tracking URL. Report the `opened`
+        // for the exact delivery shown, then route its redirect. A delivery-only URL is handled even
+        // though it has no redirect. Non-Customer.io URLs pass through unchanged.
+        guard let destination = CustomerIO.liveActivities.handleWidgetUrl(url) else {
```
mahmoud-elmorabea: The refactor dropped the guard-rail comment that was here: *"Activities render from iOS 16.2, so this must not be gated any higher — only the in-app routing below needs 17.2."*

`handleWidgetUrl` is still correctly outside the `#available(iOS 17.2, *)` check, but nothing records why any more. The natural future cleanup — wrapping the whole body in the 17.2 check, since `routeToLiveActivities` needs it — compiles clean and silently stops reporting `opened` for everyone on 16.2–17.1. No test catches that. Please keep the comment.

Apps/CocoaPods-FCM/src/App.swift
```diff
@@ -32,48 +34,33 @@ struct MainApp: App {
...
-                        switch command {
-                        case "login":
-                            userManager.logout() // will force the app's UI to navigate back to login screen
-                        case "dashboard":
-                            settingsScreen = nil // as long as user is logged in, this will make dashboard show
-                        case "settings":
-                            var siteId: String?
-                            var cdpApiKey: String?
+                .onOpenURL { incomingURL in
+                    // A Customer.io Live Activity URL reports the exact opened delivery before
```
mahmoud-elmorabea: Scope check: this PR narrows the scene lifecycle API, but this file only got a control-flow reformat (`if let` → `guard`, `host!` → `host ?? ""`) and lost its integration docs in the process — the "how to implement deep links in a SwiftUI app" note, the universal-link domain, the `cocoapods-fcm://` scheme, the pop-up behavior, and the debugging link.

This is the reference SwiftUI sample customers read to learn the integration. A reader of the new version can't tell which URLs it accepts. Either restore the comments or revert the file — the reformat isn't buying anything here.

Sources/DataPipeline/Integration/CioAppLifecycleCoordinator.swift
mahmoud-elmorabea: 🔴  **Behavior regression vs `main` on the cold Live Activity path.**

Pre-diff, `scene(_:willConnectTo:options:)` called `handle(urlContexts: connectionOptions.urlContexts)`, which looped and routed every context. Now any `candidateCount >= 2` returns early and runs *no* routing closure.

The case that worries me isn't two URLs — it's one URL plus one user activity. This PR newly counts `connectionOptions.userActivities` toward `candidateCount`, so a cold Live Activity tap that iOS co-delivers with a Handoff/Spotlight `NSUserActivity` now goes from "widget URL routed, `opened` reported" to "nothing happens." That's a lost delivery metric and a dead deep link, and it fails closed silently apart from a log line.

It also contradicts the warm path 30 lines down: `handleOpenURLs` routes every URL from the identical `Set<UIOpenURLContext>` shape. Two opposite policies for the same data is going to be hard to explain to a host.

Sources/DataPipeline/Integration/CioAppLifecycleCoordinator.swift
mahmoud-elmorabea: `.notificationOwnedByApplication` is only reachable when `candidateCount == 0`, so the signal is erased exactly where a host would act on it.

Cold connection with `notificationResponse != nil` **and** one URL context returns `.handled`/`.unhandled`; with two candidates it returns `.rejectedAmbiguousInput` (your own `testSceneHandleConnection_givenNotificationAndMultipleURLs_thenRejectsSceneAmbiguity` asserts this). Either way the host can't tell a notification response was also delivered.

A host switching on the result to decide "do I still need to run my own notification handling?" gets the wrong answer — which is the failure mode a single aggregate enum was meant to prevent. Consider returning a small struct (`uiOutcome` + `hadNotificationResponse`) so the two independent facts stay independent.

Sources/DataPipeline/Integration/CioAppLifecycleCoordinator.swift
mahmoud-elmorabea: This guard is unreachable. Getting here requires `candidateCount == 1` with both arrays empty, which means `shortcutItem != nil` by construction. The `else { return .noActivation }` branch can never execute, and it reads as a real fallback — worth collapsing to `activation.shortcutItem!` with a comment, or restructuring the three checks into a `switch` so the exhaustiveness is visible.

Sources/DataPipeline/Integration/CioAppLifecycleCoordinator.swift
mahmoud-elmorabea: `Array(urlContexts)` iterates in `Set` hash order, which is per-process seeded. When a warm callback coalesces two URLs that both mutate UI, which screen the user ends up on differs run to run on the same build.

The doc comment normalizes this ("in unspecified order") but a host has no hook to make it deterministic. Sorting by `url.absoluteString` before routing would at least make it reproducible.

Sources/DataPipeline/Integration/CioAppLifecycleTypes.swift
mahmoud-elmorabea: We ship source (no `-enable-library-evolution`), so this public enum is effectively frozen the moment it's released: any customer can `switch` over it exhaustively without `@unknown default`, and adding a sixth case in a minor version breaks their build.

The five cases plainly anticipate growth, and this surface has already been reshaped once in this PR. Either make it a struct with static members, or document that hosts should read `wasHandled` and include `@unknown default` if they switch.

Apps/APN-UIKit/APN UIKit/SceneDelegate.swift
mahmoud-elmorabea: Heads up that this makes the sample demonstrate the new API incorrectly, through no fault of the PR.

`handleUniversalLinkDeepLink` returns `doesMatchUniversalLink(url)`, which compares against `URL(string: "http://applinks:ciosample.page.link/spm")`. That string has a non-numeric port, so `URL(string:)` returns **nil** — I ran it on Swift 6.2 to confirm. `universalLink?.host` is nil, so the comparison is always `false`.

That bug is pre-existing on `main` and this PR doesn't touch `DeepLinksHandlerUtil.swift`. But before this PR the Bool was discarded; now it becomes the value of a *public SDK contract*. On a successful cold universal-link launch the sample posts `showDeepLinkScreenOnDashboard` (navigation did happen) while `handleConnection` reports `.unhandled` / `wasHandled == false`. Any host copying this pattern and using `wasHandled` to gate a fallback will double-handle.

Worth fixing the helper here (`URL(string: "https://ciosample.page.link/spm")`) so the reference integration tells the truth.

Apps/APN-UIKit/APN UIKit/SceneDelegate.swift
mahmoud-elmorabea: Two seats now call `route(userActivity:)`: the coordinator's `continueUserActivity` closure (line 32, from `connectionOptions.userActivities`) and this one. `scene(_:continue:)` never enters the coordinator, so the coordinator *cannot* dedupe — and its own doc comment says repeated identical values in later callbacks are treated as new occurrences.

On any launch path where iOS puts the activity in `connectionOptions.userActivities` and *also* calls `scene(_:continue:)`, `handleUniversalLinkDeepLink` fires twice → two `showDeepLinkScreen*` notifications → duplicate popup. Pre-diff the cold path ignored `userActivities` entirely, so this exposure is new to this PR.

I'm not certain iOS does both on any given path, but since would serve as guidance to consumers, let's verify this

Tests/DataPipeline/APITest.swift
mahmoud-elmorabea: This omits `handleConnection(options:routeURL:continueUserActivity:performShortcut:)` — the method customers actually integrate against.

Per this file's own header ("If a public function of the SDK is modified, this test class will not successfully compile"), dropping `public` from `handleConnection`, renaming `options:` → `connectionOptions:`, or reordering its four closures all compile green today. The `api-docs` check compares generated text, not call-site compilability, so it doesn't cover this either. One more line here closes it.

Tests/DataPipeline/Integration/CioSceneLifecycleCoordinatorTests.swift
mahmoud-elmorabea: All 16 tests go through the internal generic overloads via `makeSceneActivation` / `handleOpenURLs([URL])`. The public wrappers are never called, so these four field reads have zero coverage:

```swift
urlActivations: Array(connectionOptions.urlContexts),
userActivities: Array(connectionOptions.userActivities),
shortcutItem: connectionOptions.shortcutItem,
hasNotificationResponse: connectionOptions.notificationResponse != nil
```

Swapping `urlContexts` for `userActivities` there, or inverting the `!= nil`, keeps all 16 green while breaking every real cold launch.

I accept `UIScene.ConnectionOptions` has no public initializer, so this may be genuinely untestable — the generic seam is a reasonable response to that. But then the adapter is our unverified surface and it should be as boring as possible, plus called from `APITest` so at least the signature is pinned.

Apps/CocoaPods-FCM/src/App.swift
mahmoud-elmorabea: Scope check: this PR narrows the scene lifecycle API, but this file only got a control-flow reformat (`if let` → `guard`, `host!` → `host ?? ""`) and lost its integration docs in the process — the "how to implement deep links in a SwiftUI app" note, the universal-link domain, the `cocoapods-fcm://` scheme, the pop-up behavior, and the debugging link.

This is the reference SwiftUI sample customers read to learn the integration. A reader of the new version can't tell which URLs it accepts. Either restore the comments or revert the file — the reformat isn't buying anything here.

Shahroz16:
I removed the public coordinator and enum entirely. The PR now has no production SDK or public API change; its net diff is only the pre-existing APN sample URL-matcher fix plus three direct tests. Flutter now adapts UIScene callbacks in the wrapper and calls the already released `CustomerIO.liveActivities.handleWidgetUrl` primitive through native 4.7.4, so there is no #1215 pin or cross-repo coordinator dependency. All of the coordinator-specific threads are therefore outdated. Could you please re-review the reduced two-file PR when convenient?

mahmoud-elmorabea (approved):

PR #1216, opened 2026-08-17
ci: verify Xcode 27 nightly launches

Shahroz16:
## Summary

- Extends the existing Xcode 27 preview workflow to install and launch both native APN and CocoaPods FCM simulator products.
- Requires the expected executable to remain alive for 10 seconds and publishes structured launch diagnostics when it does not.
- Uses the merged shared simulator-launch action from customerio/mobile-ci-tools at main.

## Release impact

CI-only. This PR changes no SDK source, package metadata, or release artifact. The preview workflow runs nightly, by manual dispatch, and when the workflow itself changes.

## Validation

- actionlint and git diff --check pass.
- MessagingPushAPN builds successfully on the booted iOS 26.5 simulator.
- make format, strict lint, pre-commit, and pre-push hooks pass with zero violations.
- The rewritten exact-head APN and FCM preview jobs are the authoritative hosted Xcode 27 validation.

No unit-test classes remain in this scope; the removed cross-repository watchdog and its Python tests were intentionally split out.

## Boundary

This proves simulator compilation, installation, launch, executable identity, and short process survival. It does not prove lifecycle callback delivery, real APN/FCM delivery, signing, physical-device behavior, or App Store acceptance.

---

> [!NOTE]
> **Low Risk**
> CI-only workflow and documentation changes; no SDK, signing, or runtime product code is modified.
> 
> **Overview**
> The **Xcode 27 preview compatibility** workflow no longer stops at unsigned simulator builds. After each fixture compiles, it resolves the built `.app`, installs it on the simulator via the shared **`launch-simulator-app`** action, and requires the process to stay alive for **10 seconds** (`LAUNCH_SURVIVAL_SECONDS`).
> 
> The matrix now carries per-fixture **`app-name`** values so both native APN and CocoaPods FCM products resolve to the correct bundle path. In-workflow **GitHub step summary** blocks for compile/setup/toolchain outcomes are removed in favor of the shared action’s launch classification and an explicit **launch log** uploaded with build (and pod) logs; artifact upload **fails if no log files exist**.
> 
> Docs describe the new smoke scope (compile → install → launch → short survival), **`main`** concurrency so manual runs replace stale nightlies, and the review invariant that **compile-only passes can miss launch-time lifecycle failures**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c504562d1410d9bffcd406c2d23ce1163c14a09f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Shahroz16:
Hosted validation is green at 67dbedf: the Xcode 27 APN and CocoaPods FCM jobs both compiled, installed, launched, and survived 10 seconds; the cross-repository nightly watchdog tests also passed. The unrelated full-suite async timeouts were rerun and the exact rerun passed.\n\nThis remains draft until mobile-ci-tools #16 merges, the action reference changes to `@main`, and all three wrapper launch workflows report complete runs so the watchdog cannot accept compile-only results.

.github/workflows/xcode27-nightly-watchdog.yml
```diff
@@ -0,0 +1,56 @@
...
+      - name: Verify scheduled runs completed successfully
+        if: github.event_name != 'pull_request'
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+        run: |
+          python3 scripts/ci/check_xcode27_nightly_runs.py \
+            --target customerio/customerio-ios:ios-toolchain-compatibility.yml \
+            --target customerio/customerio-flutter:ios-toolchain-compatibility.yml \
+            --target customerio/customerio-expo-plugin:ios-toolchain-compatibility.yml \
+            --max-age-hours 26
```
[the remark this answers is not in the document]
Shahroz16: Good catch. Fixed in 21b1a4d by minting a GitHub App installation token explicitly scoped to customerio-ios, customerio-flutter, and customerio-expo-plugin, then passing that token to the watchdog. The repository-scoped github.token is no longer used for cross-repository Actions reads. Unit tests and actionlint pass.

.github/workflows/xcode27-nightly-watchdog.yml
```diff
@@ -0,0 +1,56 @@
...
+
+      - name: Notify Mobile team of a failed or missing nightly
+        if: failure() && github.event_name == 'schedule'
+        uses: slackapi/slack-github-action@70cd7be8e40a46e8b0eced40b0de447bdb42f68e # v1.26.0
```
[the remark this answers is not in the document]
Shahroz16: Agreed. Fixed in 21b1a4d by making the Slack text accurately describe either an unhealthy/missing nightly or an internal watchdog failure, so classification-test and script failures are no longer misattributed.

.github/workflows/xcode27-nightly-watchdog.yml
```diff
@@ -0,0 +1,135 @@
...
+            }
+        env:
+          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_NOTIFY_RELEASES_WEBHOOK_URL }}
+          SLACK_WEBHOOK_TYPE: INCOMING_WEBHOOK
+
+      - name: Fail unhealthy watchdog run
+        if: always() && github.event_name != 'pull_request' && steps.verify.outputs.exit-code != '0'
+        env:
+          WATCHDOG_EXIT_CODE: ${{ steps.verify.outputs.exit-code }}
+        run: exit "${WATCHDOG_EXIT_CODE:-1}"
```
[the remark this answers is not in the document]
Shahroz16: This is intentional fail-closed watchdog behavior. If checkout, unit tests, or another prerequisite fails before `verify`, nightly health is genuinely unknown. The `failure() && exit-code == ""` Slack condition reports exactly that infrastructure state, and the final `always()` step ensures scheduled/manual monitoring does not accidentally finish green without a successful verification result. Pull-request runs are explicitly excluded from the final failure step. Restricting these steps to cases where `verify` produced an output would recreate the missing-run blind spot this follow-up is meant to close.

.github/workflows/ios-toolchain-compatibility.yml
```diff
@@ -135,12 +171,123 @@ jobs:
...
+            echo "**Fixture:** $FIXTURE"
+            echo '**Toolchain family:** Xcode 27 / iOS SDK 27'
+            echo "**Next step:** $next_step"
+          } >> "$GITHUB_STEP_SUMMARY"
+
+      - name: Collect launch diagnostics
+        if: >-
+          failure() &&
+          steps.launch.outcome == 'failure' &&
+          (steps.launch.outputs.failure-reason == 'launch-failed' ||
```
mahmoud-elmorabea: **This is the blocking one.** Gating on `steps.launch.outputs.failure-reason` contradicts the shared action's documented contract.

`mobile-ci-tools/README.md:33-35` (the action's own docs):

> A failed composite action is not required to propagate its mapped outputs, so diagnostics upload must not depend on the `log-path` output. If the action uses `continue-on-error: true`, gate that upload on `steps.<id>.outcome == 'failure'` because the job-level `failure()` predicate remains false.

And `action.yml` maps `failure-reason` off the same inner composite step as `log-path`, so the warning covers both.

Walk the nightly that matters: `APN UIKit` is rejected at launch under the iOS 27 SDK. The `launch` step exits 1, its mapped outputs aren't published, this condition evaluates against `''` and skips, no `.ips` is copied, the diagnostics directory is never created, and `Record launch failure` falls into its `''` arm and prints `launch-smoke-inconclusive`. The one artifact identifying the crash is silently absent on the one run the workflow exists for.

You already treat empty output as reachable at line 183. Gate this on `steps.launch.outcome == 'failure'` instead and it works regardless of output propagation.

.github/workflows/ios-toolchain-compatibility.yml
```diff
@@ -135,12 +171,123 @@ jobs:
...
+              ;;
+            did-not-survive)
+              classification=fixture-exited-early
+              next_step='inspect the launch diagnostics for the crash or process-identity change'
+              ;;
+            invalid-input|invalid-app|sdk-mismatch)
+              classification=launch-contract-invalid
+              next_step='inspect the built product metadata and smoke-test inputs'
+              ;;
+            *)
```
mahmoud-elmorabea: `unexpected-error` is the one documented `failure-reason` value this `case` doesn't cover, and it's reachable after install+launch.

The action documents the full vocabulary in `action.yml:44-46` and `README.md:27-31`: `invalid-input`, `invalid-app`, `sdk-mismatch`, `runtime-unavailable`, `runtime-selection-failed`, `simulator-boot-failed`, `install-failed`, `launch-failed`, `did-not-survive`, `unexpected-error`, `none`. You cover all of them except `unexpected-error` — `launch.sh` calls `fail unexpected-error` from three sites plus its global ERR trap, and some of those are in the `ps`-inspection path that runs *after* the app is installed and launched.

Two bad outcomes when it fires: the summary prints `::warning::unrecognized launch failure reason: unexpected-error` — which is false, it's a documented value — and because it's not in the `Collect launch diagnostics` allowlist on line 227, the crash report for an app that was actually running gets discarded.

Worth adding an arm. Also consider that a `*)` arm hard-coding an upstream vocabulary while referencing the action at `@main` will drift silently — pinning to a tag or SHA makes the coupling explicit (`__tests__` in the Expo PR actively asserts the `@main` literal, which locks the drift in rather than catching it).

.github/workflows/ios-toolchain-compatibility.yml
```diff
@@ -98,8 +105,36 @@ jobs:
...
+        env:
+          FIXTURE: ${{ matrix.fixture }}
+        run: touch "$RUNNER_TEMP/${FIXTURE}-launch-start"
+
+      - name: Install and launch fixture
+        id: launch
+        uses: customerio/mobile-ci-tools/github-actions/ios/launch-simulator-app/v1@main
+        with:
+          app-path: ${{ env.APP_PRODUCT_PATH }}
+          expected-ios-major: '27'
```
mahmoud-elmorabea: `expected-ios-major: '27'` hard-requires an iOS 27 iPhone simulator runtime, and nothing in this job asserts or installs one.

`report-toolchain` runs `simctl list runtimes` for the *record*, and asserts SDK majors — not runtime availability. So if the `xcode-27` preview image ships Xcode 27 with only the iOS 26.x runtime pre-installed (the 27 runtime being a separate `xcodebuild -downloadPlatform iOS` fetch), report-toolchain passes, then `launch.sh`'s runtime selector finds no `SimRuntime.iOS-27` device and exits `runtime-unavailable` → `simulator-infrastructure-unavailable` → red, every night, with no compile-only fallback.

I can't verify the image contents from here, which is why I'm asking rather than asserting. But your Validation section describes a build on the booted **iOS 26.5** simulator, so the 27 launch path looks unexercised. Two things would settle it: a `workflow_dispatch` run showing green, and either an explicit runtime-availability assertion with a clear message or a `-downloadPlatform` step so the job provisions what it requires.

.github/workflows/ios-toolchain-compatibility.yml
```diff
@@ -165,11 +313,53 @@ jobs:
...
+            echo
+            echo '**Classification:** fixture-workflow-failed'
+            echo "**Fixture:** $FIXTURE"
+            echo '**Next step:** inspect the first failing step; no product compatibility result was produced'
+          } >> "$GITHUB_STEP_SUMMARY"
+
       - name: Upload build logs
+        id: upload-logs
         if: always()
+        continue-on-error: true
```
mahmoud-elmorabea: `continue-on-error: true` here removes the last signal that diagnostics were lost. Combined with the pre-existing `if-no-files-found: ignore`, a missing artifact becomes indistinguishable from a healthy run: `outcome=failure`, `conclusion=success`, job green, and the only trace is a `diagnostics-upload-failed` block appended under a `launch-passed` block.

That matters because nearly every summary this PR adds ends in "inspect the uploaded build log" / "inspect the launch diagnostics" — instructions now pointing at something that may not exist, on a run nobody was alerted about. If the goal is "an upload hiccup shouldn't mask the launch result," the launch result is already in the step summary; I'd rather the upload failure stayed red.

.github/workflows/ios-toolchain-compatibility.yml
```diff
@@ -14,9 +14,12 @@ permissions:
   contents: read
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.ref }}
```
mahmoud-elmorabea: Adding `github.event_name` to the concurrency group means a manual dispatch no longer cancels an in-flight nightly on the same ref — `schedule-refs/heads/main` and `workflow_dispatch-refs/heads/main` are now separate groups.

So dispatching while the 05:17 nightly is queued or running gives you four concurrent 90-minute `xcode-27` jobs contending for the preview label, and each one now boots a simulator and holds it through the survival window rather than just compiling. Our sample-app checks already queue behind an org-wide macOS runner cap, so this makes the dispatched run *more* likely to sit behind the nightly it was meant to replace — and the doc you're editing specifically warns that job timeouts don't cover queue time.

If the intent was to stop PR runs from cancelling nightlies, `github.ref` alone already does that, since `pull_request` runs get `refs/pull/N/merge`. Same change is in customerio-expo-plugin#391.

.github/workflows/ios-toolchain-compatibility.yml
```diff
@@ -98,8 +105,36 @@ jobs:
...
+          fi
+          echo "APP_PRODUCT_PATH=$app_path" >> "$GITHUB_ENV"
+
+      - name: Record launch diagnostic boundary
+        shell: bash
+        env:
+          FIXTURE: ${{ matrix.fixture }}
+        run: touch "$RUNNER_TEMP/${FIXTURE}-launch-start"
+
+      - name: Install and launch fixture
```
mahmoud-elmorabea: The survival window is spent running a real sample app configured with literal placeholder credentials, which puts a second plausible cause behind every `did-not-survive`.

`Prepare placeholder build environment` copies `BuildEnvironment.sample.swift`, whose values are the strings `CUSTOMERIO_WORKSPACE_CDP_API_KEY` / `CUSTOMERIO_WORKSPACE_SITE_ID`. So those 10 seconds are SDK init plus failing/retrying CDP requests with a malformed write key. If the SDK ever adds a precondition on key shape, or the placeholder path throws during device-token registration, the classification is `fixture-exited-early` and the summary says "inspect the launch diagnostics for the crash" — an SDK-config crash read as an iOS 27 lifecycle incompatibility, in the one workflow whose entire value is correct attribution.

A discriminator would fix this cheaply: grep the collected log for the UIScene rejection signature (`UIScene life cycle is required` / `_UIApplicationEvaluateRuntimeIssueForNoSceneLifecycleAdoption`) and classify on that rather than on "the process died." That also addresses the same gap I raised on the Expo PR's policy doc.

.github/workflows/ios-toolchain-compatibility.yml
```diff
@@ -135,12 +171,123 @@ jobs:
...
+          if [[ -d "$crash_reports" ]]; then
+            expected_executable="${APP_NAME%.app}"
+            actual_executable="$(/usr/libexec/PlistBuddy \
+              -c 'Print :CFBundleExecutable' "$APP_PRODUCT_PATH/Info.plist" 2>/dev/null || true)"
+            if [[ -n "$actual_executable" ]]; then
+              if [[ "$actual_executable" != "$expected_executable" ]]; then
+                echo "::warning::expected executable $expected_executable but built $actual_executable; collecting reports for the built executable"
+              fi
+              found_crash_report=false
+              for _ in {1..10}; do
```
mahmoud-elmorabea: The `{1..10}` retry count and the "within 10 seconds" warning text duplicate `LAUNCH_SURVIVAL_SECONDS`, and the coupling is accidental — those two windows measure unrelated things.

This loop needs to cover the simulator→host `.ips` flush, which happens *after* the action's own retries and after its EXIT trap runs `simctl shutdown`. That has nothing to do with how long we require the process to stay alive. So raising `LAUNCH_SURVIVAL_SECONDS` to 30 leaves the loop at 10 spins with a now-wrong message, and a slow flush past the fixed window yields `::warning::no crash report matching ...` on precisely the crash this workflow was built to capture.

Worth its own named constant sized to the flush. Same for `docs/dev-notes/IOS-TOOLCHAIN-COMPATIBILITY-CI.md:3`, which hardcodes "remains alive for ten seconds" while this PR parameterizes the value.

.github/workflows/ios-toolchain-compatibility.yml
```diff
@@ -126,6 +161,7 @@ jobs:
         env:
           FIXTURE: ${{ matrix.fixture }}
         run: |
+          set -euo pipefail
```
mahmoud-elmorabea: `shell: bash` already runs `bash --noprofile --norc -eo pipefail {0}`, so the only thing `set -euo pipefail` adds to these summary-only steps is `-u`.

That inverts their purpose. These steps exist to explain why a job failed; making them abort on any unset variable means a dropped `env:` entry or an empty `${{ matrix.* }}` interpolation kills the step that would have printed `fixture-compile-failed`, leaving a red job with no classification at all. Same at lines 275, 290, and 306.

I'd drop `-u` from the reporting steps specifically (keep it in the ones doing real work), or default every interpolation with `${VAR:-unknown}`.

.github/workflows/ios-toolchain-compatibility.yml
```diff
@@ -112,12 +147,12 @@ jobs:
           : "${XCODE_BUILD:?missing xcode-build output}"
           : "${IOS_SIMULATOR_SDK:?missing iphonesimulator-sdk output}"
           {
-            echo '## Native fixture compile'
+            echo '## Native fixture compatibility'
```
mahmoud-elmorabea: Four smaller items, grouped:

**1. Duplicate success summary.** On success the action already appends `## iOS simulator launch smoke / Classification: launch-passed / Survival window: 10s / Scope: install, launch, and process survival only; no callback, push-delivery, signing, or App Store claim` (`launch.sh`). This step then appends `## Native fixture compatibility` with the same classification and near-identical scope prose. Only the toolchain line is new — the survival number and scope wording are now maintained in two repos and will drift.

**2. `APP_PRODUCT_PATH` via `GITHUB_ENV`** (line 118) leaks into all eleven later steps, and `Collect launch diagnostics` reads it under `set -u` without declaring it in its own `env:` — unlike `FIXTURE` and `APP_NAME`. If the gate is relaxed (including to the `steps.launch.outcome == 'failure'` form I'm asking for above, which also covers `invalid-input` where no path was resolved), `-u` aborts the collection. `echo "app-path=…" >> "$GITHUB_OUTPUT"` plus an `env:` entry scopes it properly.

**3. Hardcoded `app-name` in the matrix** (line 33) duplicates `PRODUCT_NAME` state that already lives in the Xcode projects. A target rename keeps the compile green and turns a real result into `built-app-not-found` with next-step text pointing at build settings — which reads as infra, not a rename. Globbing `Build/Products/*-iphonesimulator/*.app` or reading `FULL_PRODUCT_NAME` removes the column.

**4. `rm -rf` of the derived-data dir** (lines 80 and 97) can't fire: `RUNNER_TEMP` is fresh per job on hosted runners, so the stale-product case it guards is unreachable. If it's defending a future self-hosted `xcode-27` runner, one place with a comment beats two copies.

.github/workflows/ios-toolchain-compatibility.yml
mahmoud-elmorabea: 🟠 `mobile-ci-tools/README.md:33-35`: "A failed composite action is not required to propagate its mapped outputs... gate that upload on `steps.<id>.outcome == 'failure'`." `failure-reason` is mapped off the same inner step as `log-path`.

So on a real crash this skips, no `.ips` is collected, and line 183's `''` arm reports `launch-smoke-inconclusive`. Gate on `steps.launch.outcome == 'failure'`.

.github/workflows/ios-toolchain-compatibility.yml
mahmoud-elmorabea: 🟠  `unexpected-error` is the only documented `failure-reason` not covered (`action.yml:44-46`); `launch.sh` reaches it from three `fail` sites plus its ERR trap, some after install+launch.

Cost: a documented value prints as "unrecognized", and it's excluded from the line-227 allowlist so the crash report is dropped.

.github/workflows/ios-toolchain-compatibility.yml
mahmoud-elmorabea: With the existing `if-no-files-found: ignore`, this makes a failed upload green and indistinguishable from a healthy run — while every summary says to inspect the artifact.

.github/workflows/ios-toolchain-compatibility.yml
mahmoud-elmorabea: `schedule-` and `workflow_dispatch-` are now separate groups, so a dispatch no longer cancels an in-flight nightly and both contend for the preview label.

`github.ref` alone already keeps PR runs from cancelling nightlies (`refs/pull/N/merge`).

.github/workflows/ios-toolchain-compatibility.yml
mahmoud-elmorabea: `{1..10}` and the "within 10 seconds" text duplicate `LAUNCH_SURVIVAL_SECONDS`, but this window covers the `.ips` flush — unrelated to survival time. Raising the constant leaves both wrong.

`docs/dev-notes/IOS-TOOLCHAIN-COMPATIBILITY-CI.md:3` hardcodes "ten seconds" too.

.github/workflows/ios-toolchain-compatibility.yml
mahmoud-elmorabea: `shell: bash` already implies `-eo pipefail`, so this only adds `-u` — which lets an empty interpolation kill the step that exists to explain the failure. Same at 275, 290, 306.

.github/workflows/ios-toolchain-compatibility.yml
mahmoud-elmorabea: `launch.sh` already appends a `launch-passed` block with survival window and scope. Only the toolchain line here is new, so the scope prose now lives in two repos.

.github/workflows/ios-toolchain-compatibility.yml
```diff
@@ -98,8 +105,36 @@ jobs:
...
+        id: resolve-app
+        shell: bash
+        run: |
+          set -euo pipefail
+          app_path="$RUNNER_TEMP/${{ matrix.fixture }}-xcode-27/Build/Products/Debug-iphonesimulator/${{ matrix.app-name }}"
+          if [[ ! -d "$app_path" ]]; then
+            echo "::error::built simulator app not found at $app_path"
+            exit 1
+          fi
+          echo "APP_PRODUCT_PATH=$app_path" >> "$GITHUB_ENV"
```
mahmoud-elmorabea, line 114: `GITHUB_ENV` leaks this into all later steps, and `Collect launch diagnostics` reads it under `set -u` without declaring it in `env:` — unlike `FIXTURE`/`APP_NAME`. Relaxing that gate (incl. the line-227 fix) then aborts on unbound variable.

A step output plus an `env:` entry scopes it.

.github/workflows/ios-toolchain-compatibility.yml
```diff
@@ -72,6 +77,7 @@ jobs:
         shell: bash
         run: |
           set -o pipefail
+          rm -rf -- "$RUNNER_TEMP/${{ matrix.fixture }}-xcode-27"
```
mahmoud-elmorabea: `RUNNER_TEMP` is fresh per job on hosted runners, so the stale-product case this guards can't occur. Duplicated at 97.

mahmoud-elmorabea (approved):
No blockers 👍 

Marked two comments as 🟠 that would be worth fixing before merging 😇

.github/workflows/ios-toolchain-compatibility.yml
Shahroz16: Resolved by the simplification in c504562d: the custom `Collect launch diagnostics` shell step and its `set -u` reads were removed entirely. `APP_PRODUCT_PATH` is now consumed only as the shared action input `${{ env.APP_PRODUCT_PATH }}`, so there is no unbound shell variable or later custom diagnostic consumer.

PR #1217, opened 2026-08-18
ci: reject conditional CocoaPods deployment targets (MBL-2278)

Shahroz16:
## What changed

- Fail before mutation when the selected CocoaPods build-setting precedence contains qualified `IPHONEOS_DEPLOYMENT_TARGET[...]` keys.
- Cover target, project, and xcconfig conditional settings.
- Document why adding an unconditional override cannot safely normalize a conditional matrix.

## Why

Xcode can prefer a conditional target value over the helper's unconditional value, or a new target value can lower a higher conditional project floor. The previous audit inspected only the unconditional key and could report success for the wrong effective floor.

## Validation

- Ruby helper suite: 33 tests, 190 assertions
- `MessagingPushAPN` iOS simulator build
- format, strict lint, diff check, pre-commit and pre-push hooks

## Release and merge policy

This CI/build-tooling PR must be **squash merged only** under its `ci:` PR title. Merge or rebase would expose the branch’s `fix:` commit to semantic-release and could trigger an unintended SDK release. The squashed change does not alter SDK runtime behavior.

---

> [!NOTE]
> **Low Risk**
> CI/build-tooling only; no SDK runtime changes, with behavior limited to failing early when conditional deployment targets are present.
> 
> **Overview**
> The CocoaPods deployment-target helper now **fails before any project mutation** when it finds qualified `IPHONEOS_DEPLOYMENT_TARGET[...]` keys at the precedence it is evaluating—on target build settings, project configurations, and parsed xcconfigs (including whitespace before `[`). That closes a gap where Xcode could still apply a conditional floor while the helper only audited the unconditional key, so normalization or audit could report success for the wrong effective minimum.
> 
> Errors tell integrators to replace the conditional matrix with one numeric, unconditional value (Podfile/podspec for generated Pods targets, not hand-editing `Pods.xcodeproj`). The standalone audit fails the same way before printing its table. Docs describe this behavior; the Ruby script header now states this repo owns the copy under local tests. A broad Ruby test suite covers fail-closed paths and the case where an explicit target floor correctly ignores lower-precedence conditional project/xcconfig values.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 32a60941198bdfc76e19ececdd339dcfbb458894. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


scripts/cocoapods_deployment_target.rb
```diff
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
-# Keep this helper byte-identical across customerio-ios, customerio-flutter, and
-# customerio-reactnative so every wrapper applies the same CocoaPods contract.
+# This copy is owned by customerio-ios and covered by its local test suite.
```
mrehan27, line 3: It looks like this file is duplicated across multiple repos, and the comment suggests it is originally owned by the iOS repo. I think we should leverage `mobile-ci-tools` repo instead and reuse single copy from there, making it easier to maintain and validate.

mrehan27 (approved):

scripts/cocoapods_deployment_target.rb
Shahroz16: This local copy is intentional. `mobile-ci-tools` is CI-only and is not available when a customer Podfile executes. Flutter and React Native must ship a package-local Ruby helper that their installed Podfiles can require without fetching another repository, and native iOS keeps its local copy as the implementation and test owner. Moving the runtime helper into `mobile-ci-tools` would either break that installation boundary or introduce a network/runtime dependency into customer builds. We will keep the copies versioned with their respective packages and test the same contract in each consumer.

PR #1218, opened 2026-08-18
chore(geofence): re-arm monitored conditions on foreground entry after a long suspension

mrehan27:
## What

Re-arms every owned geofence condition in place when the app enters the foreground after 6+ hours without a rebuild (CLMonitor path, iOS 17+).

## Why

locationd's per-fence promotion state can wedge while a process stays suspended for days: the daemon's own position fixes place the device inside a fence, yet the condition record stays unsatisfied and no event is delivered (field-observed on a fence that had promoted reliably at the same spot before). A rebuild of the condition makes the OS re-evaluate from scratch and emit a corrective for any crossing its stale record missed.

Cold launch already recovers this via the adopt re-arm — but a healthy long-lived process can stay suspended for days and never cold-launch, so the recovery users get by force-killing the app now also happens on a plain open.

## How

- `willEnterForeground` observer; a timestamp gate (stamped at enqueue) limits rebuilds to once per 6 h, and the launch adopt stamps it too, so a fresh process never double-rebuilds.
- The rebuild re-adds each condition with `assuming:` = stored baseline, so it is event-silent when nothing changed; a genuine missed crossing surfaces as the OS corrective and flows through the existing dedup baseline.
- Ownership and records are read at drain time, and a condition is skipped unless its record matches the staged registration (settled-geometry guard) — re-arming a mid-transition condition from either snapshot would impose geometry the sync layer then never converges back from.

## Verification

- State-space model (v6) over the registration pipeline: the guard is load-bearing — the unguarded variant provably wedges the OS set after an adopt racing an in-flight reshape; the guarded one introduces no new violation families.
- 322/322 LocationGeofence tests.
- Field: first foreground after ~54 h suspension re-armed all conditions, event-silent (baselines agreed), and a second open 11 min later was correctly suppressed by the gate.
- Wedged-recovery wiring (simulator, synthetic): with a stale baseline forged into the store and the device parked 15 m inside the fence edge, the foreground re-arm delivered the missed enter exactly once, 19 ms after firing. A real OS wedge cannot be manufactured on demand; the same corrective mechanism has recovered real missed crossings in the field via the launch adopt path.

## Notes

- No public API changes.
- Follow-ups tracked separately: heal-refusal observability, sync-skip-tier heal coverage.

---

> [!NOTE]
> **Medium Risk**
> Touches CoreLocation monitoring lifecycle and can emit corrective enter/exit events after long suspension; guarded by interval and geometry checks but still affects geofence delivery reliability.
> 
> **Overview**
> Adds a **foreground recovery path** for the `CLMonitor` geofence monitor when locationd’s per-fence promotion state can wedge during multi-day suspension without a cold launch.
> 
> On `UIApplication.willEnterForeground`, if at least **`foregroundRearmInterval` (6 hours)** has passed since the last OS rebuild and there are owned regions, the monitor **remove/re-adds** each settled condition with `assuming:` set to the stored baseline—same idea as launch adopt re-arm, but for long-lived processes. **`lastRearmAt`** is stamped at enqueue (and on adopt) so rapid foreground opens don’t stack rebuilds; launch adopt prevents double work on fresh processes.
> 
> Re-arm only runs when **persisted records match staged `registeredConditions`** (settled-geometry guard), avoiding races with in-flight reshapes that could wedge sync forever. Adds **`geofenceForegroundRearm`** logging and observer cleanup in `deinit`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4d864658dd73fd5c3fb6c058a16bfcd47a0be4fa. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor+Rearm.swift
```diff
@@ -44,4 +47,67 @@ extension CLMonitorGeofenceMonitor {
...
+                          radius: radius,
+                          transitionTypes: record.transitionTypes
+                      )
+                else { continue }
+                await monitor.remove(identifier)
+                let condition = CLMonitor.CircularGeographicCondition(
+                    center: CLLocationCoordinate2D(latitude: center.latitude, longitude: center.longitude),
+                    radius: radius
+                )
+                await monitor.add(condition, identifier: identifier, assuming: record.lastState == .enter ? .satisfied : .unsatisfied)
```
Shahroz16, line 107: Could we rebase this onto #1220 and stamp this foreground add in conditionReadds? This remove/add can trigger the same immediate stale-belief replay that #1220 is handling. Without the stamp, the periodic recovery can emit a false transition before Core Location performs its fresh re-evaluation.

Shahroz16 (commented):
I left one integration concern around the foreground re-add and the contradiction gate.

Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor+Rearm.swift
mrehan27: Agreed on the mechanism — this add site can replay a wedged belief before the fresh evaluation, and the combined field build already carries the one-line stamp for it. Rather than stacking this PR on #1220 (both are otherwise independent), the plan is: whichever of the two merges second picks up the integration — if #1220 merges first, this branch adds the stamp when it rebases onto main. #1220's description carries the same note. Happy to convert to a stack instead if you'd rather see the stamp in this diff now.

Shahroz16 (approved):
Thanks for confirming. Approving this as-is since it will not be released independently. I’ll look for the conditionReadds stamp in whichever of #1218 and #1220 lands second.

Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor+Rearm.swift
mrehan27: Done as planned — #1220 merged first, this branch is now updated onto main and 064fc121 adds the stamp to the foreground re-add (plus the +Fixes.swift split the combined file needs for the length cap). Both ran on the combined field build all weekend, where this exact rearm path resurrected a fence set locationd had silently purged.

Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor+Rearm.swift
```diff
@@ -46,4 +49,69 @@ extension CLMonitorGeofenceMonitor {
...
+                self.conditionReadds[identifier] = ConditionReadd(start: readdStart, added: Date(), center: center, radius: radius)
+                self.knownConditionIdentifiers.insert(identifier)
+                rearmed += 1
+            }
+            guard rearmed > 0 else { return }
```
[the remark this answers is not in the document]
mrehan27: Accurate mechanism, but the consequence doesn't materialize, so leaving as-is. The settled-geometry guard only skips a condition when its record and staged registration disagree — which by that guard's design (state-space model v6) means queued pipeline ops are already about to impose fresh geometry at the OS for it. Those remove/adds rebuild the daemon's promotion record exactly as the rearm would, so an all-skipped pass is one where the rebuild arrives via the sync's own adds, and advancing the gate is then correct. The nil-geometry skip class is likewise re-registered by the sync a foreground drives anyway. Stamping at drain (or only when rearmed > 0) would reintroduce what the enqueue stamp deliberately prevents: rapid foreground cycles queuing multiple wholesale rebuild passes, where each redundant re-add can absorb an in-flight crossing — a real cost, against a no-damage 6h delay.

PR #1219, opened 2026-08-21
chore(geofence): deliver crossings the OS never promotes by healing baselines at sync

mrehan27:
## What

After every sync pass, a queued heal compares each registered-unchanged region's stored dedup baseline against a fresh fix (resolved when the heal drains) and synthesizes the crossing the OS never delivered when they unambiguously contradict.

## Why

iOS region monitoring can leave a device dwelling inside an armed fence without ever generating the enter: promotion needs a fix accurate to within the device's distance from the fence edge, which passive indoor positioning often cannot supply. Field-observed repeatedly, including a 10-minute dwell 350 m inside a healthy, actively-swept 1 km fence with zero events (Android delivered the same crossing), and a missed enter+exit pair on a 30–40 km/h drive-through. Without recovery, the miss also poisons the dedup baseline so the *next* real crossing of that fence is swallowed as a duplicate.

## How

- `BaselineHealDecision`: synthesize only when the fix is recent (≤ movement-fix max age) and its distance from the fence edge exceeds `max(horizontalAccuracy, 20 m)` — a coarse fix can never fabricate an event; ambiguity always leaves the baseline alone.
- The heal runs as a pipeline operation **enqueued behind the calling sync's own registration ops**, so baselines are read after those rewrites drain. The state-space model confirmed this ordering (v5): the read-at-sync variant races its own sync's queued rewrites; the enqueued variant removes ~21k lost-crossing orderings and adds only cooldown-absorbed duplicates.
- Delivery reuses `recordMonitorEvent`, so the baseline flip and transition-type filter are identical to the OS event path — a later OS delivery of the same crossing dedups against the healed baseline (exactly-once preserved).

## Verification

- 336/336 LocationGeofence tests (branch rebased onto current main).
- State-space model v5 over the registration pipeline (ordering verdict above).
- Simulator A/B on the non-promoting teleport-enter: 3 minutes inside an armed fence produced nothing without the heal; with it, the enter delivered on the next pass.
- Field campaign: 10 recovered real OS-missed crossings across three drive days (including a missed ENTER delivered on arrival), zero false syntheses, exactly-once held throughout; two heals raced late OS promotions by 12–20 s and the duplicates were absorbed silently.

## Notes

- No public API changes.
- Shares `BaselineHealDecision.swift` (+ its tests and the margin constant) with the contradiction-gate PR (#1220), which reuses the same decision primitive. Both PRs are self-contained; whichever merges second gets a trivial rebase and its diff shrinks by that file.

## Review round (2026-08-25)

Three review findings, all confirmed against the code and fixed in separate commits:

- **A heal's fix could predate the baseline it contradicts** — a genuine OS crossing recorded while the heal waited in the queue (or within the fix's 30 s allowed age) made the heal judge the new baseline from an old position, synthesizing the reverse transition and dedup-flipping away the real one. `MonitorRegionRecord` now stamps when its `lastState` was written, and `recordMonitorEvent` takes the evidence's timestamp, rejecting the write atomically inside the storage actor when the baseline is newer — closing the snapshot→write window as well as the queue wait. Pre-stamp records decode with a nil stamp and never block a heal.
- **Geometry could reshape between heal staging and drain** — a later sync stages a reshape synchronously while its storage rewrite queues behind the heal, so the heal compared the new circle against the old circle's baseline (the same staging-vs-drain seam the gate PR closed on its side). Each candidate's circle is now captured at enqueue and re-verified at drain against both the staged map and the stored record; mismatches skip (the reshape reseeds that baseline anyway).
- **The heal could silently skip on launch/manual/foreground syncs** — it read `bestKnownFix()` (this monitor's cache + resolver), but those syncs acquire their fix upstream and hand over coordinates only, so the cache could be empty or stale at exactly the wedged-promotion recovery moments the heal exists for. The fix is now resolved at drain time through the movement resolver: at most one one-shot request per sync (on the operation pipeline, never the event path), free on movement syncs whose pass just resolved through the same resolver, and the result carries a real timestamp and accuracy — strictly fresher than a sync-time snapshot would be by drain, and it feeds the baseline-age guard above. This deviates from carrying the sync's own fix through `setMonitoredRegions` deliberately: that fix would itself be up to 30 s stale at drain, and the plumbing would widen the monitor protocol for no fresher evidence.

Verification: 336/336 LocationGeofence tests including 5 new storage tests pinning the recency guard, stamp preservation across unchanged re-registrations, stamp reset on reshape, and nil-stamp fail-open for legacy records.

---

> [!NOTE]
> **Medium Risk**
> Changes when geofence transitions are delivered and how dedup baselines advance, but behind explicit accuracy/age guards, pipeline ordering, and evidence-timestamp checks to limit false syntheses and races with real OS events.
> 
> **Overview**
> **Adds a post-sync “baseline heal”** on the iOS 17+ `CLMonitor` path: for regions left registered with unchanged geometry, a queued pipeline step compares the persisted dedup baseline to a **drain-time** location fix and, when `BaselineHealDecision` says the device is unambiguously on the other side of the fence, **synthesizes the missing enter/exit** through the same `recordMonitorEvent` / `onTransition` path as real OS events.
> 
> **Safety and ordering:** synthesis only runs for recent fixes whose distance from the fence edge exceeds `max(horizontalAccuracy, 20 m)`; ambiguous fixes are ignored. The heal is **enqueued after** the calling sync’s registration ops, captures each candidate’s circle at enqueue and re-checks it at drain, and resolves location via `movementFixResolver` so launch/manual syncs aren’t skipped for lack of a local cache.
> 
> **Storage:** `MonitorRegionRecord` gains `lastStateChangedAt`; `recordMonitorEvent` accepts `onlyIfBaselinePredates` so a heal whose fix predates a genuine OS crossing (e.g. while queued) is rejected with `.suppressedNewerBaseline` instead of flipping the baseline backward. Unchanged re-registration preserves the stamp; geometry changes reset it; legacy nil stamps fail open.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 28e78c69ad024e764fb7503a65942ca8d1dd40b8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor+Registration.swift
```diff
@@ -177,6 +184,8 @@ extension CLMonitorGeofenceMonitor {
             // initial-enter decision must not count a region the OS never took.
             if ownedRegionIdentifiers.contains(region.identifier) { added.insert(region.identifier) }
         }
+        // Enqueued after the adds above so the heal drains behind this sync's own ops.
+        enqueueBaselineHeal(candidates: healCandidates, fix: bestKnownFix())
```
Shahroz16: Could we carry the actual sync fix into the monitor instead of calling bestKnownFix() here? The refresh path reduces the acquired fix to latitude and longitude, while bestKnownFix() reads this monitor’s separate CLLocationManager or a previous movement request. On a first, manual, or app-launch sync this can be nil or unrelated, so the heal can silently skip even though the sync has the fresh fix it needs. Preserving its timestamp and horizontal accuracy would also keep the decision tied to this pass.

Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor+BaselineHeal.swift
```diff
@@ -0,0 +1,50 @@
...
+    /// flip and transition-type filter are identical to the OS event path, and a later OS
+    /// delivery of the same crossing dedups against the healed baseline.
+    func enqueueBaselineHeal(candidates: [String], fix: CLLocation?) {
+        guard let fix, CLLocationCoordinate2DIsValid(fix.coordinate), !candidates.isEmpty else { return }
+        enqueueMonitorOperation { [weak self] _ in
+            guard let self else { return }
+            let records = await self.storage.getMonitorRegionRecords()
+            for identifier in candidates.sorted() {
+                guard self.ownedRegionIdentifiers.contains(identifier),
+                      let condition = self.registeredConditions[identifier],
```
Shahroz16: Could we capture the candidate’s expected geometry and verify it still matches both registeredConditions and the stored record when this operation drains? A later sync can stage a reshape synchronously before this older heal runs, while that reshape’s storage rewrite remains queued behind it. This can compare the new circle with the old baseline and synthesize the wrong transition.

Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor+BaselineHeal.swift
```diff
@@ -0,0 +1,50 @@
...
+                else { continue }
+                let center = CLLocation(latitude: condition.center.latitude, longitude: condition.center.longitude)
+                guard let transition = BaselineHealDecision.synthesizedTransition(
+                    distanceFromCenter: fix.distance(from: center),
+                    radius: condition.radius,
+                    horizontalAccuracy: fix.horizontalAccuracy,
+                    // Age at drain time, so a delayed drain disqualifies the fix instead of
+                    // trusting a snapshot that has gone stale in the queue.
+                    fixAge: -fix.timestamp.timeIntervalSinceNow,
+                    lastState: record.lastState
```
Shahroz16, line 58: Could we also make sure the baseline has not changed since this heal was staged, perhaps with a revision or last-event timestamp? A genuine OS crossing can update the record while this operation is waiting. If the captured fix predates that crossing, the heal can interpret the new baseline using the old position and emit the reverse transition. This is reachable when driving across a small fence within the 30-second freshness window.

Shahroz16 (commented):
I left three comments around which fix the heal uses and two queue-ordering cases that can produce an incorrect transition.

Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor+Registration.swift
mrehan27: Confirmed the gap — launch/manual/foreground syncs acquire their fix upstream and hand over coordinates only, so bestKnownFix() could be empty or stale at exactly the recovery moments the heal exists for. Fixed in 03ab5d2a with one deviation from the suggestion: instead of plumbing the sync's fix through setMonitoredRegions (it would be up to 30 s stale by the time the queued heal drains, and widens the monitor protocol), the heal now resolves its fix at drain time through the movement resolver — at most one one-shot request per sync on the operation pipeline (never the event path), free on movement syncs whose pass just resolved through the same resolver, and the result carries a real timestamp and accuracy that also feed the new baseline-age guard. Strictly fresher evidence than a sync-time snapshot; same intent.

Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor+BaselineHeal.swift
mrehan27: Confirmed — a later sync stages the reshape synchronously while its storage rewrite queues behind the heal, so the heal could judge the old baseline against the new circle. Same staging-vs-drain seam Bugbot's finding closed on the gate's side. Fixed in b6c1795f: each candidate's circle is captured when the heal is enqueued (synchronously with the sync's unchanged-diff) and re-verified at drain against both registeredConditions and the stored record; a mismatch skips the candidate, whose baseline the reshape reseeds anyway.

Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor+BaselineHeal.swift
mrehan27: Confirmed, and it's the nastier of the two orderings — the reverse synthesis also dedup-flips away the user's real crossing. Fixed in 2df63b5a with the timestamp approach you suggested: MonitorRegionRecord stamps when lastState was written (seed or state change; belief re-emissions of an unchanged state don't move it), and recordMonitorEvent takes the evidence's timestamp and rejects the write atomically inside the storage actor when the baseline is newer — so the check also closes the smaller snapshot-to-write window, not just the queue wait. Records persisted before the stamp existed decode nil and never block a heal. Storage tests pin the guard, stamp preservation across unchanged re-registrations, stamp reset on reshape, and the legacy fail-open.

Shahroz16 (approved):
The latest changes address my comments around fix selection, geometry changes, and baseline ordering.

PR #1220, opened 2026-08-21
chore(geofence): refuse OS transitions a fresh fix geometrically contradicts

mrehan27:
## What

Refuses an OS-delivered geofence transition when it lands inside a short window after our own (re)add of that condition AND a fresh fix confidently contradicts it. Events outside the window are untouched.

## Why

CLMonitor keeps a per-identifier belief that survives app uninstalls and re-evaluates lazily. On every add (first registration, launch adopt, rearm), the daemon instantly replays that belief against our staged `assuming:` — **without a fresh evaluation** — and its own real evaluation follows only ~3 s later. When the belief is wrong (stale by hours from crossings observed while the app was dead, left over from a previous install, or the default "unsatisfied" for never-seen identifiers), that replay is delivered as a real transition:

- false ENTER for a fence 7.8 km away on a fresh install's first sync (field, twice: two different fences on two days, one of them registered mid-drive);
- false enters for two fences 1.1 km / 5.4 km away on a plain cold-launch adopt after 22 h suspended (field);
- false-EXIT storm when a user's first login happens inside fences — 6 fences ⇒ 12 junk CDP events in 3 s (deterministic sim repro).

Each false event also flips the dedup baseline, so the *next real* crossing of that fence is swallowed as a duplicate. Reproduced deterministically on simulator (~90 s/cycle); every alternative SDK-side cause was eliminated by instrumentation before pinning the mechanism.

## How

- The monitor stamps each condition's (re)add — when it began, when the add returned, and the circle it imposed. The gate vets only events dated inside `[start, added + contradictionGateReplayWindow]` (10 s; measured replays land 0.003–3.3 s after the add) — a normal crossing, before the re-add or after the window, is never delayed by a fix request nor at any risk of refusal.
- In-window events are checked against a fresh fix via `BaselineHealDecision` (introduced here) with `lastState` = the incoming transition: non-nil means the fix says the opposite, beyond `max(horizontalAccuracy, 20 m)` from the edge. Refusal happens BEFORE the baseline advances, so the daemon's own ~3 s re-evaluation lands as a duplicate against the untouched baseline and is absorbed — net zero events, baseline stays true.
- Legitimate in-window correctives (recovery of a crossing the OS observed but never delivered) agree with the fix and pass unchanged.
- Fail-open everywhere: no fix, stale fix, coarse accuracy, or within the ambiguity margin ⇒ deliver. A real event is never dropped for lack of a fix.

## Verification

- **State-space model v8.2** (125,952 paths per arm over belief × truth × baseline × assuming × fix-availability × heal × buffered pre-add catch-ups): zero fabricated deliveries whenever a usable fix exists (no-gate delivers 1,312 in the same paths); out-of-window crossings provably untouched; every single-crossing in-window refusal recovered; exactly-once and baseline convergence hold. The model earned its keep twice: it caught that a fix predating a fast real crossing could refuse it (fixed by the replay-window design), and end-to-end rig runs caught a burst-ordering hole in an earlier guard implementation (fixed and re-proven).
- **Simulator rig, end-to-end on this exact code**: false-exit storm at first-login-inside — 6/6 refused, all 6 real initial enters delivered; false-enter burst on reinstall-far — 6/6 refused, zero events tracked; real drive-through crossings delivered normally.
- 347/347 LocationGeofence tests, including 16 gate-decision tests pinning refusal/fail-open/window/cooldown semantics.
- The production incidents above were each cleaned up minutes later by the baseline heal — this PR stops them at the source instead.

## Notes

- No public API changes. CLMonitor path (iOS 18+) only; the classic path's exposure is unverified (no field observation there).
- Accepted residue, by design: with no usable fix the gate fails open (a replay can pass); a real round trip completed entirely inside the 10 s window and judged only by a pre-crossing cached fix can be suppressed (same class of quick round trips circles already miss natively).
- Shares `BaselineHealDecision.swift` (+ tests and margin constant) with the baseline-heal PR (#1219), which uses the same primitive for the opposite failure (deliver what the OS wrongly withheld). Both PRs are self-contained; whichever merges second gets a trivial rebase and its diff shrinks by that file.
- Integration note with #1218 (foreground rearm): that PR adds a third re-add site, which must also stamp the replay window — a one-line addition for whichever of the two merges second (verified on the combined field build).
- Follow-up tracked separately: launch adopts can run twice back-to-back (reconcile rewire), doubling replay opportunities — a recency gate there is a cheap independent hardening.

## Review round (2026-08-25)

Two review findings, both confirmed against the code and fixed in separate commits:

- **Replay window had no lower bound** — any event dated before the re-add (negative delta) qualified for gating, so a genuine catch-up buffered in `pendingEvents` across an adopt re-add could be refused off a cached pre-crossing fix, and its refusal would dedup away the paired corrective. The stamp now records a start taken *before* the remove/add pair (a bare non-negative check against the add's return would leak replays stamped inside the remove→add gap) and the window is two-sided. Model v8.2 adds the buffered-catch-up class as a new dimension: the two-sided window refuses none of them, and an upper-only negative control fails exactly there — the model is provably sensitive to the bound.
- **Serial fix resolution could stall the event consumer** — the events loop awaits `process` serially, so the resolver's in-flight coalescing never engages; with no usable fix obtainable, each fence in a burst started its own 10 s request. A resolution that comes back without a fresh fix is now remembered for `movementFixMaxAge`: a burst pays at most one request and later events fail open immediately (a late fix from that request still reaches the cache and serves them).

Re-validated end-to-end on this exact code: false-enter burst 6/6 refused (refusals 2 ms apart — one shared fix, no repeated requests); false-exit storm 6/6 refused with all 6 real initial enters delivered; real drive-in delivered all 6 enters untouched; fresh-install-inside produced zero junk.

---

> [!NOTE]
> **Medium Risk**
> Changes core geofence event delivery and dedup baseline timing on CLMonitor; incorrect gating could suppress real crossings or let false replays through when fixes are unavailable (designed to fail open).
> 
> **Overview**
> Adds a **contradiction gate** on the CLMonitor path so OS transitions that are really stale daemon “belief replays” right after a condition (re)add can be dropped before the dedup baseline advances.
> 
> Each registration/rearm now records a **`ConditionReadd`** stamp (timing + circle geometry). For **10s** after an add, events dated in that replay window can be vetted with a fresh location fix (reusing **`MovementFixResolver`**, with one request per burst and fail-open on missing/stale/coarse fixes). **`BaselineHealDecision`** decides whether the fix unambiguously disagrees with the incoming enter/exit (edge distance beyond `max(accuracy, 20m)`); if so the event is refused and logged, leaving the baseline intact so the daemon’s later re-evaluation dedups silently. Events outside the window, the movement trigger, and ambiguous geometry behave as before.
> 
> New unit tests cover heal/gate geometry, replay-window bounds, and gate-fix request cooldown.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5542eebf20492ed54e6143d7276a6ee331b6a66e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor+ContradictionGate.swift
+++ b/Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor+ContradictionGate.swift
@@ -1,0 +1,67 @@
+import CioInternalCommon
+import CoreLocation
+import Foundation
+
+/// Contradiction gate for OS-delivered transitions, split out to keep the monitor's event
+/// plumbing readable (same convention as `+Registration`).
+///
+/// CLMonitor re-adds (registration, adopt, rearm) replay the daemon's per-identifier belief as an
+/// immediate event computed WITHOUT a fresh evaluation. The belief can be stale by hours
+/// (crossings observed while the app was dead), predate the install (it persists across
+/// uninstalls), or default to unsatisfied for never-seen identifiers — producing false enters for
+/// far fences and false-exit storms when registering while inside (both field-observed; the
+/// daemon's own re-evaluation follows ~3 s later, measured).
+///
+/// The gate therefore vets only events landing inside a short window after our own (re)add of
+/// that identifier — where belief replays live — and refuses one that a fresh fix unambiguously
+/// contradicts, BEFORE the dedup baseline advances, so the daemon's re-evaluation lands as a
+/// duplicate against the untouched baseline and is absorbed silently. Everything outside the
+/// window flows exactly as before: a normal crossing is never delayed by a fix request, and a
+/// legitimate in-window corrective (rearm/wedged-crossing recovery) agrees with the fix and
+/// passes. Undecidable fixes — none, stale, or within the ambiguity margin — fail OPEN.
+@available(iOS 17.0, *)
+extension CLMonitorGeofenceMonitor {
+    /// True when the event lands inside the identifier's replay window AND a trustworthy fix
+    /// confidently contradicts it (`BaselineHealDecision` with `lastState` = the incoming
+    /// transition — a non-nil result means the fix says the opposite of what the OS delivered).
+    func isEventContradictedByFreshFix(identifier: String, transition: GeofenceTransition) async -> Bool {
+        guard let readdedAt = conditionReaddTimestamps[identifier],
+              Date().timeIntervalSince(readdedAt) <= GeofenceConstants.contradictionGateReplayWindow,
+              let condition = registeredConditions[identifier]
+        else { return false }
+        let gateFix = await resolveGateFix()
+        guard let gateFix, CLLocationCoordinate2DIsValid(gateFix.coordinate) else { return false }
+        let center = CLLocation(latitude: condition.center.latitude, longitude: condition.center.longitude)
+        let distanceFromCenter = gateFix.distance(from: center)
+        guard BaselineHealDecision.synthesizedTransition(
+            distanceFromCenter: distanceFromCenter,
+            radius: condition.radius,
+            horizontalAccuracy: gateFix.horizontalAccuracy,
+            fixAge: -gateFix.timestamp.timeIntervalSinceNow,
+            lastState: transition
+        ) != nil else { return false }
+        logger.geofenceEventRefusedByContradiction(
+            identifier: identifier,
+            transition: transition,
+            distanceFromCenter: distanceFromCenter,
+            radius: condition.radius,
+            accuracy: gateFix.horizontalAccuracy
+        )
+        return true
+    }
+
+    /// Freshest fix obtainable for the gate: reuses the movement resolver (one-shot request when
+    /// the cache is stale, coalesced across concurrent events — a replay burst re-adds many
+    /// conditions at once and every event shares one resolution — cached-fix fallback on timeout).
+    /// A cached fix is acceptable as-is: the gate only runs moments after an add that itself
+    /// positioned the staging off the same cache, so it cannot meaningfully predate the state the
+    /// event claims. Returns via `bestKnownFix()` so the freshest of the cache and the request
+    /// wins; the caller's decision applies the fix-age guard to the result.
+    private func resolveGateFix() async -> CLLocation? {
+        await withCheckedContinuation { continuation in
+            movementFixResolver.resolve(cached: bestKnownFix()) { [weak self] _ in
+                continuation.resume(returning: self?.bestKnownFix())
+            }
+        }
+    }
+}
--- a/Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor+ContradictionGate.swift
+++ b/Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor+ContradictionGate.swift
@@ -23,9 +23,23 @@ import Foundation
 extension CLMonitorGeofenceMonitor {
     /// A drained (re)add: when it landed at the OS and the circle it imposed there.
     struct ConditionReadd {
-        let date: Date
+        /// Stamped before the remove/add pair is issued. A replay is created by the add, so no
+        /// replay can be dated earlier — while an event dated before it (e.g. a catch-up buffered
+        /// in `pendingEvents` across the re-add) is provably not this add's replay.
+        let start: Date
+        /// Stamped after `add` returned; the replay window extends from here. Kept separate from
+        /// `start` because a replay can be dated inside the remove→add gap, before `add` returns.
+        let added: Date
         let center: LocationData
         let radius: Double
+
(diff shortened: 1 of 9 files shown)


Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor+ContradictionGate.swift
```diff
@@ -0,0 +1,80 @@
...
+    /// Geometry comes from the drained add's stamp, not `registeredConditions`: that map updates
+    /// synchronously when a reshape is staged, so during its staging→drain gap an event the OS
+    /// computed on the old circle must still be judged against the old circle.
+    /// The window compares the EVENT's date to the add, not the processing time: an event can sit
+    /// in `pendingEvents` until the bootstrap binds the handler, and a belief replay must stay
+    /// gated no matter how late it drains — while a real crossing minutes after the add stays
+    /// ungated no matter how fast it is processed.
+    func isEventContradictedByFreshFix(identifier: String, transition: GeofenceTransition, eventDate: Date) async -> Bool {
+        guard let readd = conditionReadds[identifier],
+              eventDate.timeIntervalSince(readd.date) <= GeofenceConstants.contradictionGateReplayWindow
```
Shahroz16: Could we bound both sides of this window? Any event dated before readd.date has a negative delta, so it currently qualifies no matter how old it is and can be judged against a newer circle. Because the stamp is written after await monitor.add, simply adding a non-negative check could miss immediate replays created before add returns. Capturing a start before the add and a deadline after it, then testing a queued pre-add event, would cover both cases.

Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor+ContradictionGate.swift
```diff
@@ -0,0 +1,80 @@
...
+    /// Freshest fix obtainable for the gate: reuses the movement resolver (one-shot request when
+    /// the cache is stale, coalesced across concurrent events — a replay burst re-adds many
+    /// conditions at once and every event shares one resolution — cached-fix fallback on timeout).
+    /// A cached fix is acceptable as-is: the gate only runs moments after an add that itself
+    /// positioned the staging off the same cache, so it cannot meaningfully predate the state the
+    /// event claims. Returns via `bestKnownFix()` so the freshest of the cache and the request
+    /// wins; the caller's decision applies the fix-age guard to the result.
+    private func resolveGateFix() async -> CLLocation? {
+        await withCheckedContinuation { continuation in
+            movementFixResolver.resolve(cached: bestKnownFix()) { [weak self] _ in
```
Shahroz16, line 98: Could we share one fix attempt across a re-add burst, including caching a failed or timed-out result for that window? The events loop and pending drain both await process serially, so MovementFixResolver never receives concurrent calls to coalesce. With no usable fix, each fence starts another 10-second request, which can stall the single consumer for the number of fences times 10 seconds and delay or overflow later real events.

Shahroz16 (commented):
I left two comments around the replay-window bound and serial fix resolution.

Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor+ContradictionGate.swift
mrehan27: Confirmed, including the subtlety: a bare non-negative check against the post-add stamp would leak replays dated inside the remove→add gap. Fixed in 5b8615fe exactly as you outlined — a start captured before the remove/add pair plus the add's return, window = [start, added + 10 s]. The state-space model gained the buffered pre-add catch-up as a dimension (v8.2): the two-sided window refuses none of them across 125,952 paths per arm, and an upper-only negative control fails precisely on that class, so the model is now sensitive to this bound. Rig re-run on the updated code: storm and burst still 6/6 refused, real drive-in enters all delivered.

Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor+ContradictionGate.swift
mrehan27: Confirmed — the serial events loop means the resolver's coalescing never engages here (the doc comment claiming otherwise was wrong and is corrected). Fixed in 5542eebf: a resolution that comes back without a fresh fix is remembered for movementFixMaxAge, so a burst pays at most one request and later events skip straight to the cached read and fail open immediately; a late fix from the failed request still lands in latestFix where the cache read serves the rest of the burst. In the rig's burst the 6 refusals now land 2 ms apart off one shared fix.

Shahroz16 (approved):
The latest changes address my comments around the replay window and repeated fix resolution.

PR #1222, opened 2026-08-23
fix(in-app): deallocate subscribers on main thread

Shahroz16:
## Summary

- run in-app subscriber unsubscription on MainActor
- keep the task-owned subscriber and its WebKit-owning object graph on the main thread through final release
- add a regression test that verifies deinitialization happens on the main thread when the task is the last owner

## Validation

- regression test demonstrated failure before the fix and passed afterward
- focused regression test passed 20 repeated iterations
- InAppMessageStateTests passed
- Customer.io-Package build-for-testing passed
- MessagingInApp Debug and Release simulator builds passed
- SwiftLint passed with 0 violations
- APN-UIKit Maestro flow passed on iPhone 17 Pro simulator, iOS 26.5: live modal displayed, Continue dismissed it, and a subsequent custom event persisted with no crash signatures

## Linear

[MBL-2294: Crash on modal in-app message dismiss, WKWebView owning object](https://linear.app/customerio/issue/MBL-2294/crash-on-modal-in-app-message-dismiss-wkwebview-owning-object)

---

> [!NOTE]
> **Medium Risk**
> Threading change to all in-app store subscription paths; low behavioral surface area but touches modal/WebKit teardown where incorrect threading previously caused production crashes.
> 
> **Overview**
> Fixes a crash when modal in-app messages are dismissed by ensuring **subscribe** and **unsubscribe** work in `InAppMessageStoreManager` runs on the **main actor**. Those APIs now wrap store calls in `Task { @MainActor [store, subscriber] in ... }` so when the returned `Task` is the last strong owner of a subscriber tied to UI/WebKit, **deinitialization happens on the main thread**.
> 
> Adds regression tests with a deinit-tracking subscriber that assert main-thread teardown when the subscribe or unsubscribe task completes and releases the subscriber.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9e171a56da79cac79ddedb056eefe6252f413231. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Sources/MessagingInApp/State/InAppMessageManager.swift
+++ b/Sources/MessagingInApp/State/InAppMessageManager.swift
@@ -105,7 +105,9 @@ class InAppMessageStoreManager: InAppMessageManager {
     /// In order to keep the subscriber alive, the caller must keep a strong reference to the subscriber.
     @discardableResult
     func subscribe(subscriber: InAppMessageStoreSubscriber) -> Task<Void, Never> {
-        Task { await store.subscribe(subscriber) }
+        Task { @MainActor [store, subscriber] in
+            await store.subscribe(subscriber)
+        }
     }
 
     /// Subscribe to store updates with given subscriber and comparator.
@@ -120,6 +122,8 @@ class InAppMessageStoreManager: InAppMessageManager {
         comparator: @escaping (InAppMessageState, InAppMessageState) -> Bool,
         subscriber: InAppMessageStoreSubscriber
     ) -> Task<Void, Never> {
-        Task { await self.store.subscribe(subscriber, comparator) }
+        Task { @MainActor [store, subscriber] in
+            await store.subscribe(subscriber, comparator)
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


Sources/MessagingInApp/State/InAppMessageManager.swift
```diff
@@ -95,7 +95,9 @@ class InAppMessageStoreManager: InAppMessageManager {
 
     @discardableResult
     func unsubscribe(subscriber: InAppMessageStoreSubscriber) -> Task<Void, Never> {
-        Task { await store.unsubscribe(subscriber) }
+        Task { @MainActor [store, subscriber] in
+            await store.unsubscribe(subscriber)
+        }
```
mrehan27, line 100: Should we apply the same to `store.subscribe` in this class as well?

mrehan27 (approved):

Sources/MessagingInApp/State/InAppMessageManager.swift
Shahroz16: Yes, good catch. The subscribe tasks had the same transient strong-ownership risk. I moved both subscribe overloads onto MainActor with explicit store and subscriber captures, and added a regression that failed before the change and now passes, including 20 repeated iterations.

PR #1223, opened 2026-08-24
chore: align native UIScene migration sample and guidance

Shahroz16:
## Summary

- move the APN-UIKit sample Customer.io-initiated links to `SDKConfigBuilder.deepLinkCallback`
- keep OS-delivered universal links and URL contexts owned by `UISceneDelegate`
- make the sample claim only routes it handles and restore the visible universal-link destination
- document callback threading, lifetime, fallback, and scene ownership in public API comments and the internal developer note
- isolate sample routing tests with injected storage and notification delivery

The shipped SDK change is documentation-only. Runtime changes are confined to the APN-UIKit sample.

Customer-facing UIScene migration guidance belongs in the public documentation site and is tracked by MBL-2235.

## Validation

- strict SwiftLint and SwiftFormat
- full `Customer.io-Package` simulator build
- APN-UIKit sample build
- focused `APN UIKitTests/Tests`
- latest documentation-only follow-up passed strict local lint and will run the normal PR CI suite


docs/ios-uiscene-migration.md
mrehan27, line 1: Is this document intended as customer facing guidance? Wouldn't it be better suited for our docs site instead?

mrehan27 (approved):

docs/ios-uiscene-migration.md
Shahroz16: Good point. This is customer-facing guidance, so the public docs site should be the canonical source. I removed the repository-local guide and README link in bb8d0c4f. This PR now keeps only the executable sample, the public API comments, and the internal developer note. Publication of the customer guide is tracked in MBL-2235: https://linear.app/customerio/issue/MBL-2235/publish-ios-27-lifecycle-migration-guidance-and-coordinated-sdk

docs/dev-notes/DEEPLINKS-AND-NOTIFICATIONS.md
```diff
@@ -190,7 +190,8 @@ Tier 3: UIKitWrapper.open(url: url)
 ```
 
 **Notes:**
-- `deepLinkCallback` is registered via `@_spi(Internal) setDeepLinkCallback(_:)` — used by React Native and Flutter wrappers. Not intended for direct host app use.
+- Native hosts register `deepLinkCallback` through the public `SDKConfigBuilder.deepLinkCallback(_:)` API. React Native and Flutter wrappers install their native routing adapters through the internal `setDeepLinkCallback(_:)` SPI.
```
mahmoud-elmorabea, line 193: 🔴 Inverted. `grep setDeepLinkCallback` = zero hits in both wrapper repos, and customerio-flutter#399 / customerio-reactnative#641 both use the public `SDKConfigBuilder.deepLinkCallback`. Should say all three register through the public API; the SPI has no caller (if that's right, worth saying so).

If we plan to use a SPI,, is there a reason? There is already a public setter

Apps/APN-UIKit/APN UIKit/Util/DeepLinksHandlerUtil.swift
```diff
@@ -1,71 +1,82 @@
...
+        notificationCenter: NotificationCenter = .default
+    ) {
+        self.storage = storage
+        self.notificationCenter = notificationCenter
+    }
+
+    // Only claims the sample app's `apn-uikit://settings` route.
     func handleAppSchemeDeepLink(_ url: URL) -> Bool {
-        handleDeepLinkAction(url)
+        guard url.scheme?.lowercased() == "apn-uikit" else { return false }
```
mahmoud-elmorabea, line 33: 🟠 This guard drops an `https` destination unwrapped from a Live Activity redirect — `SceneDelegate.handle(urlContexts:)` is the only caller and never routes to `handleUniversalLinkDeepLink`.

docs/dev-notes/DEEPLINKS-AND-NOTIFICATIONS.md
Shahroz16: You are right about the wrapper call sites: Flutter #399 and React Native #641 configure the public builder and do not call the SPI directly. The SPI is not caller-less, though. `CustomerIO.initialize` uses it internally to transfer the public builder callback into `DeepLinkUtil` across SDK module boundaries. I corrected the internal note and SPI test wording in #1225; no wrapper or SPI API change is needed.

Apps/APN-UIKit/APN UIKit/Util/DeepLinksHandlerUtil.swift
Shahroz16: This identifies a valid sample-routing gap, but the `handleAppSchemeDeepLink` guard itself is intentional because that method only owns app-scheme routes. The issue was that `SceneDelegate` sent every unwrapped destination to it. `handleWidgetUrl` still recorded opened attribution before routing, and SDK core plus direct universal-link handling were unaffected. #1225 shares the existing scheme-aware helper with the scene path used by both cold and warm URL contexts and adds focused HTTPS, app-scheme, and decline regression coverage.

PR #825, opened 2024-10-09
chore: add resource loader for user agent generation

mrehan27:
part of [MBL-540](https://linear.app/customerio/issue/MBL-540/react-native)

### Changes

- Added `SdkResourceLoader` class to load resources from bundled SDK and generate `SdkClient` based on it
- Added tests for `SdkResourceLoader`

#### Native iOS (Main Target):

```
Customer.io iOS Client/3.4.1 (arm64; iOS 18.0) io.customer.ios-sample.apn-spm.APN-UIKit/1.0
```

#### Native iOS (NSE):

```
Customer.io iOS Client/3.4.1 (arm64; iOS 18.0) io.customer.ios-sample.apn-spm.APN-UIKit.richpush/1.0
```

#### React Native iOS (Main Target):

```
Customer.io ReactNative Client/3.9.0 (arm64; iOS 18.0) io.customer.ami/1.0
```

#### React Native iOS (NSE):

```
Customer.io ReactNative Client/3.9.0 (arm64; iOS 18.0) io.customer.ami.richpush/1.0
```

PR #622, opened 2024-03-11
docs: [vision-os-sample-app][6/n] Common Views

Ahmed-CIO:
This is a PR in a series of PRs to build sample app that runs Identify and Track in VisionPro
The app is built to be an interactive walk through guidance for developers who are integrating Customer.io Swift SDK in their VisioPro apps.
For more context:

- See the project [one pager](https://www.notion.so/custio/Spacial-Identify-and-Tracking-6587d848bbde494095c0585237326ed3?pvs=4)
- See [Linear Issue](https://linear.app/customerio/issue/MBL-115/sample-app-for-visionpro)

This PR adds punch of generic SwiftUI view
- `MainLayoutView` this view will be used to wrap all content and handle bunch of the navigation logic. Including displaying the `Logo`, the side menu, and error messages
- `LineView` custom view for flexible divider
- `FloatingTitleTextView` custom text field to ensure the field placeholder becomes labels when the field is filled.

- Open `VisionOs.xcproject`
- Hit CMD+R to run the project and make sure you have VisionOS simulator installed and selected

PR #1236, opened 2026-08-29
ci: add fast and full PR validation lanes

Shahroz16:
## Summary
- keep routine PR lint, tests, and API validation fast
- run sample distribution, SDK size comparison, and Maestro E2E only when `ci:full` is applied
- preserve main, scheduled, and manual validation behavior
- isolate unrelated label events so they cannot cancel a requested full run

## Validation
- `xcodebuild build-for-testing` succeeded on an iPhone 17 Pro simulator
- changed workflow YAML passes differential Actionlint validation
- no release or deploy workflow changed


Shahroz16:
## Revised CI decision after correctness review

Exact head: `75807a6e`

The review feedback was correct: semantic-release intent is not a correctness boundary. A `chore:`, `ci:`, or `test:` PR can break compilation, dependency resolution, public API, SDK size, or a sample app. This head no longer uses PR or commit semantics to remove those checks.

- Correctness stays automatic and path-based: tests, lint, API compatibility, SDK size, stable APN and FCM compile-and-launch, plus Xcode 27 compatibility for sample, package, podspec, deployment-target, and workflow integration paths.
- `ci:full` gates true Maestro end-to-end coverage. Schedules and manual dispatch remain available.
- PR title and commit semantics control only signed sample distribution side effects.
- Stable SDK release and deploy workflows are unchanged.
- The newly introduced Xcode setup action is pinned to an immutable SHA.

This PR changes CI integration paths, so it intentionally exercised the conservative lane even with no `ci:full` label. Final hosted result: **18 successful, 3 intentional skips, 1 neutral, 0 failures**. APN and FCM stable launches, both Xcode 27 fixtures, SDK size, tests, API checks, and lint passed. Signed distribution and Maestro skipped as designed.

Measured on this final run, the two Xcode 27 preview jobs consumed **13m08s of Mac runner time**. A routine `Sources/**`-only PR keeps stable APN, stable FCM, and SDK size, but does not spend those two preview jobs.

Claude Code Opus rejected the earlier semantic lane split because iOS source coverage was incomplete and Mac work was duplicated. After the path-based redesign it returned **APPROVE**, with no blocker or high-severity findings. All changed workflows pass actionlint, the local package build passed, and the final hosted head is green.

.github/scripts/validate-release-policy.sh
```diff
@@ -0,0 +1,89 @@
+#!/usr/bin/env bash
```
mrehan27, line 1: **P2** — this script and `reusable-release-policy.yml` are byte-identical across all five SDK repos (sha256 match), with nothing keeping them in sync. `customerio/mobile-ci-tools` already hosts shared reusable workflows — this change calls one. Moving it there is mechanical with zero drift today; after merge, every policy fix becomes a five-repo change.

.github/workflows/maestro-e2e.yml
```diff
@@ -34,11 +35,37 @@ on:
...
+
+  # Metadata edits rerun full so a newer skipped check cannot replace a failure.
   prepare:
     name: Select iOS suites
+    needs: release_policy
+    if: >-
+      always() &&
+      (vars.MOBILE_E2E_ENABLED == 'true' || github.event_name == 'workflow_dispatch') &&
+      (github.event_name != 'pull_request' ||
+      ((github.event.action == 'edited' ||
```
mrehan27: **P2** — `github.event.action == 'edited' || github.event.action == 'labeled'` forces the full lane regardless of classification, so a title typo fix or any unrelated label runs the expensive lane on a non-release PR. Gating `labeled` on `github.event.label.name == 'ci:full'` keeps the intent without that case.

.github/workflows/maestro-e2e.yml
```diff
@@ -34,11 +35,37 @@ on:
 
 concurrency:
   group: maestro-ios-${{ github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```
mrehan27, line 45: **P2** — concurrency policy diverges across the five repos:

- android: `cancel-in-progress: ${{ github.event_name == 'pull_request' }}`
- ios, flutter: same, plus `|| startsWith(github.ref, 'refs/heads/feature/')`
- reactnative: ungated `cancel-in-progress: true`
- expo-plugin: both forms, in the same PR

Worth converging, or a comment if the differences are deliberate.

.github/workflows/compile-changed-sample-apps.yml
```diff
@@ -0,0 +1,167 @@
...
+    name: Compile APN sample
+    needs: detect-changes
+    if: needs.detect-changes.outputs.apn == 'true'
+    runs-on: macos-26
+    timeout-minutes: 60
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install Xcode
+        uses: maxim-lobanov/setup-xcode@v1.6.0
```
mrehan27: **P2** — `maxim-lobanov/setup-xcode@v1.6.0` is a third-party action pinned by tag; every other third-party action here is SHA-pinned (first-party `actions/*` use major tags). Newly introduced, not on main. Also at :142.

mrehan27 (commented):
Cross-repo review — the same change across five repos.

The lane gating looks right: omitting `always()` so a skipped `prepare` transitively blocks E2E, and moving the dependabot guard to `github.event.pull_request.user.login` (on a `labeled` event `github.actor` is the labeler, not the author).

Main ask: the release-policy script and reusable workflow are byte-identical in all five repos today (sha256 match), so extracting them to `mobile-ci-tools` is free now and a five-repo change after merge.

.github/workflows/compile-changed-sample-apps.yml
Shahroz16: Addressed in `75807a6e`: all three newly introduced `maxim-lobanov/setup-xcode` references now pin the immutable SHA `60606e260d2fc5762a71e64e74b2174e8ea3c8bd` for v1.6.0. The affected workflow passes actionlint and the composite action manifest parses cleanly.

.github/scripts/validate-release-policy.sh
Shahroz16: Agreed that five copies create drift risk. I am deliberately not extracting this classifier in these PRs because it now gates signed/published side effects, and this rollout's safety boundary is to avoid adding a new cross-repository dependency to release-adjacent behavior. Existing `mobile-ci-tools@main` usage does not make that change free: a shared policy should be versioned, pinned, canaried in one SDK, and then migrated across callers. The five local copies are byte-identical and fixture-tested for this rollout; extraction is a focused follow-up.

.github/workflows/maestro-e2e.yml
Shahroz16: Reviewed across all five. The differences follow trigger scope rather than different PR policy: superseded PR runs cancel everywhere; iOS/Flutter also cancel feature-branch pushes; Android preserves non-PR sample/release runs; Expo's full wrapper is PR-only; and the React Native scene workflow no longer cancels scheduled/manual/reusable runs. Standardizing the surface syntax here would change existing non-PR behavior, which is outside this PR's release-safety boundary.

.github/workflows/build-sample-apps.yml
```diff
@@ -17,17 +18,37 @@ on:
         default: false
 
 concurrency:
+  # Ignore unrelated label events without cancelling the active validation run.
   group: ${{ github.workflow }}-${{ github.ref }}
```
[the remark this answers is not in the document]
Shahroz16: Current head `75807a6e` uses this concurrency group as an exact-SHA, fail-closed replacement: metadata events repeat the SDK-size correctness job instead of replacing a prior failure with a skipped check. Unrelated labels cannot start signed distribution because the job requires `event.label.name == ci:full`, and they cannot start Maestro for the same reason. This can restart the size job on a rare metadata event, but it does not remove validation coverage or create a false green.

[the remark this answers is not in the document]
Shahroz16: You are right. My earlier reply verified the job guards but missed that an unrelated label could cancel an already-running `ci:full` distribution or Maestro run. Fixed in `61f3ecc8`: unrelated labels and body-only edits now use a separate `metadata` concurrency group, while `ci:full`, title/base changes, and new commits stay in `validation`. The SDK-size check still repeats fail-closed on metadata, but those runs cannot update the shared PR size comment, avoiding a stale-report race found in the first Opus review. All changed workflows pass `actionlint`, and the corrected event model received an explicit Opus approval.

.github/workflows/maestro-e2e.yml
Shahroz16: Addressed on current head `61f3ecc8`. A bare metadata event no longer requests the full lane: release classification still controls eligibility, `ci:full` must be the label that fired the labeled event, and unrelated labels/body edits are isolated in the `metadata` concurrency group.

.github/workflows/build-sample-apps.yml
```diff
@@ -17,17 +18,44 @@ on:
...
-  cancel-in-progress: true
+  # Unrelated labels and body-only edits must not cancel an in-flight release or
+  # ci:full run. They share a separate group, where SDK-size validation repeats.
+  group: >-
+    ${{ github.workflow }}-${{ github.ref }}-${{
+      github.event_name == 'pull_request' &&
+      ((github.event.action == 'labeled' && github.event.label.name != 'ci:full') ||
+      (github.event.action == 'edited' &&
+      github.event.changes.title.from == '' &&
+      github.event.changes.base.ref.from == '')) &&
```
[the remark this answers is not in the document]
Shahroz16: Addressed in `69ba2571`. The concurrency and iOS comment-writer guards now test the presence of `github.event.changes.title` / `.base` objects directly. Body-only edits therefore use `metadata` and cannot write the shared size comment; title edits and base retargets use `validation`. `actionlint` passes and Opus approved the explicit truth table.

.github/workflows/ios-toolchain-compatibility.yml
```diff
@@ -23,18 +32,84 @@ env:
...
+                  ;;
+                # The APN SwiftPM fixture compiles every source module except
+                # MessagingPushFCM and Migration. The CocoaPods FCM fixture
+                # includes those two modules as well as the shared SDK graph.
+                Sources/MessagingPushFCM/*|Sources/Migration/*)
+                  fcm=true
+                  ;;
+                Sources/*)
+                  apn=true
+                  ;;
```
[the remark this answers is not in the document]
Shahroz16: Verified against the same Bash case semantics used by the runner. This is pattern matching, not pathname expansion, so slash is not special and star does match nested components. Representative paths selected as expected: Sources/MessagingInApp/Gist/Network/BaseNetwork.swift selected APN; Sources/MessagingPushFCM/Integration/Nested/File.swift selected FCM; Sources/Migration/Nested/File.swift selected FCM. The historical replay also included nested Sources paths. No selector gap here, so I am resolving this as a false positive.

Shahroz16:
## Cross-SDK historical selector audit

This is the centralized audit note for iOS #1236, Android #825, React Native #649, Flutter #402, and Expo #399.

I replayed each final path-selection policy over the 20 most recently merged PRs and 20 consecutive release-to-release diffs in each repository: 100 PR diffs and 100 release intervals total. The first replay found real under-selection rather than simply confirming the design:

- 9 iOS source-only PRs would not have selected Xcode 27 preview validation.
- React Native #622 changed the example app but would not have selected the FCM consumer.

Both gaps are fixed. The final replay has zero misses against the documented correctness-path mapping across the 100 PR diffs, and every one of the 100 release intervals selects its relevant release, native, packaging, and compatibility paths. Internal selectors use `git diff --no-renames --name-only -z`, so a rename exposes both its deleted source path and added destination path. Missing SHAs fail red instead of producing an empty matrix.

The replay validates path decisions, not old binaries or label-state expressions. I separately exercised the `ci:full`, unrelated-label, retained-label-on-synchronize, nested-path, deletion, rename, empty-diff, fork, manual, and scheduled cases. The exact current heads were then run in hosted CI.

Historical job-cell selection, not runner-minute measurement:

- iOS stable samples: 26 selected variants versus 38 if both ran on every non-documentation PR, 31.6% fewer. Xcode 27 preview: 20 versus 38, 47.4% fewer. Source changes always get a representative preview fixture; packaging and integration changes select every affected fixture; nightly and manual runs select both.
- Android keeps unit, instrumentation, API, lint, binary-size, sample, and the real AGP floor automatic. The controlled snapshot plus compatibility measurement remains 23.4 to 14.6 runner-minutes, 37.7% less load, while replacing four false-green AGP jobs with one real combined build.
- React Native samples: 57 selected variants versus 60 broad variants. Full scene E2E now runs automatically for direct routing/runtime-bridge paths, 8 of the last 20 PRs and 6 of 20 release intervals, versus 13 of 20 under the old broad list. `ci:full` remains the escape hatch and continues validating later commits; nightly remains the backstop.
- Flutter samples: 35 selected variants versus 38 broad variants across the 19 file-changing PRs; one historical PR was an empty diff.
- Expo compatibility: 96 selected variants versus 120 broad variants, 20% fewer; pnpm integration selected 12 of 20 PRs.

Final hosted rollup on the exact pushed heads: 114 passed checks, 6 intentional skips, 0 pending, 0 failures. iOS APN/FCM stable launches, both Xcode 27 fixture launches, SDK size, tests and API passed. React Native APN/FCM signed samples and the 20-minute warm/terminated scene E2E passed. Android, Flutter and Expo are fully green. Expo had one first-attempt npm package-parse failure before native compilation; the unchanged retry passed that step and the full iOS native validation, so it is recorded as transient rather than hidden.

Claude Code Opus initially returned NO-GO because label-only React Native E2E removed automatic routing coverage. After the affected-path plus `ci:full` hybrid and fail-closed fixes, its review of the final pushed heads returned GO with no blockers.

Stable SDK release and deploy workflows remain unchanged.

mahmoud-elmorabea (approved):

Shahroz16:
On the shared release-policy extraction: agreed as a follow-up after these PRs merge. Keeping the identical local copies in this stack avoids making five active PRs depend on a sixth, unmerged mobile-ci-tools change. Once the behavior is merged and stable, extraction can happen atomically with pinned shared-workflow references.

PR #1237, opened 2026-08-30
chore: classify and log in-app message failures internally

mahmoud-elmorabea:
## MBL-2310 · iOS stack — step 2 of 3

| Step | PR | |
| --- | --- | --- |
| 1 | #1231 — `chore:` fire the in-app error callback once on engine timeout |  |
| **2** | **#1237 — `chore:` classify and log in-app message failures internally** | **← you are here** |
| 3 | #1238 — `fix:` surface the failure reason on `errorWithMessage` | cuts the iOS release |

Merge bottom-up. Every `chore:` step is internal only — no public type, no protocol or interface change, and the committed API surface is untouched — so nothing user-visible ships if an unrelated release cuts while they sit on main. The final `fix:` carries the whole feature and is the only step that triggers a release.

**Android mirror:** customerio/customerio-android#823 → customerio/customerio-android#826 → customerio/customerio-android#827 → customerio/customerio-android#828. Both platforms land the same API and must release before the wrapper PRs (Flutter, React Native) can start.

---
Gives every in-app failure a reason and keeps the detail the SDK was throwing away. Internal only — nothing public changes yet.

Stacked on #1231. Groundwork for MBL-2310; the next PR puts the reason on `errorWithMessage`.

### Why

All nine failure sites in `EngineWeb` collapsed into a bare `delegate?.error()`, so why a message failed survived only in whatever each site happened to log. Two of the most useful sources were dropped entirely:

- **The renderer's own `errorMessage`.** The 3.0 bundle sends `{ target, errorMessage }` with the engine `error` event — e.g. `Unable to find "step-2" in payload.` `EngineWeb` discarded `eventProperties`. That string is the only first-hand account of why a message would not render.
- **The `NSError` behind a failed navigation.** `didFail` and `didFailProvisionalNavigation` both received one and ignored it, losing the description and the `NSURLError` code — which is the most likely real-world failure, the renderer being unreachable.

### Change

An internal `InAppMessageError` (`reason`, `detail`, `code`) with five coarse reasons, each mapping to a different thing an integrator would actually do: `network`, `timeout`, `renderFailed`, `webViewCrashed`, `internalError`.

All nine sites now go through one `reportFailure` choke point that classifies and logs before notifying the delegate, so the next PR changes the delegate hand-off in exactly one place. `EngineEventHandler.getErrorProperties` parses the renderer payload. The 5s bootstrap timeout is now a named constant rather than a literal.

`MessageManager.routeError` is classified too, though renderer 3.0 never emits `routeError` — noted in a comment so nobody counts it as live coverage.

### No public API movement

Every new declaration is internal — `InAppMessageError`, `InAppMessageErrorReason`, `getErrorProperties`, and `bootstrapTimeoutSeconds` (a member of a public class, so internal by default). `api-docs/` is untouched. `EngineWebDelegate` is public API, which is exactly why widening it waits for the `fix:` PR.

### Verification

`InAppMessageErrorTests` covers the renderer payload parser across all four combinations of `errorMessage`/`target` plus the no-parameters case, the `NSError` mapping, and log formatting.

`MessagingInAppTests/InAppMessageErrorTests` + `MessagingInAppTests/EngineWebTimeoutTests` — 9 tests, 0 failures. `MessagingInApp` scheme builds clean.

### DIFF
--- a/Sources/MessagingInApp/Gist/EngineWeb/EngineEventHandler.swift
+++ b/Sources/MessagingInApp/Gist/EngineWeb/EngineEventHandler.swift
@@ -49,6 +49,29 @@ enum EngineEventHandler {
         extractRoute(properties: properties)
     }
 
+    /// Pulls the renderer's own description of a failure out of an `error` event.
+    ///
+    /// The renderer sends `{ target, errorMessage }` — e.g. `Unable to find "step-2" in payload.`
+    /// That string is the only first-hand account of why a message would not render, and it was
+    /// being dropped on the floor.
+    static func getErrorProperties(properties: EngineEventProperties) -> String? {
+        guard let parameters = properties["parameters"] else { return nil }
+
+        let errorMessage = parameters["errorMessage"] as? String
+        let target = parameters["target"] as? String
+
+        switch (errorMessage, target) {
+        case (.some(let message), .some(let target)):
+            return "\(message) (target: \(target))"
+        case (.some(let message), .none):
+            return message
+        case (.none, .some(let target)):
+            return "Engine reported an error for target: \(target)"
+        case (.none, .none):
+            return nil
+        }
+    }
+
     static func getRouteLoadedProperties(properties: EngineEventProperties) -> String? {
         extractRoute(properties: properties)
     }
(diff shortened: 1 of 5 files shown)


Shahroz16 (approved):

Shahroz16 (approved):
Re-reviewed the rebased head and the full current stack. The patch is unchanged from the previously approved version, the focused in-app suite passes, and the exact child stack is green. The lone failure on this PR is the unrelated `SynchronizedEquatableTests` timing test.

PR #1238, opened 2026-08-30
fix: surface the failure reason on in-app error callback

mahmoud-elmorabea:
## MBL-2310 · iOS stack — step 3 of 3

| Step | PR | |
| --- | --- | --- |
| 1 | #1231 — `chore:` fire the in-app error callback once on engine timeout |  |
| 2 | #1237 — `chore:` classify and log in-app message failures internally |  |
| **3** | **#1238 — `fix:` surface the failure reason on `errorWithMessage`** | **← you are here** |

Merge bottom-up. Every `chore:` step is internal only — no public type, no protocol or interface change, and the committed API surface is untouched — so nothing user-visible ships if an unrelated release cuts while they sit on main. The final `fix:` carries the whole feature and is the only step that triggers a release.

**Android mirror:** customerio/customerio-android#823 → customerio/customerio-android#826 → customerio/customerio-android#827 → customerio/customerio-android#828. Both platforms land the same API and must release before the wrapper PRs (Flutter, React Native) can start.

---
Puts the reason an in-app message failed onto `errorWithMessage`. Closes the MBL-2310 work on iOS.

Stacked on #1237. Paired with android#828.

### Why

`errorWithMessage` delivered only `messageId` and `deliveryId`, so an integrator learned that a message failed but never why. The reason existed on device in debug logs only, and nothing is recorded server-side, so diagnosing meant shipping a debug-logging build.

### API

```swift
public enum InAppMessageErrorReason: String {
    case network, timeout, renderFailed, webViewCrashed, internalError
}

public struct InAppMessageError: Error, Equatable {
    public let reason: InAppMessageErrorReason
    public let detail: String?   // renderer's own message, or the NSError description
    public let code: Int?        // NSURLError code / HTTP status where one exists
}

func errorWithMessage(message: InAppMessage, error: InAppMessageError)
```

Reasons are coarse on purpose: each maps to a different thing an integrator would actually do — check connectivity, look at a slow renderer, report the message to us, or file an SDK bug. `detail` is diagnostic text for logs, not for parsing. `InAppMessageError` conforms to `Error` so hosts can log or throw it idiomatically.

### Additive

The new requirement ships with a protocol-extension default that forwards to `errorWithMessage(message:)`, so existing conformers keep compiling **and keep receiving failures**. Same for the two other public protocols this had to widen, `GistDelegate` and `EngineWebDelegate`.

The old callback is deliberately **not** deprecated in this release — `@available(*, deprecated)` would emit a warning in every existing host for a purely additive change.

`api-docs/CioMessagingInApp.api` moves by **11 insertions and zero deletions or modifications**.

### One wrinkle worth flagging for review

The generated mock derives its member names from a method's base name, so the overload collided with the existing `errorWithMessage` — `_errorWithMessageCallsCount` was declared twice and would not compile. The stencil already supports `// sourcery: Name = "…"` for exactly this, so the new method is annotated `errorWithMessageAndError` and the mock's members are named from that. No stencil change, no other module affected.

### Verification

- `InAppListenerSourceCompatTests` locks source compatibility the same way `InboxProtocolSourceCompatibilityTests` does: a conformer implementing only the pre-change members, so the file stops compiling if a default is ever removed. It also asserts the default actually **forwards** rather than silently swallowing, which is what keeps old listeners working.
- Existing negative assertions were strengthened to cover the new member too — otherwise they would have passed vacuously.
- New test pins `.internalError` for a failure arriving through the reason-less entry point.
- `MessagingInAppTests` — 499 tests, 0 failures.
- `Apps/APN-UIKit` demos the new callback, tracking `reason`, `detail`, and `code`. Built locally, since `build-sample-apps` no longer runs on PRs.

### DIFF
--- a/Sources/MessagingInApp/Gist/EngineWeb/EngineWeb.swift
+++ b/Sources/MessagingInApp/Gist/EngineWeb/EngineWeb.swift
@@ -51,6 +51,10 @@ public class EngineWeb: NSObject, EngineWebInstance {
 
     /// A failure raised before a delegate was attached, held until one is.
     private var pendingFailure: InAppMessageError?
+
+    /// Set once the first failure is accepted. An engine instance renders one message, so it fails
+    /// at most once.
+    private var hasReportedFailure = false
     var webView = WKWebView()
 
     public var view: UIView {
@@ -153,7 +157,21 @@ public class EngineWeb: NSObject, EngineWebInstance {
     /// `MessageManager` owns the resulting `messageLoadingFailed` dispatch — see `forcedTimeout`.
     /// A failure raised before the delegate is attached is held and delivered on assignment, so no
     /// path here reports into the void.
+    ///
+    /// The first failure wins and every later one is dropped. WebKit can deliver more than one
+    /// failure callback for a single load, and the bootstrap timer is still armed when a failure
+    /// arrives — the dispatch it triggers is asynchronous, so without latching, a host could be told
+    /// `.network` and then `.timeout` about the same message. That is deterministic for inline
+    /// messages, whose dismissal never reaches `cleanEngineWeb()`.
     private func reportFailure(_ error: InAppMessageError) {
+        guard !hasReportedFailure else { return }
+        hasReportedFailure = true
+
+        // The timer retains self through its target, so leaving it armed also keeps this instance
+        // alive for the rest of the timeout after the message has already failed.
+        _timeoutTimer?.invalidate()
+        _timeoutTimer = nil
+
         logger.logWithModuleTag(
             "Message \(currentMessage.describeForLogs) failed: \(error.describeForLogs)",
             level: .error
--- a/Sources/MessagingInApp/Gist/EngineWeb/EngineWeb.swift
+++ b/Sources/MessagingInApp/Gist/EngineWeb/EngineWeb.swift
@@ -11,6 +11,15 @@ public protocol EngineWebDelegate: AnyObject {
     func routeLoaded(route: String)
     func sizeChanged(width: CGFloat, height: CGFloat)
     func error()
+    /// Called when a message fails to load or render, with the reason it failed.
+    func error(_ error: InAppMessageError)
+}
+
+public extension EngineWebDelegate {
+    /// Defaulted so conformers written against the reason-less callback keep compiling.
+    func error(_ error: InAppMessageError) {
+        self.error()
+    }
 }
 
 protocol EngineWebInstance: AutoMockable {
@@ -136,7 +145,7 @@ public class EngineWeb: NSObject, EngineWebInstance {
             "Message \(currentMessage.describeForLogs) failed: \(error.describeForLogs)",
             level: .error
         )
-        delegate?.error()
+        delegate?.error(error)
     }
 
     @objc
--- a/Sources/MessagingInApp/Gist/EngineWeb/EngineWeb.swift
+++ b/Sources/MessagingInApp/Gist/EngineWeb/EngineWeb.swift
@@ -39,7 +39,18 @@ public class EngineWeb: NSObject, EngineWebInstance {
     /// How long the engine gets to report `bootstrapped` before the message is treated as failed.
     static let bootstrapTimeoutSeconds: TimeInterval = 5.0
 
-    public weak var delegate: EngineWebDelegate?
+    public weak var delegate: EngineWebDelegate? {
+        didSet {
+            // `loadMessage()` runs from `init`, before the delegate exists, so a failure raised
+            // there has nobody to report to yet. Deliver it as soon as one is attached.
+            guard delegate != nil, let failure = pendingFailure else { return }
+            pendingFailure = nil
+            delegate?.error(failure)
+        }
+    }
+
+    /// A failure raised before a delegate was attached, held until one is.
+    private var pendingFailure: InAppMessageError?
     var webView = WKWebView()
 
     public var view: UIView {
@@ -140,12 +151,18 @@ public class EngineWeb: NSObject, EngineWebInstance {
     /// Single exit for every failure in this class: classify, log, then tell the delegate.
     ///
     /// `MessageManager` owns the resulting `messageLoadingFailed` dispatch — see `forcedTimeout`.
+    /// A failure raised before the delegate is attached is held and delivered on assignment, so no
+    /// path here reports into the void.
     private func reportFailure(_ error: InAppMessageError) {
         logger.logWithModuleTag(
             "Message \(currentMessage.describeForLogs) failed: \(error.describeForLogs)",
             level: .error
         )
-        delegate?.error(error)
+        guard let delegate = delegate else {
+            pendingFailure = error
+            return
+        }
+        delegate.error(error)
     }
 
     @objc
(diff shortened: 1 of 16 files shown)


Sources/MessagingInApp/Type/InAppMessageError.swift
```diff
@@ -5,7 +5,7 @@ import Foundation
 /// The cases are deliberately coarse: each one maps to a different thing an integrator would do
 /// about it — check connectivity, look at a slow renderer, report the message content to us, or
 /// file an SDK bug. Finer detail belongs in ``InAppMessageError/detail``.
-enum InAppMessageErrorReason: String {
+public enum InAppMessageErrorReason: String {
```
Shahroz16, line 13: Could we document that callers should include an @unknown default or default branch when switching on this enum? The package is distributed from source, so adding another reason later will break exhaustive switches unless we set that expectation now. I left the equivalent note on Android as well.

Sources/MessagingInApp/Type/InAppMessageError.swift
```diff
@@ -21,19 +21,19 @@ enum InAppMessageErrorReason: String {
...
-struct InAppMessageError: Error, Equatable {
+public struct InAppMessageError: Error, Equatable {
     /// The coarse category. Branch on this.
-    let reason: InAppMessageErrorReason
+    public let reason: InAppMessageErrorReason
 
     /// Human-readable detail from the layer that failed — a `WKWebView` error description, or the
     /// message the renderer itself reported. Free-form and unstable: log it, don't parse it.
-    let detail: String?
+    public let detail: String?
```
Shahroz16, line 37: Could we make the safety guidance for detail a little stronger before this becomes public? Some values come directly from the renderer, so I would say it is for local diagnostics and should not be forwarded verbatim to analytics. The sample currently tracks error-detail, which encourages exactly that pattern.

Sources/MessagingInApp/Type/InAppMessageError.swift
```diff
@@ -21,19 +21,19 @@ enum InAppMessageErrorReason: String {
...
 
     /// The underlying platform error code where the failing layer had one, e.g. an `NSURLError`
     /// code or an HTTP status. `nil` when the failure carried no numeric code.
-    let code: Int?
+    public let code: Int?
```
Shahroz16, line 45: I do not think iOS can currently populate an HTTP status here. The implementation only maps NSError from the navigation failure callbacks and never inspects WKNavigationResponse, so the network case's ‘host returned an error status’ wording and this HTTP-status example promise parity with Android that we do not have. Could we describe this as an NSURLError or underlying platform code until response handling is added?

Sources/MessagingInApp/Type/InAppMessageError.swift
mahmoud-elmorabea: Done. One deviation worth calling out: I documented a plain `default:` rather than `@unknown default:`.

The SDK is distributed as source through SPM and CocoaPods, so it builds without library evolution and the compiler treats this enum as frozen at the integrator's build. `@unknown default:` is the tool for resilient enums and would be the wrong advice here — a plain `default:` is what actually keeps a host compiling across an added case. Happy to switch if you read it differently.

Same guidance added on Android in #828 (`else ->`).

Sources/MessagingInApp/Type/InAppMessageError.swift
mahmoud-elmorabea: Done, and you were right that the sample was undercutting the guidance. `detail` is now documented as local diagnostics only — log it, don't parse it, and don't forward it verbatim to analytics or crash reporting.

APN-UIKit now logs the detail and tracks only `error-reason` and `error-code`. Same change to `kotlin_compose` in #828.

Sources/MessagingInApp/Type/InAppMessageError.swift
mahmoud-elmorabea: You're right, and this was my error. iOS cannot produce an HTTP status here: an error response still completes navigation, so it never reaches `didFail` / `didFailProvisionalNavigation`, and we don't implement `decidePolicyFor navigationResponse`. The doc was promising Android parity that does not exist.

Fixed in both places — `code` now says it is the `NSError` code behind a failed navigation, typically an `NSURLError`, with an explicit note that it is not an HTTP status; and the `network` case no longer mentions a host error status.

Sources/MessagingInApp/Gist/EngineWeb/EngineWeb.swift
```diff
@@ -131,12 +151,18 @@ public class EngineWeb: NSObject, EngineWebInstance {
...
     /// `MessageManager` owns the resulting `messageLoadingFailed` dispatch — see `forcedTimeout`.
+    /// A failure raised before the delegate is attached is held and delivered on assignment, so no
+    /// path here reports into the void.
     private func reportFailure(_ error: InAppMessageError) {
         logger.logWithModuleTag(
             "Message \(currentMessage.describeForLogs) failed: \(error.describeForLogs)",
             level: .error
         )
-        delegate?.error()
+        guard let delegate = delegate else {
```
Shahroz16, line 179: Could we make the first failure terminal here? Dispatching `messageLoadingFailed` is asynchronous, so a navigation, renderer, or content-process failure can leave the bootstrap timer alive long enough to report a second `.timeout`; repeated WebKit failure callbacks can also reach the host more than once. This is deterministic for inline messages because their failure does not drive the modal-state cleanup that invalidates the timer. Invalidating and niling `_timeoutTimer`, plus latching the first reported failure before logging or delegating, would keep the new callback once per engine instance. A test that reports two different failures should pin the behavior.

Shahroz16 (changes requested):
The three original API and documentation points are resolved. One remaining callback-deduplication issue needs to be fixed before the final stack is ready.

Sources/MessagingInApp/Gist/EngineWeb/EngineWeb.swift
mahmoud-elmorabea: Fixed, and you were right on every part of it — including the inline detail.

Verified before changing anything: `_timeoutTimer` was invalidated in exactly three places — when scheduling a new one, in `cleanEngineWeb()`, and on `.bootstrapped`. `reportFailure` never touched it. And `cleanEngineWeb()` is only reached from the modal teardown path, so for inline this was not a race at all.

`reportFailure` now latches the first failure and invalidates the bootstrap timer. One extra reason to kill the timer: `Timer.scheduledTimer(target:)` retains self, so leaving it armed also kept the engine alive for the rest of the timeout after the message had already failed.

Test added as you suggested, reporting two different failures. The negative control describes the bug better than I can — with the latch removed it fails as:

```
XCTAssertEqual failed: ("2") is not equal to ("1")
XCTAssertEqual failed: ("…timeout") is not equal to ("…network")
XCTAssertEqual failed: ("nil") is not equal to ("Optional(-1009)")
```

Two callbacks, and the `.timeout` overwrote both the real cause and its `NSURLError` code. Suite is 501 tests, 0 failures.

Shahroz16 (approved):
The first-failure latch closes the duplicate callback path, including inline renderers, and preserves the original failure reason. The focused 21-test pass is clean and the full exact-head CI is green.

Shahroz16 (approved):
Re-reviewed the restacked head. The six PR commits are patch-equivalent to the approved version, and the only inherited change is the timer-test cleanup from #1231. Looks good.

PR #1239, opened 2026-08-31
chore(geofence): add the polygon geometry kernel

mrehan27:
Part of [MBL-2290](https://linear.app/customerio/issue/MBL-2290/ios-add-polygon-wire-contract-and-geometry-foundation)

Part 1 of the polygon geofence work. Pure geometry with **no consumer yet** — nothing
reads `vertices` until the next PR, so this changes no runtime behaviour at all.

### What's here

**`PolygonRegion`** — point-in-polygon (ray cast, half-open rule so a crossing shared by two edges
counts once) and signed edge distance, on a local equirectangular projection around the vertex
centroid. The projection is shared verbatim with the Android SDK; at fence scale its error is far
below GPS accuracy and both SDKs agree within the cross-SDK fixtures' 0.5 m tolerance.

**`Geofence.vertices`** — the field a ring will land in, plus decode that tolerates caches written
by SDK versions predating it.

**The sign convention is positive-inside, the opposite of the circle path.**
`PolygonRegion.signedEdgeDistance` is positive inside; the circle path's edge distance
(`distanceFromCenter - radius`, what `BaselineHealDecision` consumes) is negative inside.

That is not a preference — it is pinned by the geometry fixtures already shared with Android and
cannot be flipped unilaterally. It is documented at the declaration, and every consumer has to
convert rather than assume. During the spike a consumer was written against the wrong assumption and
produced inverted verdicts that neither side's unit tests could see, because each was internally
consistent. The test that pins the relationship in terms of physical position rather than sign lands
with its consumer in a later PR.

### Testing

365/365 `LocationGeofenceTests` pass on iOS 26. Format and lint clean. No public API surface, so no
docs baseline change.

New coverage: kernel unit tests and the cross-SDK geometry fixtures.

### Stack

1. `mbl-2290-a-polygon-geometry-kernel` — geometry kernel ← **this PR**
6. `mbl-2291-d-polygon-tripwire-state` — tripwire state (#1247)
7. `mbl-2291-e-polygon-tripwire-planting` — tripwire planting + budget (#1248)

### DIFF
--- a/Sources/LocationGeofence/Geometry/PolygonRegion.swift
+++ b/Sources/LocationGeofence/Geometry/PolygonRegion.swift
@@ -37,6 +37,8 @@ struct PolygonRegion {
     /// A repeat describes a zero-length edge, which contributes nothing to either containment or
     /// edge distance; collapsing it keeps the count comparable to the server's own cap, which is
     /// stated in unique vertices.
+    /// Canonicalises and projects a ring already known to be well formed. O(n): the degeneracy
+    /// checks live in `init(validating:)`, not here, because callers rebuild a region per evaluation.
     init?(vertices: [LocationData]) {
         var open: [LocationData] = []
         open.reserveCapacity(vertices.count)
@@ -47,17 +49,13 @@ struct PolygonRegion {
             open.removeLast()
         }
         guard open.count >= 3 else { return nil }
-        self.vertices = open
 
         let lat0 = open.map(\.latitude).reduce(0, +) / Double(open.count)
         let lon0 = open.map(\.longitude).reduce(0, +) / Double(open.count)
         let latitudeRadians = lat0 * Self.degreesToRadians
         let longitudeRadians = lon0 * Self.degreesToRadians
         let cosLatitude = cos(latitudeRadians)
-        self.referenceLatitudeRadians = latitudeRadians
-        self.referenceLongitudeRadians = longitudeRadians
-        self.cosReferenceLatitude = cosLatitude
-        self.projected = open.map {
+        let planar = open.map {
             Self.project(
                 $0,
                 referenceLatitudeRadians: latitudeRadians,
@@ -65,6 +63,82 @@ struct PolygonRegion {
                 cosReferenceLatitude: cosLatitude
             )
         }
+        self.vertices = open
+        self.referenceLatitudeRadians = latitudeRadians
+        self.referenceLongitudeRadians = longitudeRadians
+        self.cosReferenceLatitude = cosLatitude
+        self.projected = planar
+    }
+
+    /// Decode-time construction: additionally rejects rings that cannot be monitored sensibly —
+    /// zero area, or self-intersecting. Separate from `init(vertices:)` because that one runs on
+    /// hot paths (once per polygon per wake and per evaluation) while these checks are O(n²), so
+    /// they belong at the API boundary where they run once per sync.
+    init?(validating vertices: [LocationData]) {
+        self.init(vertices: vertices)
+        guard Self.enclosesArea(projected), !Self.selfIntersects(projected) else { return nil }
+    }
+
+    /// A ring below this is a line or a point. It can never contain anything, so it would hold an OS
+    /// monitoring slot and pull the shared wake circle toward its floor while never firing.
+    private static let minimumAreaSquareMeters = 1.0
+
+    private static func enclosesArea(_ ring: [Point]) -> Bool {
+        var twiceArea = 0.0
+        for i in 0 ..< ring.count {
+            let a = ring[i]
+            let b = ring[(i + 1) % ring.count]
+            twiceArea += a.x * b.y - b.x * a.y
+        }
+        return abs(twiceArea) / 2 >= minimumAreaSquareMeters
+    }
+
+    /// True when two non-adjacent edges cross OR touch. Even-odd ray casting reports both lobes of a
+    /// bow-tie as inside, so such a ring delivers enters for ground the polygon never covered, and a
+    /// pair of lobes joined at a single point is the same shape with the crossing degenerated to a
+    /// vertex. Touching counts, matching Android — a ring repeating a vertex non-consecutively
+    /// (a…b…a…c) survives canonicalisation on both SDKs and must then be dropped on both.
+    private static func selfIntersects(_ ring: [Point]) -> Bool {
+        let n = ring.count
+        guard n >= 4 else { return false }
+        for i in 0 ..< n {
+            let a1 = ring[i], a2 = ring[(i + 1) % n]
+            // j starts at i+2 to skip the shared-vertex neighbour; the i == 0 guard skips the
+            // closing edge, which shares a vertex with the first.
+            for j in stride(from: i + 2, to: n, by: 1) where !(i == 0 && j == n - 1) {
+                let b1 = ring[j], b2 = ring[(j + 1) % n]
+                if Self.segmentsMeet(a1, a2, b1, b2) { return true }
+            }
+        }
+        return false
+    }
+
+    /// Orientations below this are treated as collinear. Projected coordinates are metres, so the
+    /// cross product is m²: this is exact-zero with room for float error, not a tolerance.
+    private static let orientationEpsilon = 1e-9
+
+    private static func segmentsMeet(_ p1: Point, _ p2: Point, _ p3: Point, _ p4: Point) -> Bool {
+        let o1 = orientation(p1, p2, p3)
+        let o2 = orientation(p1, p2, p4)
+        let o3 = orientation(p3, p4, p1)
+        let o4 = orientation(p3, p4, p2)
+        if o1 * o2 < 0, o3 * o4 < 0 { return true }
+        if abs(o1) <= orientationEpsilon, within(p1, p3, p2) { return true }
+        if abs(o2) <= orientationEpsilon, within(p1, p4, p2) { return true }
+        if abs(o3) <= orientationEpsilon, within(p3, p1, p4) { return true }
+        if abs(o4) <= orientationEpsilon, within(p3, p2, p4) { return true }
+        return false
+    }
+
+    /// Whether `point` lies in the bounding box of the `a`–`b` segment, used only once the three are
+    /// known to be collinear.
+    private static func within(_ a: Point, _ point: Point, _ b: Point) -> Bool {
+        point.x >= min(a.x, b.x) && point.x <= max(a.x, b.x)
+            && point.y >= min(a.y, b.y) && point.y <= max(a.y, b.y)
+    }
+
+    private static func orientation(_ a: Point, _ b: Point, _ c: Point) -> Double {
+        (b.x - a.x) * (c.y - a.y) - (b.y - a.y) * (c.x - a.x)
     }
 
     func contains(_ location: LocationData) -> Bool {
--- a/Sources/LocationGeofence/Geometry/PolygonRegion.swift
+++ b/Sources/LocationGeofence/Geometry/PolygonRegion.swift
@@ -4,14 +4,13 @@ import Foundation
 /// Planar geometry for a polygon geofence, evaluated on a local equirectangular projection
 /// around the polygon's vertex centroid.
 ///
-/// Projection (shared verbatim with the Android SDK; the cross-SDK fixtures assume it):
-/// `x = R·Δlon·cos(lat₀)`, `y = R·Δlat` (radians, R = 6371000 m). At fence scale (≤ ~10 km)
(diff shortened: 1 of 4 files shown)


Sources/LocationGeofence/Geometry/PolygonRegion.swift
```diff
@@ -0,0 +1,144 @@
...
+    init?(vertices: [LocationData]) {
+        var open: [LocationData] = []
+        open.reserveCapacity(vertices.count)
+        for vertex in vertices where vertex != open.last {
+            open.append(vertex)
+        }
+        if let first = open.first, let last = open.last, open.count > 1, first == last {
+            open.removeLast()
+        }
+        guard open.count >= 3 else { return nil }
```
mahmoud-elmorabea, line 63: This only rejects rings with fewer than 3 distinct consecutive vertices, so collinear (zero-area) and self-intersecting rings build a non-nil region. Three collinear points give a fence that occupies a monitored slot and can never fire (`contains` is false everywhere, no log). A bow-tie `[a, c, b, d]` produces two disjoint "inside" lobes under even-odd.

Android's `PolygonGeometry.from` rejects collinear, self-intersecting, and concave rings (per the v1 convex-only decision), so the same workspace fence would behave differently per platform. #1240's `validatedPolygonVertices` doesn't add these checks either.

Fine as a follow-up, but worth tracking so the two SDKs converge.

Sources/LocationGeofence/Geometry/PolygonRegion.swift
```diff
@@ -0,0 +1,144 @@
...
+    /// `static` so the initializer can project the ring before `self` is fully formed, and so the
+    /// formula lives in exactly one place.
+    private static func project(
+        _ location: LocationData,
+        referenceLatitudeRadians: Double,
+        referenceLongitudeRadians: Double,
+        cosReferenceLatitude: Double
+    ) -> Point {
+        Point(
+            x: earthRadiusMeters * (location.longitude * degreesToRadians - referenceLongitudeRadians) * cosReferenceLatitude,
```
mahmoud-elmorabea, line 196: No Δlon wrap here, and the reference longitude (line 54) is an arithmetic mean of raw longitudes. A ring straddling the antimeridian projects to a ~40,000 km strip.

Example: `[(0,179.999),(0,-179.999),(0.001,-179.999),(0.001,179.999)]` → lon0 = 0, vertices land at x = ±20,014,975 m. `contains(0.0005, 179.9995)` (inside) returns false; `contains(0.0005, 0.0)` (other side of the planet) returns true. The covering check in #1240 uses geodesic distance so the ring passes validation.

Low probability but silent and inverted. One-line fix: normalize Δlon into (-180, 180], or reject such rings the way Android does. Follow-up is fine.

Sources/LocationGeofence/Model/Geofence.swift
```diff
@@ -39,6 +45,13 @@ struct Geofence: Codable, Equatable, Sendable {
         self.lastUpdated = lastUpdated
         self.geosetIds = geosetIds
         self.metadata = metadata
+        self.vertices = vertices
+    }
+
+    /// Geometry kernel for a polygon geofence; `nil` for circles. Built on demand — callers on a
+    /// hot path should hold the result rather than re-deriving it per fix.
+    var polygonRegion: PolygonRegion? {
```
mahmoud-elmorabea, line 59: This returns nil both for a circle and for a polygon whose stored ring fails `PolygonRegion.init?`, but the doc says nil means circle (and the `vertices` doc above claims validation "at the API boundary" that this PR doesn't add).

`PolygonMembershipResolver.handleTransition` on the wiring branch already does `guard let polygon = geofence.polygonRegion else { trackTransition(...) }`, so an unbuildable ring would be forwarded as a covering-circle transition. Not reachable from a live sync today, but any future tightening of `init?` makes already-cached rings fail on decode (`init(from:)` doesn't re-validate) and silently demote to circles until the next sync.

Suggest making the two nil cases distinguishable (shape enum or `isPolygon`), or documenting that nil must not be read as "circle". Follow-up ok.

Sources/LocationGeofence/Geometry/PolygonRegion.swift
```diff
@@ -0,0 +1,144 @@
+import CioInternalCommon
+import Foundation
+
+/// Planar geometry for a polygon geofence, evaluated on a local equirectangular projection
+/// around the polygon's vertex centroid.
+///
+/// Projection (shared verbatim with the Android SDK; the cross-SDK fixtures assume it):
```
mahmoud-elmorabea: Doc accuracy: Android's `PolygonGeometry` uses a per-segment projection around the query point with `cos(meanLatitude)` + `normalizeLongitude` and a raw lat/lon ray cast with a BOUNDARY relation. No Android branch has a centroid projection or the `test-vectors-geometry-latlon.json` the fixtures header references, so "shared verbatim" isn't true today.

Exact ports of both kernels agree to ~0.0005 m on the 23 fixtures (≤1 km), so the fixtures are fine. But the stated 0.5 m agreement up to ~10 km is numerically false: rotated convex polygons diverge by up to 1.21 m at lat 31° and 3.54 m at lat 60°.

Suggest rewording to state the actual relationship and dropping/restating the 10 km claim. Follow-up ok.

Sources/LocationGeofence/Geometry/PolygonRegion.swift
```diff
@@ -0,0 +1,144 @@
...
+/// `x = R·Δlon·cos(lat₀)`, `y = R·Δlat` (radians, R = 6371000 m). At fence scale (≤ ~10 km)
+/// the projection error is far below GPS accuracy, and both SDKs agree within the fixtures'
+/// 0.5 m tolerance.
+///
+/// A point exactly on the boundary counts as outside (signed distance 0 with a negative sign
```
mahmoud-elmorabea: Doc accuracy: the half-open ray cast (`(a.y > p.y) != (b.y > p.y)`, `p.x < xCross`) classifies points on the west/south edges and the SW vertex as **inside**, and only east/north edges as outside.

Against `square400` with probe points reusing the exact vertex literals: west-edge midpoint → `contains == true`, sd = +0.0; south-edge midpoint → true; SW vertex → true; east/north midpoints and the other three vertices → false. No test asserts the boundary rule.

Suggest rewording lines 12-14 to describe the half-open rule (or drop the sentence). Follow-up ok.

Sources/LocationGeofence/Geometry/PolygonRegion.swift
```diff
@@ -0,0 +1,144 @@
...
+        }
+    }
+
+    func contains(_ location: LocationData) -> Bool {
+        isInside(project(location))
+    }
+
+    /// Meters to the nearest polygon edge: **positive inside, negative outside**.
+    ///
+    /// Note this is the OPPOSITE sign to the circle path's edge distance
```
mahmoud-elmorabea: Doc accuracy: `Geofence.edgeDistanceTo` is `max(0, distanceTo(location) - radius)`, clamped to 0 inside, never negative. Only `BaselineHealDecision`'s `distanceFromCenter - radius` is signed.

A consumer following this doc to feed polygons into `GeofenceDistanceFilter.nearest` would negate `signedEdgeDistance`, ranking inside-polygons negative while inside-circles rank 0, so the `maxBusinessGeofences` cap evicts the wrong regions. Not hit by anything on the wiring branch today. The correct conversion is `max(0, -signedEdgeDistance)`. Follow-up ok.

mahmoud-elmorabea (approved):

Sources/LocationGeofence/Geometry/PolygonRegion.swift
mrehan27: Verified — the only structural guard is `open.count >= 3` after collapsing consecutive duplicates, so collinear and bow-tie rings do build. Answering it the same way as the validation thread on #1240: degenerate geometry is polygon validity, which we've now handed to the server rather than re-implementing per SDK, so the SDK checks only what building a region needs. One correction: the convex-only point is stale — Android dropped it during the parity round and the shared vectors now carry concave cases.

Sources/LocationGeofence/Geometry/PolygonRegion.swift
mrehan27: Good catch, and fixed — this is exactly what #1249 does: longitudes unwrap into ±180° of a reference vertex before anything averages them, and the query point unwraps onto the ring's line rather than its own. Your `179.999 / -179.999` case is in the tests along with a rotation-invariance property test and a shared `square400_dateline` fixture.

Sources/LocationGeofence/Model/Geofence.swift
mrehan27: Agreed — nil genuinely means both "circle" and "polygon whose stored ring failed to build", and the `vertices` doc claims a validation boundary that doesn't exist. Taking it as a follow-up on the feature branch.

Sources/LocationGeofence/Geometry/PolygonRegion.swift
mrehan27: Both right. The kernels are independent ports that agree to ~0.0005 m on the fixtures, not one shared implementation, and the 0.5 m / 10 km claim is numerically false — I'll reword to the actual relationship and drop the range claim as a follow-up.

Sources/LocationGeofence/Geometry/PolygonRegion.swift
mrehan27: Right — the rule is half-open, west/south edges and the SW vertex classify as inside, and nothing asserts it. Follow-up: reword the doc to describe the actual rule and add the boundary assertions.

Sources/LocationGeofence/Geometry/PolygonRegion.swift
mrehan27: Right — `edgeDistanceTo` clamps at 0 and is never negative, so the doc is wrong and `max(0, -signedEdgeDistance)` is the correct conversion. Follow-up, grouped with the other doc fixes.

Sources/LocationGeofence/Geometry/PolygonRegion.swift
mrehan27: Correcting my earlier reply: degeneracy is no longer handed to the server. `b0dbf924` rejects zero-area and self-intersecting rings — touches included, matching Android — via `init(validating:)` at the API boundary, so the O(n²) checks stay off the per-evaluation rebuild. `758ce7f1` adds the range check for coordinates, and that one is on the cheap initializer because the longitude unwrap walks in 360° steps and never terminates past ~3.2e18.

Sources/LocationGeofence/Geometry/PolygonRegion.swift
mrehan27: Landed as #1249 (`1c5f679a`): longitudes are unwrapped into ±180° of the first vertex before anything averages them, and the query point is unwrapped onto the ring's line. Measured on a Taveuni ring — the sign flip against the same shape at Greenwich is gone, and rotation invariance is pinned by a test with a negative control.

Sources/LocationGeofence/Model/Geofence.swift
mrehan27: Landed in `b769ade0`: the doc now says nil means EITHER a circle or a polygon whose stored ring no longer builds, and to read `vertices` for the distinction. `8e88e3d5` acts on it — the covering-circle path is taken only when `vertices == nil`, and an unbuildable ring logs undecided instead of firing an enter across the whole circle.

Sources/LocationGeofence/Geometry/PolygonRegion.swift
mrehan27: Landed in `b769ade0`: the doc no longer claims a shared implementation. It says Android ports the same contract independently, and gives the measured agreement — ~0.0005 m, diverging to ~1.2 m at lat 31°.

Sources/LocationGeofence/Geometry/PolygonRegion.swift
mrehan27: Landed in `b769ade0`: the half-open rule is documented — west and south edges and the SW vertex read inside — and `PolygonRegionTests` asserts it on each boundary rather than leaving it implied.

Sources/LocationGeofence/Geometry/PolygonRegion.swift
mrehan27: Landed in `b769ade0`: the doc now states the conversion explicitly as `max(0, -signedEdgeDistance)`, and says the circle path's value is clamped at 0 so it can never go negative.

PR #1240, opened 2026-08-31
chore(geofence): read the v2 polygon wire contract

mrehan27:
Part of [MBL-2290](https://linear.app/customerio/issue/MBL-2290/ios-add-polygon-wire-contract-and-geometry-foundation)

Part 2 of the polygon geofence work. Teaches the API boundary the v2 wire shape agreed
with backend: a `shape` discriminator, GeoJSON `geometry`, and a nested `enclosing_circle` carrying
the unpadded canonical radius. A polygon still monitors as that circle, so nothing downstream
changes behaviour yet.

### What's here

- `latitude`/`longitude`/`radius` become optional and are resolved **per shape**, so a polygon
  region carrying only `enclosing_circle` decodes without throwing. That matters more than it
  looks: the region list is a plain array, so one region that throws takes the whole response with
  it — see the compatibility note below.
- An unrecognized `shape` **drops the region**. A shape we don't understand must never be monitored
  as whatever circle fields happen to be present.
- A region whose `shape` is **absent** while polygon fields are present also drops, but under a
  different reason, and the difference decides whether the cache survives. A shape the server
  *named* that this version cannot monitor means the workspace has moved on, so a refresh returning
  only those is allowed to clear stale monitors. A *missing* discriminator is a malformed response,
  so it counts as an unreadable region and the previous fence set is kept. A blank or
  whitespace-padded `shape` routes as absent for the same reason: left as a distinct string it
  reaches the unrecognized-shape branch, which is the exempt one, and clears the cache.
- Ring rules: `Polygon` only, exactly one ring, longitude-first positions, optional GeoJSON
  elevation ignored. Extra rings are holes we don't support, and honouring just the outer ring
  would report someone standing in a hole as inside the fence — so the fence drops instead.
- The vertex cap matches the server's cap on unique vertices, counted after unclosing the ring so
  the two agree.

The `shape` discriminator let this get *smaller*: intent is now explicit, so the previous
three-way absent/malformed/ring decode is gone and a polygon failing any rule simply drops.

### Three things worth a reviewer's attention

**Coverage slack is proportional, not flat.** The server computes the enclosing circle on a sphere;
`CLLocation.distance` measures WGS84. On a 7.9 km ring that is an 11 m gap. A failed coverage check
*drops* the region, so a flat 1 m slack would silently delete a fence whose circle legitimately
covers. Now that the contract calls the radius canonical — and therefore tight — this is
load-bearing rather than defensive. The Niagara-on-the-Lake fixture pins it, using the pessimistic
spherical radius deliberately.

**Why v2 is a separate endpoint rather than v1 growing a field.** Shipped 4.7.x decodes
`latitude`/`longitude`/`radius` as required, inside a non-lossy array. A polygon region under this
shape would throw and take the *entire* response with it — a user would lose every geofence,
circles included, until they moved out of range. v1 therefore never sends polygons, and must not
proxy v2's payload.

**A partially-unreadable response applies the survivors and deregisters the rest**, where before
this PR the whole sync failed and everything was kept. Deliberate — a region that cannot be read is
a region we cannot honour — but it is the one case the all-dropped guard does not cover, since that
guard fires only when *nothing* survived. Raised by @sydneycollins-cio in review.

### Testing

402/402 `LocationGeofenceTests` pass on iOS 26. Format and lint clean. No public API surface.

New coverage: decode and validation tests for every drop rule, an explicit longitude-first ordering
assertion, and the Niagara-on-the-Lake fixture — a real concave municipal boundary whose reflex
notch sits inside the covering circle but outside the town, which is the case a convex hull gets
wrong.

Payload cost was measured against this decoder before agreeing the 500-vertex cap: worst case
(100 fences × 500 vertices) is 1.2 MB raw, 292 KB deflated, 34 ms to decode and 26 ms to validate.
A typical 20 × 12 payload is 2 KB and under a millisecond.

### Stack

1. `mbl-2290-a-polygon-geometry-kernel` — geometry kernel (#1239, merged)
2. `mbl-2290-b-polygon-wire-contract` — v2 wire contract decode ← **this PR**
6. `mbl-2290-c-polygon-antimeridian` — antimeridian projection fix (#1249)
7. `mbl-2291-f-polygon-shared-wake-circle` — shared dynamic movement circle (#1250, replaces the
   closed #1247 and #1248)

### DIFF
--- a/Sources/LocationGeofence/Api/GeofenceApiResponse.swift
+++ b/Sources/LocationGeofence/Api/GeofenceApiResponse.swift
@@ -312,10 +312,7 @@ extension GeofenceApiRegion {
         guard let region = PolygonRegion(vertices: positions),
               region.vertices.count <= GeofenceConstants.maxPolygonVertexCount
         else { return nil }
-        let coverageBound = coveringRadius + max(
-            GeofenceConstants.polygonCoverageSlackMeters,
-            coveringRadius * GeofenceConstants.polygonCoverageSlackFraction
-        )
+        let coverageBound = coveringRadius + GeofenceConstants.polygonCoverageSlackMeters
         guard region.vertices.allSatisfy({ vertex in
             coveringCenter.distance(from: CLLocation(latitude: vertex.latitude, longitude: vertex.longitude)) <= coverageBound
         }) else { return nil }
--- a/Sources/LocationGeofence/Api/GeofenceApiResponse.swift
+++ b/Sources/LocationGeofence/Api/GeofenceApiResponse.swift
@@ -44,7 +44,7 @@ private struct LenientRegion: Decodable {
 
 /// Why a region on the wire never became a monitorable fence.
 enum GeofenceRegionDropReason: String, Error {
-    case unknownShape = "unrecognized shape"
+    case unknownShape = "unrecognized or inconsistent shape"
     case unusableCircle = "invalid coordinates or radius"
     case unusablePolygon = "missing or undecodable polygon geometry"
 }
@@ -265,6 +265,11 @@ extension GeofenceApiRegion {
         let resolved: ResolvedGeometry?
         let dropReason: GeofenceRegionDropReason
         switch shape?.lowercased() {
+        case nil where geometry != nil || enclosingCircle != nil:
+            // Polygon fields with no discriminator: the payload describes something this decoder
+            // cannot name. Falling through to the flat circle fields would monitor a shape the
+            // server never described. Android drops the same combination.
+            return .failure(.unknownShape)
         case nil, "circle":
             resolved = resolvedCircle()
             dropReason = .unusableCircle
--- a/Sources/LocationGeofence/Api/GeofenceApiResponse.swift
+++ b/Sources/LocationGeofence/Api/GeofenceApiResponse.swift
@@ -323,7 +323,8 @@ extension GeofenceApiRegion {
     }
 
     /// Decodes the ring into the kernel's canonical (unclosed) vertices so the cache stores one
-    /// representation. Only what building the region needs — the server validates the geometry.
+    /// representation. Validates here rather than at every use: the degeneracy checks are O(n²) and
+    /// a region is rebuilt per wake and per evaluation, so this is the one place they can run once.
     private static func polygonVertices(_ ring: [[Double]]) -> [LocationData]? {
         var positions: [LocationData] = []
         positions.reserveCapacity(ring.count)
@@ -332,7 +333,7 @@ extension GeofenceApiRegion {
             guard position.count >= 2 else { return nil }
             positions.append(LocationData(latitude: position[1], longitude: position[0]))
         }
-        return PolygonRegion(vertices: positions)?.vertices
+        return PolygonRegion(validating: positions)?.vertices
     }
 
     private static func resolveTransitionTypes(_ raw: [String]?) -> Set<GeofenceTransition> {
--- a/Sources/LocationGeofence/Api/GeofenceApiResponse.swift
+++ b/Sources/LocationGeofence/Api/GeofenceApiResponse.swift
@@ -81,6 +81,10 @@ struct GeofenceApiRegion: Decodable {
     /// The circle the server guarantees contains the polygon — the shape actually registered at
     /// the OS as the wake trigger.
     let enclosingCircle: GeofenceApiEnclosingCircle?
+    /// Whether the payload carried a non-null `geometry` or `enclosing_circle` at all, which is not
+    /// the same as either having decoded: both use `try?`, so a malformed value becomes `nil` and
+    /// would otherwise be indistinguishable from a v1 circle.
+    var carriesPolygonFields: Bool = false
     let externalId: String?
     let transitionTypes: [String]?
     /// Wire format is milliseconds since epoch.
@@ -131,6 +135,7 @@ extension GeofenceApiRegion {
         self.radius = try container.decodeIfPresent(Double.self, forKey: .radius)
         self.geometry = try? container.decodeIfPresent(GeofenceApiGeometry.self, forKey: .geometry)
         self.enclosingCircle = try? container.decodeIfPresent(GeofenceApiEnclosingCircle.self, forKey: .enclosingCircle)
+        self.carriesPolygonFields = container.holdsValue(forKey: .geometry) || container.holdsValue(forKey: .enclosingCircle)
         self.externalId = try container.decodeIfPresent(String.self, forKey: .externalId)
         self.transitionTypes = try container.decodeIfPresent([String].self, forKey: .transitionTypes)
         self.lastUpdated = try container.decodeIfPresent(Double.self, forKey: .lastUpdated)
@@ -167,6 +172,12 @@ private extension KeyedDecodingContainer {
         guard let raw = try? decode([String: LenientMetadataValue].self, forKey: key) else { return nil }
         let filtered = raw.compactMapValues(\.value)
         return filtered.isEmpty ? nil : filtered
+    }
+
+    /// Whether the key is present with a non-null value, regardless of whether that value decodes
+    /// into the type the field expects.
+    func holdsValue(forKey key: Key) -> Bool {
+        contains(key) && ((try? decodeNil(forKey: key)) == false)
     }
 
     /// Absent or null → nil, so a not-yet-rolled-out field is treated as "no value" rather than throwing.
@@ -265,10 +276,11 @@ extension GeofenceApiRegion {
         let resolved: ResolvedGeometry?
         let dropReason: GeofenceRegionDropReason
         switch shape?.lowercased() {
-        case nil where geometry != nil || enclosingCircle != nil:
+        case nil where carriesPolygonFields:
             // Polygon fields with no discriminator: the payload describes something this decoder
             // cannot name. Falling through to the flat circle fields would monitor a shape the
-            // server never described. Android drops the same combination.
+            // server never described. Android drops the same combination. Keyed on the fields being
+            // PRESENT, not on their having decoded — a malformed one is still a claim of a polygon.
             return .failure(.unknownShape)
         case nil, "circle":
             resolved = resolvedCircle()
--- a/Sources/LocationGeofence/Api/GeofenceApiResponse.swift
+++ b/Sources/LocationGeofence/Api/GeofenceApiResponse.swift
@@ -7,7 +7,46 @@ import Foundation
 /// gradually; per-field fallbacks live in `toDomain`.
 struct GeofenceApiResponse: Decodable {
     let config: GeofenceApiConfig?
+    /// How many regions the payload carried, including any that failed to decode. `geofences`
+    /// alone cannot tell "the server sent none" from "none of them survived".
+    let receivedRegionCount: Int
     let geofences: [GeofenceApiRegion]
+
+    private enum CodingKeys: String, CodingKey {
+        case config, geofences
+    }
+
+    /// One malformed region must not cost the whole response, so regions are decoded leniently and
+    /// the bad ones skipped.
+    init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        self.config = try container.decodeIfPresent(GeofenceApiConfig.self, forKey: .config)
+        let lenient = try container.decodeIfPresent([LenientRegion].self, forKey: .geofences) ?? []
+        self.receivedRegionCount = lenient.count
+        self.geofences = lenient.compactMap(\.region)
+    }
+
+    init(config: GeofenceApiConfig?, geofences: [GeofenceApiRegion]) {
+        self.config = config
+        self.geofences = geofences
+        self.receivedRegionCount = geofences.count
+    }
+}
+
+/// Decodes a region without ever throwing, so one bad element cannot fail the array around it.
+private struct LenientRegion: Decodable {
+    let region: GeofenceApiRegion?
+
+    init(from decoder: Decoder) throws {
+        self.region = try? GeofenceApiRegion(from: decoder)
+    }
+}
+
+/// Why a region on the wire never became a monitorable fence.
+enum GeofenceRegionDropReason: String, Error {
+    case unknownShape = "unrecognized shape"
+    case unusableCircle = "invalid coordinates or radius"
+    case unusablePolygon = "missing or undecodable polygon geometry"
 }
 
 struct GeofenceApiConfig: Decodable {
@@ -69,9 +108,7 @@ struct GeofenceApiGeometry: Decodable, Equatable {
 struct GeofenceApiEnclosingCircle: Decodable, Equatable {
     let latitude: Double
     let longitude: Double
-    /// Unpadded canonical radius: the smallest circle the server will claim contains the polygon.
-    /// Any platform margin is added when registering with the OS, never here — the coverage check
-    /// has to run against the value the server actually guaranteed.
+    /// The smallest circle the server claims contains the polygon, registered with the OS as-is.
     let baseRadiusM: Double
 }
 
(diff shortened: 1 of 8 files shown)


Sources/LocationGeofence/Api/GeofenceApiResponse.swift
```diff
@@ -179,29 +215,110 @@ private extension Comparable {
...
+    ///   GeoJSON spec allows and we ignore;
+    /// - every position is a valid coordinate;
+    /// - the kernel accepts the ring (≥3 distinct vertices after unclosing the closed ring);
+    /// - the distinct vertex count is within `maxPolygonVertexCount`, counted after unclosing so it
+    ///   matches the server's cap on *unique* vertices;
+    /// - the covering guarantee holds: every vertex within the enclosing radius (+ slack) of its
+    ///   center. Vertices inside a convex circle imply the whole polygon is inside, so this fully
+    ///   defends "enclosing-circle exit ⇒ polygon exit" against a server that miscomputed it.
+    /// Returns the kernel's canonical (unclosed) vertices so the cache stores one representation.
+    private static func validatedPolygonVertices(
```
mahmoud-elmorabea: This re-validates polygons on the device (coordinate validity, vertex cap, coverage check). The backend already owns these rules. Re-implementing them here means iOS, Android, and the server can disagree on the same payload, and any rule change needs an SDK release to take effect. Suggest the SDK only checks what it needs to decode and build the region, and trusts the backend for the rest.

Sources/LocationGeofence/Api/GeofenceApiResponse.swift
```diff
@@ -28,9 +28,20 @@ struct GeofenceApiPlatformConfig: Decodable {
...
-    let radius: Double
+    /// `"circle"` or `"polygon"`; absent means circle, which is what v1 payloads look like. An
+    /// unrecognized value drops the region — a shape we don't understand must never quietly fall
+    /// back to whatever circle fields happen to be present.
+    let shape: String?
+    /// Circle geometry, present on a circle region. A polygon region carries `enclosingCircle`
+    /// instead, so these are optional and their absence is resolved per shape rather than thrown.
+    let latitude: Double?
+    let longitude: Double?
+    let radius: Double?
```
mahmoud-elmorabea, line 83: Before this PR a bad response failed the decode and kept the cache and OS registrations. Now it produces an empty list, which wipes the cache and deregisters everything in the refresh coordinator. Worth a guard: if the server sent N regions and none survived, treat it as a fetch failure. An actually empty response should still work.

Sources/LocationGeofence/Api/GeofenceApiResponse.swift
```diff
@@ -55,9 +88,12 @@ extension GeofenceApiRegion {
...
         self.id = try container.decodeStringOrInt(forKey: .id)
         self.name = try container.decodeIfPresent(String.self, forKey: .name)
-        self.latitude = try container.decode(Double.self, forKey: .latitude)
-        self.longitude = try container.decode(Double.self, forKey: .longitude)
-        self.radius = try container.decode(Double.self, forKey: .radius)
+        self.shape = try container.decodeIfPresent(String.self, forKey: .shape)
+        self.latitude = try container.decodeIfPresent(Double.self, forKey: .latitude)
+        self.longitude = try container.decodeIfPresent(Double.self, forKey: .longitude)
+        self.radius = try container.decodeIfPresent(Double.self, forKey: .radius)
+        self.geometry = try? container.decodeIfPresent(GeofenceApiGeometry.self, forKey: .geometry)
```
mahmoud-elmorabea, line 141: The `try?` only covers geometry and enclosing circle. A wrong-typed shape, radius, or timestamp on one region still fails the whole response, which the doc comment above says shouldn't happen. `LenientMetadataValue` already does this right; the region array could use the same approach.

Sources/LocationGeofence/Api/GeofenceApiResponse.swift
```diff
@@ -43,9 +54,31 @@ struct GeofenceApiRegion: Decodable {
...
+    /// the contract guarantees exactly one; further rings would be holes, which we don't support.
+    let coordinates: [[[Double]]]
+}
+
+/// The server-computed circle containing a polygon.
+struct GeofenceApiEnclosingCircle: Decodable, Equatable {
+    let latitude: Double
+    let longitude: Double
+    /// Unpadded canonical radius: the smallest circle the server will claim contains the polygon.
+    /// Any platform margin is added when registering with the OS, never here — the coverage check
```
mahmoud-elmorabea: Nothing adds a margin at OS registration. Both monitors clamp down to the OS max, so a large covering circle can end up smaller than its polygon. Either fix the comment or pull the downstream drop-not-clamp guard into this PR.

Sources/LocationGeofence/Api/GeofenceApiResponse.swift
```diff
@@ -179,29 +215,110 @@ private extension Comparable {
...
+        let resolved: ResolvedGeometry?
+        switch shape?.lowercased() {
+        case nil, "circle":
+            resolved = resolvedCircle()
+        case "polygon":
+            resolved = resolvedPolygon()
+        default:
+            resolved = nil
+        }
+        guard let resolved else { return nil }
```
mahmoud-elmorabea: All drop paths end up logged as "invalid coordinates or radius" (`geofenceInvalidRegionDropped`). Unknown shape and decode failures should say so.

Sources/LocationGeofence/Api/GeofenceApiResponse.swift
mrehan27: Agreed and done — dropped the per-vertex coordinate check, the vertex cap and the coverage check; the SDK now only checks what building a region needs (a position with ≥2 elements, a ring with ≥3 distinct vertices). Android is mirroring it and the shared payload vectors are re-baselined: 4 cases flip to accepted, and the 3 tight-fit cases are deleted since they only discriminated through the coverage check.

Sources/LocationGeofence/Api/GeofenceApiResponse.swift
mrehan27: Confirmed from source and fixed — the response now carries `receivedRegionCount`, and refresh returns `.fetchFailed(.decoding)` when the server sent regions and none survived, so the cache and OS registrations stay put. A genuinely empty list still applies normally.

Sources/LocationGeofence/Api/GeofenceApiResponse.swift
mrehan27: Fixed — regions decode through a wrapper that never throws, so a wrong-typed radius or timestamp costs that region only. Worth noting for the Android port: a plain `try?` per element in an unkeyed container can spin forever, because a failed decode doesn't reliably advance the index.

Sources/LocationGeofence/Api/GeofenceApiResponse.swift
mrehan27: Fixed the comment — nothing adds a margin, so the claim was simply wrong. Left the drop-not-clamp guard in #1246 where it already lives rather than pulling it down into this PR.

Sources/LocationGeofence/Api/GeofenceApiResponse.swift
mrehan27: Fixed — added `GeofenceRegionDropReason` (`unknownShape` / `unusableCircle` / `unusablePolygon`) and the drop log now names the actual cause.

Sources/LocationGeofence/Api/GeofenceApiResponse.swift
```diff
@@ -179,29 +254,87 @@ private extension Comparable {
...
-    func toDomain() -> Geofence? {
-        guard radius > 0,
-              CLLocationCoordinate2DIsValid(CLLocationCoordinate2D(latitude: latitude, longitude: longitude))
-        else { return nil }
-        return Geofence(
+    func toDomain() -> Result<Geofence, GeofenceRegionDropReason> {
+        let resolved: ResolvedGeometry?
+        let dropReason: GeofenceRegionDropReason
+        switch shape?.lowercased() {
+        case nil, "circle":
```
Shahroz16, line 293: Could we limit the missing-shape fallback to true v1 circle records? A payload that includes geometry or enclosingCircle but omits shape still takes this circle branch whenever flat coordinates are present, so iOS can monitor and report a circle the server did not describe. Android drops this exact combination as “shape discriminator is missing or inconsistent.” Checking both shape-only fields are nil before resolvedCircle() would keep the SDKs aligned, with one malformed-payload test.

Shahroz16 (commented):
One wire-contract edge case still needs tightening.

Sources/LocationGeofence/Api/GeofenceApiResponse.swift
Shahroz16: One reason coordinate validity still belongs on-device is that the current antimeridian unwrap subtracts or adds 360 in a loop until the longitude is in range. A finite value such as 1e12 can take billions of iterations during response parsing and again when the cached polygon is rebuilt. That is a construction and runtime-safety precondition, rather than duplicating server polygon policy. Could we retain CLLocationCoordinate2DIsValid for ring positions, and preferably use arithmetic wrapping, with a large-longitude regression test?

Sources/LocationGeofence/Api/GeofenceApiResponse.swift
mrehan27: Done — a region carrying `geometry` or `enclosing_circle` without `shape` now drops as `.unknownShape`. Keyed on the field being present rather than on it decoding, since both use `try?` and a malformed one would otherwise read as a plain v1 circle.

Sources/LocationGeofence/Api/GeofenceApiResponse.swift
mrehan27: Follow-up: the asymmetry had a second effect you didn't have to spell out — a malformed `geometry` nils out and the region takes the v1 circle branch. Presence is now tracked separately from decode success.

Sources/LocationGeofence/Api/GeofenceApiResponse.swift
mrehan27: Both landed in #1239, which is now merged — `init?(vertices:)` (the cheap initializer every cache rebuild goes through, not just decode) rejects any ring position failing `CLLocationCoordinate2DIsValid`, and `init_givenOutOfRangeCoordinate_expectRejectedWithoutWalking` covers 1e12 and 1e300. I kept the loop rather than switching to arithmetic wrapping, but `unwrapLongitude` now guards `isFinite && abs(longitude) <= 360` itself, so it terminates in at most two iterations whatever the caller does — same bound, and termination stays a property of the function. Happy to switch to `remainder` if you'd still prefer that form.

Sources/LocationGeofence/Api/GeofenceApiResponse.swift
Shahroz16: Could we distinguish a missing shape discriminator from an explicitly unsupported shape here? `shape == nil` with `geometry: {}` now returns `.unknownShape`, which `readableRegions` exempts from the unreadable count. I reproduced `{"geofences":[{"id":"broken","latitude":1,"longitude":2,"radius":100,"geometry":{}}]}` returning refresh success and clearing the existing cache. This is a malformed payload, not a workspace moving to a new shape. A separate unreadable drop reason would preserve the cache while still allowing an explicit unsupported shape to clear stale monitors.

Shahroz16 (commented):
The previous decode and circle-fallback fixes are in place. I found one remaining cache-preservation gap and added the failing payload to the existing thread. Holding approval here and on the dependent #1244/#1245/#1246/#1249/#1250 stack until that is addressed.

Sources/LocationGeofence/Api/GeofenceApiResponse.swift
mrehan27: Good catch, and confirmed from your payload — fixed in `77bec678`. The two cases now have separate drop reasons: `unknownShape` stays for a shape the server NAMED that this version does not implement (workspace moved on, stale monitors should go, still exempt from the all-dropped guard), and a new `undescribedShape` covers polygon fields with no discriminator, which counts as unreadable and preserves the cache. `refresh_givenPolygonFieldsWithoutShape_expectFetchFailureAndCacheKept` uses your exact JSON; reverting only the reason split makes it fail with the cache cleared, so the guard is doing the work rather than something incidental.

Shahroz16 (approved):
The original cache-preservation regression now passes, and the decode and circle-fallback fixes still hold. My earlier concerns in this PR are addressed.

Sources/LocationGeofence/Api/GeofenceApiResponse.swift
```diff
@@ -28,9 +72,24 @@ struct GeofenceApiPlatformConfig: Decodable {
...
+    let radius: Double?
+    /// GeoJSON boundary of a polygon region.
+    let geometry: GeofenceApiGeometry?
+    /// The circle the server guarantees contains the polygon — the shape actually registered at
+    /// the OS as the wake trigger.
+    let enclosingCircle: GeofenceApiEnclosingCircle?
+    /// Whether the payload carried a non-null `geometry` or `enclosing_circle` at all, which is not
+    /// the same as either having decoded: both use `try?`, so a malformed value becomes `nil` and
+    /// would otherwise be indistinguishable from a v1 circle.
+    var carriesPolygonFields: Bool = false
```
sydneycollins-cio: Nit: carriesPolygonFields defaults to false in the memberwise init, so a hand-built region with geometry claims no polygon fields. Deriving it from geometry presence would stop drift.

Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator+Refresh.swift
```diff
@@ -106,6 +110,33 @@ extension GeofenceSyncCoordinatorImpl {
...
     }
 
+    /// Decodes the response's regions, or fails when the payload turns out to be unreadable rather
+    /// than empty. A response whose regions we could not READ is a broken payload, not "this user
+    /// has no geofences": reading it as the latter wipes the cache and deregisters everything. A
+    /// workspace that has moved entirely to a shape this SDK does not support is the opposite case
+    /// — the response read fine and nothing in it is monitorable, so failing forever would freeze
+    /// stale fences behind a refresh that can never succeed. An actually empty list applies
+    /// normally.
+    private func readableRegions(from response: GeofenceApiResponse) -> Result<[Geofence], GeofenceSyncError> {
```
sydneycollins-cio, line 120: Non-blocking: a partially-unreadable response now applies survivors and deregisters the failed fences, where pre-PR the whole sync failed and kept everything. Intentional, but it's the one case the all-dropped guard doesn't cover. Worth a line in the PR description.

Sources/LocationGeofence/Logger+Geofence.swift
```diff
@@ -5,9 +5,17 @@ import Foundation
 private let geofenceTag = "Geofence"
 
 extension Logger {
-    func geofenceInvalidRegionDropped(_ identifier: String) {
+    func geofenceInvalidRegionDropped(_ identifier: String, reason: GeofenceRegionDropReason) {
         error(
-            "Geofence '\(identifier)' dropped — invalid coordinates or radius, not registerable with the OS",
+            "Geofence '\(identifier)' dropped — \(reason.rawValue), not registerable with the OS",
+            geofenceTag,
```
sydneycollins-cio, line 11: Nit: the "not registerable with the OS" suffix mislabels undescribedShape, which is a payload problem, not an OS limit. Drop the suffix.

sydneycollins-cio (approved):
LGTM.

One doc note: the description still documents the coverage check and vertex cap, both removed when policy moved server-side, and frames a separate v2 endpoint this diff doesn't touch. Worth updating so the merged record matches the code.

Sources/LocationGeofence/Api/GeofenceApiResponse.swift
mrehan27: Fixed in `a293a1f4`, though not by deriving it — `geometry != nil` is decode *success*, and both polygon fields decode with `try?`, so a malformed geometry would read as a v1 circle again, which is the case this flag was added for. Dropped the default instead so the memberwise init has to state it; both polygon fixtures were indeed claiming false.

Sources/LocationGeofence/Logger+Geofence.swift
mrehan27: Dropped in `93f49c36`. It was wrong for `unknownShape` too — its own doc says the workspace moved on — so 2 of the 4 reasons rather than 1.

Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator+Refresh.swift
mrehan27: Added to the PR description. Deliberate: the all-dropped guard fires only when *nothing* survived, so this partial case sits outside it by design.

PR #1244, opened 2026-08-31
chore(geofence): add the polygon membership layer

mrehan27:

Part 3 of the polygon geofence work. The belief layer only — what the SDK thinks the device's
membership is for each polygon, and how that belief turns into a crossing. Nothing writes to it
yet; the resolver arrives in the next PR.

### What's here

**`PolygonMembership` / `PolygonMembershipDecision`** — the verdict and the accuracy gate that
produces it. A fix decides membership only when the device is farther from the boundary than its
own accuracy, floored at 20 m. Undecidable leaves the stored belief untouched rather than guessing,
which is what "no event without gated geometric confirmation" means in code.

**`GeofenceStorage+PolygonMembership`** — the persisted belief, with the first-decision rule that
matters: a polygon with no record is *undecided*, not outside, so the first decisive fix placing
the device inside is a genuine enter. Establishing "outside" for the first time is not a crossing.

Belief is pruned when a polygon leaves the registered set, on the same rule the monitor records
follow: it is no longer being evaluated, so a retained belief would go stale and suppress the enter
owed when the device returns. Clearing a monitor region record drops the belief with it, for the
same reason — kept across an unmonitored gap, an inside belief swallows the next real enter as
no-change if the device left while nothing was watching. A forced reseed does the same: the
deferred clear is skipped when a re-registration overtakes it, so the two halves of what an
identifier believes reseed together.

An evaluation that *confirms* the stored belief refreshes its evidence stamp, not only one that
changes it. Re-proving a belief is newer evidence for it, and holding the stamp at the last change
let a later evaluation carrying older opposite evidence clear the ordering guard and deliver a
crossing the newer fix had just disproved.

A belief is only ever *created* for a polygon in the registered set, so an evaluation that raced a
prune cannot resurrect a fence the OS is no longer watching and deliver an enter for it.

The sign convention. `PolygonRegion.signedEdgeDistance` is positive inside, the circle path's edge
distance is negative inside. `SignConventionTests` pins the relationship in terms of physical
position rather than sign, because during the spike a consumer was written against the wrong
assumption and produced inverted verdicts that neither side's unit tests could see.

### Testing

429/429 `LocationGeofenceTests` pass on iOS 26. Format and lint clean, no public API surface.

Unit tests for the gate, the storage transitions, the sign convention, the unmonitored-clearing
rule, and the refusal to create a belief for an unregistered id.

### Stack

1. `mbl-2290-a-polygon-geometry-kernel` — geometry kernel (#1239, merged)
2. `mbl-2290-b-polygon-wire-contract` — v2 wire contract decode (#1240, merged)
3. `mbl-2291-a-polygon-membership-layer` — membership belief + storage ← **this PR**
6. `mbl-2290-c-polygon-antimeridian` — antimeridian projection fix (#1249)
7. `mbl-2291-f-polygon-shared-wake-circle` — shared dynamic movement circle (#1250)

### DIFF
--- a/Sources/LocationGeofence/Polygon/PolygonMembershipDecision.swift
+++ b/Sources/LocationGeofence/Polygon/PolygonMembershipDecision.swift
@@ -1,0 +1,29 @@
+import Foundation
+
+/// Decides what a fix establishes about a device's membership in a polygon.
+///
+/// The polygon counterpart of `BaselineHealDecision`, and deliberately the same guards applied to a
+/// polygon's signed edge distance instead of a circle's: the fix must be recent, its accuracy
+/// usable, and its distance from the boundary must exceed the ambiguity margin. This is the single
+/// place a polygon verdict is formed, so "no event without gated geometric confirmation" is
+/// enforced in one function rather than at each call site.
+enum PolygonMembershipDecision {
+    /// The membership the fix establishes, or `nil` when it cannot decide — a stale fix, an
+    /// unusable accuracy, or a boundary inside the ambiguity band. Never returns `.unknown`:
+    /// that is what the caller records in response to `nil`.
+    ///
+    /// - Parameter signedEdgeDistance: metres to the polygon boundary in `PolygonRegion`'s
+    ///   convention — **positive inside**, which is the opposite of the circle path's edge
+    ///   distance. Converting here rather than at call sites keeps the mismatch in one place.
+    static func resolvedMembership(
+        signedEdgeDistance: Double,
+        horizontalAccuracy: Double,
+        fixAge: TimeInterval
+    ) -> PolygonMembership? {
+        guard fixAge >= 0, fixAge <= GeofenceConstants.movementFixMaxAge else { return nil }
+        guard horizontalAccuracy > 0 else { return nil }
+        let margin = max(horizontalAccuracy, GeofenceConstants.baselineHealMinEdgeMargin)
+        guard abs(signedEdgeDistance) > margin else { return nil }
+        return signedEdgeDistance > 0 ? .inside : .outside
+    }
+}
--- a/Sources/LocationGeofence/Polygon/PolygonMembershipDecision.swift
+++ b/Sources/LocationGeofence/Polygon/PolygonMembershipDecision.swift
@@ -9,12 +9,11 @@ import Foundation
 /// enforced in one function rather than at each call site.
 enum PolygonMembershipDecision {
     /// The membership the fix establishes, or `nil` when it cannot decide — a stale fix, an
-    /// unusable accuracy, or a boundary inside the ambiguity band. Never returns `.unknown`:
-    /// that is what the caller records in response to `nil`.
+    /// unusable accuracy, or a boundary inside the ambiguity band.
     ///
     /// - Parameter signedEdgeDistance: metres to the polygon boundary in `PolygonRegion`'s
     ///   convention — **positive inside**, which is the opposite of the circle path's edge
-    ///   distance. Converting here rather than at call sites keeps the mismatch in one place.
+    ///   distance. The sign test below is written for this convention; nothing is converted.
     static func resolvedMembership(
         signedEdgeDistance: Double,
         horizontalAccuracy: Double,
--- a/Sources/LocationGeofence/Storage/GeofenceStorage+PolygonMembership.swift
+++ b/Sources/LocationGeofence/Storage/GeofenceStorage+PolygonMembership.swift
@@ -56,6 +56,13 @@ extension GeofenceStorage {
 }
 
 extension GeofenceState {
+    /// Drops one identifier's belief, for a reseed that has already decided the stored circle
+    /// baseline can no longer be trusted: kept, a belief of "inside" formed before an unmonitored
+    /// gap swallows the next real enter as no-change.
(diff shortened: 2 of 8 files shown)


Sources/LocationGeofence/Storage/GeofenceStorage.swift
```diff
@@ -204,6 +204,7 @@ actor GeofenceStorage {
         state.movementTriggerCenter = nil
         state.monitoredGeofenceIds = nil
         state.monitorRegionRecords = nil
+        state.polygonMembership = nil
```
mahmoud-elmorabea, line 212: Polygon belief is cleared here but not in `clearMonitorRegionRecord` / `forceReseed`. When CLMonitor reports the covering circle unmonitored, the circle baseline is reseeded but the polygon record stays. If the user left the polygon while unmonitored, the next real enter is suppressed as no-change. Suggest clearing the polygon record on that path too.

Also worth adding polygon membership to the doc list above since it's now user-scoped.

Sources/LocationGeofence/Storage/GeofenceStorage+PolygonMembership.swift
```diff
@@ -0,0 +1,60 @@
...
+        onlyIfBeliefPredates evidenceTimestamp: Date? = nil,
+        now: Date = Date()
+    ) -> PolygonMembershipOutcome {
+        var state = loadFromDisk() ?? GeofenceState()
+        var records = state.polygonMembership ?? [:]
+        let existing = records[identifier]
+        if let evidenceTimestamp, let existing, existing.lastChangedAt > evidenceTimestamp {
+            return .suppressedNewerDecision
+        }
+        guard let existing else {
```
mahmoud-elmorabea, line 50: When there is no record, the evidence-timestamp check above is skipped (`existing` is bound first) and the monitored set isn't consulted. An evaluation in flight for a polygon that was just pruned recreates a record for an unregistered polygon and can deliver an enter. Suggest refusing to create a record for an id not in `monitoredGeofenceIds`.

Sources/LocationGeofence/Polygon/PolygonMembershipDecision.swift
```diff
@@ -0,0 +1,29 @@
...
+/// Decides what a fix establishes about a device's membership in a polygon.
+///
+/// The polygon counterpart of `BaselineHealDecision`, and deliberately the same guards applied to a
+/// polygon's signed edge distance instead of a circle's: the fix must be recent, its accuracy
+/// usable, and its distance from the boundary must exceed the ambiguity margin. This is the single
+/// place a polygon verdict is formed, so "no event without gated geometric confirmation" is
+/// enforced in one function rather than at each call site.
+enum PolygonMembershipDecision {
+    /// The membership the fix establishes, or `nil` when it cannot decide — a stale fix, an
+    /// unusable accuracy, or a boundary inside the ambiguity band. Never returns `.unknown`:
```
mahmoud-elmorabea: `PolygonMembership` has no `.unknown` case, and its header says a nil result must leave the existing belief untouched. This sentence says the opposite. Suggest deleting it.

Sources/LocationGeofence/Polygon/PolygonMembershipDecision.swift
```diff
@@ -0,0 +1,29 @@
...
+    /// that is what the caller records in response to `nil`.
+    ///
+    /// - Parameter signedEdgeDistance: metres to the polygon boundary in `PolygonRegion`'s
+    ///   convention — **positive inside**, which is the opposite of the circle path's edge
+    ///   distance. Converting here rather than at call sites keeps the mismatch in one place.
```
mahmoud-elmorabea: No conversion happens here; the positive-inside sign is used as-is. A reader looking for the sign flip could add one and invert every verdict. Suggest rewording to say the sign test below is written for this convention.

mahmoud-elmorabea (approved):

Sources/LocationGeofence/Storage/GeofenceStorage.swift
mrehan27: Both done. `forceReseed` now drops the belief alongside the circle baseline, and the sign-out doc names polygon belief. Tests cover the reseed dropping it and a routine re-registration keeping it.

Sources/LocationGeofence/Storage/GeofenceStorage+PolygonMembership.swift
mrehan27: Fixed — the create path refuses an id outside `monitoredGeofenceIds` and returns `.suppressedUnmonitored`.

Sources/LocationGeofence/Polygon/PolygonMembershipDecision.swift
mrehan27: Deleted.

Sources/LocationGeofence/Polygon/PolygonMembershipDecision.swift
mrehan27: Reworded — it now says the sign test below is written for this convention and nothing is converted.

Shahroz16 (approved):
The membership storage and pruning changes look good, and the earlier cache blocker in the base is fixed. The focused tests pass; no remaining concern in this PR.

mahmoud-elmorabea (approved):

PR #1245, opened 2026-08-31
chore(geofence): decide polygon membership from covering-circle events

mrehan27:

Part 4 of the polygon geofence work. The resolver that turns OS covering-circle events into
customer-facing polygon transitions. Not wired to the monitor yet — that is the next PR.

### What's here

The OS monitors a polygon's covering circle and knows nothing about the polygon, so membership is a
second fact established on device. Two invariants define correctness, and neither may be relaxed:

- an ENTER requires a gated fix placing the device inside the polygon;
- an EXIT requires either a covering-circle exit — polygon ⊆ circle, so leaving one leaves the
  other — or a gated fix placing the device outside.

Everything else is silent. Circle geofences pass straight through to the tracker, so the
contradiction gate, baseline heal and dedup baseline that serve them stay unaware polygons exist.

Foreground evaluation covers the one case no OS event reaches: a device already standing inside a
polygon when monitoring begins has crossed nothing, and standing still produces no movement pass
either.

A geofence missing from the cache is **forwarded, not dropped** — a sync can race an event, or the
OS can still hold a condition the cache dropped. Treating it as a circle is the behaviour that
predates polygons, and losing a real crossing is worse than one shaped like its covering circle.

The resolver owns its own `MovementFixResolver` rather than reading the monitor's, for the same
reason each monitor has one: the transition handler carries only coordinates while the gate needs
accuracy and age, and on a wrapper cold wake `CustomerIO.initialize` has not run.

It requests **ten-metre** accuracy, not the hundred the circle path uses. A verdict needs the device
farther from the boundary than the fix is accurate, so a hundred-metre fix could never decide
anything for a polygon near the minimum monitored size — those fences would sit registered and never
fire in either direction.

The cached fix is chosen by **age, not by source**. CoreLocation's own cache advances between
passes, driven by other clients in the process, so preferring this resolver's last delivered fix
outright requested a fix already in hand and, when that request failed, fell back to the older of
the two — the newest-of-the-two rule the monitors' `bestKnownFix` already applied, invalid
coordinates dropped the same way.

A cold process has delivered no fix of its own, so a failed request used to fall back to nothing and
lose the crossing permanently — the monitor has already advanced its dedup baseline. CoreLocation's
cached fix is the fallback now, still subject to the age gate.

A **stored ring that no longer builds is not a circle**: forwarding it would fire a customer enter
anywhere inside the covering circle, annulus included. Only `vertices == nil` takes the circle path.
That guard sits on the enter branch only — an exit needs no ring, so refusing one would strand the
belief at inside and suppress every later exit and re-enter.
And the covering exit carries the event date, so a replayed or synthesized exit cannot overwrite a
newer enter and leave the device believed outside while it sits inside. The date is required, not
optional: it is the only thing ordering that write, and a caller allowed to omit it writes unordered.

**A covering exit is refused when the circle crossed is not the one the fence still has.** The
certainty is polygon ⊆ *its own* covering circle. A refresh replaces ring and circle together, so
leaving the old circle says nothing about the new ring — the device can be standing inside it. This
does not self-heal: the spurious exit stamps the belief with the OS event's date, and the repairing
fix is then refused as an older decision, so the belief sticks outside. Refusing the exit instead
leaves the belief for a fix to decide against the current ring; refusing rather than clearing is
what stops a device that never left from being re-entered. A nil circle is accepted — a cold wake
cannot say which circle an event was for, and an event that cannot be shown stale is treated as
current.

**That comparison happens inside the write, not before the hop.** The resolver matched the event's
circle against the fence it had already loaded, then hopped to the storage actor — and a refresh
landing in that gap left `outside` recorded for a device inside the replacement polygon. The circle
now travels into `recordPolygonMembership` and is compared in the same `loadFromDisk` that writes,
exactly as the evaluated ring is. The pre-hop check was deleted rather than kept alongside it: two
comparisons against geometry read at different instants is the defect, not the cure. Raised by
@Shahroz16 in review.

The comparison uses the radius **as registered**. Both monitors register
`min(radius, maximumRegionMonitoringDistance)`, so comparing against the fence's raw radius would
never match an over-cap fence; `MonitoredCircle` therefore carries the cap it was clamped against
and reuses `matchesRegistered`'s clamp and tolerances rather than assuming the float round trip
through CoreLocation is exact. The clamp half is belt-and-braces at the shipped state — #1246 drops
over-cap polygons before registration, so they never reach this comparison — but the comparison
should not depend on a guard living in a later PR. The tolerance half is live either way.

**A verdict is decided from the ring current when it is decided, never the pass's copy.** Resolving
a fix suspends, and a refresh can replace a fence under the same id while it does — the pass would
then deliver a crossing for a shape the workspace has moved off, with no user switch involved.
`evaluate` takes an id and reads the fence itself afterwards, so a stale ring is not in scope to be
used by mistake. Registration is read from the same load: sampling one before the await and the
other after is how the two come to disagree. It costs one state decode per polygon, which buys
verdicts at most one hop stale instead of a whole location request stale.

That closed the window across the location request but left a second one: the verdict is computed
on the main actor and written by a separate hop onto the storage actor, and a replacement can commit
in that gap. The evaluated ring now travels into `recordPolygonMembership` and is compared inside
the same `loadFromDisk` that writes — the one place read and write cannot be separated. A third
check in the resolver would only be a third thing that can go stale. It compares the ring itself
rather than `lastUpdated`, which the server owns and may not bump on a replacement. Registration
needs no equivalent: `recordRegistration` prunes the belief in the same actor call that narrows the
set.

**The user is re-checked after the fix resolves, and again immediately before the emit.** The
tracker stamps whoever is current when it is entered, and the membership write is an await of its
own, so a check placed only at the verdict leaves the emit unguarded. The write is deliberately left
unguarded: a belief states geometry, true whoever is signed in, while an emit is an attribution, and
attribution is what a switch invalidates. The foreground observer has no caller to take an expected
user from, so it samples one when it fires.

The foreground pass resolves **one** fix above the loop. Resolving per polygon meant an empty cache
issued a fresh timed request for every registered polygon, holding the main actor for as long as
that took and still deciding nothing. Foregrounds also arrive in bursts, so a pass already running
wins and the skip is logged.

### Testing

468/468 `LocationGeofenceTests` pass on iOS 26. Format and lint clean, no public API surface.

Unit tests over the enter/exit/undecidable/no-fix matrix, foreground evaluation, the transition-type
filter, the unbuildable-ring path, and an exit older than the stored belief.

For the two hazards above, each test holds the pass inside its fix request so the event lands
strictly between the read and the verdict, and each has a control that must still deliver:
a fence replaced under the same id, a fence dropped from the catalog, a fence unregistered, a user
switching at fix delivery, and a user switching after the write but before the emit. Every one was
negative-controlled — reverting the guard makes it fail, and the failure is the wrong event being
delivered, not just an assertion.

The unregistered-mid-fix case pins the outcome rather than the mechanism, and says so: storage also
refuses that write because `recordRegistration` prunes the belief with the set. The resolver's own
check is what stops it depending on a storage rule that covers only the create path.

Four storage tests cover the write-time geometry window: a verdict whose ring was replaced mid-write
is refused, the unchanged-ring control still writes, a covering exit still lands after the exact
replacement that refuses a fix-derived verdict, and a fence dropped from the catalog refuses.

The exit-circle tests build the event circle **from the OS side**, not from the fence. The two
earlier ones built it from `geofence.monitoredCircle` — comparing the expression against itself —
which is why neither the replaced-circle case nor the clamped-radius case was visible. A further
case pins an over-cap fence's exit still being delivered.

Four storage tests cover the write-time circle window: a circle replaced since the event was raised
is refused, the unchanged circle still delivers, a fence gone from the cache is refused, and an
over-cap fence's clamped circle still delivers. **Negative-controlled** — removing the guard fails
those *and* the resolver-level replaced-circle case, which is what proves the check is carried by
the atomic write rather than by a leftover.

### Stack

4. #1245 `mbl-2291-b` membership resolver ← **this PR**

### DIFF
--- a/Sources/LocationGeofence/Logger+Geofence.swift
+++ b/Sources/LocationGeofence/Logger+Geofence.swift
@@ -154,6 +154,22 @@ extension Logger {
         info("Synthesized \(transition.rawValue) for region \(identifier): fresh fix contradicts stored baseline (OS never delivered the crossing)", geofenceTag)
     }
 
+    /// Positive record of an OS-delivered business transition. Without it a native delivery is
+    /// only identifiable by the ABSENCE of a synthesized line, which makes the OS promotion rate
+    /// inferred rather than measured.
+    func geofenceOsTransitionReceived(identifier: String, transition: GeofenceTransition) {
+        debug("OS delivered \(transition.rawValue) for region \(identifier)", geofenceTag)
+    }
+
+    func geofencePolygonTransition(identifier: String, transition: GeofenceTransition, confirmedByFix: Bool) {
+        let basis = confirmedByFix ? "a gated fix" : "the covering-circle exit"
+        info("Polygon \(transition.rawValue) for region \(identifier), confirmed by \(basis)", geofenceTag)
+    }
+
+    func geofencePolygonUndecided(identifier: String, reason: String) {
+        debug("Polygon membership undecided for region \(identifier): \(reason)", geofenceTag)
+    }
+
     func geofenceEventRefusedByContradiction(identifier: String, transition: GeofenceTransition, distanceFromCenter: Double, radius: Double, accuracy: Double) {
         info("Refused OS \(transition.rawValue) for region \(identifier): a fresh fix contradicts it (distance \(Int(distanceFromCenter)) m, radius \(Int(radius)) m, accuracy \(Int(accuracy)) m)", geofenceTag)
     }
--- a/Sources/LocationGeofence/Logger+Geofence.swift
+++ b/Sources/LocationGeofence/Logger+Geofence.swift
@@ -166,6 +166,12 @@ extension Logger {
         info("Polygon \(transition.rawValue) for region \(identifier), confirmed by \(basis)", geofenceTag)
     }
 
+    /// A whole-set pass declined because one is already running. Logged so a foreground that
+    /// produced no verdicts is distinguishable from one that never ran.
+    func geofencePolygonPassSkipped(reason: String) {
+        debug("Skipped polygon evaluation pass: \(reason)", geofenceTag)
+    }
+
     func geofencePolygonUndecided(identifier: String, reason: String) {
         debug("Polygon membership undecided for region \(identifier): \(reason)", geofenceTag)
     }
--- a/Sources/LocationGeofence/Logger+Geofence.swift
+++ b/Sources/LocationGeofence/Logger+Geofence.swift
@@ -156,7 +156,8 @@ extension Logger {
 
     /// Positive record of an OS-delivered business transition. Without it a native delivery is
     /// only identifiable by the ABSENCE of a synthesized line, which makes the OS promotion rate
-    /// inferred rather than measured.
+    /// inferred rather than measured. Logged at the monitors' dispatch sites, where provenance is
+    /// known: the resolver's entry point also receives synthesized heals, which would inflate it.
     func geofenceOsTransitionReceived(identifier: String, transition: GeofenceTransition) {
         debug("OS delivered \(transition.rawValue) for region \(identifier)", geofenceTag)
     }
--- a/Sources/LocationGeofence/Polygon/PolygonMembershipResolver.swift
+++ b/Sources/LocationGeofence/Polygon/PolygonMembershipResolver.swift
@@ -31,6 +31,7 @@ final class PolygonMembershipResolver {
     private let fixResolver: MovementFixResolver
     private let logger: Logger
     private var foregroundObserverToken: NSObjectProtocol?
+    private var isEvaluatingAllPolygons = false
 
     init(
         storage: GeofenceStorage,
@@ -65,7 +66,7 @@ final class PolygonMembershipResolver {
     /// condition the cache has dropped) is forwarded rather than dropped: treating it as a circle
     /// is the behaviour that predates polygons, and losing a real crossing is worse than a
     /// covering-circle-shaped one.
-    func handleTransition(identifier: String, transition: GeofenceTransition, occurredAt: Date? = nil) async {
+    func handleTransition(identifier: String, transition: GeofenceTransition, occurredAt: Date) async {
         logger.geofenceOsTransitionReceived(identifier: identifier, transition: transition)
         guard let geofence = await cachedGeofence(id: identifier), geofence.vertices != nil else {
             // Uncached, or a genuine circle: forward untouched, the behaviour that predates polygons.
@@ -84,7 +85,7 @@ final class PolygonMembershipResolver {
             // fix. It is still ORDERED against the stored belief: a synthesized or replayed exit
             // arriving after a newer enter would otherwise swallow it and leave the device believed
             // outside while it sits inside.
-            await apply(.outside, to: geofence, evidence: occurredAt)
+            await apply(.outside, to: geofence, evidence: occurredAt, confirmedByFix: false)
         case .enter:
             await evaluate(geofence: geofence, polygon: polygon)
         }
@@ -94,15 +95,34 @@ final class PolygonMembershipResolver {
     ///
     /// The one case no OS event covers: a device already inside a polygon when monitoring begins
     /// has crossed nothing, so the OS has nothing to report, and a device standing still produces
-    /// no movement pass either. Foregrounding is the remaining signal, and it is free — the first
-    /// evaluation resolves a fix and the rest read it from the cache.
+    /// no movement pass either. Foregrounding is the remaining signal, and it is free — one fix
+    /// serves the whole pass.
+    ///
+    /// Foregrounds arrive in bursts, so a pass already running wins: a second concurrent scan reads
+    /// the same storage and the same fix and can only duplicate the location work.
     func evaluateAllPolygons() async {
+        guard !isEvaluatingAllPolygons else {
+            logger.geofencePolygonPassSkipped(reason: "a pass is already running")
+            return
+        }
+        isEvaluatingAllPolygons = true
+        defer { isEvaluatingAllPolygons = false }
         let registered = await storage.getRegisteredBusinessIds()
         let polygons = await storage.getCachedGeofences()
             .filter { registered.contains($0.id) && $0.vertices != nil }
+        guard !polygons.isEmpty else { return }
+        // One request for the whole pass. Resolving per polygon would issue a fresh timed request
+        // for every one of them whenever the cache stays empty, holding the main actor for minutes
+        // and still deciding nothing.
+        guard let fix = await resolveFix() else {
+            for geofence in polygons {
+                logger.geofencePolygonUndecided(identifier: geofence.id, reason: "no usable fix")
+            }
+            return
+        }
         for geofence in polygons {
             guard let polygon = geofence.polygonRegion else { continue }
-            await evaluate(geofence: geofence, polygon: polygon)
+            await evaluate(geofence: geofence, polygon: polygon, fix: fix)
         }
     }
 
@@ -122,7 +142,15 @@ final class PolygonMembershipResolver {
     }
 
     private func evaluate(geofence: Geofence, polygon: PolygonRegion) async {
-        guard let fix = await resolveFix(), CLLocationCoordinate2DIsValid(fix.coordinate) else {
+        guard let fix = await resolveFix() else {
+            logger.geofencePolygonUndecided(identifier: geofence.id, reason: "no usable fix")
+            return
+        }
+        await evaluate(geofence: geofence, polygon: polygon, fix: fix)
+    }
+
+    private func evaluate(geofence: Geofence, polygon: PolygonRegion, fix: CLLocation) async {
+        guard CLLocationCoordinate2DIsValid(fix.coordinate) else {
             logger.geofencePolygonUndecided(identifier: geofence.id, reason: "no usable fix")
             return
         }
@@ -139,14 +167,21 @@ final class PolygonMembershipResolver {
             )
             return
         }
-        await apply(membership, to: geofence, evidence: fix.timestamp)
+        await apply(membership, to: geofence, evidence: fix.timestamp, confirmedByFix: true)
     }
 
     /// Applies a membership verdict and delivers the crossing when it changes the stored belief.
     /// `evidence` is when the crossing happened — a fix's timestamp, or the OS event's date for a
-    /// covering-circle exit. `nil` only when the caller has no date, which leaves the write
-    /// unordered against the stored belief.
-    private func apply(_ membership: PolygonMembership, to geofence: Geofence, evidence: Date?) async {
+    /// covering-circle exit. Required, not optional: it is what orders the write against the stored
+    /// belief, and a caller allowed to omit it could silently write an unordered one. `confirmedByFix`
+    /// says which of the two it was, since both carry a date and the date alone cannot tell the log
+    /// how membership was decided.
+    private func apply(
+        _ membership: PolygonMembership,
+        to geofence: Geofence,
+        evidence: Date,
+        confirmedByFix: Bool
+    ) async {
         let outcome = await storage.recordPolygonMembership(
             membership,
             forIdentifier: geofence.id,
@@ -158,7 +193,7 @@ final class PolygonMembershipResolver {
         logger.geofencePolygonTransition(
             identifier: geofence.id,
             transition: transition,
-            confirmedByFix: evidence != nil
+            confirmedByFix: confirmedByFix
         )
         await transitionEmitter.trackTransition(geofenceId: geofence.id, transition: transition)
     }
--- a/Sources/LocationGeofence/Polygon/PolygonMembershipResolver.swift
+++ b/Sources/LocationGeofence/Polygon/PolygonMembershipResolver.swift
@@ -1,0 +1,192 @@
+import CioInternalCommon
+import CoreLocation
+import Foundation
+#if canImport(UIKit)
+import UIKit
+#endif
+
+/// Turns OS covering-circle events into customer-facing polygon transitions.
+///
+/// The OS monitors a polygon's server-guaranteed covering circle and knows nothing about the
+/// polygon itself, so membership is a second fact that has to be established on device. This sits
+/// between the monitor and the event tracker and is the only component that owns it. Circle
+/// geofences pass straight through: the contradiction gate, baseline heal and dedup baseline that
+/// serve them are deliberately left unaware polygons exist.
+///
+/// Two invariants define correctness here, and neither may be relaxed:
+/// - an ENTER requires a gated fix placing the device inside the polygon;
+/// - an EXIT requires either a covering-circle exit — polygon ⊆ circle, so leaving the circle
+///   provably leaves the polygon — or a gated fix placing the device outside.
+///
+/// Everything else is silent. That is what "no event without gated geometric confirmation" means
+/// in code, and why an undecidable fix leaves the stored belief untouched rather than guessing.
+///
+/// Owns its own `MovementFixResolver` rather than reading the monitor's: the transition handler
+/// carries only coordinates, while the gate needs accuracy and age, and on a wrapper cold wake
+/// `CustomerIO.initialize` has not run — the same reasoning that gives each monitor its own.
+@MainActor
+final class PolygonMembershipResolver {
+    private let storage: GeofenceStorage
+    private let transitionEmitter: GeofenceTransitionEmitting
+    private let fixResolver: MovementFixResolver
+    private let logger: Logger
+    private var foregroundObserverToken: NSObjectProtocol?
+
+    init(
+        storage: GeofenceStorage,
+        transitionEmitter: GeofenceTransitionEmitting,
+        logger: Logger,
+        fixResolver: MovementFixResolver? = nil
+    ) {
+        self.storage = storage
+        self.transitionEmitter = transitionEmitter
+        self.logger = logger
+        self.fixResolver = fixResolver ?? MovementFixResolver(
+            logger: logger,
+            backgroundTaskRunner: GeofenceBackgroundTime.runner(name: "io.customer.geofence.polygon-fix")
+        )
+        registerForegroundEvaluation()
+    }
+
+    deinit {
+        if let foregroundObserverToken {
+            NotificationCenter.default.removeObserver(foregroundObserverToken)
+        }
+    }
+
+    /// Routes a business-geofence transition the OS delivered. A circle fence is forwarded to the
+    /// tracker unchanged; a polygon's covering-circle event is interpreted against membership.
+    ///
+    /// A geofence missing from the cache (a sync raced this event, or the OS still holds a
+    /// condition the cache has dropped) is forwarded rather than dropped: treating it as a circle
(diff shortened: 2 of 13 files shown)


Sources/LocationGeofence/Polygon/PolygonMembershipResolver.swift
```diff
@@ -0,0 +1,192 @@
...
+    /// `evidence` is the fix's timestamp, or `nil` for the covering-circle exit, which is
+    /// geometric rather than fix-derived and so is never outranked by a newer belief.
+    private func apply(_ membership: PolygonMembership, to geofence: Geofence, evidence: Date?) async {
+        let outcome = await storage.recordPolygonMembership(
+            membership,
+            forIdentifier: geofence.id,
+            onlyIfBeliefPredates: evidence
+        )
+        guard case .deliver(let transition) = outcome,
+              geofence.transitionTypes.contains(transition)
```
mahmoud-elmorabea, line 260: Consequence: a polygon set to enter-only fires once per install and never again; an exit-only polygon never fires at all. The filtered edge is dropped at the monitor (covering circles are registered with the customer's transition types), so the belief never advances. Suggested fix for #1246: register polygon covering circles with both types and let this filter be the only customer-facing one.

Sources/LocationGeofence/Polygon/PolygonMembershipResolver.swift
```diff
@@ -0,0 +1,192 @@
...
+    private func cachedGeofence(id: String) async -> Geofence? {
+        await storage.getCachedGeofences().first { $0.id == id }
+    }
+
+    /// Freshest fix obtainable, requesting one when the cache is stale. Mirrors the gate's
+    /// resolution: the completion's coordinates are discarded in favour of `latestFix`, which
+    /// carries the accuracy and timestamp the decision needs.
+    private func resolveFix() async -> CLLocation? {
+        await withCheckedContinuation { continuation in
+            fixResolver.resolve(cached: fixResolver.latestFix) { [weak self] _ in
```
mahmoud-elmorabea: Consequence: on a cold-process wake, if the location request fails or times out, that enter is lost for good. The monitor has already advanced its dedup baseline so it won't re-emit; recovery only happens if the user foregrounds while still inside. `latestFix` is only what this resolver delivered itself. Suggest passing `manager.location` as the cached fallback, matching the gate.

Sources/LocationGeofence/Polygon/PolygonMembershipResolver.swift
```diff
@@ -0,0 +1,192 @@
...
+    init(
+        storage: GeofenceStorage,
+        transitionEmitter: GeofenceTransitionEmitting,
+        logger: Logger,
+        fixResolver: MovementFixResolver? = nil
+    ) {
+        self.storage = storage
+        self.transitionEmitter = transitionEmitter
+        self.logger = logger
+        self.fixResolver = fixResolver ?? MovementFixResolver(
```
mahmoud-elmorabea, line 54: Consequence: polygons near the 100 m minimum never fire in either direction on Wi-Fi or cell fixes, and nothing under ~40 m across can fire even with perfect GPS. The default resolver requests hundred-meter accuracy while the decision needs the point to be farther from the edge than the fix accuracy. Suggest requesting best or ten-meter accuracy for polygon evaluation.

Sources/LocationGeofence/Polygon/PolygonMembershipResolver.swift
```diff
@@ -0,0 +1,192 @@
...
+            )
+            return
+        }
+        await apply(membership, to: geofence, evidence: fix.timestamp)
+    }
+
+    /// Applies a membership verdict and delivers the crossing when it changes the stored belief.
+    /// `evidence` is the fix's timestamp, or `nil` for the covering-circle exit, which is
+    /// geometric rather than fix-derived and so is never outranked by a newer belief.
+    private func apply(_ membership: PolygonMembership, to geofence: Geofence, evidence: Date?) async {
```
mahmoud-elmorabea: Consequence: if iOS stops monitoring a polygon and the user leaves during that window, the next real enter is silently dropped and the next exit fires without a matching enter. The circle baseline reseeds on `.unmonitored`; the polygon belief does not. Already raised on #1244; fix is in storage.

Sources/LocationGeofence/Polygon/PolygonMembershipResolver.swift
```diff
@@ -0,0 +1,192 @@
...
+              let polygon = geofence.polygonRegion
+        else {
+            await transitionEmitter.trackTransition(geofenceId: identifier, transition: transition)
+            return
+        }
+        switch transition {
+        case .exit:
+            // polygon ⊆ covering circle, so leaving the circle is geometric certainty and needs
+            // no fix.
+            await apply(.outside, to: geofence, evidence: nil)
```
mahmoud-elmorabea: Consequence: a synthesized or spurious exit can fire seconds after a legitimate enter, and the OS's corrective enter is then swallowed. The user sits inside with the SDK believing they are outside, and the enter cooldown blocks recovery for an hour. The handler carries no event date, so this exit is trusted unconditionally. Suggest carrying the event date and passing it as evidence.

Sources/LocationGeofence/Polygon/PolygonMembershipResolver.swift
```diff
@@ -0,0 +1,192 @@
...
+    ///
+    /// The one case no OS event covers: a device already inside a polygon when monitoring begins
+    /// has crossed nothing, so the OS has nothing to report, and a device standing still produces
+    /// no movement pass either. Foregrounding is the remaining signal, and it is free — the first
+    /// evaluation resolves a fix and the rest read it from the cache.
+    func evaluateAllPolygons() async {
+        let registered = await storage.getRegisteredBusinessIds()
+        let polygons = await storage.getCachedGeofences()
+            .filter { registered.contains($0.id) && $0.vertices != nil }
+        for geofence in polygons {
```
mahmoud-elmorabea, line 138: Consequence: with many polygons registered and no fix available, every foreground can hold the main actor for minutes, issuing one location request per polygon, and still decide nothing. The cache only helps when the first request succeeds. Suggest resolving once above the loop and bailing on nil, plus an in-flight guard.

Sources/LocationGeofence/Polygon/PolygonMembershipResolver.swift
```diff
@@ -0,0 +1,192 @@
...
+    /// tracker unchanged; a polygon's covering-circle event is interpreted against membership.
+    ///
+    /// A geofence missing from the cache (a sync raced this event, or the OS still holds a
+    /// condition the cache has dropped) is forwarded rather than dropped: treating it as a circle
+    /// is the behaviour that predates polygons, and losing a real crossing is worse than a
+    /// covering-circle-shaped one.
+    func handleTransition(identifier: String, transition: GeofenceTransition) async {
+        logger.geofenceOsTransitionReceived(identifier: identifier, transition: transition)
+        guard let geofence = await cachedGeofence(id: identifier),
+              let polygon = geofence.polygonRegion
```
mahmoud-elmorabea: Consequence: a cached polygon whose ring no longer builds would fire a customer enter anywhere inside its covering circle, with no log. Latent today. `Geofence.polygonRegion` doc says nil is not proof of a circle. Suggest taking the circle branch only when `vertices == nil`.

Sources/LocationGeofence/Logger+Geofence.swift
```diff
@@ -154,6 +154,22 @@ extension Logger {
         info("Synthesized \(transition.rawValue) for region \(identifier): fresh fix contradicts stored baseline (OS never delivered the crossing)", geofenceTag)
     }
 
+    /// Positive record of an OS-delivered business transition. Without it a native delivery is
+    /// only identifiable by the ABSENCE of a synthesized line, which makes the OS promotion rate
+    /// inferred rather than measured.
+    func geofenceOsTransitionReceived(identifier: String, transition: GeofenceTransition) {
```
mahmoud-elmorabea, line 161: Consequence: anyone measuring how often iOS delivers crossings itself gets a count inflated by the SDK's own synthesized events. This is logged at the top of `handleTransition`, which also receives baseline-heal transitions. Suggest logging at the two OS dispatch sites where provenance is known.

Sources/LocationGeofence/Polygon/PolygonMembershipResolver.swift
mrehan27: Fixed in #1246 exactly as suggested: polygon covering circles register with both edges, and this filter is the only customer-facing one.

Sources/LocationGeofence/Polygon/PolygonMembershipResolver.swift
mrehan27: Fixed — CoreLocation's own cached fix is the fallback now, still age-gated. #1250 narrows it in one direction: a forced-fresh request never answers from a held fix.

Sources/LocationGeofence/Polygon/PolygonMembershipResolver.swift
mrehan27: Fixed — the resolver's `MovementFixResolver` requests ten-metre accuracy instead of the circle path's hundred.

Sources/LocationGeofence/Polygon/PolygonMembershipResolver.swift
mrehan27: Fixed in #1244 storage, on both paths — `clearMonitorRegionRecord` and `forceReseed`.

Sources/LocationGeofence/Polygon/PolygonMembershipResolver.swift
mrehan27: Fixed — the covering exit is ordered against the stored belief by the event date, and the date is non-optional through the handler so no caller can omit it.

Sources/LocationGeofence/Polygon/PolygonMembershipResolver.swift
mrehan27: Both done: one fix resolved above the loop with a bail on nil, plus an in-flight guard that logs the skip. Tests assert one request for a three-polygon pass, and one for two concurrent passes.

Sources/LocationGeofence/Polygon/PolygonMembershipResolver.swift
mrehan27: Fixed — the circle branch is taken only when `vertices == nil`. A stored ring that no longer builds logs undecided and forwards nothing.

Sources/LocationGeofence/Logger+Geofence.swift
mrehan27: Fixed — moved to the two OS dispatch sites in the monitors, where provenance is known.

Sources/LocationGeofence/Polygon/PolygonMembershipResolver.swift
```diff
@@ -0,0 +1,242 @@
...
+        _ membership: PolygonMembership,
+        to geofence: Geofence,
+        evidence: Date,
+        confirmedByFix: Bool
+    ) async {
+        let outcome = await storage.recordPolygonMembership(
+            membership,
+            forIdentifier: geofence.id,
+            onlyIfBeliefPredates: evidence
+        )
```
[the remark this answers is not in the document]
mrehan27: Fixed — `recordPolygonMembership` stores `evidenceTimestamp ?? now`, so `lastChangedAt` is the fix's own timestamp rather than the write time, on both the create and the change path. `recordPolygonMembership_givenEvidenceNewerThanPriorEvidence_expectAccepted` pins it: a write landing 60 s after the fix that justified it no longer rejects a later verdict whose own fix is genuinely newer.

Sources/LocationGeofence/Polygon/PolygonMembershipResolver.swift
```diff
@@ -0,0 +1,192 @@
...
+        guard case .deliver(let transition) = outcome,
+              geofence.transitionTypes.contains(transition)
+        else { return }
+        logger.geofencePolygonTransition(
+            identifier: geofence.id,
+            transition: transition,
+            confirmedByFix: evidence != nil
+        )
+        await transitionEmitter.trackTransition(geofenceId: geofence.id, transition: transition)
+    }
```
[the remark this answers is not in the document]
mrehan27: Working as intended. A covering-circle exit proves the device left the circle, not that it was ever inside the polygon — the annulus case is exactly a device that enters the circle, never crosses the ring, and leaves again. With no prior belief no enter was ever delivered, so emitting an exit here would produce an unpaired exit; `suppressedInitialOutside` is what keeps enter/exit balanced.

Sources/LocationGeofence/Polygon/PolygonMembershipResolver.swift
```diff
@@ -0,0 +1,242 @@
...
+        // One request for the whole pass. Resolving per polygon would issue a fresh timed request
+        // for every one of them whenever the cache stays empty, holding the main actor for minutes
+        // and still deciding nothing.
+        guard let fix = await resolveFix() else {
+            for geofence in polygons {
+                logger.geofencePolygonUndecided(identifier: geofence.id, reason: "no usable fix")
+            }
+            return
+        }
+        for geofence in polygons {
```
Shahroz16, line 138: Could we revalidate the registered polygon after the location request returns, before recording or emitting a verdict? I reproduced a refresh replacing a fence under the same ID while a fix was pending: the resumed pass evaluated the old ring and emitted an enter even though the current polygon excluded the point. This also happens without switching users. Checking the current polygon/version when applying the result, with a refresh-during-fix regression, would cover this. I'd like to close this before approving the resolver stack.

Shahroz16 (commented):
One resolver correctness issue remains; details inline.

Sources/LocationGeofence/Polygon/PolygonMembershipResolver.swift
mrehan27: Done in `8ccb96f6`, structurally rather than as a check: `evaluate` takes an id and reads the fence after the fix, so a caller's `PolygonRegion` is never in scope to decide from. Registration comes from the same load — sampling geometry after the await and registration before it is how the two come to disagree.

Tests hold the pass inside its fix request and change the catalog between the read and the verdict: replaced ring, dropped fence, unregistered fence, each with a control that must still deliver. Reverting the re-read fails them with a delivered enter from the stale ring.

One honest note on the unregistered case: it passes with and without the resolver change, because `recordRegistration` prunes the belief with the set so storage refuses the write as unmonitored. That test pins the outcome, not the mechanism — the re-read is what stops the resolver depending on a storage rule that only covers the create path.

Shahroz16: Thanks, re-reading after the fix closes the original case. Could we also make the membership write conditional on the geometry it evaluated? In a coordinator regression, a refresh committed a same-ID replacement between `getRegisteredGeofence` and `recordPolygonMembership`. The replacement excluded the fix, but the old verdict still stored INSIDE and emitted ENTER. That gap includes waiting to resume on MainActor, so it is not necessarily brief. Passing the evaluated geometry revision into the atomic membership write and rejecting a mismatch would close this window.

Shahroz16 (commented):
The refresh-during-location-request fix is verified. One narrower geometry read/write issue remains in the existing discussion, so I am holding approval.

Sources/LocationGeofence/Polygon/PolygonMembershipResolver.swift
mrehan27: Done in `0601f657`, the shape you suggested. `recordPolygonMembership` now takes `onlyIfRingMatches`, and the evaluated ring is compared against the stored one **inside the same `loadFromDisk` that writes** — the one place read and write cannot be separated — returning `.suppressedGeometryChanged` on a mismatch. A third check in the resolver would only be a third thing that can go stale.

It compares the ring itself rather than `lastUpdated`: the server owns that field and a replacement that failed to bump it would pass a check written against it.

The covering-circle exit passes no ring on purpose — containment holds whatever the ring is, so leaving the circle is a verdict no replacement can invalidate. A test pins that it still delivers after the exact replacement that refuses a fix-derived verdict.

Four storage tests: replaced-ring refused, unchanged-ring control still writes, covering exit still lands after that replacement, fence dropped from the catalog refused.

Sources/LocationGeofence/Polygon/PolygonMembershipResolver.swift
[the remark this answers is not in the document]
mrehan27: Not a defect — the premise is wrong. A covering-circle exit proves the device left the **circle**, not that it was ever inside the **polygon**: the annulus between the two is exactly the region where both are true. With no stored belief no enter was ever delivered, so emitting here would produce a lone exit with no matching enter.

The circle path does the same thing: `recordMonitorEvent` with no record returns `.suppressedNoBaseline` and only establishes the baseline. Adopting this suggestion would make polygons behave differently from circles, in the wrong direction.

Documented at `GeofenceStorage+PolygonMembership.swift` (a polygon with no record is undecided, not outside) and pinned by `handleTransition_givenPolygonExitWhileNotInside_expectSilent`.

Reviewed independently by a second session, which reached the same conclusion from source.

Sources/LocationGeofence/Polygon/PolygonMembershipResolver.swift
Shahroz16: Thanks, the atomic ring check closes the fix-derived case. Could we carry the covering-circle check into the membership write too? In a storage-boundary regression, I matched the event circle to the cached fence, replaced both the ring and circle under the same ID, then made the same unguarded write used by this exit path. It returned `deliver(.exit)` and stored OUTSIDE even though the point was inside the replacement polygon. Leaving the old circle only proves outside for polygons it still encloses. Comparing the event circle against the current fence inside the storage write, with a refresh-between-check-and-write regression, would close this remaining window.

Shahroz16 (commented):
The atomic ring check closes the fix-derived race. A covering-circle exit still has a separate read/check/write window, reproduced at the storage boundary in the existing discussion, so I am holding approval for that path.

Sources/LocationGeofence/Polygon/PolygonMembershipResolver.swift
mrehan27: Closed in `4f2dc236`. The circle now travels into `recordPolygonMembership` and is compared inside the same `loadFromDisk` that writes, so there is no read/check/write window; the pre-hop check is removed rather than kept alongside it, since two comparisons against geometry read at different instants was the defect. Nil still means the producer could not say which circle was crossed and is treated as current.

Shahroz16: Thanks, the circle comparison now runs inside the same storage operation as the membership write. The replaced-circle regression and unchanged/clamped controls pass. This closes the resolver's storage-window issue.

Shahroz16 (approved):
The circle comparison now runs atomically with the membership write, closing the remaining storage-window concern. The updated stack passed 71 focused resolver and storage tests on the simulator, and the independent review found no new issue in this fix. This resolver change looks ready.

PR #1246, opened 2026-08-31
chore(geofence): route monitored transitions through the polygon resolver

mrehan27:

Part 5 of the polygon geofence work. Wires the resolver into the monitor, and closes two
polygon-specific hazards in the paths around it.

### What's here

**Binder** — business transitions now dispatch to the resolver, which forwards circles to the
tracker unchanged and interprets a polygon's covering-circle event against membership.

**Initial enter-when-inside excludes polygons.** That test compares the anchor against `radius`,
which for a polygon is its *covering circle* — so a device sitting in the annulus would have been
handed an enter it never earned. Newly-registered polygons go to the gated evaluation instead,
which is also the only thing that reports a device already standing inside one.

**A polygon whose covering circle exceeds the OS limit is dropped, not clamped.** Both monitors
clamp silently, and a clamped circle no longer contains the polygon — which turns the
covering-circle exit from geometric certainty into a false exit, and lets the device enter through
a part of the polygon no wake covers. Circles keep clamping: for them the monitored circle *is* the
fence, so a smaller one only reports later. The drop happens **before ranking**, so a refused
polygon no longer occupies a business slot a usable candidate could take.

**A polygon's covering circle registers with both edges**, whatever the customer's transition types
are. The circle is machinery, not the fence: registering it enter-only starved the resolver of the
exit it needs to advance belief, so an enter-only polygon fired once per install and an exit-only
one never fired at all. The customer's filter applies to the polygon verdict instead. Real circles
keep the customer's types.

**Transitions carry the event date** — the OS event's, or the fix timestamp for a synthesized heal —
so a late or replayed one can be ordered against what the SDK believes. Not optional: every dispatch
site already had a date, and a consumer handed none writes unordered. The OS-received log moved to
the dispatch sites at the same time, where provenance is known; logged in the resolver it counted
the SDK's own synthesized heals as OS deliveries.

**Newly registered polygons are evaluated as a batch**, sharing one fix. Resolving per polygon
spends the whole location-request timeout once per polygon on the main actor whenever no fix is
obtainable, and decides nothing — the same defect the foreground pass had.

**The user is re-checked after the fix resolves**, not only before evaluation starts. Resolving
suspends, and a switch in that window clears user-scoped state — the resumed task would rewrite the
old user's belief and stamp any event to whoever signed in. The OS-delivered path deliberately does
not re-check: a sign-out clears the registration set, which already refuses the write, and a switch
that re-registers the same polygon leaves the new user genuinely inside it.

**The batch holds ids, not rings.** One fix serves every polygon in it, so a refresh has the whole
request in which to replace geometry under those ids — the same hazard the resolver closes in #1245,
reaching this path because the batch is built before the fix. It carries `[String]` and each verdict
reads its own fence afterwards. The pre-filter still builds the ring and discards it, deliberately:
it runs before the request, so rejecting an unbuildable ring there saves a whole fix request.

**Events carry the circle they were raised for.** #1245 refuses a covering-circle exit whose circle
the fence no longer has, but nothing supplied that circle, so on every real crossing the check saw
nil and never fired. CLMonitor now reports the condition it registered, the classic path the region
the delegate handed it — captured at intake, because a queued event can drain after the region was
replaced — and the baseline heal the condition it synthesized against. Each producer also reports
the cap it clamped that radius against, so the comparison can reconstruct what was actually
registered.

### Testing

476/476 `LocationGeofenceTests` pass on iOS 26. Format and lint clean.

Each behaviour is tested with a circle alongside the polygon as the negative control, so the
polygon's silence is the exclusion working rather than a dead assertion — including that a real
circle still registers with the customer's own transition types, and that a cap-sized candidate
takes the slot a refused polygon would have held.

### Stack

5. #1246 `mbl-2291-c` monitor wiring ← **this PR**

### DIFF
--- a/Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator+InitialEnter.swift
+++ b/Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator+InitialEnter.swift
@@ -21,11 +21,22 @@ extension GeofenceSyncCoordinatorImpl {
         expectedUserId: String,
         anchor: LocationData
     ) {
-        let newInside = candidates.filter { region in
+        let newlyRegistered = candidates.filter { region in
             osRegistration.registeredIds.contains(region.id)
                 && !previouslyRegisteredIds.contains(region.id)
+        }
+        // A polygon must NOT be judged by this containment test: `radius` is its covering circle, so
+        // a device in the annulus would emit an enter it never earned. Newly-registered polygons go
+        // to the resolver's gated evaluation instead, which is also the only thing that reports the
+        // device already standing inside one — there is no crossing for the OS to deliver.
+        let newPolygons = newlyRegistered.filter { $0.vertices != nil }
+        let newInside = newlyRegistered.filter { region in
+            region.vertices == nil
                 && region.transitionTypes.contains(.enter)
                 && region.distanceTo(anchor) <= min(region.radius, osRegistration.maxMonitoringRadius)
+        }
+        if !newPolygons.isEmpty {
+            evaluateNewPolygons(newPolygons, expectedUserId: expectedUserId)
         }
         guard !newInside.isEmpty else { return }
         // Deliver off the refresh gate (like the binder does for real crossings) so a slow send can't
@@ -37,6 +48,20 @@ extension GeofenceSyncCoordinatorImpl {
                 // current — so stop the batch the moment identity changes.
                 guard contextStore.currentUserId == expectedUserId else { return }
                 await transitionEmitter.trackTransition(geofenceId: region.id, transition: .enter)
+            }
+        }
+    }
+
+    /// Hands newly-registered polygons to the gated evaluation. Resolved from the graph at the call
+    /// site rather than stored: the resolver is a `@MainActor` singleton and this coordinator is
+    /// one of its own dependencies, so a stored reference back would make the two initialize each
+    /// other.
+    private func evaluateNewPolygons(_ polygons: [Geofence], expectedUserId: String) {
+        Task { @MainActor [contextStore] in
+            let resolver = DIGraphShared.shared.polygonMembershipResolver
+            for polygon in polygons {
+                guard contextStore.currentUserId == expectedUserId else { return }
+                await resolver.evaluateMembership(geofenceId: polygon.id, reason: "new polygon")
             }
         }
     }
--- a/Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator+InitialEnter.swift
+++ b/Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator+InitialEnter.swift
@@ -61,7 +61,13 @@ extension GeofenceSyncCoordinatorImpl {
             let resolver = DIGraphShared.shared.polygonMembershipResolver
             for polygon in polygons {
                 guard contextStore.currentUserId == expectedUserId else { return }
-                await resolver.evaluateMembership(geofenceId: polygon.id, reason: "new polygon")
+                // Also re-checked inside, after the fix resolves: that await is the window where a
+                // user switch would otherwise land an event on the wrong profile.
+                await resolver.evaluateMembership(
+                    geofenceId: polygon.id,
+                    reason: "new polygon",
+                    isStillCurrent: { contextStore.currentUserId == expectedUserId }
+                )
             }
         }
     }
--- a/Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator+InitialEnter.swift
+++ b/Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator+InitialEnter.swift
@@ -58,17 +58,14 @@ extension GeofenceSyncCoordinatorImpl {
     /// other.
     private func evaluateNewPolygons(_ polygons: [Geofence], expectedUserId: String) {
         Task { @MainActor [contextStore] in
-            let resolver = DIGraphShared.shared.polygonMembershipResolver
-            for polygon in polygons {
-                guard contextStore.currentUserId == expectedUserId else { return }
-                // Also re-checked inside, after the fix resolves: that await is the window where a
-                // user switch would otherwise land an event on the wrong profile.
-                await resolver.evaluateMembership(
-                    geofenceId: polygon.id,
-                    reason: "new polygon",
-                    isStillCurrent: { contextStore.currentUserId == expectedUserId }
-                )
-            }
+            guard contextStore.currentUserId == expectedUserId else { return }
+            // Also re-checked inside, per polygon, after the fix resolves: that await is the window
+            // where a user switch would otherwise land an event on the wrong profile.
+            await DIGraphShared.shared.polygonMembershipResolver.evaluateMembership(
+                geofenceIds: polygons.map(\.id),
+                reason: "new polygon",
+                isStillCurrent: { contextStore.currentUserId == expectedUserId }
+            )
         }
     }
 }
--- a/Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator+OsRegistration.swift
+++ b/Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator+OsRegistration.swift
@@ -48,7 +48,20 @@ extension GeofenceSyncCoordinatorImpl {
                 transitionTypes: [.exit]
             ))
         }
-        desired.append(contentsOf: businessRegions.map { region in
+        // A polygon whose covering circle exceeds the OS cap must be DROPPED, not clamped. Both
+        // monitors clamp silently, and a clamped circle no longer contains the polygon — which
+        // turns the covering-circle exit from geometric certainty into a false exit, and lets the
+        // device enter through a part of the polygon no wake covers. Circles keep clamping: for
+        // them the monitored circle IS the fence, so a smaller one only reports later.
+        let maximumRadius = monitor.maximumMonitoringRadius
+        let registrable = businessRegions.filter { region in
+            guard region.vertices != nil, region.radius > maximumRadius else { return true }
+            logger.geofencePolygonExceedsMonitoringLimit(
+                identifier: region.id, radius: region.radius, limit: maximumRadius
+            )
+            return false
+        }
+        desired.append(contentsOf: registrable.map { region in
             GeofenceRegionRequest(
                 identifier: region.id,
                 center: LocationData(latitude: region.latitude, longitude: region.longitude),
--- a/Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator+OsRegistration.swift
+++ b/Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator+OsRegistration.swift
@@ -53,6 +53,8 @@ extension GeofenceSyncCoordinatorImpl {
         // turns the covering-circle exit from geometric certainty into a false exit, and lets the
         // device enter through a part of the polygon no wake covers. Circles keep clamping: for
         // them the monitored circle IS the fence, so a smaller one only reports later.
+        // Callers drop these before ranking so the slot is reused; this is the last-line guard for
+        // anything that reached here anyway, and it logs so the skip is never silent.
         let maximumRadius = monitor.maximumMonitoringRadius
         let registrable = businessRegions.filter { region in
             guard region.vertices != nil, region.radius > maximumRadius else { return true }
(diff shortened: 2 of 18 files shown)


Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator+InitialEnter.swift
```diff
@@ -40,4 +51,18 @@ extension GeofenceSyncCoordinatorImpl {
...
+    /// Hands newly-registered polygons to the gated evaluation. Resolved from the graph at the call
+    /// site rather than stored: the resolver is a `@MainActor` singleton and this coordinator is
+    /// one of its own dependencies, so a stored reference back would make the two initialize each
+    /// other.
+    private func evaluateNewPolygons(_ polygons: [Geofence], expectedUserId: String) {
+        Task { @MainActor [contextStore] in
+            let resolver = DIGraphShared.shared.polygonMembershipResolver
+            for polygon in polygons {
+                guard contextStore.currentUserId == expectedUserId else { return }
+                await resolver.evaluateMembership(geofenceId: polygon.id, reason: "new polygon")
```
Shahroz16: Could we carry expectedUserId through the resolver and re-check it after the location await, before changing membership or emitting? This guard runs only before evaluateMembership, which can suspend for a fresh fix. If the user switches in that window, cleanup clears user-scoped state, then this task resumes, recreates the old polygon belief, and any enter is stamped to the new user. A suspended-fix user-switch test would pin this, and the movement-wide evaluation added later needs the same boundary.

Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator+OsRegistration.swift
```diff
@@ -48,7 +48,20 @@ extension GeofenceSyncCoordinatorImpl {
                 transitionTypes: [.exit]
             ))
         }
-        desired.append(contentsOf: businessRegions.map { region in
+        // A polygon whose covering circle exceeds the OS cap must be DROPPED, not clamped. Both
+        // monitors clamp silently, and a clamped circle no longer contains the polygon — which
+        // turns the covering-circle exit from geometric certainty into a false exit, and lets the
+        // device enter through a part of the polygon no wake covers. Circles keep clamping: for
+        // them the monitored circle IS the fence, so a smaller one only reports later.
+        let maximumRadius = monitor.maximumMonitoringRadius
```
Shahroz16, line 58: Could we exclude these oversized polygons before nearest selection, or top the set back up after this drop? At the moment nearest(... limit: maxBusinessGeofences) runs first, so a dropped polygon still consumes one of the business slots. With 19 selected regions and one oversized polygon, iOS registers only 18 even if a valid 20th candidate is nearby. A cap-sized candidate test would catch the coverage loss.

Shahroz16 (commented):
Two lifecycle and capacity cases still look open.

Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator+InitialEnter.swift
mrehan27: Done — `evaluateMembership` takes an `isStillCurrent` closure re-checked after the fix resolves. The OS-delivered path deliberately does not: sign-out clears the registration set, which already refuses the write, and a switch that re-registers the same polygon leaves the new user genuinely inside it.

Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator+OsRegistration.swift
mrehan27: Done — oversized polygons are excluded before ranking, so a refused one no longer holds a slot. Covered by a cap-sized-candidate test.

mrehan27:
bugbot run

Shahroz16 (commented):
The registration-slot and post-fix user checks look good. Approval is still held on the inherited #1240 cache issue. The hosted iOS smoke check is also red at the identify/custom-event-persisted assertion: https://github.com/customerio/customerio-ios/actions/runs/33910990872/job/101147119364. Could we rerun that and investigate if it repeats? The 478 local geofence tests passed, but they don’t cover this hosted smoke path.

Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator+InitialEnter.swift
Shahroz16: The user-switch guard passes now. There's a separate stale-catalog case in evaluateMembership: a same-user refresh can replace the polygon while resolveFix is pending, and the retained pending list still emits from the old ring. I reproduced this with the identity guard enabled. Could we apply the [current-polygon/version validation discussed in #1245](https://github.com/customerio/customerio-ios/pull/1245#discussion_r3950915974) here too? I'm holding approval until that's covered.

Shahroz16 (commented):
The earlier cache and registration-slot fixes check out. The remaining resolver issue is linked in the thread. Lint currently fails while downloading its tool (HTTP 504); could we rerun that too?

Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator+InitialEnter.swift
mrehan27: Done in `9a4d9495`. The batch carries `[String]` and each verdict reads its own fence after the shared fix — the same cure as #1245, which this path needed separately because the batch is built before the request.

The pre-filter still builds the ring and discards it, deliberately: it runs before the request, so rejecting an unbuildable ring there saves a whole fix request, which the foreground pass has no equivalent saving for. Comment says so now.

Covered by a replaced-ring test on `evaluateMembership` plus a control that must still deliver.

Shahroz16: Thanks, the batch fix passes the original replaced-ring regression. The remaining issue is now the smaller geometry read/write window in the shared resolver, not the list retained across the location request. I have kept the repro and suggested fix in the [existing #1245 discussion](https://github.com/customerio/customerio-ios/pull/1245#discussion_r3950915974); this path inherits it.

Shahroz16 (commented):
The batch now carries IDs and the refresh-during-fix regression passes. Approval is still held for the shared resolver write-boundary issue in #1245.

Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator+InitialEnter.swift
mrehan27: Covered by `0601f657` on #1245, which this path inherits as you expected — the membership write is now conditional on the geometry it evaluated, compared inside the same `loadFromDisk` that writes. Nothing further was needed here; the batch fix in `9a4d9495` (now `86eaeb85` after the restack onto the merged feature branch) closed this path's own half.

mrehan27: Correction to the SHA above: after the restack onto the merged feature branch the batch fix `9a4d9495` is now **`e69f46f1`**, not `86eaeb85` — `86eaeb85` is an unrelated commit on this branch (carrying the event's circle to the resolver). The substance is unchanged.

Sources/LocationGeofence/Polygon/PolygonMembershipResolver.swift
```diff
@@ -105,10 +105,62 @@ final class PolygonMembershipResolver {
...
+            for geofenceId in pending {
+                logger.geofencePolygonUndecided(identifier: geofenceId, reason: "no usable fix")
+            }
+            return
+        }
+        for geofenceId in pending {
+            await evaluate(geofenceId: geofenceId, fix: fix, isStillCurrent: isStillCurrent)
+        }
+    }
+
```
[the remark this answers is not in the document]
mrehan27: Not reachable. On the path described, the stale write is refused rather than persisted: a user switch runs `clearUserScopedState`, which nils **both** `monitoredGeofenceIds` and `polygonMembership`. The resumed task therefore hits the create path, whose guard is `state.monitoredGeofenceIds?.contains(identifier) == true`, and returns `.suppressedUnmonitored`. The update branch is unreachable for the same reason — the clear removed `existing`.

The unguarded write is deliberate and documented at `PolygonMembershipResolver.swift`: a belief states geometry, which is true whoever is signed in, while an emit is an attribution — and attribution is what a switch invalidates, which is why the guard sits before the emit. Circle baselines are user-agnostic in the same way.

Reviewed independently by a second session, which also could not construct a reachable sequence.

Shahroz16 (commented):
The fix-derived geometry guard is now in place. I found a separate event-circle provenance issue in the CLMonitor wiring, detailed inline. This also inherits the [covering-circle write race in #1245](https://github.com/customerio/customerio-ios/pull/1245#discussion_r3950915974), so I am holding approval until those exit paths are covered.

Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor+Registration.swift
```diff
@@ -227,4 +227,16 @@ extension CLMonitorGeofenceMonitor {
             transitionTypes: transitionTypes
         )
     }
+
+    /// The circle this monitor registered for `identifier`, which is the one the OS raised the
+    /// event against. Nil before adoption repopulates the map on a cold wake — the consumer then
+    /// cannot tell a replaced circle from the current one, and treats the event as current.
+    func eventCircle(for identifier: String) -> MonitoredCircle? {
+        registeredConditions[identifier].map {
```
Shahroz16, line 235: Could we preserve the circle associated with the event instead of reading the staged registration here? `startMonitoring` updates `registeredConditions` before its queued OS remove/add drains, and dispatch reads this map after awaiting storage. An EXIT from the old circle can therefore be labeled with the replacement circle, pass the resolver's match check, and write OUTSIDE while the device is inside the new polygon. The contradiction gate already documents this staging gap. Carrying and validating the event's circle or registration generation through dispatch, including buffered events across re-add, with a reshape-during-dispatch regression would cover it.

mrehan27: Confirmed, and fixed in #1262. The ledger now stamps `liveFrom` when the queued remove+add actually drains, and an event raised before that resolves to the previous generation, so an exit from the old circle is no longer labelled with its replacement. It took three commits because the previous generation was also never populated on the one path a geometry change takes — `setMonitoredRegions` releases before re-registering, so there was nothing to supersede — and `RegisteredConditionLedger` was extracted so the sequence can be tested at all.

Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor.swift
```diff
@@ -307,12 +307,12 @@ final class CLMonitorGeofenceMonitor: NSObject, GeofenceRegionMonitoring, @preco
...
             movementFixResolver.resolve(cached: bestKnownFix()) { [weak self] location in
                 self?.logger.geofenceOsTransitionReceived(identifier: identifier, transition: transition)
-                self?.onTransition?(identifier, transition, location)
+                self?.onTransition?(identifier, transition, location, event.date, self?.eventCircle(for: identifier))
             }
             return
         }
         logger.geofenceOsTransitionReceived(identifier: identifier, transition: transition)
-        onTransition?(identifier, transition, currentLocationData())
+        onTransition?(identifier, transition, currentLocationData(), event.date, eventCircle(for: identifier))
```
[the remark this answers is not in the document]
mrehan27: Real, and fixed later in the stack in #1262 rather than here: `liveFrom` is stamped when the queued remove+add drains, and an event raised before that resolves to the previous generation instead of the staged one.

Shahroz16 (commented):
Rechecked the updated stack. The storage-write fix in #1245 is addressed, and this PR's own patch is unchanged. I'm still holding approval on event attribution: the staged circle can differ from the one that produced the OS event. The proposed fix in #1262 still loses the live circle when two registrations stage before the queue drains, which can drop a genuine exit. Could we close that remaining case before landing the wiring? Details are in the [existing follow-up](https://github.com/customerio/customerio-ios/pull/1262#discussion_r3973131951).

Shahroz16 (approved):
The event-attribution concern is now fixed in follow-up #1262 at 4f9c3ba7, including the two-staging case. I verified that fix against the original regression and the storage path. This PR's reviewed delta is unchanged, so I'm clearing its hold with that fix included in the stack.

PR #1247, opened 2026-08-31
chore(geofence): persist polygon tripwires

mrehan27:

Part 6 of the polygon geofence work. Tripwire state only — the model, its storage, and the
identifier namespace. **Nothing plants one yet**; that is the next PR.

### What's here

A tripwire is the device-centred circle planted inside a covering circle so the OS wakes the SDK
once the device has moved far enough for the membership verdict to change. Between the polygon
boundary and the covering circle the OS is silent, and a crossing there would go unnoticed until
some unrelated wake.

Tripwires are recognized by identifier prefix rather than a registry, so a cold-wake event can be
classified before any state has loaded. Customer geofence ids are server-assigned int64s, so they
cannot collide with the prefix.

Two prune rules, and the second is the subtle one. A tripwire is dropped when its polygon leaves
the registered set, on the same rule the membership belief follows. But its **monitor record** is
keyed by the tripwire identifier rather than the geofence id, so the record prune has to name it
explicitly — otherwise every sync would strip the baseline of a live condition and dedup away the
EXIT the tripwire exists to deliver. The state-space model catches exactly this as a negative
control: with the record prune keyed off business ids alone it reports 1.7 M suppressed crossings.

### Testing

Storage round-trip, both prune rules, and user-scoped clear.

### Stack

1. `mbl-2290-a-polygon-geometry-kernel` — geometry kernel (#1239)
6. `mbl-2291-d-polygon-tripwire-state` — tripwire state ← **this PR**
7. `mbl-2291-e-polygon-tripwire-planting` — tripwire planting + budget (#1248)

### DIFF
--- a/Sources/LocationGeofence/GeofenceInternalIdentifier.swift
+++ b/Sources/LocationGeofence/GeofenceInternalIdentifier.swift
@@ -1,0 +1,29 @@
+import Foundation
+
+/// Region identifiers the SDK registers for its own control flow rather than for a customer
+/// geofence. Internal regions are never tracked as transitions and never emit an initial enter.
+///
+/// Two kinds exist: the single movement trigger, whose EXIT drives re-ranking, and the per-polygon
+/// tripwires planted inside a covering circle to wake the SDK for a membership re-evaluation.
+/// Tripwires are recognized by prefix rather than a registry, so a cold-wake event can be
+/// classified before any state has loaded. Customer geofence ids are server-assigned int64s, so
+/// they cannot collide with the prefix.
+enum GeofenceInternalIdentifier {
+    private static let tripwirePrefix = "cio_tripwire_"
+
+    /// The tripwire identifier for a polygon geofence.
+    static func tripwire(for geofenceId: String) -> String {
+        tripwirePrefix + geofenceId
+    }
+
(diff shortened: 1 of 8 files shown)


mrehan27:
Closing — the per-polygon tripwire is replaced by a dynamically sized movement circle in #1250.

The approach came from @mahmoud's review: instead of planting one region per active polygon, the single movement trigger shrinks to the nearest polygon boundary. One region instead of N, and it removes the problem this pair was working around rather than mitigating it — Bugbot's finding here was that tripwires eat business slots, and #1248 answered by reserving slots. #1250 makes the contention not exist, so `maxBusinessGeofences` is honoured exactly.

Not merging these first: reviewing code that is about to be deleted wastes reviewer time, and the redesign is additive on top of #1246 rather than a rewrite of these diffs.

Carried forward into #1250:
- the fresh-fix requirement on wake evaluation (originally #1248) — the simulator found it was silently unwired and swallowing a polygon ENTER
- `GeofenceInternalIdentifier` internal-region handling, which is unrelated to tripwire planting and stays in #1246
- the 100 m floor and its rationale, now on the movement trigger

Branches are kept, so anything else needed can be recovered from them.

PR #1248, opened 2026-08-31
chore(geofence): plant polygon tripwires and register them

mrehan27:

Part 7 of the polygon geofence work, and the last of this ticket. Plants tripwires, composes them
into the OS-monitored set, and teaches both monitors what they are.

### What's here

A tripwire is planted whenever an evaluation places the device inside a covering circle, moved as
the device moves, and retired on leaving. Its radius reaches the polygon boundary, so leaving it is
exactly when the last verdict stops being safe to trust.

**An OS covering-circle enter outranks our own fix.** Measured in the field: a 15.5 s-old cached
fix — well inside `movementFixMaxAge` — placed the device ~400 m from where it was, outside a
circle it had just entered, and retired the tripwire for the whole 12 minutes it then spent in the
annulus.

**A tripwire is never planted for a polygon the device is outside.** The foreground path evaluates
*every* registered polygon, including ones kilometres away, so planting unconditionally burns the
20-region budget on wakes that cannot happen. This was a real field bug: a 1205 m tripwire planted
for a polygon 1.7 km away.

**The business cap shrinks by the number of planted tripwires**, since both draw from the same
20-region OS budget.

**Both monitors now treat tripwires as internal regions.** Three places assumed otherwise:
registration assumed a region with no usable fix was outside (a tripwire is on the device, so it
must assume inside), the contradiction gate vetted tripwire events and could refuse the very EXIT
they exist to deliver, and that EXIT dispatched on a possibly-stale cached fix rather than a
freshened one.

### Testing

Unit tests over plant / move / retire, the stale-fix case above, the containment gate, and
registration composition. Neither monitor is unit-testable without a host, so the identifier is
what is pinned there.

The composition rules are also checked by an exhaustive model (18,398 compositions: polygon count ×
every inside/outside pattern × circle count × business cap). The shipping design violates nothing;
ungated planting violates in 7,974 cases, and dropping the business-cap adjustment overflows the
budget to 39 regions in the saturated case.

### Stack

1. `mbl-2290-a-polygon-geometry-kernel` — geometry kernel (#1239)
6. `mbl-2291-d-polygon-tripwire-state` — tripwire state (#1247)
7. `mbl-2291-e-polygon-tripwire-planting` — tripwire planting + budget ← **this PR**

### DIFF
--- a/Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator+InitialEnter.swift
+++ b/Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator+InitialEnter.swift
@@ -56,6 +56,11 @@ extension GeofenceSyncCoordinatorImpl {
     /// site rather than stored: the resolver is a `@MainActor` singleton and this coordinator is
     /// one of its own dependencies, so a stored reference back would make the two initialize each
     /// other.
+    ///
+    /// The tripwire this plants may not reach the OS on this pass — its re-registration needs the
+    /// refresh gate, which the sync calling us still holds. It is persisted either way, so the next
+    /// sync carries it. A device standing still has no next sync, which is what foreground
+    /// evaluation is for.
     private func evaluateNewPolygons(_ polygons: [Geofence], expectedUserId: String) {
         Task { @MainActor [contextStore] in
             let resolver = DIGraphShared.shared.polygonMembershipResolver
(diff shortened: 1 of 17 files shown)


mrehan27:
Closing — the per-polygon tripwire is replaced by a dynamically sized movement circle in #1250.

The approach came from @mahmoud's review: instead of planting one region per active polygon, the single movement trigger shrinks to the nearest polygon boundary. One region instead of N, and it removes the problem this pair was working around rather than mitigating it — Bugbot's finding here was that tripwires eat business slots, and #1248 answered by reserving slots. #1250 makes the contention not exist, so `maxBusinessGeofences` is honoured exactly.

Not merging these first: reviewing code that is about to be deleted wastes reviewer time, and the redesign is additive on top of #1246 rather than a rewrite of these diffs.

Carried forward into #1250:
- the fresh-fix requirement on wake evaluation (originally #1248) — the simulator found it was silently unwired and swallowing a polygon ENTER
- `GeofenceInternalIdentifier` internal-region handling, which is unrelated to tripwire planting and stays in #1246
- the 100 m floor and its rationale, now on the movement trigger

Branches are kept, so anything else needed can be recovered from them.

PR #1249, opened 2026-09-02
chore(geofence): evaluate polygons crossing the antimeridian correctly

mrehan27:
Part of [MBL-2290](https://linear.app/customerio/issue/MBL-2290)

Found by @mahmoud in review of the polygon stack.

## The bug

`PolygonRegion` picked its projection reference by averaging vertex longitudes. A ring spanning the antimeridian holds values near both `+180` and `-180`, and their mean is `0` — Greenwich. An 800 m fence then projects a hemisphere wide and every verdict inverts.

Measured on a Taveuni (Fiji) ring, against the identical shape moved away from the wrap:

| | `contains` | `signedEdgeDistance` |
|---|---|---|
| on the antimeridian | `false` | `-4256.8 m` |
| same shape at Greenwich | `true` | `+4256.8 m` |

Exact sign flip. `-4256 m` is far past the ambiguity margin, so the gate records `.outside` with full confidence and logs no `geofencePolygonUndecided` — **the fence silently never fires and nothing in the logs says why**.

Nothing upstream catches it: the payload is valid, and the wrap only bites once the ring is projected.

## The fix

Unwrap longitudes into ±180° of the first vertex before anything averages them, and unwrap the query point onto **the ring's** line rather than its own — a fix at `-179.99` belongs beside a ring unwrapped to `+180.01`, not a full turn from it.

+33 / −5 production lines.

## Tests

Two properties rather than one fixture:

- **Translation invariance** — the antimeridian ring versus its twin at Greenwich; verdicts and distances must agree to 0.001 m across three sample points.
- **Rotation invariance** — the unwrap's reference is the first vertex, so rotating the ring must describe the same fence.

**Negative control:** with the unwrap disabled both tests fail, including a case reporting `contains=true` for a point that is outside — a false ENTER, which is worse than the miss above.

## A second antimeridian defect, found by parity audit

Same meridian, different mechanism, found while auditing Android's canonicalisation rather than in review.

A ring may legally close with the **opposite sign** to the one it opened with — `+180` and `-180` name one meridian and GeoJSON permits either. Positions were compared as raw values, so those two did not collapse. The closing vertex survived canonicalisation as a zero-length edge, `selfIntersects` read that as a crossing, and `init?(validating:)` returned nil — the region dropped at decode and **the fence never registered at all**. Android canonicalises this away, so the same payload worked there and silently did not here.

Confirmed by test before fixing, which exposed a second problem in the same output: the two initialisers **disagreed**. The cheap `init?(vertices:)` accepted the five-vertex ring with its zero-length edge; only the validating one rejected it. Anything rebuilt from cache through the hot path would have evaluated a degenerate ring.

Positions now compare equal when latitudes match and longitudes name the same meridian, reusing the `unwrapLongitude` the projection already depends on.

Four tests: the opposite-sign closure collapses to four vertices on **both** initialisers, a same-sign closure still collapses, two genuinely distinct positions either side of the meridian are not collapsed, and an **out-of-range longitude is rejected rather than collapsed away**. **Negative controls:** without the collapse the opposite-sign test fails while its controls pass; with the validity guard below the collapse instead of above it, the out-of-range test fails.

That last one is the order the guards run in. `360` unwraps onto `0`, so testing "same meridian" before validating would let an invalid position be silently dropped and the ring build as if the server had never sent it. The validity check therefore runs first, and `PolygonRegion.init?(validating:)` now states the invariant it rests on: it is the only writer of `Geofence.vertices`, so a ring from any other source has not been through these checks.

## Cross-SDK

The projection is a contract shared with Android, so the fixtures carry a new **`square400_dateline`**: the same meter-space square synthesized around an antimeridian anchor, with every expected value identical to `square400` by construction. An implementation with this bug fails the twin while the original still passes. Android needs the same fix to stay in parity.

## Termination

The walk itself now refuses a non-finite or out-of-range longitude. Both callers already reject
those (#1239 range-checks ring positions, the resolver gates the fix), so behaviour is unchanged —
it makes termination a property of the walk rather than of its callers. Past roughly 3.2e18,
subtracting 360 stops changing a `Double` and the loop would never end.

## Considered and not fixed

`GeofenceRegionMonitoring.swift:46` compares region centres with `abs(center.longitude - self.center.longitude)`, which at the wrap reads two points ~20 m apart as ~360 apart. Not reachable: both sides of that comparison derive from the same stored server value, which never flips sign. Left alone rather than widening scope.

## Stack

6. #1249 `mbl-2290-c` antimeridian fix ← **this PR**

Based on `mbl-2291-c` rather than the geometry kernel.

482/482 `LocationGeofenceTests` pass on iOS 26. Format and lint clean.

### DIFF
--- a/Sources/LocationGeofence/Geometry/PolygonRegion.swift
+++ b/Sources/LocationGeofence/Geometry/PolygonRegion.swift
@@ -187,7 +187,12 @@ struct PolygonRegion {
         }
     }
 
+    /// The walk is bounded by its inputs: `init?(vertices:)` refuses out-of-range ring positions and
+    /// `evaluate` refuses an invalid fix, so both sides arrive inside ±180°. The guard keeps that a
+    /// property of this function rather than of its callers — a value far enough out of range stops
+    /// changing when 360 is subtracted from it, and the loop would never end.
     private static func unwrapLongitude(_ longitude: Double, near reference: Double) -> Double {
+        guard longitude.isFinite, abs(longitude) <= 360, reference.isFinite else { return longitude }
         var value = longitude
         while value - reference > 180 {
             value -= 360
--- a/Sources/LocationGeofence/Geometry/PolygonRegion.swift
+++ b/Sources/LocationGeofence/Geometry/PolygonRegion.swift
@@ -62,12 +62,15 @@ struct PolygonRegion {
         }
         guard open.count >= 3 else { return nil }
 
-        let lat0 = open.map(\.latitude).reduce(0, +) / Double(open.count)
-        let lon0 = open.map(\.longitude).reduce(0, +) / Double(open.count)
+        // Unwrap before averaging: a ring spanning the antimeridian holds values near both +180
+        // and -180, whose mean is Greenwich, and the fence then projects a hemisphere wide.
+        let unwrapped = Self.unwrapLongitudes(open)
+        let lat0 = unwrapped.map(\.latitude).reduce(0, +) / Double(unwrapped.count)
+        let lon0 = unwrapped.map(\.longitude).reduce(0, +) / Double(unwrapped.count)
         let latitudeRadians = lat0 * Self.degreesToRadians
         let longitudeRadians = lon0 * Self.degreesToRadians
         let cosLatitude = cos(latitudeRadians)
-        let planar = open.map {
+        let planar = unwrapped.map {
             Self.project(
                 $0,
                 referenceLatitudeRadians: latitudeRadians,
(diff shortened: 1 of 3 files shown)


Shahroz16 (commented):
The antimeridian changes look good, but this still inherits the [resolver issue in #1245](https://github.com/customerio/customerio-ios/pull/1245#discussion_r3950915974), so I'm holding approval. CI also timed out in the Common synchronization and Gist queue tests; I haven't tied those failures to this patch. Could we rerun the failed tests and investigate if they repeat?

Shahroz16 (commented):
The antimeridian change looks good, including equivalent +180/-180 positions and the geometry regressions. I found no issue in this patch itself, but am holding approval while the [shared resolver write-boundary issue in #1245](https://github.com/customerio/customerio-ios/pull/1245#discussion_r3950915974) remains in the parent stack.

Shahroz16 (commented):
The antimeridian change is unchanged by the restack, and the focused geometry tests pass. I found no additional issue in that change itself, but this still inherits the [covering-circle write race in #1245](https://github.com/customerio/customerio-ios/pull/1245#discussion_r3969425638) and [event-circle provenance issue in #1246](https://github.com/customerio/customerio-ios/pull/1246#discussion_r3969426080), so I am holding approval for those shared paths.

Shahroz16 (commented):
The rebase leaves this PR's own geometry changes unchanged, and I found no new issue in that delta. Approval is still held on the inherited event-attribution issue from #1246. Its proposed fix in #1262 has a remaining queued-registration case that can drop a genuine exit, tracked [here](https://github.com/customerio/customerio-ios/pull/1262#discussion_r3973131951).

Shahroz16 (approved):
The geometry changes remain unchanged from the earlier review. I've verified the event-attribution fix in follow-up #1262 at 4f9c3ba7, which clears the shared hold. This looks ready with that fix included in the stack.

PR #1253, opened 2026-09-05
chore(geofence): record the fetched fence catalog, and stamp each log file with its own header

mahmoud-elmorabea:
iOS half of the pair with customerio-android#842. Two diagnostics follow-ups plus a flake fix, all found while using the sink on a real drive.

### 1. Record the fetched fence catalog

A capture named fences only by opaque id. Nobody reading a log could tell which geoset a crossing belonged to, and a replay could not place the circles at all — the scenario transform emitted `fixture.api.fetch body:null` because the geometry is nowhere in the log.

`GeofenceApiRegion` already carries `name`, `latitude`, `longitude`, `radius` and `geosetIds`, and we discarded all of it. Each successful fetch now emits one record per fence, behind the existing `CIOGeofenceDiagnostics` gate:

```
ev=fence.cataloged io=in id=11125 name=Momo_Dubai_Test gs=4471 lat=25.10967 lon=55.18406 rad=150 tt=enter,exit
```

Re-fetching the geometry from the workspace at replay time is **not** equivalent: fences move, so a drive replayed months later would silently run against today's circles rather than the ones it crossed — and a capture from a customer or an opted-in tester has no workspace to query.

Names are sanitized like any other value; a workspace-authored name can contain spaces, commas and `=`, each of which would break the parser's split. `gs` and `tt` join the composed-key set alongside `ids`, since their commas are structure rather than untrusted input.

Gated **whole** rather than gated-tail, and so deliberately absent from the `invocations` table that asserts the gated-tail contract — `proseHalf_expectIdenticalWhicheverWayTheGateIsSet` catches the difference immediately. These records carry no prose worth emitting on their own, and moving the detail into the prose to satisfy that table would leak coordinates with the gate off. Their own tests pin the same guarantees.

### 2. Rebuild the file header per file

The header was built once at launch and re-written verbatim on every open, so a process outliving a day stamped each rolled-over file with the original session's timestamp — and, since the iOS header carries `tz`, the original timezone.

That second part is reachable *because of* the recent timezone fix: a zone change now correctly opens a new file named for the new zone, and until this change stamped it with the old one.

Observed on a device: four consecutive daily files, every one reading `ts=2026-09-02T23:53:44.454`.

### 3. Fix a test flake introduced in #1243

`GeofenceDiagnostics.overrideForTesting` is process-global, and swift-testing's `.serialized` orders tests *within* a suite, not across them. `GeofenceMonitorOwnershipTests` forcing the gate on overlapped `GeofenceLogTailTests` asserting it was off — `everyRecord_givenDiagnosticsOff_expectNoDiagnosticKeyAnywhere` failed roughly one run in four with 83 issues. **This flake is on `main` today.** Both suites now take a shared lock around any body that touches the gate.

### Verification

- 374 geofence tests pass, up from 369
- swiftformat and `swiftlint --strict` clean on every changed file
- APN-UIKit builds
- End to end: injecting the new records into the real Dubai drive capture turns `fixture.api.fetch body:null` into the full geometry for both fetches

### DIFF
--- a/Apps/APN-UIKit/APN UIKit/Diagnostics/DiagnosticLog.swift
+++ b/Apps/APN-UIKit/APN UIKit/Diagnostics/DiagnosticLog.swift
@@ -66,7 +66,7 @@ final class DiagnosticLog: @unchecked Sendable {
         startMono = DiagnosticClock.monotonicNanos()
 
         // The SDK's diagnostic tail is enabled by the CIOGeofenceDiagnostics Info.plist key.
-        writer.open(header: fileHeaderLine())
+        writer.open { [weak self] in self?.fileHeaderLine() ?? "" }
         deviceState.start(onChange: makeDeviceStateHandler())
         installDispatcher()
 
--- a/Apps/APN-UIKit/APN UIKit/Diagnostics/DiagnosticLog.swift
+++ b/Apps/APN-UIKit/APN UIKit/Diagnostics/DiagnosticLog.swift
@@ -66,7 +66,7 @@ final class DiagnosticLog: @unchecked Sendable {
         startMono = DiagnosticClock.monotonicNanos()
 
         // The SDK's diagnostic tail is enabled by the CIOGeofenceDiagnostics Info.plist key.
-        writer.open { [weak self] in self?.fileHeaderLine() ?? "" }
+        openWriter()
         deviceState.start(onChange: makeDeviceStateHandler())
         installDispatcher()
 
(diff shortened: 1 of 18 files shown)


mrehan27 (approved):
Cross-repo with customerio-android#842. Splitting `delivery.*` out is the important half — logging only on delivery success meant an offline device read as though it had missed crossings it actually handled correctly.

Checked every non-logger file for behaviour change and found none: the `guard case .deliver` paths and the `return logger.geofenceInfo(...)` sites keep the original control flow.

Issue #17, opened 2021-08-18
Make US region default region


levibostian:
In the initialization of the SDK, make the US region the default. 

# Why are we doing this for our customers? 

1. Most customers today use the US region. 
2. Making the US region the default is consistent with our other SDKs. 
3. In the future, we may remove the requirement to provide to us a Region anyway (as the SDK will just [look up the region that you have set for your workspace](https://customer.io/docs/api/#operation/getRegion)). Therefore, set a default to make migration easier in the future. 

# How do we test that this change is successful? 

# How are we building this? 

- [x] Change `CustomerIO` class `public init()` and `static func initialize()` to have a default parameter: `..., region: Region = Region.US)`

PR #1255, opened 2026-09-06
ci: run iOS lint on Linux instead of a macOS runner

mahmoud-elmorabea:
Points Lint at customerio/mobile-ci-tools#17. Lint needs no Xcode yet holds one of the org's five macOS slots.

Compared against the macOS job on the same commit: same files formatted, same autocorrections, same clean diff, 0 violations. 1m19s against 23s in-job, versus a median 22 minute wait for a slot.

Flip to `@main` once #17 merges. `customerio-ios-fcm` follows.

The two red test checks are an unrelated known flake; this PR only touches a lint workflow.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -11,4 +11,12 @@ concurrency:
 
 jobs:
   lint:
+    # Points at the branch, not @main, while we prove the Linux runner reports exactly what the
+    # macOS runner does. Flip to ios-lint-linux.yml@main once mobile-ci-tools#<PR> merges.
+    uses: customerio/mobile-ci-tools/.github/workflows/ios-lint-linux.yml@ci/ios-lint-ubuntu
+
+  # TEMPORARY, delete before this PR leaves draft. Runs today's macOS lint on the same commit
+  # so the two jobs can be compared side by side. They are expected to agree on everything:
+  # the same files formatted, the same autocorrections, and no net diff.
+  lint-macos-baseline:
     uses: customerio/mobile-ci-tools/.github/workflows/ios-lint.yml@main
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -11,12 +11,5 @@ concurrency:
 
 jobs:
   lint:
-    # Points at the branch, not @main, while we prove the Linux runner reports exactly what the
-    # macOS runner does. Flip to ios-lint-linux.yml@main once mobile-ci-tools#<PR> merges.
+    # Branch pin, not @main, until mobile-ci-tools#17 merges. Flip it then.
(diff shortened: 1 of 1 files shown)


Shahroz16 (approved):
The Linux run checks all 756 files with the pinned tools and passes. Please merge customerio/mobile-ci-tools#17 and switch this reference to `@main` before merging this PR.

.github/workflows/lint.yml
```diff
@@ -11,4 +11,4 @@ concurrency:
 
 jobs:
   lint:
-    uses: customerio/mobile-ci-tools/.github/workflows/ios-lint.yml@main
+    uses: customerio/mobile-ci-tools/.github/workflows/ios-lint-linux.yml@main
```
Shahroz16, line 14: Thanks for switching this to the merged workflow. Could we resolve the Linux formatter failure before merging? The [latest native run](https://github.com/customerio/customerio-ios/actions/runs/34888553119/job/104125179502) exits 139 in `swiftformat . --swiftversion 5.3` after installing the pinned 0.61.1 binary, so formatting never completes and lint is skipped. The earlier passing run does not validate this head. We need a complete green format/lint run with these tools, or to keep this caller on macOS until that works.

Shahroz16 (commented):
The shared workflow is merged and this now points to main, but I cannot renew approval while the migrated lint job itself fails on this head. The current failure is in formatting, not the unrelated tests mentioned in the description.

mahmoud-elmorabea:
**Closing. The migration works; SwiftFormat's Linux build isn't reliable enough to gate on.**

The lint job itself was verified against the macOS job on the same commit: same files formatted, same autocorrections, same clean diff, zero violations. That part is sound.

The blocker is upstream. SwiftFormat's Linux binary segfaults at random on this tree, measured on real x86_64 runners:

| Build | Failures |
| --- | --- |
| 0.61.1 (pinned) | 4 in 71 attempts |
| 0.63.0 | 2 in 20 attempts |
| macOS, same version and tree | 0 in 15 |

Independent of the tree, the container, the source revision and `RLIMIT_STACK`.

Cause: the official Linux builds are statically linked against musl, whose default thread stack is 128 KB. Upstream diagnosed this in nicklockwood/SwiftFormat#2669 and shipped a fix in 0.63.0, but it sets `GNU_STACK`, which sizes only the main thread. SwiftFormat formats files on pthread workers, which still get musl's default. Our measurements match that prediction, and it explains why single files never crash while whole directories do.

A bounded retry does work. Over 30 rounds it absorbed the one crash that occurred and converged every time, with three attempts projecting to roughly one failure in 5,600 runs. Deferred rather than adopted: a standing workaround for someone else's crash sits awkwardly in a gate that is meant to be trusted.

Worth revisiting once upstream sets a worker-thread stack size. `ios-lint-linux.yml` is merged in mobile-ci-tools and stays there ready for that; both repos remain on the macOS lint in the meantime.

PR #1258, opened 2026-09-08
fix(geofence): refuse re-delivered CLMonitor events, and put the wrapper behind OS seams

mahmoud-elmorabea:
Trimmed to what the recorded-drive replay harness actually drives, plus one behaviour fix. Replaces the earlier pipeline/ledger/fix-requester extractions, which no test substituted.

### Behaviour fix: re-delivered CLMonitor events

On device, CoreLocation delivers the same crossing event two to three times: identical identifier, state and `date`, never on the simulator. The wrapper's only defence was the stored per-region baseline. For the movement trigger, the first copy's own movement pass re-seeds that baseline, so a copy arriving after the re-seed read as a new exit and started a second movement pass and sync. Recorded drives won that race by 4–9 ms every time; #1268 has the test that fails on `main`.

The fix is two arguments on the storage call the wrapper already makes: the OS event date as the evidence bound the baseline heal already uses, and as the baseline's own stamp so the comparison stays in one clock. A stale copy is refused as `newer_baseline`; a prompt one still reads as `no_state_change`. Replaying the recorded drives changes no output.

### Seams

- `GeofenceOSSeams`: protocols for the condition monitor and the location authority, with the live CoreLocation adapters. The `CLMonitorGeofenceMonitor` init takes them plus a clock; every `Date()` in its extensions reads the clock. Movement-fix requests stay on the pre-existing closure.
- `GeofenceRefreshTrigger`: the sync-trigger decisions leave `GeofenceModuleState`, which reached them only through `DIGraphShared`.
- `GeofenceStorage` and `MovementFixResolver` take the clock for age measurement.

### Logging contract

Every geofence record states `io=in|out|obs`. Outputs are exactly `registration.applied`, `transition.accepted` and `module.reset`; decisions about inputs (drops, suppressions, synthesised enters) are observations. New inputs: `identity.changed`, `permission.changed`, `location.fix` at every cache read. Callbacks carry the OS event date; timing fields log at full precision. `GeofenceDiagnostics.overrideForTesting` is a `@TaskLocal`.

### Not here

The replay harness and its scenario tests. This is the SDK half that makes them possible. `LocationGeofenceTests`: 387 passed.

Fixes MBL-2309


mahmoud-elmorabea:
Superseded by the stack on `testing/geofencing-replay-harness`: #1269 (duplicate-callback fix) → #1270 (log io contract) → #1271 (OS seams and clock) → #1272 (refresh trigger). Same content, split by idea and with narrative comments stripped.

PR #1259, opened 2026-09-09
chore(geofence): keep polygon belief across an OS eviction

mrehan27:

When CLMonitor reports a polygon's covering circle unmonitored, the circle baseline is reseeded and — until this PR — the polygon belief was dropped with it, on both the `.unmonitored` path and the `forceReseed` re-registration behind it.

**This reverses part of what @mahmoud-elmorabea asked for in the #1244 review, deliberately.** That request weighed one of three cases; the other two were not on the table then, and one of them only became reachable once #1250 started re-evaluating polygons after registration. The table below is the whole argument — please push back on it rather than on the diff.

## Why the belief has to survive

Whether the device crossed during the gap is **unknowable**: the OS was not watching. Every rule is therefore a guess, and dropping guesses "it left".

| device during the gap | drop (before) | keep (this PR) |
|---|---|---|
| stayed inside | looks brand new → **duplicate enter** | no-change, silent ✓ |
| left, stayed out | create path → **exit never delivered** | **exit delivered** ✓ |
| left and returned | correct enter | both edges missed |

Dropping is wrong in the two common columns. The asymmetry that makes it bite is that a *brand-new* polygon found inside deliberately delivers an enter — the polygon counterpart of enter-when-inside — so a discarded belief turns a device that never moved into a second enter. Circles don't have this: no baseline plus an event is `.suppressedNoBaseline`, silent.

**Keeping never produces a false event; dropping does.** The column it gives up degrades to a *missed* event, and it is the rarest — it needs the device to leave *and* return with no evaluation pass in between.

## The belief is not stale while the OS is away

An eviction touches monitor-local state only. It never changes `monitoredGeofenceIds`, and `evaluateAllPolygons` filters on the registered set rather than on OS monitoring state — so **movement-trigger wakes and foreground passes keep evaluating the polygon with real fixes throughout the gap**. The belief being kept may be seconds old, not a guess left over from before the eviction. This is also what narrows the lost column: it requires no pass to have run between the two crossings, meaning the device moved less than the wake radius and the app never came to the foreground.

## Tests

518/518 `LocationGeofenceTests` pass on iOS 26. Lint and format clean. +7 / −15 production lines.

Storage covers each column and the changed guard shape; two resolver tests drive the real sequence end to end and assert what a customer receives.

The resolver tests originally exercised only half of an eviction: their helper wrote no monitor record, so `clearMonitorRegionRecord` found nothing to remove and returned early. It seeds one now — reverting only the clear's belief drop fails both resolver tests, where before it changed nothing there. Raised by @Shahroz16 in review.

**Negative-controlled:** restoring either drop fails only that site's tests, and the resolver case fails *by emitting the duplicate enter* rather than by a bare assertion.

Two of the new tests were vacuous on the first attempt and the fix is worth knowing about: the harness builds its fix before the test writes its belief, so the fix predated the belief, the write was refused as `.suppressedNewerDecision`, and `emitter.isEmpty` passed because nothing executed. The stayed-inside case now also asserts the evidence stamp advanced, which an unexecuted pass cannot satisfy.

## Not verified

- **Strict concurrency** could not be run: `Common/Util/UIKitWrapper.swift:34` errors under `-strict-concurrency=complete` and `Common` is a dependency of every module, so nothing downstream compiles. `GeofenceStorage` is an `actor` and this change only removes state mutation.
- **No device or simulator run** — reproducing this needs a real CLMonitor budget eviction.

## Stack

8. #1259 `mbl-2291-h` belief survives eviction ← **this PR**

### DIFF
--- a/Tests/LocationGeofence/Geofence/Polygon/PolygonMembershipResolverTests.swift
+++ b/Tests/LocationGeofence/Geofence/Polygon/PolygonMembershipResolverTests.swift
@@ -1164,10 +1164,17 @@ struct PolygonMembershipResolverTests {
     /// What the OS reporting a covering circle unmonitored actually does to storage: the deferred
     /// clear, then the next registration reseeding the circle baseline.
     private func evictCoveringCircle(_ setup: Setup, id: String) async {
+        let center = LocationData(latitude: 0, longitude: 0)
+        // Seeded first: the clear only removes a monitor record that exists, so without this the
+        // `.unmonitored` half is a no-op and only the reseed would be under test.
+        await setup.storage.recordMonitorRegistration(
+            identifier: id, transitionTypes: [.enter, .exit], initialState: .enter,
+            center: center, radius: 300
+        )
         await setup.storage.clearMonitorRegionRecord(identifier: id)
         await setup.storage.recordMonitorRegistration(
             identifier: id, transitionTypes: [.enter, .exit], initialState: .exit,
-            center: LocationData(latitude: 0, longitude: 0), radius: 300, forceReseed: true
+            center: center, radius: 300, forceReseed: true
         )
     }
 }
--- a/Tests/LocationGeofence/Geofence/Polygon/PolygonMembershipResolverTests.swift
+++ b/Tests/LocationGeofence/Geofence/Polygon/PolygonMembershipResolverTests.swift
@@ -1116,4 +1116,58 @@ struct PolygonMembershipResolverTests {
         #expect(await setup.emitter.snapshot().map(\.transition) == [.enter])
         #expect(await setup.storage.getPolygonMembership()["1"]?.membership == .inside)
     }
+
+    // MARK: - Belief across an OS eviction
+
+    /// The eviction sequence end to end: the `.unmonitored` clear, then the re-registration that
+    /// reseeds the circle baseline, then a pass. The device never left, so the surviving belief
+    /// makes this a no-change and the customer gets no second enter for a visit already reported.
+    @Test
+    func evaluateAllPolygons_givenEvictionWhileStillInside_expectNoDuplicateEnter() async {
+        let setup = await makeSetup(fix: fix(latitude: 0, longitude: 0))
+        await registerPolygons(setup, ids: ["1"])
+        // Dated behind the fix: `makeSetup` builds the CLLocation first, so a belief stamped now
+        // would postdate it and the pass would be refused as a newer decision without evaluating.
+        let before = Date().addingTimeInterval(-60)
+        _ = await setup.storage.recordPolygonMembership(
+            .inside, forIdentifier: "1", onlyIfBeliefPredates: before
+        )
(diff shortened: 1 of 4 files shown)


Tests/LocationGeofence/Geofence/Polygon/PolygonMembershipResolverTests.swift
```diff
@@ -1062,4 +1062,58 @@ struct PolygonMembershipResolverTests {
...
+        await setup.resolver.evaluateAllPolygons()
+
+        #expect(await setup.emitter.snapshot().map(\.transition) == [.exit])
+        #expect(await setup.storage.getPolygonMembership()["1"]?.membership == .outside)
+    }
+
+    /// What the OS reporting a covering circle unmonitored actually does to storage: the deferred
+    /// clear, then the next registration reseeding the circle baseline.
+    private func evictCoveringCircle(_ setup: Setup, id: String) async {
+        await setup.storage.clearMonitorRegionRecord(identifier: id)
```
Shahroz16, line 1174: Could we seed a monitor registration before clearing it here? `registerPolygons` only writes the registered IDs and cache, so this clear returns without removing anything. Adding that setup would make these resolver tests cover the actual eviction sequence as well as the force-reseed path. The storage test already covers removal, so this is a non-blocking test improvement.

Shahroz16 (commented):
Keeping polygon belief across eviction looks sound, and all 67 focused storage/resolver tests passed locally. Registration removal and sign-out still clear the belief. The inline test-setup suggestion is non-blocking; I am holding approval only because this head inherits the [covering-circle write race in #1245](https://github.com/customerio/customerio-ios/pull/1245#discussion_r3969425638) and [event-circle provenance issue in #1246](https://github.com/customerio/customerio-ios/pull/1246#discussion_r3969426080).

Tests/LocationGeofence/Geofence/Polygon/PolygonMembershipResolverTests.swift
mrehan27: Fixed in `310987fe`. The helper now seeds a monitor registration before the clear, so the `.unmonitored` half is actually exercised instead of returning early — reverting the clear's belief drop now fails both resolver tests, where before it changed nothing there.

Shahroz16 (commented):
Thanks for seeding the monitor record before eviction. That closes the earlier test-setup suggestion, and all 71 focused resolver and storage tests passed on the simulator. I found no new issue in this delta. Approval is still held on the inherited event-attribution issue from #1246, with the remaining case in its proposed #1262 fix tracked [here](https://github.com/customerio/customerio-ios/pull/1262#discussion_r3973131951).

Shahroz16 (approved):
The eviction tests now seed the monitor record and pass. I've also verified the shared event-attribution fix in follow-up #1262 at 4f9c3ba7, so this no longer needs to wait on that concern. Looks ready with the follow-up included in the stack.

PR #1260, opened 2026-09-09
test: bump deadlock detection timeout from 30s to 60s (MBL-1706)

sydneycollins-cio:
Bumps the timeout for the concurrent deadlock detection test from 30s to 60s. The test spawns 2000 concurrent DispatchGroup operations to verify the Synchronized equality operator doesn't deadlock. On shared CI runners under load, these can exceed 30s and cause false positives. A genuine deadlock still trips 60s.

This test failed ~25 times in 77 runs since Jul 14, most recent on main Sep 9. We'll compare the next 2 weeks against the ~3 failures/week baseline. If flakes persist, fallback is reducing iterations to 500.

[MBL-1706](https://linear.app/customerio/issue/MBL-1706)

### DIFF
--- a/Tests/Common/Synchronized/SynchronizedTests.swift
+++ b/Tests/Common/Synchronized/SynchronizedTests.swift
@@ -238,12 +238,9 @@ struct SynchronizedEquatableTests {
     }
 
     // Calls a == b and b == a concurrently. Without the ObjectIdentifier lock-ordering in ==,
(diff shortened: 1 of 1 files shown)


Shahroz16 (approved):
The timeout remains bounded and the opposing lock-order exercise is unchanged. All five focused equality tests passed locally, including the concurrent case, and CI is green. The planned follow-up will tell us whether this reduces the flake rate.

PR #1261, opened 2026-09-09
test: remove real-network waits from GistQueueNetworkTest

sydneycollins-cio:
Removes the remaining waitForExpectations(timeout: 5.0) calls from GistQueueNetworkTest. These tests verify synchronous identifier-validation logic, but were waiting for incidental URLSession round-trips that added CI flakiness without testing anything extra.

This test class flaked 9 times in 8 weeks on CI due to network latency. Tests no longer wait on real networking; synchronous validation assertions remain intact. Note: BaseNetwork still initiates URLSession requests under the hood. Stubbing the network layer is a follow-up outside this PR.

[MBL-2407](https://linear.app/customerio/issue/MBL-2407)

### DIFF
--- a/Tests/MessagingInApp/Gist/Network/GistQueueNetworkTest.swift
+++ b/Tests/MessagingInApp/Gist/Network/GistQueueNetworkTest.swift
@@ -14,6 +14,9 @@ class GistQueueNetworkTest: UnitTest {
 
     // MARK: - User Identifier Validation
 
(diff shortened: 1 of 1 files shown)


Shahroz16 (approved):
Removing the unused network waits leaves the synchronous validation assertions intact. All seven focused tests passed locally and CI is green. Could we adjust the PR description to say the tests no longer wait for real networking? BaseNetwork still starts URLSession requests; stubbing those can stay a separate follow-up.

PR #1262, opened 2026-09-09
chore(geofence): attribute a covering-circle event to the circle it was raised against

mrehan27:

#1245 refuses a covering-circle exit whose circle the fence no longer has, and #1246 supplies that circle from each monitor. The guard was inert. This makes it work.

## What was wrong

Three defects, each breaking attribution differently.

**The previous generation was never populated.** `previousConditions` had one writer: the branch that registers over a live entry. `setMonitoredRegions` calls `releaseOwnership` first, which removed the entry, so the re-registration found nothing to supersede. On the one path a geometry change actually takes, every stale event resolved to nil, and nil is treated as current.

**Attribution used the staging time.** Registration records geometry synchronously, but the OS keeps evaluating the old circle until the queued remove+add drains. An event raised in that window postdates the new registration and still belongs to the old circle. `+ContradictionGate` already documents this gap, which is why it stamps `ConditionReadd` at drain time.

**A second staging discarded the live circle.** Found by @Shahroz16. Live A, stage B, then stage A again before B's add drains: the second retire replaced previous-A with never-live B. An exit the OS raised against A was labelled B, the atomic circle guard rejected it against current A, and membership stayed inside — a genuine exit dropped.

## What it does now

The ledger tracks what the OS is evaluating, not what was last asked for. Per identifier:

- `staged` — the latest registration, which geometry diffs compare against. Cleared by `retire`.
- `queued` — staged and awaiting their own OS add, FIFO, because `enqueueMonitorOperation` chains each operation behind the last.
- `live` and `previouslyLive` — what the OS evaluates now, and what it evaluated before.

`confirm(stagedAt:)` promotes the generation whose staging time matches, so the circle that goes live is the oldest queued, not the newest staged. Attribution reads the live generations only. Adoption stamps `liveFrom: .distantPast`, since the OS was already evaluating that circle before this process existed.

Two boundary changes: after a bare `retire`, and after `startMonitoring` bails on blocked permission or invalid coordinates, attribution returns the live circle where it used to return nil. The OS still holds that circle until the queued removal drains, so the comparison is now made instead of skipped.

## Why a new type

`CLMonitorGeofenceMonitor` builds a `CLLocationManager` and cannot be created in a unit test. The previous tests handed the selection its inputs directly, so they could not observe that production never produced them. That test shape is why the first defect shipped past review.

## Tests

538/538 `LocationGeofenceTests` on iOS 26. Lint and format clean.

Nineteen ledger tests drive sequences rather than feeding a selector: release-then-re-register, register-over-the-top, events either side of a drain, three generations draining in turn, two replacements staged before either drains, retire, adoption, forget, forgetAll, the staged/live split, and the `registeredAt`-excluded equality the baseline heal depends on.

**Negative control.** Restoring the committed ledger under the new tests fails exactly the three that pin the new behaviour, then passes again on restore. Reproduced independently in a second session.

**Review follow-up: a stale event must not fail open.** Attribution returned nil both when nothing was known for an identifier and when the event predated every generation held. A consumer reads nil as "cannot say" and takes the event as current, which skips the geometry guard on the storage write — so an exit raised against a circle two replacements ago stored `outside` for a device standing inside the polygon that replaced it, delivering a spurious exit and then a spurious enter once the next pass corrected the belief. The two cases are now separate: nothing ever live for the identifier is a cold wake and still fails open; a generation that went live and no longer covers the event is refused. Reachable because attribution resolves after the contradiction gate's fix request (10 s timeout) and the storage write, while registrations drain on their own queue.

The alternative — falling back to the previous generation unconditionally, as before this PR's Bugbot fix — was built and run too. It closes the two-generation case but not the one-generation case, where there is no previous generation to name.

Two tests separate the three behaviours, and neither is redundant. The exit test builds real generations, maps them through the production initialiser and asserts the belief survives; under the behaviour this PR shipped it fails with the customer-visible symptoms, an exit delivered and the belief flipped to `outside`. It does not separate the unconditional fall-back, which the geometry guard catches anyway once a circle is named — that one is caught by the ledger test for a single live generation.

**Review follow-up: the stamp was on the wrong side of the add.** `liveFrom` was taken after `await monitor.add` returned. The OS begins evaluating as the add lands and dates its corrective event then, so every corrective event was dated before the generation that produced it — attribution handed back the circle just replaced, the geometry guard refused it, and the crossing the staged `assuming:` snapshot exists to re-synthesize was swallowed. On every geometry change, not only the first registration. `ConditionReadd` already keeps `start` and `added` apart for this same window; the ledger's stamp ignored it. Introduced in this PR by `e565f47c`, which moved attribution off staging time. Now stamped between the remove and the add.

Expiry also requires a generation to have been replaced. With no previous generation an older event was raised against a condition the OS held before this process registered anything, whose circle may be the one in force now, so refusing it would drop a genuine crossing.

**Test determinism.** One resolver test stamped its two fixes at separate instants, so a runner slow enough to invert their order made it decide from the cache. Both now come from one instant. Reproduced by forcing the delay, and confirmed fixed the same way.

## Not verified

- **Strict concurrency** cannot be run in this repo: `Common/Util/UIKitWrapper.swift:34` errors under `-strict-concurrency=complete` and `Common` is a dependency of every module.
- **No device run.** Reproducing any of these windows needs a real CLMonitor re-registration with a queued drain.
- **The stamp placement is not executed by any test.** It is one line in `startMonitoring`, inside a class that cannot be instantiated in a unit test. The ledger tests pin the boundary it relies on — an event at exactly `liveFrom` resolves to the new generation — but that the stamp is taken before the add rather than after is verified only by a second session's probe.

## Stack

9. #1262 `mbl-2291-g` event circle generation ← **this PR**


Shahroz16 (commented):
The single-replacement cases pass, but I reproduced a remaining two-staged-replacements case that drops a genuine exit. The 82 existing focused tests passed; the added ledger/storage regression fails as described inline. I am holding approval for that generation-retention window.

Sources/LocationGeofence/Monitor/RegisteredConditionLedger.swift
```diff
@@ -0,0 +1,114 @@
...
+    mutating func confirm(_ identifier: String, stagedAt: Date, at liveFrom: Date) {
+        guard var live = current[identifier], live.registeredAt == stagedAt, live.liveFrom == nil else { return }
+        live.liveFrom = liveFrom
+        current[identifier] = live
+    }
+
+    /// Gives up the claim on a condition while keeping its circle for attribution: an event the
+    /// daemon already raised against it can still be dequeued after this.
+    mutating func retire(_ identifier: String) {
+        if let released = current.removeValue(forKey: identifier) { previous[identifier] = released }
```
Shahroz16: Could we retain the confirmed-live circle separately from pending registrations? I reproduced live A, stage B, then stage A again before B's OS add drains. The second retire replaces previous A with unconfirmed B, so an EXIT raised while the OS still monitors A is reported as B. The atomic circle guard then rejects that genuine exit against current A and leaves membership INSIDE. Keeping the live generation until the OS actually replaces it, and confirming the matching queued generation even after another has staged, with this two-staging regression would close the gap.

mrehan27: Fixed in `4f9c3ba7`, both halves. The ledger now keeps the live generation separately from queued ones: `retire` only drops the claim, so a staged circle the OS never took can no longer displace it. And `confirm(stagedAt:)` promotes the generation whose staging time matches even when a later one has staged, so with several queued the circle that goes live is the oldest queued rather than the newest staged — the OS queue is serial, so that is the order it takes them in.

Your case is pinned by `circle_givenTwoReplacementsStagedBeforeEitherDrains_expectTheCircleTheOsHolds`, and the FIFO promotion by `confirm_givenThreeStagedAndTheyDrainInTurn_expectEachBecomesLiveInOrder`. Restoring the old ledger under the new tests fails exactly those plus `retire_expectClaimDroppedAndAttributionKept`, and passes again on restore.

One behaviour change worth flagging since you traced this path: after a bare `retire`, and after `startMonitoring` bails on blocked permission or invalid coordinates, attribution now returns the live circle where it returned nil before. The OS still holds that circle until the queued removal drains, so the consumer now compares instead of accepting blindly.

Tests/LocationGeofence/Geofence/Monitor/RegisteredConditionLedgerTests.swift
```diff
@@ -0,0 +1,265 @@
...
+
+        #expect(ledger.circle(for: "1", raisedAt: drainedAt.addingTimeInterval(1))?.center.longitude == 0.005)
+    }
+
+    /// A drain promotes the generation it belongs to, not the newest staged one. Three can be
+    /// queued at once, and the queue is serial, so the circle the OS takes next is the oldest
+    /// queued — reading the newest instead attributes events to a circle the OS has not reached.
+    @Test
+    func confirm_givenThreeStagedAndTheyDrainInTurn_expectEachBecomesLiveInOrder() {
+        var ledger = RegisteredConditionLedger()
```
Shahroz16, line 174: As a follow-up, could we cover staging A, forgetting the identifier, staging B, then receiving A's old confirmation? That callback must not mark B live before B's own add drains. The keyed guard handles this today; a regression would preserve that distinction through eviction and re-registration.

Sources/LocationGeofence/Monitor/RegisteredConditionLedger.swift
Shahroz16: Thanks, both parts are addressed in 4f9c3ba7. The original two-staging regression now keeps A as the live circle and delivers its exit through the real storage guard. I also checked that each queued generation becomes live when its own add drains. This closes the reported issue.

Shahroz16 (approved):
The queued/live separation closes the reported exit-loss case, and FIFO confirmation now follows the registration that actually drains. I reran the original ledger/storage regression: it fails on the old commit and passes here. All 87 focused simulator tests pass, and the independent review is complete. The small test suggestion can follow separately. This verifies the modeled sequencing and storage behavior, not real-device CLMonitor scheduling.

Sources/LocationGeofence/Monitor/RegisteredConditionLedger.swift
```diff
@@ -0,0 +1,149 @@
...
+        if let live = entry.live, let liveFrom = live.liveFrom, raisedAt >= liveFrom {
+            return live
+        }
+        // Older than the generation before the current one too. Nil is fail-open — a consumer
+        // treats it as current and accepts the event — where naming the oldest circle held would
+        // fail its geometry guard instead. Neither is reachable while events dequeue in order, so
+        // the documented contract decides it rather than a safety argument.
+        guard let previous = entry.previouslyLive, let previousFrom = previous.liveFrom,
+              raisedAt >= previousFrom
+        else { return nil }
```
Shahroz16: Could we distinguish a known-expired generation from the cold-start nil case here? In a ledger/real-storage regression, A's exit waits while B and C finish registering under the same identifier. This now returns nil, skipping the geometry guard and storing OUTSIDE for replacement polygon C even though the device is inside it. Before this change the same case was refused. Event order doesn't exclude it: processing awaits the fix/storage paths while registrations drain independently. Retaining the needed circle, or requiring a fresh polygon verdict when attribution has expired, would avoid treating that old exit as current.

Shahroz16 (commented):
The earlier queued/live fix remains addressed, but I can't renew approval on this head because the new lower-bound fallback can accept an old exit against a replacement polygon. The added ledger/real-storage regression passes before that change and fails here; the focused Opus review also checked the counterexample. This models the interleaving, not a real-device CLMonitor race. The 88 existing focused simulator tests pass.

Separately, could we timestamp the fresh outside fix when requestFreshFix answers in evaluateNewlyRegistered_givenCachedInsideAndFreshOutside_expectOnlyTheFreshVerdict, then rerun CI? The fixture creates it before async setup, so a delay over five seconds makes it older than the cached inside fix. A six-second delay reproduces both current CI assertions; stamping the response on delivery makes all 47 resolver tests pass with that same delay. This timing issue is inherited from the base, not caused by this ledger change, and can be fixed separately.

Sources/LocationGeofence/Monitor/RegisteredConditionLedger.swift
Shahroz16: Thanks, the two-replacement false exit is addressed, and the timestamp fixture now passes with the same six-second delay. Could we narrow the new expired check so a first-ever registration doesn't expire an event raised before its add returns? confirm stamps liveFrom after await monitor.add, and ConditionReadd already accounts for event dates inside that add window. With no previous generation, I reproduced an EXIT for the same circle being classified expired, after the monitor baseline has advanced, so the resolver emits nothing and leaves membership INSIDE. Keeping this case as noneHeld until previouslyLive exists makes the added regression pass while retaining the two-replacement refusal (49 resolver tests pass with that local diagnostic change). Please cover that first-add boundary too. This is a modeled ledger/resolver/storage interleaving, not a real-device reproduction.

Shahroz16 (commented):
The previous false-exit case and fixture timing issue are addressed. I am holding renewed approval only for the first-registration attribution case described in the existing thread. Opus identified it, and the added resolver/storage regression fails on this head and passes with the narrower expiration check. All 97 existing focused simulator tests pass. The CI test rerun is green; iOS smoke is still running.

Tests/LocationGeofence/Geofence/Monitor/RegisteredConditionLedgerTests.swift
mrehan27: Added as `confirm_givenTheIdentifierWasForgottenBeforeTheDrain_expectTheReplacementStaysQueued`: stages A, forgets the id, stages B, then delivers A's confirmation, and asserts nothing is live while B stays the staged generation.

Checked it actually catches the regression — with `confirm` promoting the newest queued generation instead of the keyed one, it fails.

Sources/LocationGeofence/Monitor/RegisteredConditionLedger.swift
mrehan27: You were right, and it turned out wider than the first registration.

The root cause was where the stamp was taken. `liveFrom` was read after `await monitor.add` returned, but the OS starts evaluating as the add lands and dates its corrective event then — so every corrective event was dated before the generation that produced it. On a replacement that attributed it to the circle just replaced, the geometry guard refused it, and the crossing was lost on every geometry change, not only on a first add. `ConditionReadd` already keeps `start` and `added` apart for exactly that window; the ledger's stamp sat on the wrong side of it. Introduced in this PR by `e565f47c`, which moved attribution off staging time.

Fixed in `47be22b1` by stamping between the remove and the add. Expiry now also requires a generation to have been replaced, so your case — an event older than a first-ever add — is `noneHeld` and still treated as current.

Pinned by `attribution_givenAnEventAtTheInstantTheAddWasIssued_expectTheNewGeneration` and `attribution_givenAFirstAddAndAnEventBeforeItWasIssued_expectNoneHeld`; both fail if the boundary or the guard moves. 538/538 locally.

One gap worth stating: the stamp placement itself has no unit test. It is one line in `startMonitoring`, which cannot be instantiated in a unit test. The ledger tests pin the boundary it relies on, but that the stamp is taken before the add rather than after was verified only by a second session's probe. Noted in the description.

PR #1263, opened 2026-09-11
chore(geofence): record polygon geometry in the fence catalog

mrehan27:
Part of [MBL-2418](https://linear.app/customerio/issue/MBL-2418).

The fence catalog records `lat`/`lon`/`rad` per fence so a drive can be replayed later. A polygon carries none of those on the wire — it has `enclosingCircle` and `geometry` instead — so all three were dropped and a polygon catalogued as id, name, geoset and transitions: unplaceable. This appeared when main's diagnostics merged onto the polygon branch and started running against data they predate.

## What it does

- `sh` = `circle` | `polygon`
- `lat`/`lon`/`rad` fall back to the enclosing circle for a polygon. `rad` is the backend's radius, unpadded
- `nv` = vertex count, `ring` = outer ring as `lat_lon` pairs (SDK order; the wire is GeoJSON, reversed)
- `ring` joins `composedKeys`, otherwise sanitizing folds the pair separators into the intra-pair ones and the list parses as one run-together token

`ring` is for placement only, never a membership input: it truncates at 64, and a partial ring is still a valid-looking polygon. `nv` is authoritative so a consumer can detect truncation and refuse.

A coordinate is rendered before invalid regions are dropped, so a malformed one records as `bad` rather than 309 digits.

Field names agreed with Android, who have the same gap — theirs catalogues a polygon as a placeable circle ~1 km too large, so it fails silently where ours fails visibly.

## Tests

568/568 `LocationGeofenceTests`. Five catalog tests: polygon, circle, truncation marker with `nv` intact, ring that failed to decode (still placeable by its circle), and an absurd coordinate.

## Stack

1. #1263 `mbl-2418-a` polygon fence catalog ← **this PR**
2. #1264 `mbl-2418-b` polygon logs onto the tail convention + `GeofenceLogTail` coverage

### DIFF
--- a/Sources/LocationGeofence/Api/GeofenceApiRegion+Catalog.swift
+++ b/Sources/LocationGeofence/Api/GeofenceApiRegion+Catalog.swift
@@ -22,11 +22,11 @@ extension GeofenceApiRegion {
     /// element before joining, so it is the only intra-pair character that survives the pipeline.
     var catalogRing: [String]? {
         guard let ring = geometry?.coordinates.first, !ring.isEmpty else { return nil }
-        return ring.compactMap { position in
-            guard position.count >= 2 else { return nil }
-            // Rendered before `usableRegion` drops invalid coordinates, so a malformed payload can
-            // reach this. `%.5f` turns 1e300 into 309 digits, and a ring of those is tens of KB in
-            // one line — that the server sent garbage is the useful record, not the garbage.
+        // Rendered before `usableRegion` drops invalid coordinates, so a malformed payload reaches
+        // this. Every position is kept, bad ones as `bad`: `%.5f` turns 1e300 into 309 digits, and
+        // dropping them instead would leave `nv` counting vertices the ring does not show.
+        return ring.map { position in
+            guard position.count >= 2 else { return "bad_bad" }
             return "\(Self.catalogCoordinate(position[1], max: 90))_\(Self.catalogCoordinate(position[0], max: 180))"
         }
     }
--- a/Sources/LocationGeofence/Api/GeofenceApiRegion+Catalog.swift
+++ b/Sources/LocationGeofence/Api/GeofenceApiRegion+Catalog.swift
@@ -13,7 +13,12 @@ extension GeofenceApiRegion {
 
     var catalogRadius: Double? { radius ?? enclosingCircle?.baseRadiusM }
 
-    /// Outer ring as `lat_lon` pairs — the SDK's order, not GeoJSON's reversed one.
+    /// Outer ring as `lat_lon` pairs — the SDK's order, not GeoJSON's reversed one — and the
+    /// CANONICAL ring: consecutive duplicates collapsed and the closing vertex dropped, which is
+    /// what membership is computed against and what Android's `nv` counts. Counting the wire ring
+    /// instead would report one extra vertex for every closed GeoJSON ring, and a consumer applying
+    /// the "fewer pairs than `nv` means truncated" rule would then refuse every correct polygon
+    /// from one platform.
     ///
     /// Placement and eyeballing only, never a membership input: the field truncates, and a partial
     /// ring is still a valid-looking polygon. `nv` is the authoritative vertex count.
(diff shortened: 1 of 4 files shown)


Tests/LocationGeofence/Geofence/GeofenceLogTailTests.swift
```diff
@@ -502,6 +502,31 @@ struct GeofenceLogTailTests {
 
     // MARK: - Fence catalog
 
+    /// A ring in GeoJSON order (longitude first) and CLOSED, as the wire sends it — the closing
+    /// position repeats the first, which the catalog must drop.
+    private func catalogPolygonRegion(id: String = "22250", vertices: Int = 4) -> GeofenceApiRegion {
+        var ring = (0 ..< vertices).map { i -> [Double] in
+            [55.18 + Double(i) / 10000, 25.10 + Double(i) / 10000]
```
Shahroz16: Could we use a non-degenerate quad in the accepted-polygon fixture and keep the diagonal as a separate rejected-ring case? These points are collinear, so the kernel rejects the ring and the catalog tests exercise the fallback rather than the new `toDomain()` branch. A valid quad would pin both paths. This is a coverage follow-up, not an approval hold.

Sources/LocationGeofence/Api/GeofenceApiRegion+Catalog.swift
```diff
@@ -0,0 +1,66 @@
+import CioInternalCommon
+import Foundation
+
+/// What the fence catalog records, resolved per shape so a polygon is placeable.
+extension GeofenceApiRegion {
+    var catalogShape: String { carriesPolygonFields ? "polygon" : "circle" }
```
Shahroz16: For the diagnostics follow-up, could we derive `sh` and the circle fields from the normalized `shape` discriminator, as the mapper and Android #859 do? An explicit `shape: "circle"` with stray geometry is monitored as a circle but cataloged here as a polygon; a polygon carrying flat fields also reports those instead of its enclosing circle. The ordinary payload works, but matching the mapper would keep these mixed-field captures interpretable.

Shahroz16 (approved):
The catalog and geofence tests passed on the iOS simulator. The canonical-ring fix is present at this head, and #1264 handles the separate dropped-region logging fix. I left two non-blocking follow-ups for fixture coverage and catalog consistency.

Tests/LocationGeofence/Geofence/GeofenceLogTailTests.swift
mrehan27: Fixed in 110293fb, and you were right that it was more than coverage. The vertices were collinear, so the ring encloses no area, `PolygonRegion(validating:)` rejected it, and every catalog test fell through to the fallback — the branch that reads the kernel's list had no coverage at all.

The fixture is a regular ring now. I also added a test that discriminates: the two paths only disagree across the antimeridian, where the kernel's `samePosition` unwraps longitude and exact equality does not, so a ring closing at +180 that opened at -180 gives nv=4 through the kernel and nv=5 through the fallback. Disabling the kernel branch fails it with nv=5.

That also falsified a comment I had written claiming the two canonicalisations cannot disagree on an accepted fence. They can; it is safe only because accepted fences take the kernel path first. Corrected in the same commit.

Sources/LocationGeofence/Api/GeofenceApiRegion+Catalog.swift
mrehan27: Done in 70d5f6e7. Both cases were real: an explicit `shape: "circle"` carrying stray geometry was catalogued as a polygon, and a `shape: "polygon"` carrying flat fields reported rad=300 instead of its enclosing circle's 900 — so the catalog described a different fence from the monitored one.

`shape` is now normalized exactly as `toDomain` does it (trim, lowercase, blank as absent) and drives both the shape and the circle fields. `sh` became a closed set of four, one per mapper branch: circle, polygon, undescribed, unknown. An unrecognized shape stays groupable and the raw token is already on `registration.rejected`.

PR #1264, opened 2026-09-11
chore(geofence): put polygon records on the structured diagnostics convention

mrehan27:
Closes [MBL-2418](https://linear.app/customerio/issue/MBL-2418).

Polygon logging predates main's tail convention, so nine records emitted prose only — invisible to replay and to `GeofenceLogTail`.

## What it does

- Tails on the nine polygon records: `polygon.transition`, `.verdict`, `.undecided`, `.undelivered`, `.wake.pass`, `.pass.skipped`, `.evaluation.requested`, plus `movement.radius.chosen`. All registered in the contract table and the declared vocabulary
- Reasons are token enums, not free strings. One previously interpolated two measurements into `why=`, so no two records shared a token; `edge` and `acc` are now their own keys
- `geofenceAllRegionsDropped` → `api.fetch.unreadable`. "Payload unreadable, cache survives" is a different fact from "no fences"
- An over-limit polygon reports `registration.rejected`, not a new family — same question, same answer

## Merge artifacts fixed

- `geofenceInvalidRegionDropped` was overloaded: main's tailed version had zero callers, the untailed one had the only caller. Collapsed, keeping the tail
- `GeofenceRegionDropReason` keyed `why=` off the human sentence, so rewording it would silently change the token. Added `logToken`, matching `GeofenceSyncSkipReason`
- The OS-dispatch record keeps its place, now as `os.callback.dispatched`. Main's `os.callback.received` covers its original purpose, but for a circle fence nothing else sits between that and `transition.accepted` — without it a crossing killed by the gate or the dedup is indistinguishable from one never delivered

## Tests

584/584 `LocationGeofenceTests`, lint clean. The contract table and vocabulary set caught every record added without being registered.

## Stack

1. #1263 `mbl-2418-a` polygon fence catalog
2. #1264 `mbl-2418-b` polygon log tails ← **this PR**


Sources/LocationGeofence/Polygon/PolygonMembershipResolver.swift
```diff
@@ -319,11 +321,11 @@ final class PolygonMembershipResolver {
         guard case .deliver(let transition) = outcome,
               geofence.transitionTypes.contains(transition)
         else {
-            logger.geofencePolygonNotDelivered(identifier: geofence.id, outcome: "\(outcome)")
+            logger.geofencePolygonNotDelivered(identifier: geofence.id, reason: .outcome(outcome))
```
Shahroz16, line 322: Could we give the transition-type filter its own reason, such as `transition_type_not_registered`? For an enter-only polygon, an outside decision can reach this branch as `.deliver(.exit)`, so the record becomes `polygon.undelivered why=deliver` and hides the actual reason nothing was sent. Splitting the guard and covering the filtered-transition case would fix the diagnostic without changing delivery behavior. I am treating this as a non-blocking logging follow-up.

Shahroz16 (approved):
The geofence simulator tests passed, including the user-change cases, and the dropped-region call now reaches the structured logger. The canonical-ring fix already exists in #1263 and can come through the stack update. I left one non-blocking diagnostic-reason follow-up; it does not change transition delivery.

Sources/LocationGeofence/Polygon/PolygonMembershipResolver.swift
mrehan27: Fixed in 3fad8154, with the token you suggested. Split the guard so the transition filter reports `why=transition_type_not_registered` instead of inheriting the write outcome. Delivery is unchanged.

Confirmed the case first: an enter-only polygon decided outside reaches the boundary as `.deliver(.exit)`, so the combined guard logged `why=deliver` on a record that exists because nothing was sent. New test covers it; restoring the combined guard fails it, and the failure output shows the old `why=deliver`.

PR #1265, opened 2026-09-11
chore(geofence): stop one bad row discarding the whole pending queue

mrehan27:
Part of [MBL-2413](https://linear.app/customerio/issue/MBL-2413).

The pending geofence queue decoded as a single array, so one undecodable row discarded every other row. A read failure also returned an empty list, so the next append overwrote a queue that was intact on disk.

Rows now decode one at a time and skipped rows are counted. A read reports `unreadable` separately from empty, and `append`/`remove` refuse to write over it. `append` returns an outcome rather than a Bool, so a refusal is no longer logged as a failed write.

The state that loses the queue is a read that fails while a write still succeeds — an atomic write renames a temp file, so it needs permission on the directory, not the target. Data Protection is the suspected route on device; that part is unverified.

**Evidence:** reverting the append guard makes `append_givenAnUnreadableFile_expectRefusedAndTheQueueUntouched` fail with the queue file's bytes changed. LocationGeofenceTests 386/386, lint clean.

## Related

One ticket, one defect shape, two modules. Separate PRs because they sit on different bases and neither depends on the other.

1. #1265 geofence pending queue → `feature/geofence-polygons` ← **this PR**
2. #1266 Common push delivery store → `main`

### DIFF
--- a/Sources/LocationGeofence/Event/GeofenceEventTracker.swift
+++ b/Sources/LocationGeofence/Event/GeofenceEventTracker.swift
@@ -205,6 +205,9 @@ final class GeofenceEventTracker: @unchecked Sendable {
         cooldownKey: String
     ) async {
         switch write {
+        // Unreachable: the one call site guards on `.persisted`. Kept total rather than narrowed
+        // so the compiler still lists the arms, but it must stay a no-op only while that guard
+        // holds — reaching here would release no cooldown and log nothing.
         case .persisted: return
         case .writeFailed:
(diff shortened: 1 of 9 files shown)


Shahroz16 (approved):
The mixed-row, unreadable-file and tracker tests passed on the simulator. Preserving unreadable bytes and declining a fresh unpersisted crossing is the intended tradeoff here; treating an arbitrary read error as permission to overwrite would undo that protection. This validates the file-permission model, not Data Protection behavior on a locked device.

PR #1266, opened 2026-09-11
chore(push): stop one bad row discarding the whole pending delivery queue

mrehan27:
Part of [MBL-2413](https://linear.app/customerio/issue/MBL-2413). Same defect shape as #1265, in the store that backs push delivery — found because the geofence store was modelled on this one and copied the collapse with the shape.

The app-group file decoded as one array, so one undecodable row discarded the rest, and any read failure returned an empty list — the NSE's append then overwrote rows that were intact on disk. The NSE is where this bites: it runs on pushes that can arrive before the first unlock after a reboot, while the file is still protected.

Rows now decode one at a time and are counted. A read reports `unreadable` separately from empty, and `append`/`remove`/`removeAll` refuse to write over it.

Three calls worth a reviewer's opinion:
- `read()` is a new requirement on the public `PendingPushDeliveryStore`, so it is source-breaking for an external conformer. No default is possible — it could only be `.rows(loadAll())`, which is the bug.
- `loadAll()` stays public with no production caller. Deprecating it would warn in customer builds, so that is a product call rather than part of this fix.
- Geofence's `append` returns a 3-case outcome; this one stays `Bool`, because a failed append here still runs the delivery HTTP and only loses the retry backstop.

**Evidence:** reintroducing the collapse makes `test_append_givenAnUnreadableFile_expectRefusedAndTheQueueUntouched` fail with the file's bytes changed. 488 passed / 0 failed / 5 skipped across Common and MessagingPush, lint clean, api-docs baseline regenerated.

## Related

One ticket, one defect shape, two modules. Separate PRs because they sit on different bases and neither depends on the other.

1. #1265 geofence pending queue → `feature/geofence-polygons`
2. #1266 Common push delivery store → `main` ← **this PR**

### DIFF
--- a/Tests/Common/Service/Push/PendingPushDeliveryStoreTests.swift
+++ b/Tests/Common/Service/Push/PendingPushDeliveryStoreTests.swift
@@ -128,6 +128,20 @@ final class CioAppGroupPendingPushDeliveryStoreTests: UnitTest {
         XCTAssertTrue(logged("could not be read, leaving it intact"))
     }
 
+    /// `remove(id:)` is the NSE success path, so it is the one that runs on every delivered push.
+    func test_remove_givenAnUnreadableFile_expectRefusedAndTheQueueUntouched() throws {
+        let store = makeStore()
+        let metric = makeMetric()
+        XCTAssertTrue(store.append(metric))
+        let before = try Data(contentsOf: metricsFile)
+        try FileManager.default.setAttributes([.posixPermissions: 0o000], ofItemAtPath: metricsFile.path)
+
+        XCTAssertFalse(store.remove(id: metric.id))
+
+        try FileManager.default.setAttributes([.posixPermissions: 0o644], ofItemAtPath: metricsFile.path)
+        XCTAssertEqual(try Data(contentsOf: metricsFile), before)
+    }
+
     func test_removeAll_givenAnUnreadableFile_expectRefusedAndTheQueueUntouched() throws {
         let store = makeStore()
         let metric = makeMetric()
--- a/Tests/Common/Service/Push/PendingPushDeliveryStoreTests.swift
+++ b/Tests/Common/Service/Push/PendingPushDeliveryStoreTests.swift
@@ -75,17 +75,22 @@ final class CioAppGroupPendingPushDeliveryStoreTests: UnitTest {
         logger.errorReceivedInvocations.contains { $0.message.contains(needle) }
     }
 
-    /// The second row has no `device_id`. Every field on the metric is non-optional, so schema
-    /// evolution alone reaches this — and one of them used to discard the other rows with it.
-    private static let oneGoodOneBadRow = """
-    [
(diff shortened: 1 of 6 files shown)


Tests/Common/Service/Push/PendingPushDeliveryStoreTests.swift
```diff
@@ -0,0 +1,173 @@
...
+    private func logged(_ needle: String) -> Bool {
+        logger.errorReceivedInvocations.contains { $0.message.contains(needle) }
+    }
+
+    /// The second row has no `device_id`. Every field on the metric is non-optional, so schema
+    /// evolution alone reaches this — and one of them used to discard the other rows with it.
+    private static let oneGoodOneBadRow = """
+    [
+      {"id":"6C7E1B2A-0000-4000-8000-000000000001","delivery_id":"d1","device_id":"token","event":"delivered","timestamp":"2023-11-14T22:13:20.000Z"},
+      {"id":"6C7E1B2A-0000-4000-8000-000000000002","delivery_id":"d2","event":"delivered","timestamp":"2023-11-14T22:13:21.000Z"}
```
Shahroz16, line 83: Could we build this fixture with the same ISO-8601 encoder settings as the store, then remove `device_id` from the bad row? The hand-written timestamps include fractional seconds that the store's encoder does not emit. These tests pass on iOS 26.5, but using persisted-format bytes would keep this focused on row isolation rather than runtime-specific date parsing. Non-blocking test follow-up.

Shahroz16 (approved):
The 14 pending-store and startup-flush tests passed on the simulator, including preservation of unreadable bytes and valid neighboring rows. The new read result is wired through the startup caller. I left a non-blocking fixture suggestion; locked-device Data Protection still needs device validation.

Tests/Common/Service/Push/PendingPushDeliveryStoreTests.swift
mrehan27: Agreed, and not yet done — I will build the fixture with the store's own encoder settings rather than hand-written timestamps carrying fractional seconds it never emits, and drop `device_id` from the bad row. You are right that as written it leans on runtime date parsing rather than row isolation.

This PR is parked pending a look from Mahmoud, since the store is his from #1030, so the fix will land before it moves.

Shahroz16: Thanks, c64c1a3 addresses this. Both mixed-row tests pass with the encoder-generated dates, so I have no further changes to ask for here.

Shahroz16 (approved):
Re-reviewed the fixture update. It now matches the store's ISO-8601 encoder settings, and all 32 focused simulator tests passed, including the pending-store, startup-flush and SSE retry tests. The current CI run still has a timeout in the unchanged SSE retry test, so that needs a green rerun before merging. Locked-device Data Protection remains outside this simulator validation.

PR #1267, opened 2026-09-11
chore(geofence): log the two fields the timing constants are calibrated from

mrehan27:
Part of [MBL-2415](https://linear.app/customerio/issue/MBL-2415).

The ticket is blocked on field calibration, and two of the four constants could not be calibrated from a drive because the inputs were never logged. This adds them ahead of a weekend field run.

## What it does

- `spd` on `movement.fix.resolved`. Speed rode only on `os.callback.received`, which samples the moments the OS chose to wake us — a different population from the fixes a movement pass actually uses, which is what the wake margin is sized against. Omitted rather than reported as `-1` when the fix carries no speed; CoreLocation's `-1` is not stationary.
- `contradiction.evaluated` — every event arriving while a re-add record is held, with `dly` (event date minus the add) and `win` (whether the current bound covered it). The gate recorded only refusals, so an event just outside `contradictionGateReplayWindow` and one that was never a replay both logged nothing. The constant's own comment cites 0.003–3.3 s measured in reproductions; this is what confirms or moves it from real traffic.

No behaviour change — both are diagnostics, gated by `CIOGeofenceDiagnostics`.

The other two constants need nothing: `movement.fix.resolved` already logs `age` on both sides of the 30 s threshold, so the live-fix bound is measurable as-is.

## Tests

564/564 `LocationGeofenceTests`, lint clean.

**Evidence:** dropping the negative-speed guard makes `movementFixResolved_givenNoSpeedOnTheFix_expectTheKeyOmitted` fail with `spd=-1.0` — a fabricated value that would skew any average taken over the sample.

**Not covered:** `contradiction.evaluated`'s placement. The contract table pins the record's shape and vocabulary, but `CLMonitorGeofenceMonitor` cannot be instantiated in a unit test, so nothing asserts the gate emits it before the window check rather than after. That is the one thing worth a careful read.

## Related

Touches `GeofenceLogTailTests.swift` in the same place as #1265 — both hoist `declaredVocabulary` out of the test to stay under the function-length cap, with identical text. Whoever merges second may see a trivial overlap there.

### DIFF
--- a/Sources/LocationGeofence/Logger+Geofence.swift
+++ b/Sources/LocationGeofence/Logger+Geofence.swift
@@ -277,13 +277,18 @@ extension Logger {
         )
     }
 
-    func geofenceMovementFixResolved(ageSeconds: TimeInterval, requested: Bool) {
+    /// `spd` rides here and not only on `os.callback.received`: the wake margin is sized from how
+    /// far the device travels between passes, and this is the fix a pass actually uses. Speed at
+    /// OS-callback time is a different population — it only samples moments the OS chose to wake us.
+    func geofenceMovementFixResolved(ageSeconds: TimeInterval, requested: Bool, speed: CLLocationSpeed? = nil) {
         let source = requested ? "freshly requested" : "cached"
         debug(
             "Movement pass using \(source) fix, age \(String(format: "%.1f", ageSeconds))s"
                 + geofenceTail("movement.fix.resolved", .input, [
                     ("age", GeofenceLog.num(ageSeconds)),
-                    ("prov", requested ? "requested" : "cached")
+                    ("prov", requested ? "requested" : "cached"),
+                    // Negative means the fix carries no speed, which is not the same as stationary.
+                    ("spd", speed.flatMap { $0 >= 0 ? GeofenceLog.num($0) : nil })
                 ]),
             geofenceTag
         )
--- a/Sources/LocationGeofence/Logger+Geofence.swift
+++ b/Sources/LocationGeofence/Logger+Geofence.swift
@@ -268,87 +268,12 @@ extension Logger {
         )
     }
 
-    // MARK: - Movement trigger
-
-    func geofenceMovementTrigger(tier: HandleMovementTier) {
-        debug(
-            "Movement trigger EXIT: \(tier.rawValue)"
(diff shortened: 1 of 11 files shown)


Sources/LocationGeofence/Logger+Geofence.swift
```diff
@@ -277,13 +277,18 @@ extension Logger {
...
+    func geofenceMovementFixResolved(ageSeconds: TimeInterval, requested: Bool, speed: CLLocationSpeed? = nil) {
         let source = requested ? "freshly requested" : "cached"
         debug(
             "Movement pass using \(source) fix, age \(String(format: "%.1f", ageSeconds))s"
                 + geofenceTail("movement.fix.resolved", .input, [
                     ("age", GeofenceLog.num(ageSeconds)),
-                    ("prov", requested ? "requested" : "cached")
+                    ("prov", requested ? "requested" : "cached"),
+                    // Negative means the fix carries no speed, which is not the same as stationary.
+                    ("spd", speed.flatMap { $0 >= 0 ? GeofenceLog.num($0) : nil })
```
Shahroz16: Before using these samples to calibrate the wake margin, could we add a caller-purpose token or another reliable way to separate them? The same resolver emits this record for movement, contradiction-gate, baseline-heal and polygon requests, while `prov` only distinguishes cached from requested. Mixing those populations can skew the speed distribution. This can follow separately from the instrumentation PR.

Shahroz16 (approved):
The geofence simulator tests passed. The guard split preserves the existing contradiction decision, and negative or unavailable speed remains omitted. I left a follow-up about distinguishing speed samples before using them to choose timing constants; this approval is for the instrumentation, not for calibrated constants.

Sources/LocationGeofence/Logger+Geofence.swift
mrehan27: Fixed in 0c78cee0, and this was the one worth catching. Five call sites share the resolver — movement, contradiction gate, baseline heal, pending events, polygon — so the speed samples were five populations in one, while the wake margin is calibrated from movement alone. Shipping as-is would have produced weekend data that looked usable and was not.

Added `for=<caller>`. No default parameter, so a new call site cannot join the sample silently — the compiler found all 18 existing call sites. The async path is labelled by the request's initiator, since later callers coalesce onto it and log nothing, so one request still yields one record.

PR #1268, opened 2026-09-11
test(geofence): expose the movement-trigger duplicate-callback race

mahmoud-elmorabea:
Adds a test suite that fails on `main` by design, to show the bug is pre-existing.

On device, CoreLocation delivers the same `CLMonitor` event two to three times (identical identifier, state and date; never on the simulator). The wrapper's only defence is the stored per-region baseline. For the movement trigger, the first copy's own movement pass re-seeds that baseline, so a second copy arriving after the re-seed is accepted as a new exit and starts another movement pass and sync. On recorded drives the second copy won that race by 4–9 ms every time; nothing guarantees it.

The wrapper's event entry is private and `CLMonitor.Event` cannot be constructed, so the tests exercise the exact storage call the wrapper makes, in the order the OS and the movement pass produce them:

- before the re-seed → suppressed (passes; how every drive got lucky)
- after the re-seed → **delivered** (fails; the bug)
- after the re-seed, given the OS event's own date → suppressed (passes; the baseline can tell when it is told)

No production change. The fix lands separately.

### DIFF
--- a/Tests/LocationGeofence/Geofence/Monitor/CLMonitorDuplicateCallbackRaceTests.swift
+++ b/Tests/LocationGeofence/Geofence/Monitor/CLMonitorDuplicateCallbackRaceTests.swift
@@ -1,0 +1,85 @@
+@testable import CioInternalCommon
+@testable import CioLocationGeofence
+import Foundation
+import Testing
+
+/// CoreLocation re-delivers the same `CLMonitor` event two to three times on device: identical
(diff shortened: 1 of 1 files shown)


mahmoud-elmorabea:
Closing: the fix landed in the stack as #1269, with storage-level regression tests that carry the OS event date. This suite exercised storage without the date, so it stays red by construction and served its purpose as the proof that the race pre-dated the harness.

PR #1269, opened 2026-09-11
fix(geofence): refuse re-delivered CLMonitor events by their date

mahmoud-elmorabea:
On device, CoreLocation delivers the same crossing event two to three times, identical down to the event date. The wrapper's only defence was the stored per-region baseline, and for the movement trigger the first copy's own movement pass re-seeds that baseline, so a copy landing after the re-seed read as a new exit and started a second movement pass and sync. Recorded drives won that race by 4–9 ms every time; customerio-ios#1268 (now closed) was the proof-of-bug PR that failed on main.

The fix is two arguments on the storage call the wrapper already makes: the OS event date as the evidence bound the baseline heal already uses, and as the baseline's own stamp so the comparison stays in one clock. The heal stamps its write the same way. Three storage tests pin the outcomes before and after the re-seed.

Two further commits fix pre-existing test defects in the same area. Two bootstrap tests slept for a detached re-run, and `GeofenceModule.initialize()` fires the bootstrap against the shared DI graph without awaiting it, so parallel suites wired each other's mocks; they now wait on the outcome, and every suite touching the shared graph is serialized under one parent. The module-setup suite used the trigger's cache read as a barrier for an arm that is written after the read returns, and awaited without a bound, so a fix delivered in that gap hung the whole run; every awaited stream now finishes after five seconds and that test re-delivers the fix until one is consumed.

Part of MBL-2309
Contributes to MBL-2421

### DIFF
--- a/Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor+BaselineHeal.swift
+++ b/Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor+BaselineHeal.swift
@@ -62,10 +62,13 @@ extension CLMonitorGeofenceMonitor {
                 // arrived from the OS on this path. This is also the only site that can return
                 // `.suppressedNewerBaseline` — the OS path passes no evidence timestamp — so
                 // without this the token exists but can never print.
+                // Stamped with the fix's time, not the drain time: OS events are judged against this
+                // stamp by their own date, so both writers must record evidence time.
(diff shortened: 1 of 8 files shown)


Shahroz16 (approved):
The OS event date now stays with the baseline update, and the repeat-after-reseed cases are covered. I checked this against the later lifecycle changes in the stack too. CI is green.

PR #1271, opened 2026-09-11
refactor(geofence): put the CLMonitor wrapper behind OS seams and an injected clock

mahmoud-elmorabea:
The CLMonitor wrapper talked to CoreLocation directly, so the object a test substitutes was also the object holding the decisions. Two protocols now cover every OS call it makes, one for condition monitoring and one for authorization and cached position, with the live CoreLocation adapters behind them. Every timing decision in the wrapper, the movement-fix resolver and storage reads an injected clock instead of `Date()`, so fix ages and baseline stamps come from one source.

Behaviour is unchanged; the existing suites are the coverage. The two CoreLocation adapters, the condition monitor and the authority, are device-only code with no unit coverage, as the same code was on main. The unmonitored-condition branch moves into the registration extension to keep the wrapper under the lint ceiling.

Part of MBL-2309
Contributes to MBL-2421


Shahroz16 (approved):
I traced the live adapters and injected clock through registration, authorization and fix resolution. The focused wrapper, storage and resolver tests pass in the integrated stack. Real CoreLocation behavior still needs device coverage, as noted in the PR.

PR #1272, opened 2026-09-11
refactor(geofence): extract the sync-trigger decisions from module state

mahmoud-elmorabea:
Whether a geofence sync runs at all, and from which position, lived in `GeofenceModuleState`, which reached those rules only through `DIGraphShared` and `CustomerIO.location`. They move to `GeofenceRefreshTrigger`, which takes the two OS-facing reads as closures, so the identified-user gate, the registration-centre-over-cache anchor choice and the arm-once-per-fix rule can be exercised directly. Module state keeps the wiring and the event-bus subscriptions.

Behaviour is unchanged. The event-bus observers hold the trigger strongly rather than through `[weak self]`; it lives for the process and holds no reference back to module state. The new suite waits on outcomes rather than fixed sleeps, using the helper from the first PR in this stack.

Part of MBL-2309
Contributes to MBL-2421


Shahroz16 (approved):
I checked the launch, identify, reset and first-fix wiring through the extracted trigger. The focused trigger and module tests pass in the integrated stack, and this head's CI is green.

PR #1273, opened 2026-09-13
chore(geofence): record when each condition lands at and leaves the OS

mahmoud-elmorabea:
Every `CLMonitor.add` and `CLMonitor.remove` in the wrapper now emits a record — `registration.added id=` and `registration.removed id= op=`, both observations. Android has carried the pair since the log-io contract landed, and its own logger already notes that iOS has no counterpart; this closes that.

The gap is not cosmetic. Adds and removes run inside `enqueueMonitorOperation`, so conditions reach CoreLocation one at a time behind whatever the queue already held, and the storage write that reseeds a condition's dedup baseline sits at the head of each add's own operation. On one recorded relaunch that queue was still draining 170 ms after `registration.applied` was logged, and two callbacks landing in that window were judged against a baseline the device had not written yet. Nothing recorded when any of those calls returned, so a replay answers them instantly and decides differently.

Per identifier rather than per batch, unlike Android's counts: CLMonitor takes one condition at a time, so the identifier is free and it pairs a recorded moment with the exact call a replay has to hold. `op` separates the remove that clears the way for an immediate re-add from a real drop — CLMonitor silently ignores an add over a live identifier and keeps the original circle, so every re-registration removes first.

All seven wrapper call sites carry it. The two logger methods sit in `Logger+GeofenceLifecycle` rather than beside `registration.diff` only because `Logger+Geofence` is at the 400-line limit. Logging only; no behaviour change.

Part of MBL-2309
Contributes to MBL-2421

### DIFF
--- a/Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor+BaselineHeal.swift
+++ b/Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor+BaselineHeal.swift
@@ -62,10 +62,13 @@ extension CLMonitorGeofenceMonitor {
                 // arrived from the OS on this path. This is also the only site that can return
                 // `.suppressedNewerBaseline` — the OS path passes no evidence timestamp — so
                 // without this the token exists but can never print.
+                // Stamped with the fix's time, not the drain time: OS events are judged against this
+                // stamp by their own date, so both writers must record evidence time.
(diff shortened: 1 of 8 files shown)


Shahroz16 (approved):
The registration observations are in place, and the latest commit no longer reports an unmonitored notice as an extra drop. The diagnostic contract tests pass with the stack, and CI is green.

PR #1276, opened 2026-09-14
chore(geofence): fetch nearest fences from the v2 endpoint

mrehan27:
`/geofences/nearest` resolves against `apiHost`, which is region-derived and ends in `/v1`. v1 never returns polygon payloads, so the polygon decode path has only ever run against fixtures.

The version cannot move to the host — `/track` shares `apiHost` and stays v1 — so the endpoint carries its own version and `composeUrl` strips any trailing version segment off the host. The regex is anchored at the end, so `v1.example.com/v1` keeps its hostname and a self-hosted host with no version segment still works.

## Evidence

The live v2 response run through this decoder: 9/9 regions decoded, 0 drops, 4 polygons with 4/6/3/4 vertices and their enclosing circles. Field-confirmed on device over the weekend — the airport polygon delivered enter and exit against the real API.

585/585 `LocationGeofenceTests`, lint clean, 14 production lines.

## Related

- #1266 — open, different module, no dependency

Part of [MBL-2424](https://linear.app/customerio/issue/MBL-2424/mobile-fetch-nearest-geofences-from-the-v2-endpoint)


Shahroz16 (approved):
Checked the V2 URL construction against the backend route and response contract, including preservation of the shared host and track routing. All 9 focused API tests pass. This verifies the code contract, not backend deployment availability.

PR #1277, opened 2026-09-14
chore(geofence): pin the reason token for records ev alone cannot identify

mrehan27:
`registration.rejected` covers three causes and `polygon.undelivered` eight, but the contract table only asserts that `why=` is present. Swapping two cases in either `logToken` switch passed all 585 tests — including the `queue.unreadable` vs `decode_failed` distinction the pending-queue work exists to draw.

Pins the token per case for the three switch-backed reason enums. The switches are exhaustive, so a new case is a compile error.

The five `rawValue`-backed reason enums get no equivalent test: Swift rejects duplicate raw values at compile time, so distinctness is already guaranteed there.

## Evidence

Negative control: swapped `unusable_circle` and `unusable_polygon` in production, rebuilt, ran — exactly one test failed, and it was the right one. Reverted and re-verified.

587/587 `LocationGeofenceTests`, lint clean, test-only, no production files changed.

## Related

- #1276 — open, same area, no dependency

Part of [MBL-2415](https://linear.app/customerio/issue/MBL-2415/mobile-define-geofence-timing-and-geometry-constants)


Shahroz16 (approved):
The literal wire-value and sanitization assertions pin the intended contract without relying on the formatter to generate expectations. All 40 focused log-tail tests pass.

PR #1278, opened 2026-09-14
chore(geofence): stamp transitions with the crossing time, not the send

mrehan27:
Recovered from a local-only branch built before the polygon stack was squashed, rebased onto the current feature tip.

## What

A transition was stamped when the SDK got round to sending it, not when the crossing happened. On a cold wake or a retried send those differ by the whole wake-to-verdict gap, so the customer saw the wrong event time.

Three production callers now pass a real evidence time: the OS event date for a crossing, the fix timestamp for a polygon verdict, and discovery time for a synthesized initial enter — nothing crossed anything there, so the moment we noticed is the only honest value.

## Why the userId is part of the pending key

Making the timestamp the crossing time makes `PendingGeofenceMetric.key` stable across paths, not just across retries (a retry already read the persisted timestamp). That only matters once the cooldown is gone — and the cooldown is per user and cleared on sign-out, while the queue deliberately survives it. So one crossing's evidence can be applied twice under different users. Keyed without the userId, the second user's row is dropped as a duplicate and its successful send then removes the first user's: one event lost, the other misattributed.

The key is computed, not persisted, and both sides of every comparison recompute it, so the format change needs no migration.

## Customer-visible

An event replayed after a suspension is now dated at the crossing rather than the send. That is the point of the change, but CDP will receive back-dated events where it previously received send-dated ones.

## Verification

Build clean, SwiftLint `--strict` 0 violations / 794 files, LocationGeofence + Common 902 passed / 0 failed / 3 skipped.

Reviewed by a second session against the code: no defects, and it independently confirmed the sign-out path from `ResetEvent` through `clearUserScopedState` rather than from the comment.

## Stack

- #1278 (this PR)
- #1279 clamp a belief's evidence time to the write

## Note on the ticket

MBL-2386 is titled for fix freshness and selection — the contradiction gate discarding `isFresh`, the newest-of-two rule in three copies. This PR is the crossing-time work instead, which was filed under the same branch prefix. The freshness half is untouched and still open.


Sources/LocationGeofence/Storage/PendingGeofenceMetric.swift
```diff
@@ -22,14 +22,20 @@ struct PendingGeofenceMetric: Codable, Equatable, Sendable, GeofenceMetric {
...
+    /// `userId` is load-bearing. `timestamp` is the CROSSING time, and one crossing's evidence can
+    /// be applied twice under different users: the cooldown is per user and sign-out clears it,
+    /// while this queue deliberately survives sign-out. Keyed without the userId, the second user's
+    /// row is dropped as a duplicate and its successful send then removes the first user's — one
+    /// event lost, the other misattributed. Diverges from Android's `PendingGeofenceDelivery.key`,
+    /// which carries a userId on the row but not in the key and so still has that gap.
     var key: String {
         let sec = Int(timestamp.timeIntervalSince1970)
-        let base = "\(geofenceId)_\(transition.rawValue)_\(sec)"
+        let base = "\(geofenceId)_\(transition.rawValue)_\(sec)_\(userId)"
```
Shahroz16: Could we make the composite key unambiguous as a follow-up, for example by encoding the components? For the same fence and crossing second, user `a_42` with no geoset and user `a` in geoset `42` produce the same key. That needs an unusual identity/membership change, but one row's successful send can then remove the other's pending row. This change improves the existing collision behavior; the remaining case need not hold it up.

Shahroz16 (approved):
I traced the occurrence timestamp and captured user through the emitters, pending queue, direct HTTP and EventBus delivery. All 214 focused tests pass. The key-encoding note is a follow-up to this improvement; the customer-facing future-timestamp behavior remains the explicit deferral described in #1279.

Sources/LocationGeofence/Storage/PendingGeofenceMetric.swift
mrehan27: Fixed in 5fb57684. Components are escaped before joining, so a separator inside a value cannot imitate a boundary. The key is recomputed, never persisted, so no migration is needed.

PR #1279, opened 2026-09-14
chore(geofence): clamp a belief's evidence time to the write

mrehan27:
Stacked on #1278.

## What

A polygon belief stored an evidence time taken from the fix. With a clock skewed forward, a future-dated evidence was stored as-is and then refused every later write, stranding the belief until the clock caught up. It is now clamped to the write time, so later writes win.

## Known inconsistency, deliberately left

`PolygonMembershipResolver` passes the **unclamped** evidence to the tracker, so under a skewed clock the belief is clamped while the customer's event timestamp is still in the future. Two rules for one value. Left as-is rather than widened here because clamping the customer-facing timestamp is a separate decision about what the customer should see — worth settling, but not inside this fix.

## Verification

Build clean, SwiftLint `--strict` 0 violations / 794 files, LocationGeofence + Common 903 passed / 0 failed / 3 skipped.

Split out of #1278 on review: different defect, different layer, and it changes an acceptance rule rather than a timestamp source.

## Stack

- #1278
- #1279 (this PR)


Sources/LocationGeofence/Storage/GeofenceStorage+PolygonMembership.swift
```diff
@@ -43,6 +43,9 @@ extension GeofenceStorage {
         onlyIfCircleMatches evaluatedCircle: MonitoredCircle? = nil,
         now: Date = Date()
     ) -> PolygonMembershipOutcome {
+        // A crossing cannot postdate the moment we learn of it. Unclamped, a clock set backwards
+        // stamps the belief in the future and every correcting write is refused until it catches up.
+        let evidenceTimestamp = evidenceTimestamp.map { min($0, now) }
```
Shahroz16, line 48: Could we cover recovery of an already future-dated belief as a follow-up? This protects new writes, but an existing `lastChangedAt` ahead of `now` still wins the comparison below, so normal fixes cannot repair it until the clock catches up. A persisted-state test would pin recovery for devices that ran the earlier polygon build. This is existing-state hardening, not a regression introduced here.

Shahroz16 (approved):
The clamp protects new belief writes from future-dated evidence, and all 27 focused storage tests pass. Recovery of state written by an earlier polygon build is a separate follow-up, noted inline.

Sources/LocationGeofence/Storage/GeofenceStorage+PolygonMembership.swift
mrehan27: Fixed in 258afb47. The stored stamp is now clamped on read as well as on write, so a future-dated belief can be corrected instead of winning until the clock catches up. Added a persisted-state test that seeds the record directly to disk.

PR #1280, opened 2026-09-14
chore(geofence): select a fix by one rule and trust the resolver's verdict

mrehan27:
The fix-freshness half of MBL-2386. MBL-2414 closed by pointing here: every arbiter in the geofence path is downstream of one fix, so fix selection is where the leverage is.

## Two defects

**One rule, three copies, already drifted.** `MovementFixResolver.cachedFix` and both monitors' `bestKnownFixDetail()` each hand-rolled "newest of the OS cache and the delivered fix", and they disagreed on a tie: the monitors returned the cache, the resolver returned its own fix. The value is the same either way, but `fixsrc` differs — and that field is what the drive analysis reads. Now one shared `FixSelection.newest`, with the tie going to the delivered fix, whose provenance is known rather than whatever the OS last happened to hold.

**The contradiction gate discarded the resolver's freshness verdict.** It recomputed freshness from the age of whatever `bestKnownFix()` returned *after* the request. A request that failed still leaves an OS cache that can be under `movementFixMaxAge`, so the recomputation read failure as success and left `gateFixRequestFailedAt` unarmed. The block exists so a burst with no fix obtainable doesn't stall the event consumer for one timeout per event — unarmed, every later event paid another full timeout. Cost, not correctness: the caller still age-guards through `BaselineHealDecision`.

## Coverage limits, stated rather than papered over

- `FixSelection` is covered directly, including the tie and the invalid-coordinate case.
- **The gate fix has no test.** `CLMonitor` cannot be instantiated in a unit test — it aborts the test process — so `resolveGateFix` is unreachable from the suite. Extracting the one-line arming decision to make it testable would only test a restatement of itself, so I didn't.
- Nothing *pins* that the monitors call the shared rule, and neither is unit-testable — so instead of a test, `bestKnownFixDetail()` is now a `GeofenceFixSelecting` protocol default and the monitors hold zero copies. Re-introducing one means adding a shadowing method, which shows up in review rather than drifting silently.

## Verification

Build clean, SwiftLint `--strict` 0 violations / 796 files, LocationGeofence + Common 900 passed / 0 failed / 3 skipped.

A fourth copy of the same rule — the resolver's fallback pick — is folded in too; it already settled its tie the same way, so that is a dedup, not a behaviour change.

## Not in this PR

The `latestFix` -> `lastDeliveredFix` rename (30 references across 7 files, cosmetic, would bury the two fixes) and the forced-fresh entry point owning `cached: nil`, which is a larger change to how callers get a baseline.


Shahroz16 (approved):
The shared selection preserves the intended freshness and tie rules, and the resolver verdict now reaches the cooldown decision. All 25 focused selection/resolver tests pass. The actual CLMonitor gate remains a source-reviewed boundary, not a device-tested one.

PR #1281, opened 2026-09-14
chore(geofence): pin the classic monitor's polygon event path for iOS 13-17

mrehan27:
## Polygons already work on iOS 13–17

Measured on the iOS 17.5 simulator against the real `/v2/geofences/nearest` (9 fences, 4 polygons). **No code change was needed.** In one session: v2 fetch and decode, covering circles registered, movement trigger + re-rank + dynamic wake radius, annulus correctly silent (`edge=-27`), a **polygon enter delivered** (`m=inside edge=31` → `transition.accepted` → `delivery.sent`), a circle exit delivered, initial enter-when-inside synthesized, and a cold wake — app killed, drove out, the OS relaunched it and delivered the exit.

So this PR is tests, not mechanism.

## Why the three iOS 18+ mechanisms aren't ported

Scored for **polygons**, which is this ticket. Polygon state lives in `PolygonMembershipResolver` and `GeofenceStorage+PolygonMembership`, both path-agnostic, so three of them are not "unneeded on classic" — they're not involved:

- **Contradiction gate** — nothing to refuse. Nine regions registered with the device inside one produced zero `os.callback.received`; classic emits no initial state unless you call `requestState(for:)`. Membership also never consumes the monitor baseline.
- **Re-arm** — the OS relaunches the app and delivers the crossing, which is exactly what CLMonitor fails at.
- **Dedup baseline** — membership dedups itself. The cold wake delivered the same exit twice (two OS deliveries ~20 m apart during egress, not a re-add corrective — no `registration.diff` precedes it, and classic's `adoptExistingRegions` only takes ownership). The second was absorbed by `polygon.undelivered why=no_change`, at the membership layer.
- **DROP parity is exact** — the classic path reads the crossed circle straight off the `CLCircularRegion` the delegate hands it, so `onlyIfCircleMatches` works without a ledger.

**What is genuinely uncovered is classic CIRCLES, which have no baseline or heal.** That predates this ticket and is shipped behaviour for every pre-18 user. Adding the baseline alone would be a regression, not half a fix: the only way classic produces two enters past the cooldown with no exit between is the OS missing the exit, after which the second enter is a real re-entry that classic delivers correctly today — a baseline would suppress it as `no_state_change` and strand the customer "inside". That's the 18+ failure mode heal exists to repair. Baseline and heal are a pair or neither, and that's a circles ticket.

## The tests

Four, pinning the monitor's half of the polygon path — nothing else covers this band, and the CLMonitor twin can't be built in a test at all:

- a covering-circle enter carries the crossed circle to the resolver, not `.unknown`
- the exit does too
- each buffered event carries the circle **it** crossed, not a later lookup (the one that earns its keep)
- an owned crossing emits `os.callback.received` — the "zero callbacks" evidence above is only meaningful if this path logs at all

They insert ownership directly because `startMonitoring` bails on blocked authorization in a test bundle. The event path consults exactly one thing registration establishes — `ownedRegionIdentifiers` — and geometry comes from the callback's region, so this reproduces the observable post-registration state.

## Verification limits

**iOS 17.5 only.** No 13–16 runtime is installed, and this band can't be field-tested, so 17.5 is the whole verification story. The classic `CLCircularRegion` API is unchanged since iOS 7 — that's an argument for the proxy, not evidence.

Known classic limit, unchanged by this PR: the delegate carries no event date, so `occurredAt` is stamped at receipt. After #1278 a relaunch-delivered crossing is therefore dated at relaunch, not at the crossing.

## Verification

Build clean, SwiftLint `--strict` 0 violations / 795 files, LocationGeofence 588 passed / 0 failed.

Part of [MBL-2426](https://linear.app/customerio/issue/MBL-2426/ios-support-polygon-geofences-on-ios-13-17)


Shahroz16 (approved):
The four classic-path tests pass, including per-event geometry for buffered callbacks. The red CI job is the unchanged SSE timeout; all 18 SSE tests also pass in a focused local rerun. This looks ready from the code review, with a green CI rerun still needed. These simulator checks do not prove behavior on older iOS devices.

PR #1282, opened 2026-09-15
chore(geofence): record what the contradiction gate decided when it does not refuse

mrehan27:
The gate logs its geometry only when it refuses. A gate that passed an event and a gate that had no fix to judge one both wrote nothing, so at radii below the fix accuracy — where declining is the norm — a drive cannot tell whether the gate still protects anything.

Adds `contradiction.allowed` (dist, rad, signed edge, acc, age) and `contradiction.no_fix` (with a `why`). Diagnostics only, no decision changes. Also corrects the polygon `edge` doc, which stated the sign backwards.

618 LocationGeofenceTests pass, lint clean. The signed-edge test asserts from position rather than the formula, which caught its own fixture.

Part of [MBL-2425](https://linear.app/customerio/issue/MBL-2425/mobile-evaluate-position-quality-before-accepting-circle-transitions) — these are the fields that ticket's analysis reads.


Sources/LocationGeofence/Logger+GeofenceDecision.swift
```diff
@@ -70,6 +70,64 @@ extension Logger {
...
+    /// against the record's `ev`. The measurements ride as their own keys and no `why` token
+    /// classifies them: which guard declined is derivable from `edge`, `acc` and `age`, and a
+    /// token would need a second copy of `BaselineHealDecision`'s guard sequence to produce.
+    func geofenceContradictionAllowed(
+        identifier: String,
+        transition: GeofenceTransition,
+        geometry: GateFixGeometry
+    ) {
+        debug(
+            "Allowed OS \(transition.rawValue) for region \(identifier): the fix does not contradict it (distance \(Int(geometry.distanceFromCenter)) m, radius \(Int(geometry.radius)) m, accuracy \(Int(geometry.accuracy)) m, fix age \(String(format: "%.1f", geometry.fixAge))s)"
```
Shahroz16: Could we say "the gate did not refuse it" here? This also runs for stale or inaccurate fixes, so the current wording can read like positive confirmation. The measurements and decisions are unchanged, so this is non-blocking.

Shahroz16 (approved):
The gate keeps the same decisions and now records its non-refusal outcomes with the expected signed geometry. I checked the caller and fix-selection paths, completed the Claude Code Opus review, and verified the hosted report has all 640 geofence tests passing. One non-blocking wording suggestion below.

Sources/LocationGeofence/Logger+GeofenceDecision.swift
mrehan27: Changed to "the gate did not refuse it" in 2b140b52. You are right that the old wording implied the fix agreed, which it cannot when the fix was too stale or too coarse to decide.

PR #1283, opened 2026-09-15
chore(sample): probe what CLVisit would give us for a stationary device

mrehan27:
The SDK has no wake source while the device is still: app closed, nothing moving, so no movement pass runs and baseline heal never gets a chance. A dwell crossing is missed outright — measured twice in the field on 2026-08-14 with a healthy daemon, once 896 m inside a 1 km fence for ~15 minutes.

`startMonitoringVisits` is the only iOS primitive with dwell semantics that also relaunches a terminated app. This records its arrival and departure latency and nothing else, so we can decide whether it is worth building on before any SDK code depends on it.

Sample app only. No SDK changes, off by default.

Gated on `UserDefaults`, resolved and persisted at launch, because the capture spans an OS relaunch and nothing launched that way carries an environment or launch arguments. Enable with `-cio_visit_probe YES` (or `CIO_VISIT_PROBE=1`) once; it is written to disk so the relaunched process keeps it. It cannot work before the first unlock after a reboot, where `UserDefaults.standard` is unreadable — the same window the SDK itself cannot work in. A launch in that window only reads the gate, never writes it, so it cannot overwrite a stored `YES`.

Verified: builds for device and simulator, 618 SDK tests pass, lint clean. On an iOS 26.5 simulator, `probe.visit.started` emits once per launch and the gate survives a launch with no environment set. CLVisit itself cannot fire on a simulator, so the arrival and departure records — including the latency this PR exists to measure — are unverified until a real stationary capture.

Part of [MBL-2433](https://linear.app/customerio/issue/MBL-2433/mobile-evaluate-wake-sources-for-stationary-geofence-detection)


Apps/APN-UIKit/APN UIKit/Diagnostics/VisitProbe.swift
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    private static func persistGate(_ enabled: Bool) -> Bool {
+        UserDefaults.standard.set(enabled, forKey: defaultsKey)
+        return enabled
+    }
+
+    func startIfEnabled(launchOptions: [UIApplication.LaunchOptionsKey: Any]?) {
+        launchedByLocation = launchOptions?[.location] != nil
+        guard Self.resolveAndPersistGate() else { return }
```
Shahroz16: Could we call `stopMonitoringVisits()` when the probe is explicitly disabled? After a launch with `-cio_visit_probe YES`, relaunching with `NO` only saves false and returns here. [Core Location keeps visit monitoring active across launches](https://developer.apple.com/documentation/corelocation/cllocationmanager/startmonitoringvisits()), so an off capture can still get visit-triggered wakes, just without the probe records. Stopping on the explicit-disable path, while leaving an unreadable defaults value alone, and covering the enabled-to-disabled relaunch would keep the off state meaningful.

Shahroz16 (commented):
The authorization retry and protected-defaults fixes look good. There is one remaining issue with turning the probe off, noted inline, so I'm holding approval for that path.

Apps/APN-UIKit/APN UIKit/Diagnostics/VisitProbe.swift
mrehan27: Good catch — that would have made every probe-off capture an unlogged state rather than an off one.

Fixed. `resolveGate()` now returns three cases instead of a Bool: `.enabled`, `.disabled` (a launch explicitly said off), and `.off` (no statement this launch). Only `.disabled` calls `stopMonitoringVisits()`; `.off` — including the unreadable-defaults window before first unlock — leaves a running registration alone, as you asked.

Worth flagging: my first attempt (3635e370) was wrong. The `.disabled` arm fell out of the switch into the registration path, so it stopped monitoring and immediately restarted it on the same launch. Fixed in a690fdb0.

Verified on an iOS 26.5 sim with Always granted: `=1` logs `started` only, `=0` logs `stopped` only, a bare launch with the gate off logs nothing, and a bare launch after an enable starts from disk. Negative control — `return` removed — logs `stopped` then `started`, so the check does discriminate.

Shahroz16 (approved):
Thanks, the explicit-disable path now stops visit monitoring and returns before it can restart, while an unreadable defaults value leaves the registration alone. The focused control-flow checks catch both the original bug and the missing-return version; current CI is green. This addresses my earlier concern. Actual visit timing remains for the device capture.

PR #1284, opened 2026-09-16
chore(geofence): decide polygon arrivals from the fix, not a fixed margin

mrehan27:
A polygon verdict needed clearance beyond a 20 m floor borrowed from the circle heal path. Heal synthesizes a crossing the OS never reported, so a wide deadband is cheap there. A polygon verdict is the only detection there is, so the same deadband makes a small venue undecidable: on the measured Tim Hortons ring, 24 m deep, it left 3% of the interior claimable.

The margin is now the fix's own accuracy, bounded per fence by how deep the venue is (`PolygonRegion.scale`, `2·area/perimeter`). An arrival inside that margin is not discarded: a second fix is sought, and only a usable one reading outside blocks it. An ambiguous fix reading outside is still no verdict, so a fix-derived departure stays strict and a visit can never end early.

The ceiling gates arrivals only. A device clear of the ring by more than its own accuracy is outside however thin the venue is, so that check runs ahead of the ceiling. Behind it, a 20 m-deep ring refused every departure a coarse fix could prove, leaving the visit open inside the far larger covering circle where no circle exit closes it either.

A pass settles everything one fix can decide before corroborating anything. Corroborating inline let one marginal polygon's request age the shared fix past the freshness gate for every polygon judged after it, so an arrival at one venue depended on whether an unrelated venue happened to be marginal, and on catalog order.

Decidable share of the four real rings at 5 m accuracy: 3% → 100% (Tim Hortons), 31% → 100% (Rainbow).

Verified on device geometry rather than fixtures. Sim at 3.0 m inside the real Tim Hortons ring — ambiguous at 5 m accuracy — logs `polygon.verdict id=17 m=inside edge=3 acc=5.0 cor=true` and delivers the enter, where the old rule logged nothing. A second pass at the same spot logs `why=corroboration_unnecessary` and spends no request, because the belief already says inside.

A marginal arrival is accepted unless a usable second fix positively places the device outside. It is not held pending confirmation: while the device stands still nothing re-derives a missed arrival, so refusing one loses the visit, whereas a spurious one is corrected by the next decisive fix. A timed-out, echoed or too-coarse second fix therefore commits, recording `cor=false` with a `corwhy` reason so a capture never reads it as decisive. `cor` itself stays the shared cross-SDK boolean.

The second fix must strictly postdate the fix it is confirming. That is checked against that fix's own timestamp, not against `resolveFix`'s freshness baseline: the baseline is the last fix the resolver *delivered*, and a pass answered from the location cache never records there, so an echo of the first fix used to clear it and confirm an arrival against itself. An echo is now refused and logged `corroboration_not_independent`, countable separately from location not answering. Corroboration confirms the first fix was not a one-off outlier; it is not an independent position estimate.

512 production lines added, 405 net, over our usual 250 cap. About half is the rule itself; the rest is five Bugbot findings and two review findings on the corroboration and pass paths, plus two file splits to stay under the 400-line cap. Kept here because they are defects in this PR's own new code.

Part of [MBL-2425](https://linear.app/customerio/issue/MBL-2425/mobile-evaluate-position-quality-before-accepting-circle-transitions)


Shahroz16 (commented):
The newer-fix and retry-scope fixes are in place. I'm holding approval for two introduced paths below: definite departures rejected by the venue-size ceiling, and later polygons losing usable evidence while an earlier one waits for corroboration. The first is reproduced with the actual decision/geometry sources; the second is traced through the production cache, timeout and serial pass. The completed Claude Code Opus review independently identified both. Hosted tests pass, but these cases are not covered there. Local simulator validation was unavailable because Xcode's license needs acceptance.

Sources/LocationGeofence/Polygon/PolygonMembershipDecision.swift
```diff
@@ -2,27 +2,84 @@ import Foundation
...
+        signedEdgeDistance: Double,
+        horizontalAccuracy: Double,
+        fixAge: TimeInterval,
+        venueScale: Double
+    ) -> Outcome {
+        guard fixAge >= 0, fixAge <= GeofenceConstants.movementFixMaxAge else {
+            return .undecided(.fixTooOld)
+        }
+        guard horizontalAccuracy > 0 else { return .undecided(.noUsableFix) }
+        guard horizontalAccuracy < venueScale else { return .undecided(.accuracyTooLow) }
```
Shahroz16: Could we apply the venue-size ceiling only to arrivals? For a 400 m by 20 m ring, scale is about 19 m. A fresh fix 30 m outside its long edge with 25 m accuracy used to establish outside, but now returns accuracyTooLow. I reproduced this with the actual decision and geometry sources. That point is still inside the roughly 200 m covering circle, so an existing visit stays open and a return can miss its next enter. Keeping the strict outside-clearance check ahead of this ceiling, with a regression test, would preserve departures.

Sources/LocationGeofence/Polygon/PolygonMembershipResolver+Corroboration.swift
```diff
@@ -0,0 +1,160 @@
...
+            // inside, so a second fix would cost a forced request (up to `movementFixRequestTimeout`)
+            // to reach `no_change`. A device standing near a boundary hits this on every pass.
+            guard await storage.getPolygonMembership()[geofence.id]?.membership != .inside else {
+                logger.geofencePolygonUndecided(
+                    identifier: geofence.id, reason: PolygonUndecidedReason.corroborationUnnecessary,
+                    signedEdgeDistance: signedEdgeDistance, horizontalAccuracy: fix.horizontalAccuracy
+                )
+                return nil
+            }
+            guard await corroborate(
```
Shahroz16: Could we process the single-fix verdicts before waiting for corroboration? A foreground pass can legitimately use a 25-second-old cached fix. If the first polygon needs corroboration and that request times out after 10 seconds, the serial loop hands the same now-35-second-old fix to later polygons, which reject it as fixTooOld even if it was decisive for them at the start. This makes an unrelated polygon's arrival depend on catalog order. A two-polygon timeout regression, including the reversed order, would cover this without weakening the freshness guard.

Sources/LocationGeofence/Polygon/PolygonMembershipDecision.swift
mrehan27: Fixed in 4af7d7ad (via 2adc954f). You are right, and the ceiling had no business gating departures.

The decisive-outside branch now runs before the ceiling. Clearing the ring by more than the fix accuracy is a complete argument however thin the venue is, so your 400x20 case now returns `.decided(.outside)` instead of `accuracyTooLow`.

Regression tests added, all three on the 19.05 m scale from your example:
- 30 m clear at 25 m accuracy -> `.decided(.outside)`
- 20 m clear at 25 m accuracy -> still `accuracyTooLow` (clearance is measured against accuracy, so this is not a blanket skip for anything negative)
- 8 m inside at 25 m accuracy -> still `accuracyTooLow` (arrivals unchanged)

Mutation-checked: putting the ceiling back in front fails the first test only.

Sources/LocationGeofence/Polygon/PolygonMembershipResolver+Corroboration.swift
mrehan27: Fixed in 4af7d7ad (via 2adc954f). Done the way you suggested: a pass now runs in two phases.

`classifyMembership` settles everything the pass fix alone can decide and returns marginal arrivals instead of corroborating them; `runPass` corroborates the deferred ones afterwards. No polygon can now be aged past `movementFixMaxAge` by an unrelated polygon burning the request timeout, and the freshness guard is untouched.

Two-polygon regression added in both orders, as you asked. The assertion is the ordering invariant rather than a wall-clock timeout: the decisive verdict must already be recorded when the corroboration request goes out. Restoring inline corroboration fails the marginal-first case and leaves marginal-last passing, which is the order dependence itself.

One follow-on your finding exposed, caught by Bugbot on the fix: splitting phases moved the already-inside short-circuit away from the request it protects, so a belief turning inside during phase one no longer stopped the second fix. The guard now sits immediately before `corroborate` and is re-read per polygon. No wrong events either way, since the write refused them; what was lost was the saved request.

Also worth knowing when reading captures: `age` on a `cor=true` verdict is age-at-record, not age-at-judgement, so it can legitimately exceed 30 s after a corroboration timeout. Documented at the call site.

Tests/LocationGeofence/Geofence/Polygon/PolygonMembershipResolverTests.swift
```diff
@@ -1314,4 +1314,235 @@ struct PolygonMembershipResolverTests {
...
+    private static func latitudeInsideNorthEdge(by metres: Double) -> Double {
+        0.0016 - metres / 111320
+    }
+
+    /// The high-severity case: the pass fix came from the SYSTEM cache, which
+    /// `MovementFixResolver` answers with but never records in `latestFix`. The forced
+    /// corroboration request then only had to beat `latestFix` — still nil — so CoreLocation
+    /// echoing that very same fix cleared the guard and an ambiguous arrival confirmed itself.
+    @Test
+    func evaluateMembership_givenCorroborationEchoesThePassFix_expectNoEnter() async {
```
Shahroz16: Could we add the positive half of this corroboration case? The suite now proves that an echoed or disagreeing second fix is refused, but no test asserts that a strictly newer inside fix records `inside` and delivers the enter. Without that case the new delivery path could be inverted while the tests stay green.

Shahroz16 (changes requested):
Requesting one test addition before approving this behavior change. The new suite proves that an echoed or disagreeing second fix is refused, but it never asserts the positive contract: a strictly newer inside second fix should record `inside` and deliver the enter. Without that case the corroboration path could be inverted while the tests remain green. Could we add that success case? Hosted checks are green; local simulator validation was unavailable here.

Sources/LocationGeofence/Polygon/PolygonMembershipResolver+Corroboration.swift
```diff
@@ -0,0 +1,160 @@
...
+    ///   postdate it; anything at or before it is the first fix over again, not a second opinion.
+    func corroborationFix(newerThan basis: Date) async -> CorroborationOutcome {
+        if let attempt = passCorroboration, attempt.basis == basis { return attempt.outcome }
+        let outcome: CorroborationOutcome
+        switch await resolveFix(requiringFresh: true) {
+        case .none: outcome = .unavailable
+        case .some(let fix): outcome = fix.timestamp > basis ? .obtained(fix) : .notIndependent
+        }
+        passCorroboration = (basis, outcome)
+        return outcome
```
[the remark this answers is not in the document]
Shahroz16: Checked against the current head: `resolvePassFix` clears `passCorroboration` before each pass, and the retry regression covers this path. The stale-attempt concern is addressed by 4af7d7ad.

Sources/LocationGeofence/Polygon/PolygonMembershipResolver+Pass.swift
```diff
@@ -0,0 +1,77 @@
...
+                geofenceId: geofenceId, fix: fix, isStillCurrent: isStillCurrent
+            ) {
+                deferred.append(pending)
+            }
+        }
+        for pending in deferred {
+            guard await corroborate(
+                pending.proposed, geofence: pending.geofence, polygon: pending.polygon,
+                firstFix: fix, firstEdge: pending.signedEdgeDistance
+            ) else { continue }
```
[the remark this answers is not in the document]
Shahroz16: Checked at the current head: `runPass` now re-reads the stored belief immediately before `corroborate`, so the phase-two guard is present. The remaining gap is positive success coverage, which I called out separately.

Tests/LocationGeofence/Geofence/Polygon/PolygonMembershipResolverTests.swift
mrehan27: Added in fc4ddecf. `evaluateMembership_givenSecondFixReadsInside_expectEnterDelivered` asserts the enter is delivered and the belief records `inside`.

Checked it catches the inversion you described: making `corroborate` return false at its accept branch fails this test and only this test — the other 63 stay green.

Sources/LocationGeofence/Polygon/PolygonMembershipResolver.swift
```diff
@@ -26,13 +26,21 @@ import UIKit
...
     let contextStore: BackgroundDeliveryContextStore
     let notificationCenter: NotificationCenter
     var foregroundObserverToken: NSObjectProtocol?
+
+    /// The one corroboration attempt made against the fix currently being judged, so N marginal
+    /// polygons sharing that fix cost one request rather than N — a failed attempt included.
+    /// Keyed by that fix's timestamp for CORRECTNESS (never answer for a fix it predates) and
+    /// cleared by `resolvePassFix` for LIVENESS (a cached failure must not outlive its pass). Both
+    /// are needed: a stationary device's next pass often reuses the very same cached fix.
+    var passCorroboration: (basis: Date, outcome: CorroborationOutcome)?
```
Shahroz16: Could we make this cache scoped to a pass, or serialize the fresh passes, rather than only to the fix timestamp? Each refresh launches `evaluateNewlyRegistered` and then `evaluatePolygonsAfterMovement`; both are `requiresFreshFix` passes and both bypass the `passesInFlight` guard. If the first pass times out corroboration, it stores `.unavailable` here, and the overlapping later pass can reuse that result for the same basis without making its own attempt, so a transient timeout can suppress that pass’s enter. A small overlap regression test would pin the fix.

Shahroz16 (changes requested):
Requesting one concurrency fix before approval. The positive corroboration path is now covered by the new test, and hosted checks are green. Each refresh starts both the newly-registered polygon pass and the movement pass, and both fresh passes bypass the in-flight guard. A timeout cached by the first pass can therefore be reused by the overlapping pass for the same fix without another attempt, suppressing an enter after a transient failure. Could we serialize fresh passes or scope the cached attempt to a pass, with an overlap regression test?

Sources/LocationGeofence/Polygon/PolygonMembershipResolver.swift
mrehan27: Fixed in 092933be by scoping the attempt to a pass rather than serialising — the bug is cache lifetime, not pass timing.

`PassCorroboration` is created by `runPass`, so a pass cannot reuse an attempt it did not make. The basis key stays for correctness inside the pass. `resolvePassFix` existed only to clear the shared state and is gone; its three callers use `resolveFix` directly.

Regression test drives `runPass` twice over one fix — the same shared-basis condition the overlap produces, without racing two entry points. Mutation-checked: sharing the attempt across passes drops the second request and fails that test.

You were right that the basis key alone was not enough. My own comment claimed a cached failure could not outlive its pass, which is exactly what two overlapping passes break.

Sources/LocationGeofence/Polygon/PolygonMembershipResolver+Corroboration.swift
```diff
@@ -0,0 +1,200 @@
...
+        // guard and confirm an arrival against itself.
+        let second: CLLocation
+        switch await corroborationFix(newerThan: firstFix.timestamp, cache: cache) {
+        case .obtained(let fix):
+            second = fix
+        // Two tokens, not one: an echo means the rule declined to count one fix twice, a timeout
+        // means location never answered at all. A capture has to separate them. Neither is
+        // evidence the device is outside, so both commit.
+        case .notIndependent:
+            return .unconfirmed(.corroborationNotIndependent)
```
Shahroz16, line 99: Could we update the branch contract alongside this policy change? The resolver still says an undecidable fix leaves the belief untouched, and the PR says marginal arrivals are confirmed by a second fix. This now delivers when the second request times out, echoes the first fix, or is too coarse to judge. I am not asking to reverse the calibration trade-off, but the shipped contract needs to say that a marginal inside fix is accepted unless a usable second fix contradicts it.

Sources/LocationGeofence/Logger+GeofencePolygon.swift
```diff
@@ -83,15 +83,27 @@ extension Logger {
...
                     ("m", "\(membership)"),
                     ("edge", GeofenceLog.num(signedEdgeDistance, 0)),
                     ("acc", GeofenceLog.num(horizontalAccuracy)),
-                    ("age", GeofenceLog.num(fixAge))
+                    ("age", GeofenceLog.num(fixAge)),
+                    // Matches Android's `cor` key, widened: `false` when one fix was decisive,
+                    // `true` when a second agreed, and otherwise the reason the arrival committed
+                    // without one. A capture must not read an unconfirmed marginal arrival as a
+                    // decisive one. Always `false` for a departure, which never corroborates.
+                    ("cor", corroboration.logToken)
```
Shahroz16: Could we keep `cor` as the shared boolean field and put the unconfirmed reason in a separate field, or explicitly update both producers and the consumer contract? Android pins `cor` to `true` or `false`, while this widens iOS to reason tokens. The iOS contract table does not require `cor`, and the new resolver tests assert delivery but not the emitted token, so this contract change is currently unprotected.

Shahroz16 (changes requested):
Two contract fixes remain before approval. The new policy intentionally accepts a marginal inside fix unless a usable second fix contradicts it; I am treating that as the calibration trade-off already agreed, not reopening it. But the resolver and PR still state the old confirmation invariant, and `cor` now changes from the shared boolean contract to reason strings without a pinned iOS test or a corresponding Android or consumer update. Could we align the docs and keep or deliberately version the log contract? The pass-local cache fix itself is good.

Sources/LocationGeofence/Polygon/PolygonMembershipResolver.swift
Shahroz16: Confirmed on `6d4e67a`: `PassCorroboration` is created inside `runPass`, and the overlap regression makes each pass issue its own attempt. The cross-pass cache bug is fixed.

Sources/LocationGeofence/Polygon/PolygonMembershipResolver+Corroboration.swift
mrehan27: Fixed in 8d8bff23. The resolver contract now states the asymmetry directly: a marginal inside fix is accepted unless a usable second fix places the device outside, and such a verdict records `cor=false` with a `corwhy` reason so a capture never reads it as decisive. PR body updated the same way — it was still describing the old rule, which was my miss.

Sources/LocationGeofence/Logger+GeofencePolygon.swift
mrehan27: Agreed and reverted in 8d8bff23. `cor` is the shared boolean again — true only when a second fix agreed. The reason moved to `corwhy`: iOS-only, additive, and absent entirely on a decisive verdict.

Pinned both ways in `GeofenceLogTailTests`: `cor` is now a required key on `polygon.verdict`, and `verdictCorroboration_expectCorStaysBooleanAndTheReasonRidesSeparately` asserts `cor=false`+`corwhy=no_usable_fix` for an unconfirmed arrival and `cor=true` with no `corwhy` for a confirmed one.

You were right that this was unprotected, and worse than that — widening a key Android pins was a cross-SDK contract change I made without asking anyone. A later commit also notes at the emit site that `corwhy` tokens describe the SECOND fix, so `no_usable_fix` is not misread as the judged fix being bad.

Sources/LocationGeofence/Polygon/PolygonMembershipResolver+Corroboration.swift
Shahroz16: Thanks, this now matches the block-not-gate policy we agreed to test: an unavailable or unusable second fix no longer prevents the marginal arrival, while a usable fix on the other side still does.

Sources/LocationGeofence/Logger+GeofencePolygon.swift
Shahroz16: Thanks, this preserves the shared `cor` boolean and carries the iOS-only reason separately in `corwhy`. The new contract test covers both the confirmed and unconfirmed shapes.

Sources/LocationGeofence/Logger+GeofencePolygon.swift
```diff
@@ -83,15 +83,35 @@ extension Logger {
     /// records (`contradiction.allowed`) use the same key with the opposite sign, so a parser must
     /// read `edge` against the record's `ev`. `acc` is the ambiguity margin, not a quality score —
     /// iOS has no accuracy ceiling, so a verdict holds only while `|edge|` exceeds it.
```
Shahroz16: Could we update this line now that arrivals also have a per-fence accuracy ceiling? `resolvedOutcome` returns `accuracy_too_low` when accuracy is at least the polygon scale, so saying iOS has no ceiling will point future capture readers in the wrong direction. This does not need to block the calibration experiment.

Shahroz16 (approved):
The two requested contract fixes are addressed: the docs now match the block-not-gate policy, and `cor` remains the shared boolean with the reason in additive `corwhy`. I also checked the second-fix path against the agreed calibration scope and the green exact-head CI.

Sources/LocationGeofence/Logger+GeofencePolygon.swift
Shahroz16: Thanks, the updated text now matches the real decision order: the per-fence ceiling gates inside verdicts, while a fix clearly outside its own accuracy can still settle departure before that ceiling is read. I rechecked this against the current head.

Shahroz16 (approved):
The logger contract now matches the real per-fence ceiling and departure ordering. Rechecked against the current head.

PR #1285, opened 2026-09-16
test(geofence): the recorded-scenario format and its matcher

mahmoud-elmorabea:
**iOS replay harness — 1 of 5**

`base` → **#1285** → #1286 → #1287 → #1288 → #1289

The file format the harness reads, and the matcher that compares what the SDK emitted against what was expected.

Contributes to MBL-2309
Contributes to MBL-2421


Shahroz16 (approved):
Approving the scenario format and matcher. The exact head is clean in hosted checks, and I did not find a correctness issue in the parser or count matching. This remains test-only and is correctly stacked beneath the later replay composition.

mahmoud-elmorabea:
Force-pushed one fix from device testing — flagging rather than re-requesting, since this is approved.

`Scenario.Record.fenceId` now reads `ids` as well as `id`. Android writes `ids` on an OS callback because Play Services batches several fences onto one; CoreLocation never batches, so iOS writes `id`. The reader only knew `id`, so replaying a recorded Android drive through the iOS composition reported **every** callback as an unsupported input — the run graded nothing and said so only in the unsupported list.

A single-fence `ids` is the same record spelled differently and now resolves. A batched one still returns nil: this composition cannot receive it, and replaying only its first fence would silently grade a different drive.

Found by actually running an Android drive against the iOS harness. With the fix, that drive replays and produces the same decisions Android did — including a crossing the iPhone had missed in the field, which is how we learned the SDK logic was right and the device had died mid-request.

415 tests green; `swiftlint --strict` 0 violations across 789 files.

PR #1287, opened 2026-09-16
test(geofence): compose the real SDK over the replay seams

mahmoud-elmorabea:
**iOS replay harness — 3 of 5**

`base` → #1285 → #1286 → **#1287** → #1288 → #1289

Composes the real SDK over those seams, with a smoke test that fails loudly if the SDK is never wired to the OS seam.

Contributes to MBL-2309
Contributes to MBL-2421


Shahroz16 (approved):
Approving the replay composition. Hosted checks pass and the smoke coverage verifies the SDK is wired to the real OS seam without re-implementing decision logic in the harness.

mahmoud-elmorabea:
Rebase only — no content change. Carries fixes made in #1286, #1288 and #1289 from this review round; nothing in this PR's own diff moved.

mahmoud-elmorabea:

PR #1289, opened 2026-09-16
test(geofence): replay a recorded drive and grade the SDK against it

mahmoud-elmorabea:
**iOS replay harness — 5 of 5**

`base` → #1285 → #1286 → #1287 → #1288 → **#1289**

The runner that walks a scenario through the composition, and the suite that grades the SDK against it.

**This suite reports as skipped in CI.** Its inputs are supplied only when running locally; gated with `.enabled(if:)` rather than an early return, because an early return reports as passed.

Contributes to MBL-2309
Contributes to MBL-2421


Tests/LocationGeofence/Geofence/Replay/ReplayRunner.swift
```diff
@@ -0,0 +1,352 @@
...
+        // no-ops here, and letting them define window boundaries would fragment the very windows
+        // they belong inside.
+        let stimuli = scenario.when
+            .filter { record in
+                guard record.ev == "location.fix",
+                      let source = record.fields["prov"].flatMap(GeofenceLog.FixSource.init(rawValue:))
+                else { return true }
+                return source.isArrival
+            }
+            .map(\.at)
```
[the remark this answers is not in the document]
Shahroz16: Confirmed. The runner delivers the stable time order but returns raw file order to the matcher, whose `lastIndex` grouping assumes ascending stimuli. I am requesting that the returned list use the same stable ordering, with a small out-of-order regression case.

Shahroz16 (changes requested):
Requesting one matcher alignment fix. The runner delivers `scenario.when` through `stableByTime`, but returns `stimuli` in raw file order. `ReplayMatcher.groups` assumes the stimuli are ascending, so an out-of-order capture can be delivered one way and grouped another way. Could we build `stimuli` from the same stable ordering, or sort with the same tie-break, and add a small out-of-order test? The temporary filter that drops drives without `location.fix` is also worth removing or reporting once the corpus is trusted. Hosted checks are green.

Tests/LocationGeofence/Geofence/Replay/ReplayRunner.swift
```diff
@@ -0,0 +1,352 @@
...
+            stimuli: stimuli,
+            samples: pulled,
+            carried: carriedReads(scenario, on: harness)
+        )
+
+        // When the network answered, taken from the drive rather than modelled:
+        // `fixture.api.fetch` is stamped on arrival, so its `at` is exactly the release time.
+        harness.loadBoundaryAnswers(
+            fetch: scenario.given.filter { $0.ev == "fixture.api.fetch" }.map(\.at).sorted()
+        )
```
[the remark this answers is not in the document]
Shahroz16: The coordinator gate serializes refreshes, so a second fetch cannot be in flight while the first response is pending. The response order therefore matches fetch order in this composition; I did not find a reachable out-of-order pairing here.

Tests/LocationGeofence/Geofence/Replay/ReplayRunner.swift
```diff
@@ -0,0 +1,352 @@
...
+        harness.loadPulledFixes(
+            stimuli: stimuli,
+            samples: pulled,
+            carried: carriedReads(scenario, on: harness)
+        )
+
+        // When the network answered, taken from the drive rather than modelled:
+        // `fixture.api.fetch` is stamped on arrival, so its `at` is exactly the release time.
+        harness.loadBoundaryAnswers(
+            fetch: scenario.given.filter { $0.ev == "fixture.api.fetch" }.map(\.at).sorted()
```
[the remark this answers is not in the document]
Shahroz16: Equal release timestamps do not change the timestamp sequence, and the coordinator gate keeps responses serialized. The body queue and the fetch order therefore remain paired for this runner; the separate stimulus-order issue is the one I am requesting changes for.

Tests/LocationGeofence/Geofence/Replay/ReplayRunner.swift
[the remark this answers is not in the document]
mahmoud-elmorabea: No change — agreeing with @Shahroz16 that this is not reachable, and checking it against the corpus rather than only by reading the composition.

`fixture.api.fetch` timestamps are already ascending in all 37 scenarios, so the `.sorted()` is the identity function on every file we have. Combined with the coordinator gate serializing refreshes, there is no pairing this can get wrong today.

The stimulus-ordering issue on the other thread was real and is fixed.

Tests/LocationGeofence/Geofence/Replay/ReplayRunner.swift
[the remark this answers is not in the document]
mahmoud-elmorabea: No change, same reasoning as the release-time thread, with the corpus checked: no scenario has two `fixture.api.fetch` records sharing a timestamp, so the tie-break is never exercised. Equal timestamps would not reorder the sequence anyway, and the coordinator gate keeps responses serialized.

mahmoud-elmorabea:
Addressed in 685a74b2.

`stimuli` is now built from `stableByTime(scenario.when)`, the same ordering the runner delivers in, so `ReplayMatcher.groups` gets the ascending list its `lastIndex` lookup assumes. Added `run_givenScenarioLinesOutOfOrder_expectStimuliAscending`, which is a written case rather than a waited-for one: no recorded drive is out of order today, and `stableByTime` is stable, so file order and delivery order coincide on every current scenario.

The temporary `location.fix` filter is deleted. It matched nothing — all 8 iOS scenarios carry a `location.fix` — so no drive changes status, and the silent-drop behaviour is gone.

The two fetch-ordering findings are answered on their own threads as no-change, with the corpus checked.

Corpus run is unchanged from the pre-change tip: byte-identical output, the same three pre-existing issues on one drive and nothing else. 428 tests green without the corpus.

Tests/LocationGeofence/Geofence/Replay/ReplayRunner.swift
```diff
@@ -0,0 +1,357 @@
...
+        // The stimuli that actually drive work. Pull records are excluded deliberately: they are
+        // no-ops here, and letting them define window boundaries would fragment the very windows
+        // they belong inside.
+        //
+        // Ordered the same way they are *delivered*. `ReplayMatcher.groups` finds a decision's
+        // stimulus with `lastIndex { $0 <= record.at }`, which assumes the list ascends; handing it
+        // raw file order while the runner drove `stableByTime` meant a capture whose lines were not
+        // already sorted would be graded against boundaries that never happened in that order.
+        let stimuli = Self.stableByTime(scenario.when)
+            .filter { record in
```
Shahroz16, line 68: Could we exclude records that the runner accepts as true no-ops from these boundaries too? `device.state` and `app.background` change no SDK behavior, but they currently split both the fix-provider windows and matcher groups. A battery record landing during work started by the previous real input can then move its cache read or decision into a window the SDK never had. `app.foreground` should stay because it drives the re-arm; the two inert records should not.

Tests/LocationGeofence/Geofence/Replay/ReplayRunner.swift
```diff
@@ -0,0 +1,357 @@
...
+        // value would arrive after the decision it determined. See `ReplayFixProvider`.
+        let pulled = scenario.when.compactMap { record -> ReplayFixProvider.CachedRead? in
+            guard record.ev == "location.fix",
+                  let source = record.fields["prov"].flatMap(GeofenceLog.FixSource.init(rawValue:)),
+                  !source.isArrival, source != .bus
+            else { return nil }
+            // `prov=none` is a read that found nothing — a recorded answer, not a missing record.
+            guard let latitude = record.latitude, let longitude = record.longitude,
+                  let accuracy = record.fields["acc"].flatMap(Double.init)
+            else { return source == .none ? harness.emptyPull(at: record.at) : nil }
```
Shahroz16, line 48: Could we report a recognized pull record as unsupported when its coordinates or accuracy cannot be reconstructed? This `compactMap` silently drops it, then `deliverFix` returns true because it is a pull, so the run can stay green after losing a recorded input. `prov=none` is the valid empty read; the other pull sources should fail loudly when their required fields are missing.

Shahroz16 (changes requested):
The stable ordering fix addresses the original grading bug, and exact-head CI is green. I found two remaining fail-open replay paths in the current runner, so I am keeping this changes-requested until the matcher windows exclude inert records and malformed pull records are reported instead of silently dropped.

mahmoud-elmorabea:

Shahroz16 (changes requested):
I rechecked the restacked head. The stimulus sorting fixes file-order grading, but both existing blockers remain unchanged: inert device and background records still split the provider and matcher windows, and malformed non-none pulls still disappear while deliverFix reports them handled. The focused replay suite passes because neither regression is covered yet.

Tests/LocationGeofence/Geofence/Replay/ReplayRunner.swift
mahmoud-elmorabea: Done in d63dd9b8. `device.state` and `app.background` no longer bound a window; `app.foreground` stays, since it drives the re-arm. The rule moved into `isStimulus` so the exclusion is stated once and named, rather than living in an inline filter that read as "pulls only".

Checked against the corpus rather than only by reading it: the 8 iOS drives carry 19 `app.background` records between them (and no `device.state` — iOS captures do not write those), and with all 19 no longer splitting a window, every drive grades byte-identically to the previous head — same fetch counts, same cache-read accounting, same mismatch lists. So the fragmentation you described was reachable but latent here; `run_givenInertRecords_expectNoStimulusBoundary` is what holds it now, since no recorded drive would notice a regression.

Tests/LocationGeofence/Geofence/Replay/ReplayRunner.swift
mahmoud-elmorabea: Done in d63dd9b8. A recognized pull whose coordinates or accuracy cannot be reconstructed is now appended to `unsupported` at the point the read is lost, so the run fails instead of finishing one recorded cache read short.

Reported from the `compactMap` rather than from `deliverFix`, because that is where the loss actually happens — `deliverFix` never sees the missing fields, and it still returns true for a pull, so the record is counted once rather than twice.

`prov=none` is unchanged and still returns `emptyPull`: a read that found nothing is an answer, not a missing record. Two cases hold the pair — `run_givenPullMissingPosition_expectReportedUnsupported` and `run_givenPullThatFoundNothing_expectAccepted` — so this cannot be satisfied by rejecting every position-less pull.

Tests/LocationGeofence/Geofence/Replay/ReplayRunner.swift
```diff
@@ -0,0 +1,390 @@
...
+    /// `logReceivedCallback` reads the cache and *then* logs, so a callback record and the
+    /// `location.fix` line for its own read are written microseconds apart — and a capture that
+    /// stamps to the millisecond can put them on either side of a tick. The provider uses these
+    /// only for a window the recorded reads left empty; see `ReplayFixProvider`.
+    private static func carriedReads(_ scenario: Scenario, on harness: ReplayHarness) -> [ReplayFixProvider.CachedRead] {
+        scenario.when.compactMap { record -> ReplayFixProvider.CachedRead? in
+            guard record.ev == "os.callback", record.fields["fixsrc"] != nil else { return nil }
+            guard let latitude = record.latitude, let longitude = record.longitude,
+                  let accuracy = record.fields["acc"].flatMap(Double.init)
+            else { return harness.emptyPull(at: record.at) }
```
Shahroz16: Could we apply the same fail-closed check to the callback-carried read here? A callback with a non-none fix source but missing coordinates or accuracy is currently turned into an empty pull, so the SDK receives nil and both unsupported-input and pull-accounting checks stay clean. That gives the harness the same false-green path this commit fixes for standalone pull records. The valid empty shape is fixsrc=none; could we report the other incomplete callback shapes and add the matching regression pair?

Shahroz16 (changes requested):
The two original blockers are fixed, and the four executable simulator regressions pass. One symmetric fail-open remains in callback-carried reads, so I’m keeping this changes-requested narrowly on that case.

Tests/LocationGeofence/Geofence/Replay/ReplayRunner.swift
Shahroz16: Thanks, this is resolved on d63dd9b. The runner now excludes both inert records from the shared stimulus list, and the new control keeps app.foreground as a real boundary.

Tests/LocationGeofence/Geofence/Replay/ReplayRunner.swift
Shahroz16: Thanks, the standalone pull path is fixed on d63dd9b, and the valid prov=none shape is covered separately. I found the same incomplete-record case still open for the callback-carried fallback, so I’ve kept the review narrow on that path.

Tests/LocationGeofence/Geofence/Replay/ReplayRunner.swift
mahmoud-elmorabea: Done in eb496a9d. `carriedReads` now returns what it could not rebuild alongside the reads, and `run` folds that into `unsupported`, so an incomplete callback-carried read fails the run instead of becoming an empty pull. `fixsrc=none` stays legal exactly as `prov=none` does on the standalone path, and the pair `run_givenCallbackCarriedReadMissingPosition_expectReportedUnsupported` / `run_givenCallbackCarriedReadThatFoundNothing_expectAccepted` holds both halves.

One thing worth flagging, because my first attempt got it wrong. I initially gated the whole path on the source parsing, reasoning that an unreadable `fixsrc` is the same class of loss. That dropped three valid reads and turned `first-arrival-and-crossing` red: the conformance scenarios carry `fixsrc=os_trigger`, which is the shared cross-platform vocabulary rather than one of this SDK's own `FixSource` cases. The source is now consulted only to settle whether a *missing* position is legitimate, so a complete read still carries whatever it names. That is narrower than what I first wrote and matches what you actually asked for.

Entries from this path carry their `fixsrc`. One `os.callback` can fail as a carried read and as a delivery independently, and two bare copies of the same `kind ev@at` would not say which failed.

No drive exercises this: all 135 recorded `os.callback` records carry a complete `manager_cache` read, so the corpus grades identically to d63dd9b8 and the new cases are the only thing holding the behaviour. 434 tests, with `ios-iphone15-drive5-1350` still red on its same three pre-existing issues.

Shahroz16 (approved):
Thanks, the callback-carried path now fails closed on the malformed non-`none` shape while preserving `fixsrc=none` and complete cross-platform sources. I reran the focused simulator suite at this head and all 8 cases passed. This resolves the remaining blocker.

PR #1290, opened 2026-09-16
chore(geofence): record which pass produced a polygon verdict

mrehan27:
`PolygonEvaluationReason.movement` and `.foreground` existed as log tokens but nothing ever emitted them. Only the newly-registered path passed a reason, so every verdict from the whole-set pass — the movement wake and the foreground scan both — was recorded with no provenance. Reading yesterday's field capture meant attributing verdicts to passes by guessing at their timing.

`evaluateAllPolygons` now takes the reason and logs `polygon.pass.started`, the counterpart to the existing `polygon.pass.skipped`, whose own comment says it exists so a pass that produced no verdicts can be told apart from one that never ran. Pass-level rather than one record per polygon: a long stationary capture is read by asking whether anything evaluated at all, and N lines per pass buries that.

Logged before the empty guard, so `n=0` records a pass that ran with nothing registered. That was previously a silent return.

The new parameter has no default. A caller that can omit provenance is how the two tokens came to be dead.

Logging only — no change to detection. 640 tests pass, lint clean.

Stacked on [#1284](https://github.com/customerio/customerio-ios/pull/1284) (polygon arrival rule), which targets `feature/geofence-polygons`.

[MBL-2433](https://linear.app/customerio/issue/MBL-2433)


Sources/LocationGeofence/Polygon/PolygonMembershipResolver.swift
```diff
@@ -208,7 +209,10 @@ final class PolygonMembershipResolver {
         let registered = await storage.getRegisteredBusinessIds()
         let polygons = await storage.getCachedGeofences()
             .filter { registered.contains($0.id) && $0.vertices != nil }
-        guard !polygons.isEmpty else { return }
+        // Before the empty guard on purpose: `n=0` is the record that a pass ran and had nothing
+        // to judge, which is otherwise a silent return.
+        logger.geofencePolygonPassStarted(reason: reason, count: polygons.count)
```
Shahroz16: For a follow-up, could we carry a pass ID into the verdict records, or note that overlap is still ambiguous? A movement pass can run alongside an in-flight foreground pass, so two started rows followed by interleaved verdicts still don't identify which pass produced each one. The start/count record is useful as-is; it just doesn't yet provide per-verdict attribution in that case.

Shahroz16 (approved):
Approving the logging-only delta: both production callers pass the correct reason, empty passes are visible, and hosted tests pass. The completed Claude Code Opus review found no new detection bug here. This is not approval of the full stack: the departure and pass-order issues on #1284 still need addressing before this lands. One non-blocking attribution limitation is noted below.

Sources/LocationGeofence/Polygon/PolygonMembershipResolver.swift
mrehan27: Done in 0d8be826. Each pass now gets an id from a resolver-level counter, and every verdict carries it as a `pass=` key, so interleaved verdicts attribute to their `polygon.pass.started` row.

Shahroz16 (approved):
Approving the pass provenance logging on the current head. Both production callers provide an explicit reason, including empty passes, and hosted checks pass. The `pass` field remains attached to verdict records only, which is enough to attribute decisions without changing detection.

PR #1291, opened 2026-09-16
fix(geofence): let a caller substitute the movement-fix timeout wait

mahmoud-elmorabea:
**iOS replay harness — SDK change, on top of the stack**

`base` → #1285 → #1286 → #1287 → #1288 → #1289 → **this**

The only production change in the series, kept in its own PR so the harness PRs stay test-only.

`MovementFixResolver` already accepts `waitForTimeout`, but it is constructed inside `CLMonitorGeofenceMonitor`, which neither forwards the seam nor exposes one — so no caller can reach it, and the default is a real ten-second sleep. It is now settable after construction, exactly as the neighbouring `requestFreshFix` already is. Production behaviour is unchanged.

Forwarding it as an initialiser parameter is the tidier shape and was the first attempt, but `CLMonitorGeofenceMonitor.swift` sits exactly on the 400-line `file_length` cap and cannot take one without being split.

Found by review of the harness stack. Verified against the recorded corpus: the deliberate red drive goes from three failures to two, because the replay now completes a sync it previously abandoned.

Contributes to MBL-2309
Contributes to MBL-2421


Shahroz16 (approved):
Approving the narrow timeout seam. The production default is unchanged, the current head passes hosted checks, and the known `Task.yield` replay limitation is documented with no corpus path currently depending on a fresh fix. When the replay stack is integrated with the polygon resolver, the fresh-fix path should use the same virtual boundary rather than this fallback.

mahmoud-elmorabea:
Rebase only — no content change. Carries fixes made in #1286, #1288 and #1289 from this review round; nothing in this PR's own diff moved.

mahmoud-elmorabea:

PR #1292, opened 2026-09-16
fix(deps): pin LDSwiftEventSource to 3.3.0

mahmoud-elmorabea:
`LDSwiftEventSource` 3.3.1 was published 2026-09-16 and raised its platform floors **in a patch release** — iOS 11→15, macOS 10.13→12, watchOS 4→9, tvOS 11→15. This package supports iOS 13, so it cannot build against it:

```
error: The package product 'LDSwiftEventSource' requires minimum platform version 15.0
for the iOS platform, but this target supports 13.0
```

`Package.resolved` is gitignored, so every fresh resolve picks the new version up. Every open iOS PR went red within hours of the release, and `main` would have on its next run.

Both dependency declarations are pinned, because both were affected:

- **`Package.swift`** — `.exact("3.3.0")`. `.upToNextMinor(from: "3.3.0")` would not help: 3.3.1 is a patch, so that range still admits it.
- **`CustomerIOMessagingInApp.podspec`** — `= 3.3.0`. It asked for `~> 3.3` (>= 3.3, < 4.0) and 3.3.1 is on CocoaPods trunk, so pod consumers hit the same wall. `~> 3.3.0` would not help either — in CocoaPods that means >= 3.3.0, < 3.4.0.

**Verified.** Clean resolve before the change picks 3.3.1, after it picks 3.3.0, with no other pin moving. A simulator build goes from exit 65 (the error above, on `CioMessagingInApp` and `CioMessagingInbox`) to exit 0. `pod ipc spec` reports ios 13.0 and `LDSwiftEventSource = 3.3.0`.

Unpin when our own iOS minimum reaches 15.0, or if LaunchDarkly restores the older floors in a later release.

Worth a separate conversation: `Package.resolved` being gitignored is why an upstream release could break every branch at once, and why nothing in the repo recorded which version we had all been building against.

Contributes to MBL-2309


Shahroz16 (approved):
Verified that a fresh SwiftPM resolve now selects 3.3.0 without moving the other dependencies, and the CocoaPods install also uses 3.3.0. The upstream runtime source is unchanged between these releases. Hosted builds and tests pass, and the completed Claude Code Opus review found no blocking issue.

PR #1293, opened 2026-09-17
chore(geofence): record every dropped callback and the OS's own condition list

mrehan27:
Two paths discarded an OS region callback with no record at all. Both sit before `logReceivedCallback`, so neither appears in the received count nor the dropped count — the two numbers that separate "the OS never reported it" from "we discarded it".

- An event for a condition this process does not own.
- The oldest queued event once the pre-bootstrap queue hits its cap.

Both log as `info` rather than `os.callback.dropped`: they have no receipt to net against, and filing them as drops would inflate that count.

Separately, `monitoredRegionIdentifiers` returns `ownedRegionIdentifiers`. Every registration record so far has stated our own belief, never the OS's state, so `registration.adopted n=13` means "we think thirteen". A condition we own that CLMonitor does not hold is monitored by nobody while every log says it is registered. `condition_mirror` now logs both counts each sync and names the difference; `missing` is the field that distinguishes the two cases.

Prompted by the 09-17 drive: one fence received no callbacks all day while the device sat inside its covering circle for 26 minutes and three sibling fences fired normally. Nothing we log today can say whether the OS held that fence and never promoted it, or the add never landed.

Diagnostics only, no behaviour change. 55 production lines.

Not unit-tested: CLMonitor cannot be instantiated in a unit test without aborting the test process, which is why this file has no suite today. Verified by full-SDK build and the 618-test LocationGeofence suite.

[MBL-2480](https://linear.app/customerio/issue/MBL-2480/mobile-fix-silently-dropped-geofence-callbacks)


Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor+Registration.swift
```diff
@@ -206,9 +206,41 @@ extension CLMonitorGeofenceMonitor {
...
+                ("missing", Self.identifierList(owned.subtracting(atOs))),
+                ("extra", Self.identifierList(atOs.subtracting(owned)))
+            ])
+        }
+    }
+
+    /// `nil` when empty, so the common no-drift case costs two keys rather than four. Sorted so a
+    /// capture diffs cleanly across passes.
+    private static func identifierList(_ identifiers: Set<String>) -> String? {
+        identifiers.isEmpty ? nil : identifiers.sorted().joined(separator: ",")
```
Shahroz16: Could we use `GeofenceLog.list` here and add `missing` and `extra` to `GeofenceLog.composedKeys`? The tail currently treats both keys as ordinary values and replaces commas with underscores, so two identifiers become one ambiguous token. The shared helper also preserves per-identifier sanitization and the list cap. A small tail-contract test for `condition_mirror` would catch this.

Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor+Registration.swift
```diff
@@ -206,9 +206,41 @@ extension CLMonitorGeofenceMonitor {
...
+    /// Every registration record until now asserted our own belief: `monitoredRegionIdentifiers`
+    /// returns `ownedRegionIdentifiers`, so `registration.adopted n=13` means "we think thirteen",
+    /// never "the OS holds thirteen". A condition we own but CLMonitor does not hold is monitored
+    /// by nobody while every log says it is registered — and in a capture that is indistinguishable
+    /// from the OS holding it and never promoting a crossing. `missing` is the field that separates
+    /// those two, which is the whole reason this exists.
+    ///
+    /// Enqueued rather than read inline so it observes the adds this sync just queued instead of
+    /// the state before them, and so it cannot block the caller on the monitor actor.
+    func logConditionMirrorDrift() {
```
Shahroz16: Could we return before enqueueing this when diagnostics are disabled? Every `setMonitoredRegions` call currently adds a `monitor.identifiers` actor hop and set work to the registration FIFO even though `geofenceInfo` later drops the diagnostic tail. That makes this diagnostics-only change delay later registration operations for normal SDK users.

Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor+Registration.swift
```diff
@@ -206,9 +206,41 @@ extension CLMonitorGeofenceMonitor {
...
+        enqueueMonitorOperation { [weak self] monitor in
+            guard let self else { return }
+            let atOs = Set(await monitor.identifiers)
+            let owned = self.ownedRegionIdentifiers
```
Shahroz16: Could we snapshot `ownedRegionIdentifiers` before enqueueing this operation? A second `setMonitoredRegions` call can mutate ownership synchronously while this record is still queued behind the first sync’s OS work. The closure then compares the second sync’s desired set with the first sync’s OS state and reports false `missing` or `extra` drift. Capturing the set when the log is queued keeps both sides on the same sync generation.

Shahroz16 (changes requested):
Requesting changes on three parts of the new condition-mirror record. As emitted, `missing` and `extra` lose their separators, the queued comparison can mix two sync generations, and the CLMonitor read is still added to the registration FIFO when diagnostics are disabled. Those make the diagnostic unreliable and add production work despite this being diagnostics-only. The callback and stream-resubscribe records otherwise preserve behavior.

Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor+Registration.swift
mrehan27: Fixed in a7a3b986. Both keys now go through `GeofenceLog.list` and are declared in `composedKeys`.

Confirmed by mutation before the fix: without the declaration the record emitted `missing=alpha_beta` — two conditions as one token, in the field whose only job is naming which condition the OS does not hold. `conditionMirror_expectIdentifierListsKeepTheirSeparators` pins it and fails against the old behaviour.

Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor+Registration.swift
mrehan27: Fixed in a7a3b986 — `guard GeofenceDiagnostics.isEnabled else { return }` before the enqueue, so a device with diagnostics off queues nothing at all.

Agreed this mattered: diagnostics-only should cost normal users nothing, and the actor hop was landing on the registration FIFO ahead of real monitor work regardless of the gate.

Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor+Registration.swift
mrehan27: Fixed in a7a3b986 — `ownedRegionIdentifiers` is snapshotted before the enqueue, so both sides of the comparison come from the same sync.

Peer review also checked the snapshot point itself: ownership can still change before the op drains, but every path that changes it queues its OS work behind this one, so the closure still compares this sync's ownership against this sync's OS state.

A follow-up the same review found, fixed in 9379fbad: the doc claimed `missing` answers "monitored by nobody". It does not — a condition the OS gives up on stays in `CLMonitor.identifiers` and we keep ownership deliberately, so it sits in both sets and `missing` stays empty. Comment now states only what the field catches and points at the `.unmonitored` record for the other case.

Shahroz16: Thanks for checking the startup reconcile window. Reading both sets inside the closure fixes that case, but it reopens the two-sync case from this thread. If sync A stages `A` and queues this record, then sync B changes ownership to `B` before A drains, A’s record reads OS `{A}` against ownership `{B}` and reports `missing=B, extra=A` even though both syncs are healthy. `known=A` explains the timing but does not make those fields accurate. Could we keep the comparison scoped to one sync generation and handle startup reconciliation separately?

Shahroz16 (changes requested):
The three original fixes are present, but the latest ownership read reintroduces the two-sync generation mix. A later sync can change ownership synchronously while its OS work remains queued behind the earlier sync’s mirror record, so that record reports healthy staged changes as `missing` and `extra`. The added `known` count explains the mismatch but does not make those fields truthful. Could we scope both sides to one sync generation and cover the A-then-B overlap?

Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor+Registration.swift
mrehan27: Fixed properly in 4e71a2a6 — you were right that `known` explained the mismatch without making the fields truthful.

Ownership was the wrong side to compare against in either direction: a later sync mutates it synchronously while its OS work is still queued behind this record, and `reconcileKnownConditions` unions persisted identifiers into it on the first pipeline operation. A pre-enqueue snapshot mixes with the second; an in-closure read mixes with the first. Picking either just chose which false reading to emit.

The record now compares this call's own `desired` set against `CLMonitor.identifiers`. That set is a value captured by this call and cannot move, so there is no generation to mix — the variable is removed rather than compensated for. `missing` now means exactly 'this sync asked the OS for it and the OS does not list it'.

On covering the A-then-B overlap: the comparison is extracted to `ConditionMirror.drift`, free-standing with no availability gate, with unit tests — `CLMonitor` itself cannot be instantiated in one. There is deliberately no overlap test, because ownership is no longer an input and the overlap cannot be expressed; I would rather say that than add a test that passes for the wrong reason.

`known` is gone. The diagnostics early-return you flagged earlier is still in place at the top of the method.

Shahroz16: Thanks, scoping the comparison to this sync fixes the A/B generation mix. There is still one reachable false-positive path: `desiredIdentifiers` includes registrations that `startMonitoring` deliberately rejects for blocked permission or invalid coordinates. Those identifiers are removed, never added, and then reported as `missing`, even though the SDK never asked the OS to hold them. Could we capture this sync’s accepted target set, or subtract the rejected identifiers, and add a regression case for one rejected region?

Shahroz16 (changes requested):
The generation-scoping fix addresses the original overlap problem. I am still holding approval on the rejected-registration case in the existing thread, because it currently reports an SDK refusal as OS drift. Exact-head CI is green.

Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor+Registration.swift
Shahroz16: Thanks, this now scopes `missing` to the conditions this sync actually submitted and reports the SDK-refused count separately. The target tests pin both halves, so a blocked or invalid registration no longer looks like OS drift.

Shahroz16 (approved):
The accepted target now excludes SDK-refused registrations from OS drift and keeps them visible through the separate refused count. Exact-head source review and hosted tests are clean; one sample compile remains queued.

Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor+Registration.swift
mrehan27: Fixed in `ce3a3e50`. The record now compares the accepted set — `desired` intersected with ownership — since ownership is only taken once both of `startMonitoring`'s guards pass. A refused region is no longer reported as `missing`.

Scoping alone would have made it worse, so the record also carries `refused`. Only one of the two refusal paths says anything elsewhere: unusable coordinates log per identifier, but blocked permission logs an edge-triggered tier record that names no region and stays silent on later syncs, and `geofenceRegistrationDiff` counts a refused region in neither `added` nor `removed`. Without the count, N refused registrations would have produced a completely clean record — silence instead of a wrong attribution. The key is absent when nothing was refused.

Regression tests: `accepted_givenOneRegistrationRefused_expectItIsNotReportedMissing` asserts the refused id clears `missing` and that the raw desired set still produces it, plus `target_givenEverythingAccepted_expectNothingCountedRefused` for the zero case. Both mutation-verified — they fail on different mutants.

PR #1294, opened 2026-09-17
chore(geofence): sample the condition mirror on a timer

mrehan27:
`condition_mirror` could not observe the failure it was added for. It is only emitted by `setMonitoredRegions`, a sync only runs off a wake, and the failure is the absence of wakes. On the 09-17 drive the last reading landed ten seconds before region callbacks stopped and the next came two hours later — across a window where the process was alive and taking CLVisit callbacks.

A slow sampler now runs alongside it, and every record says which occasion produced it (`at=sync|poll`).

Each sample writes two records. `mirror_beat` is written synchronously; the comparison against the OS has to run on the monitor pipeline, because `CLMonitor.identifiers` is only reachable from there. That pipeline is serial and region callbacks come off the same actor, so a probe reporting only from there goes silent in exactly the failure it exists to catch. Beats without comparisons is now a positive reading for a wedged pipeline. It reads one wedge shape, not all of them: a wedge that blocks the MainActor rather than suspending on it still looks like a healthy process.

The baseline is the ledger's staged identifiers, not `live`. `retire` keeps live generations so late events can still be attributed, so a deliberately removed fence would otherwise read as wanted for the rest of the process. `want` and `owned` are captured at one instant; the gap counts refused registrations on a sync, and the process-start divergence on a sample.

A sleep rather than a timer: neither runs while the process is suspended, but an overdue sleep resumes on the next slice of runtime something else earns us, which is the only time a sample is possible. The interval is a floor, never a guarantee.

Diagnostics-only, behind `CIOGeofenceDiagnostics`. 626 tests; the new ledger assertion fails if the baseline is keyed on `live`.

Stacked on #1293 (`mbl-2480-monitor-ingress-observability`), which is still open — merge that first.

Part of [MBL-2480](https://linear.app/customerio/issue/MBL-2480/mobile-fix-silently-dropped-geofence-callbacks)


Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor+ConditionMirror.swift
```diff
@@ -0,0 +1,158 @@
...
+    /// difference is diagnostic, not noise: the ledger begins each process empty while ownership
+    /// does not, so a condition the OS holds and we own, but which this process never staged —
+    /// an adopted record with no geometry, or the `userChangedDuringBootstrap` branch — reads as
+    /// `extra` on every sample until a sync re-registers it. With `owned` present a reader can
+    /// tell that from "the OS is holding something nobody wants".
+    ///
+    /// The gap means different things on the two occasions, so read it with `at`. On `at=poll` it
+    /// is the process-start divergence above. On `at=sync` it is refused registrations:
+    /// `startMonitoring` bails to `enqueueConditionRemoval` ahead of its ownership insert when
+    /// permission is blocked or coordinates are unusable, so the region stays in `desired` and
```
Shahroz16: Could we keep the OS comparison based on conditions this sync actually submitted, and expose rejected registrations separately? The permission and coordinate guards return before `monitor.add`, so these are SDK refusals, not conditions the OS turned down. With blocked permission, the current `desired` set still reports every region as `missing` OS drift. Capturing the accepted set for `want`/`missing` and adding a rejected count would preserve both diagnostics without conflating them.

Shahroz16 (changes requested):
The periodic beat and pipeline comparison are useful, but this currently carries forward #1293’s rejected-registration false positive. I am holding approval until the OS comparison is based on the set actually submitted to CLMonitor, with SDK refusals reported separately. Hosted CI is still finishing one Xcode preview check.

Sources/LocationGeofence/Monitor/CLMonitorGeofenceMonitor+ConditionMirror.swift
Shahroz16: Thanks, the stacked sampler now inherits the accepted target set and keeps refusals in a separate count on sync records. I rechecked the queue ordering and poll path against the current base, and the OS drift fields now keep the intended meaning.

Shahroz16 (approved):
The periodic sampler preserves the accepted/refused split and the FIFO ordering keeps each comparison aligned with its captured target. No blocking issue found on the current stacked head; remaining hosted checks are still running.

PR #1295, opened 2026-09-18
test(geofence): restore the pass-provenance resolver tests

mrehan27:
Restores three tests a strategy merge dropped.

https://linear.app/customerio/issue/MBL-2433

## What

When #1290 was merged, `git merge -X theirs` took `PolygonMembershipResolverTests.swift` wholesale and three of its tests went with it:

- `evaluateAllPolygons_givenAForegroundPass_expectTheReasonRecorded`
- `evaluateAllPolygons_givenAMovementPass_expectTheReasonRecorded`
- `evaluateAllPolygons_givenNothingRegistered_expectAPassRecordWithZero`

They assert the resolver emits `polygon.pass.started` with the right reason and count, including the zero-polygon case.

## Why it matters

The behaviour shipped fine — all three emission sites survive in `PolygonMembershipResolver.swift`. Nothing pinned them. Comment out all three and the suite still passes.

What remains is the log-tail invocation in `GeofenceLogTailTests`, which calls the logger directly with hardcoded values. It pins that the keys exist. It cannot notice that nothing calls it.

The test count could not reveal the loss either: the file was taken wholesale, so the name set was the same 166 with or without them.

## Evidence

Restored byte-identical to the pre-merge head, then mutation-verified:

| | total | passed | failed |
|---|---|---|---|
| unmutated | 2117 | 2117 | 0 |
| three emission sites commented out | 2117 | 2114 | the 3 |

Build exit 0 before the mutated run, so it was not a stale binary. No cascade — only those three fail.


Shahroz16 (approved):
The restored tests exercise the production pass-start records for foreground, movement, and the zero-polygon path, and the hosted test suite is green.

PR #1296, opened 2026-09-18
chore(geofence): refresh the catalog on any crossing, not only the trigger exit

mrehan27:
Catalog refresh depended on a single input. When that input is lost the cache stays stale with nothing to recover it.

## The problem

Refresh has exactly one trigger: the movement trigger's EXIT. Enumerated every entry point in `Sources/LocationGeofence/` — app launch/identify, and that EXIT. No timer, no background refresh, nothing periodic.

Measured on 09-18: a 38.7-minute callback silence left the device monitoring another city's fences for **3 hours**, ended only by the user opening the app. That 3 hours is a lower bound — stationary with the app unopened, it does not recover at all.

## The change

A business-fence crossing is also evidence the device moved, and those callbacks already arrive; we just ignored them for refresh purposes.

Routed through `refresh`, not `handleMovement`:
- `refresh` consults the same decision table as app launch and answers `.skip` when nothing has moved or aged, so a crossing that warrants no work costs one storage read.
- `handleMovement` assumes an EXIT happened and always re-registers, which on every crossing would re-arm the trigger continuously.

## Evidence

| | total | passed | failed |
|---|---|---|---|
| unmutated | 2116 | 2116 | 0 |
| refresh disabled (build exit 0 first) | 657 | 656 | the 1 positive test |

The two guard tests survive the mutation by design — they assert refresh does *not* fire, so removing it cannot break them. Only the test pinning the new behaviour fails.

One existing test asserted the coordinator stayed idle on business crossings, which this makes false; it now pins that `handleMovement` stays at 0 while `refresh` fires once.

Android has the same single-input structure and costed the same fix.


Sources/LocationGeofence/GeofenceMonitorBinder.swift
```diff
@@ -54,6 +56,27 @@ enum GeofenceMonitorBinder {
...
+            //
+            // Routed through `refresh`, not `handleMovement`: `refresh` consults the same decision
+            // table as app launch and answers `.skip` when nothing has moved or aged, so a crossing
+            // that warrants no work costs one storage read. `handleMovement` assumes an EXIT
+            // happened and always re-registers, which on every business crossing would re-arm the
+            // trigger continuously.
+            guard let location else { return }
+            Task {
+                await backgroundTaskRunner.withBackgroundTime {
+                    _ = await coordinator?.refresh(
```
Shahroz16, line 73: Could we resolve a current movement fix before using this callback location as the refresh anchor, and pass its freshness through? Both monitors deliberately mark business-callback locations as not fresh because they can be hours-old cached values. refresh can rank fences and persist the registration center from this point, and anchorIsLiveFix=false also expands the movement trigger to the full local radius. That can plant the trigger around a stale point and undo the tighter polygon wake radius, leaving the same stale-catalog gap this is meant to recover from. The movement-trigger path already uses MovementFixResolver, and a production-coordinator test could cover the chosen anchor and radius.

Shahroz16, line 73: Could we also make sure a trigger EXIT cannot be dropped while this new refresh owns the shared gate? A business crossing and trigger EXIT can arrive from the same movement, these independent Tasks race for refreshInProgress, and the loser returns alreadyInProgress without a retry. If handleMovement loses, the only path that freshens and re-centers the trigger is gone. A concurrent test with the real coordinator would pin that recovery path.

Shahroz16 (changes requested):
The fallback refresh is worthwhile, but these two paths can recreate the same lost-trigger condition it is meant to recover from.

Sources/LocationGeofence/GeofenceMonitorBinder.swift
mrehan27: Valid, and fixed in #1297 (kept separate as it changes `handleTransition`'s signature).

Confirmed both dispatch sites pass `locationIsFresh == false`, so the anchor could widen the trigger to the full refresh radius. #1297 carries the fix the membership pass already resolved and passes `anchorIsLiveFix: true`. Rig capture shows `movement.radius.chosen rad=100 from=fix` where the parent commit re-armed nothing.

mrehan27: Valid, and fixed in #1298 (separate, as it changes the coordinator's gate).

Confirmed there is no retry at any of the three `alreadyInProgress` sites. #1298 records the losing movement and replays it on release; refresh is deliberately not deferred, since the holder refreshes the same catalog from a fix of the same moment. Mutation test fails with the trigger stranded at the old coordinates.

mrehan27: Correction to my reply above — I said this was fixed in #1297, and that was too broad.

#1297 resolves a real fix and passes `anchorIsLiveFix: true` only in the `.circleEntered` branch. An ordinary business crossing takes `.nothingToRearm`, which still refreshes on the callback location with `anchorIsLiveFix: false`. I diffed #1298 and #1299 against the binder too; neither touches that branch.

Splitting your point in two, because they landed differently:

- Ranking, fetching and persisting the registration centre from a non-live anchor is what identify and app launch already do — `GeofenceModuleState` passes the stored registration centre with `anchorIsLiveFix: false`, and the comment there says that flag is the mitigation. Not new here.
- Widening the trigger is the real residual. When one of these crossings leads to a non-skip refresh, `wakeRadius` returns `localRefreshTriggerRadius` and replaces a tighter polygon-sized wake that was in force. Before this PR only a trigger EXIT or an app launch could do that. It is reachable when the catalog expires by time while the device sits near a boundary, which is the still-user case this work targets.

My read: accept it on this PR. The catalog freshness it buys is worth more than the wake width, a later circle entry re-arms tight through #1297, and the alternative — having a non-live refresh leave an existing trigger alone — changes the coordinator's contract for app launch as well, which is a bigger change than this PR should carry.

If you disagree I will make that change instead; say which and I will do it.

Shahroz16: Thanks for the correction. I agree the residual is a bounded wake-radius and efficiency tradeoff, not a correctness blocker here. #1297 covers the circle-entry live-fix path, and #1298 covers the movement lost behind the gate, so I'm clearing my block.

Shahroz16: Confirmed on the latest #1298 head. The losing movement is replayed, the coordinator suite passes 115/115, and the two new commits only strengthen those sequence assertions. Covered by the follow-up.

Shahroz16 (approved):
Rechecked this in stack context. #1298 covers the gate-loss path, and the remaining non-live wake-width behavior is bounded rather than a correctness blocker. Approving.

mrehan27:
Re-requesting review — this now carries a behaviour change you have not seen. You approved `fb5f1a05`; the tip is `46c7d8bc`.

I had your first point wrong twice, so the short version:

- My answer "fixed in #1297" was wrong. #1297 only covers the polygon covering-circle ENTER. Circle fences and polygon EXITs kept the raw callback anchor.
- I then argued the anchor was safe because app launch already refreshes on a non-live one. Also wrong: app launch passes the stored registration centre, and the comment there says anchoring on the location cache "ranks from a far-away old fix and clobbers the good registration with an empty set".
- `bestKnownFix()` applies no age gate, only a coordinate-validity filter, and its own doc says the OS cache can be hours old. On CLMonitor a mis-planted trigger self-heals with one corrective EXIT; on the classic path a trigger around a point the device is not inside never fires, so the wake is lost until foreground.

Fix in `8233189e`: when the caller has no live fix, `performRefresh` anchors on the registration centre rather than the passed coordinates. A time-expired catalog still refetches, and nothing can be re-ranked or re-planted from a point we cannot trust. No-op for the launch path, which already passes that centre. Mutation-checked: removing the guard fails `refresh_givenANonLiveAnchorFarFromTheRegistrationCentre_expectItDoesNotMoveAnything` and nothing else, and a second test pins that a live anchor still refetches, so the guard cannot degrade into never moving.

Also fixed: the crossing test now pins `anchorIsLiveFix == false` and passes `locationIsFresh: false` — the mock defaults it to true, so a binder hardcoding `true` passed. The no-location test is anchored on a positive barrier instead of bare yields. And two claims in the code comment were wrong: the skip path costs four cached reads, not one, and the three-hour figure now names the capture it came from.

On your second point: on this PR alone a crossing refresh can still drop a co-arriving trigger EXIT. #1298's defer-and-replay removes it and the two land in the feature branch together, so I am taking that rather than duplicating the deferral here.

659/659, lint 0.

Sources/LocationGeofence/GeofenceMonitorBinder.swift
mrehan27: Heads-up that I went further than this after you cleared it, so the thread does not read as the final state.

An independent review of the tip showed my "bounded tradeoff" framing was built on a wrong premise: `bestKnownFix()` has no age gate at all, and on the classic path a trigger planted around a point the device is not inside never fires, so it is a lost wake rather than a wider one. I fixed it instead of accepting it — `8233189e` anchors a non-live refresh on the registration centre. Details and the mutation result are in the PR comment below; re-requested since it is a behaviour change you have not seen.

Shahroz16 (approved):
Rechecked the three-commit update. Non-live business crossings now keep the established registration center, while live-fix and movement paths are unchanged. The new focused cases pass, and the adversarial re-review found no blocker. One tiny docs follow-up: the skip path now performs more than four storage reads after the extra registration-center lookup, so I would avoid pinning an exact count in that comment.

PR #1297, opened 2026-09-18
chore(geofence): re-arm the wake when a polygon's covering circle is entered

mrehan27:
Entering a polygon's covering circle leaves the device beside a boundary the OS cannot report, and the wake is sized only at registration time.

Stack: #1296 → **this** → (gate deferral, next)

## The problem

Measured in the field: a circle entry 45 m from the ring, then 11 minutes of silence. The arrival only landed when the user walked back to the car and the movement trigger fired.

The trigger in force was the full refresh radius, from a registration made kilometres earlier. Nothing re-sizes it on a circle entry.

`refresh` cannot fix this: `refreshAction` answers `.skip` unless the device moved a full refresh radius from the last registration centre, and `.skip` never touches the trigger. This revises the "routed through `refresh`, not `handleMovement`" reasoning in #1296 — `handleMovement` has a third tier, `performPolygonWakePass`, which re-arms against the nearest boundary with no ranking and no cache write.

## The change

`handleTransition` now reports whether a re-arm is due and carries the fix to size it with.

The fix must come from the membership pass, not the callback. Both monitor paths dispatch business events with `locationIsFresh == false`, and a non-live anchor makes the coordinator widen the trigger to the full refresh radius — the widest possible wake in the case needing the tightest. This is what Shahroz flagged on #1296.

Also runs `refresh` after the re-arm: `handleMovement` refetches on distance, never on age, so a time-expired catalog would otherwise stay stale.

No new region, no new permission, no background mode, and independent of the 100 m wake floor.

## Evidence

Simulator rig, same route and fence, with a control on the parent commit:

| build | circle ENTER | verdict | re-armed? |
|---|---|---|---|
| this | outside, 69 m from ring | `movement.radius.chosen rad=100 from=fix` | yes |
| parent | outside, 85 m from ring | none | no |

`registration.diff nadd=1 nrem=0 nkeep=12` — only the trigger re-added, all 12 business regions untouched, so no belief replay.

663 tests / 31 suites pass. Two mutations, both caught: passing the callback's location instead of the resolver's fix fails the provenance test; returning `.circleEntered` from the exit and no-fix paths fails both negative tests.

Not yet validated on device: the simulator delivered no movement-trigger EXIT on any route tonight, so whether the arrival is ultimately recorded is a device question.


Shahroz16 (approved):
The covering-circle ENTER now resolves a live fix before re-arming the movement trigger while preserving catalog refresh behavior. The focused LocationGeofence suite passes on the simulator. This increment looks good to me.

mrehan27:
Restacked onto the merged feature tip (`466d4cfd`). Two things changed in the process, neither behavioural:

- `PolygonMembershipResolver.swift` conflicted with #1302, which landed first. Resolved as a union: #1302's `pass:` argument on both `geofencePolygonUndecided` calls, and this PR's `.nothingToRearm` / `nil` returns. Taking either side whole would have dropped one of the two.
- This PR rewrites the comment block that #1296 corrected, so the corrections are carried forward here: the skip path costs four cached reads rather than one, and the three-hour staleness figure now names the 2026-09-18 capture.

Verified on the restacked tip: lint 0, build 0, 670/670.

PR #1298, opened 2026-09-18
chore(geofence): replay a movement pass that lost the refresh gate

mrehan27:
A business crossing and a movement-trigger EXIT routinely arrive from the same movement. Both take the coordinator's shared gate, and the loser was dropped.

Stack: #1296 → #1297 → **this**

## The problem

`refresh` and `handleMovement` share `refreshInProgress`. The loser returns `alreadyInProgress`, and there is no retry at any of the three sites.

When the loser is the movement pass, nothing re-centres the trigger. It stays on the circle the device just left, where no further EXIT can fire — the monitoring set freezes until the next app launch.

Raised by @Shahroz16 on #1296, which introduced the contention by routing business crossings into `refresh`. 8 `refresh_in_progress` events exist across our field captures, none yet next to a trigger EXIT, so the path is reachable but not observed.

## The change

Records the losing movement and replays it once the holder releases.

Only movement is deferred. A refresh that loses is redundant — the holder is refreshing the same catalog from a fix of the same moment. The movement pass is the only thing that re-arms the trigger.

Taken and cleared in one step, so a movement deferred during the replay is left for that replay's own release rather than lost or run twice. Dropped when the user changed under us, where `retryForCurrentUser` has already queued the right work.

The gate and its helpers move to their own file to stay under the file cap.

## Evidence

663 tests / 31 suites pass.

Mutation: disabling the replay fails the new test, and fails showing the trigger stranded at the original coordinates instead of following the device — the exact failure described above.

The existing `handleMovement_givenInFlightRefresh_expectAlreadyInProgress` still passes: short-circuiting is still correct, the return value is unchanged, only the recovery is new.


Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator.swift
```diff
@@ -146,9 +162,17 @@ final class GeofenceSyncCoordinatorImpl: GeofenceSyncCoordinator, @unchecked Sen
 
     func handleMovement(latitude: Double, longitude: Double, anchorIsLiveFix: Bool) async -> Result<Void, GeofenceSyncError> {
         guard acquireGate() else {
+            // Recorded, not dropped — see `deferredMovement`.
+            deferredMovement.wrappedValue = DeferredMovement(
```
Shahroz16: Could we make the failed gate check and deferred-movement publish atomic, or re-check and drain after publishing? A holder can release and drain after acquireGate() returns false but before this assignment runs. The movement is then stored with the gate already free and nothing guaranteed to drain it, leaving the trigger on the circle it just exited. A focused interleaving test around that window would cover it.

Shahroz16 (changes requested):
The replay idea is right, but the lost-wakeup window called out inline can still drop exactly the movement this PR is meant to preserve. I cannot approve this until the gate check and deferred publish are coordinated.

Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator.swift
mrehan27: Valid, fixed in 21083fe2.

`acquireGateOrDefer` now takes the gate and records the loss inside one `refreshInProgress` critical section. Either the record lands while the holder still holds the gate, so its release drains it, or the release got there first and this call takes the gate instead of deferring. Lock order is one-way: gate then deferral, and nothing takes them the other way.

`refresh`, `reset` and `applyCachedRegistration` keep the plain `acquireGate` — none of them defer.

On the test: I added two invariant tests rather than an interleaving one. The window is microseconds wide, so a thread race would pass against the broken code on nearly every run. The tests assert what the race violates instead — a call that takes the gate leaves no deferral behind, and a call that loses it leaves exactly one. Happy to add a stress test too if you would rather have one.

mrehan27: Follow-up in a91d372c: the same two-step shape was still there one line below, and my first test could not see it.

The clear that supersedes an older queued movement ran after the gate was taken, outside the section. A movement that lost the gate and recorded itself correctly could be wiped by the winner, and in that ordering the record being wiped is the newer one. Moved into the acquire branch.

The invariant test was shaped to the code — starting from nil it passed either way. It now seeds a queued movement first and fails when the clear is moved back out. Verified by mutation: moving it out fails that test and the existing supersede test.

Shahroz16 (approved):
The gate decision and deferred write now happen under the same lock, and the winner clears an older queued movement in the same critical section. I ran the focused coordinator suite on the current head, including the new regression invariants: 97/97 passed. The repository automated test suite is also green. This addresses my earlier concern.

Shahroz16 (changes requested):
Could we invalidate a drained replay when a newer movement has already run? `drainDeferredMovement` removes the old coordinates from the queue before its task starts, so a newer movement can acquire the free gate and re-center first. When the delayed replay eventually runs, it moves the trigger back to the older coordinates. I reproduced that ordering at this head with a deterministic coordinator test. A movement sequence carried into the drained task, and checked before replay, would keep the latest location authoritative.

Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator+Gate.swift
```diff
@@ -0,0 +1,126 @@
...
+            //
+            // A replay that loses the gate instead of being discarded is re-deferred with its
+            // original sequence, so the same comparison retires it at the next drain.
+            guard appliedMovementSequence.wrappedValue < pending.sequence else {
+                logger.geofenceSyncSkipped(reason: .movementOvertaken)
+                return
+            }
+            _ = await handleMovement(
+                latitude: pending.latitude, longitude: pending.longitude,
+                anchorIsLiveFix: pending.anchorIsLiveFix, sequence: pending.sequence
```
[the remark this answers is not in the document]
Shahroz16: I reproduced this on the latest head with a deterministic coordinator test. The replay passes the sequence check, a newer movement acquires and releases the gate, then the replay acquires the now-free gate and re-centres to the older coordinates. Could we make the freshness check part of the gate-acquisition decision, or re-check it after taking the gate?

Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator+Gate.swift
```diff
@@ -0,0 +1,158 @@
...
+    /// holder still holds the gate, so that holder's release drains it, or the release got there
+    /// first and this call acquires the gate instead of deferring.
+    ///
+    /// Lock order is one-way. This takes `refreshInProgress` then `deferredMovement`; nothing takes
+    /// them the other way round — `drainDeferredMovement` reads `deferredMovement` alone and fires
+    /// its replay outside the critical section.
+    func acquireGateOrDefer(_ movement: DeferredMovement) -> Bool {
+        refreshInProgress.mutating { inProgress in
+            if inProgress {
+                deferredMovement.wrappedValue = movement
```
Shahroz16: Could we preserve the highest-sequence movement when the gate is already held? I reproduced sequence 2 being queued, then a delayed replay of sequence 1 losing the gate and replacing it here. The drain subsequently re-centres to sequence 1, so the trigger ends at the older coordinates. Comparing the queued and incoming sequences before assignment would cover this ordering.

Shahroz16 (changes requested):
Thanks, the original drained-replay ordering is covered now. I reproduced two remaining cases where an older replay can still become authoritative: it can pass the freshness check before a newer movement takes the gate, and it can replace a newer movement already queued behind the gate. I added the focused cases inline.

Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator+Gate.swift
mrehan27: Valid, fixed in 74a709e4.

Highest sequence wins on defer now instead of last writer, so an older replay cannot displace a newer queued movement.

While fixing it I found the same problem one layer up, which Bugbot also flagged: the staleness test was made by the drain task before calling `handleMovement`, so a newer movement could take the gate and re-centre between the check and the acquisition. That is the same check-then-act shape this PR exists to close. Both the comparison and the acquisition now happen inside the `refreshInProgress` section. A fresh arrival can never lose that test, since its sequence postdates every applied one.

Third fix in the same commit: `noteMovementApplied` ran after every pass including failures, so a pass that re-centred nothing could retire a deferral that was still the best information available. Only on success now.

Tests: `acquireGateOrDefer_givenAQueuedNewerMovement_expectAnOlderReplayDoesNotReplaceIt` is your ordering. Plus `givenANewerMovementAlreadyApplied_expectOvertaken`, `givenAFreshArrivalAfterAnAppliedPass_expectItIsTaken`, and `handleMovement_givenTheMovementFailed_expectNoReCentreClaimed`. Mutation-checked separately — removing any one of the three fails a test and nothing else.

Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator.swift
mrehan27: Also worth flagging here: 49ebae96 moves the gated movement body from `GeofenceSyncCoordinator.swift` into `+Gate.swift`. No behaviour change — the sequence work left the file at 393 lines against a 400 cap, and restacked into #1299 it measured 413. `performMovement` becomes internal for the same split reason as its neighbours.

Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator+Gate.swift
```diff
@@ -0,0 +1,180 @@
...
+        }
+        let expectedUserId = identifiedUserId
+        let result = await performMovement(
+            expectedUserId: expectedUserId, latitude: latitude, longitude: longitude,
+            anchorIsLiveFix: anchorIsLiveFix
+        )
+        // Only on success, and before the release so a replay draining off it compares against
+        // this pass. A failed pass re-centred nothing, and claiming otherwise would retire a
+        // deferral that is still the best information available.
+        if case .success = result { noteMovementApplied(sequence) }
```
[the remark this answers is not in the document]
Shahroz16: Confirmed on the latest head. The remote fetch fails, the local fallback re-centres to the newer location, then the older replay is still accepted because the sequence was not recorded. My regression ends at longitude 0.01 instead of the locally re-armed 0.1. Could we track whether the trigger was actually re-centred separately from the remote result?

Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator+Gate.swift
```diff
@@ -0,0 +1,180 @@
...
+            inProgress = true
+            // A movement that actually runs supersedes any older one still queued: both describe
+            // the same journey and this one's coordinates are newer. Without it, a deferral
+            // recorded before this pass replays after it and moves the trigger BACK.
+            //
+            // Cleared in here, not after the guard. Outside the section it is the same two-step
+            // shape this method exists to close: a movement that lost the gate and recorded itself
+            // correctly would then be wiped by the winner, and in that ordering the record it
+            // wipes is the NEWER one — the premise above inverted.
+            deferredMovement.wrappedValue = nil
```
Shahroz16, line 210: Could we preserve a newer queued movement in the free-gate branch too? I reproduced sequence 2 waiting in the queue, then an already-drained sequence 1 replay acquiring the temporarily free gate here and clearing sequence 2. The final trigger moves to sequence 1’s older coordinates. The held-gate comparison fixes the earlier case, but this unconditional clear still demotes the queue.

Shahroz16 (changes requested):
Thanks, the two previously reported interleavings now pass. I reproduced two remaining cases at this head: a failed remote refresh can still re-arm locally without advancing the applied sequence, and an older replay taking a free gate can clear a newer queued movement. I added both exact cases inline.

Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator+Gate.swift
mrehan27: Valid, fixed in f5c53c27.

The free-gate branch now clears only what the incoming movement outranks, mirroring the held-gate comparison. `acquireGateOrDefer_givenAQueuedNewerMovement_expectAFreeGateDoesNotClearIt` is your ordering, and `givenAQueuedOlderMovement_expectAFreeGateClearsIt` pins the other direction so the fix cannot degrade into never superseding.

Same commit also takes Bugbot`s point that success was the wrong signal for "the trigger moved": a failed remote refresh re-arms from cache before returning its failure, so keying on success left `appliedMovementSequence` behind the real trigger and let an older replay through the overtaken test. `performMovement` now reports the re-centre separately from the result.

674/674 across four consecutive runs.

Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator.swift
```diff
@@ -223,7 +247,12 @@ final class GeofenceSyncCoordinatorImpl: GeofenceSyncCoordinator, @unchecked Sen
             logger.geofenceSyncSkipped(reason: .refreshInProgress)
             return .failure(.alreadyInProgress)
         }
-        defer { releaseGate() }
+        // Discarded, not drained: a movement queued behind a reset belongs to the profile this
+        // reset is clearing, and re-registering for it would undo the sign-out.
+        defer {
+            deferredMovement.wrappedValue = nil
```
Shahroz16: Could we clear the deferred movement and release the reset gate in the same critical section used to publish deferrals? I reproduced a movement landing after this clear but before release. Reset finishes with monitoring stopped, but that old-profile movement survives and a later new-user refresh drains it, moving the trigger from longitude 0.1 back to 0.01. Publishing the same movement before the clear is correctly discarded.

Shahroz16 (changes requested):
Thanks, the previous two cases are fixed on this head. One reset handoff window remains: a movement can publish after reset clears the queue but before it releases the gate, then survive into the next user’s refresh. I added the reproduced case inline.

Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator.swift
mrehan27: Valid, fixed in the latest commit.

The reset defer now discards the queue and frees the gate in one `refreshInProgress` section, via `discardDeferredAndReleaseGate`. A movement either lands before the clear and goes with it, or after the release and takes the gate on its own terms.

Same reasoning as the earlier two-step window on this PR, and the same testing approach: the gap is one statement wide, so I assert the contract — the queue is empty and the gate is genuinely free afterwards, not merely flagged — rather than racing tasks into it. `discardDeferredAndReleaseGate_expectTheQueueClearedAndTheGateFree` and `reset_givenAQueuedMovement_expectItDiscardedAndTheGateFree`. Being straight about the limit: those pin the contract, not the atomicity, so they would not fail if the two statements were moved back outside the lock. Say the word if you want a stress test instead.

676/676 across three runs.

Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator.swift
```diff
@@ -289,6 +330,9 @@ final class GeofenceSyncCoordinatorImpl: GeofenceSyncCoordinator, @unchecked Sen
             triggerRadius: effectiveConfig.localRefreshTriggerRadius
         )
         logSyncCompleted(registration, requested: (nearest.count, registerMovementTrigger), startedAt: syncStartedAt)
+        // Same reason as `refresh`: this planted the trigger, so a replay older than it must be
+        // retired rather than allowed to move it back.
+        if registerMovementTrigger { noteMovementApplied(nextMovementSequence()) }
```
Shahroz16: Could we allocate the refresh and cached-restore sequence as part of taking the gate, then carry that sequence through completion? Right now refresh takes the gate before allocating at line 121, and cached restore allocates only after OS registration here. A newer movement can queue in either gap with a lower sequence, then be discarded as overtaken by the older refresh or restore coordinates. I reproduced both paths: the trigger finishes at longitude 0.01 instead of the queued movement at 0.02. The earlier reset and replay-ordering regressions now pass; this sequence inversion is the remaining blocker I found.

mrehan27: Valid, fixed in f89e41ab. You were right and my comment on the previous commit argued the opposite — I had written that taking the sequence after the gate let a later arrival outrank the holder. It does the reverse.

`acquireGateWithSequence` now takes the gate and stamps the pass in one critical section, and the cached restore carries the sequence from entry instead of minting a fresh one after the OS registration.

On testing, being straight about what is and is not covered, because my first attempt was vacuous and I only caught it by mutation:

- The restore path IS covered. Its window spans the OS registration, so a movement can be placed exactly inside it via a new `onStartMonitoring` hook on the monitor mock. `applyCachedRegistration_givenAMovementArrivedMidRegistration_expectItOutranksTheRestore` fails when the restore allocates at the end, and nothing else fails.
- The refresh path is NOT covered, and I do not think it can be. Its acquire and allocate are adjacent synchronous statements with no suspension between them, so a test cannot land inside the window. My first test there passed with and without the fix; I kept it, relabelled as pinning the ordering rather than catching the regression, and said so in the commit. If your reproduction gets inside that window I would like to know how — it would be a better test than the one I have.

682/682, lint 0. Tip 13f034fb22582ae03043da096420d3d4ef9a0027.

Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator.swift
```diff
@@ -112,64 +142,59 @@ final class GeofenceSyncCoordinatorImpl: GeofenceSyncCoordinator, @unchecked Sen
...
-            logger.geofenceSyncSkipped(reason: .refreshInProgress)
-            return .failure(.alreadyInProgress)
-        }
-        let expectedUserId = identifiedUserId
-        let result = await performMovement(
-            expectedUserId: expectedUserId, latitude: latitude, longitude: longitude,
-            anchorIsLiveFix: anchorIsLiveFix
+        await handleMovement(
+            latitude: latitude, longitude: longitude, anchorIsLiveFix: anchorIsLiveFix,
+            sequence: nextMovementSequence()
```
Shahroz16: Could we issue the sequence for a fresh movement as part of acquireGateOrDefer too, while keeping the original sequence for replays? Fresh movements still call nextMovementSequence() before reaching the gate. I reproduced one being paused there, a refresh taking the free gate with the next sequence and re-centering, then the movement reaching a free gate but being dropped as overtaken. The base implementation would run that live movement. The new atomic allocation fixes the refresh and cached-restore inversions; this fresh-movement side of the same ordering is the remaining blocker I found.

mrehan27: Valid, fixed in fca69cec.

`acquireGateOrDefer` now mints the sequence for a fresh arrival inside the critical section and accepts a replay`s original, so "took the gate later" and "holds the later sequence" are the same statement. Only a replay can be overtaken — a fresh arrival is by definition the newest thing to reach the gate, so the comparison no longer applies to it at all.

On coverage, and I would rather state this than let "fixed" imply more than it does: **my tests do not catch a regression of this.** I mutated the fix away — putting the mint back before the gate in the three-argument `handleMovement` — and the suite stayed green. The tests I added pin the contract (a fresh arrival is never overtaken; a replay older than the applied sequence still is), and they would catch the overtaken rule being deleted, but they call the gate directly so they cannot see where the sequence was minted.

Same reason as the refresh half: the window is a suspension point inside `handleMovement` with no seam to land in. The cached-restore inversion was testable only because its window spans the OS registration, which the monitor mock can hook.

If your reproduction gets inside either window I would take that test over mine — you have now found two orderings in this area that I could not reach from a unit test, and that is the part of this PR I trust least.

684/684 across three runs, lint 0, tip fca69cec50d069628d0dc828011ecbeff53f9932.

Tests/LocationGeofence/Geofence/Coordinator/GeofenceSyncCoordinatorTests.swift
```diff
@@ -1858,6 +1858,538 @@ struct GeofenceSyncCoordinatorTests {
...
+        setup.coordinator.deferredMovement.wrappedValue = GeofenceSyncCoordinatorImpl.DeferredMovement(
+            latitude: 9, longitude: 9, anchorIsLiveFix: true, sequence: 1
+        )
+
+        let taken = setup.coordinator.acquireGateOrDefer(
+            latitude: 1, longitude: 2, anchorIsLiveFix: true,
+            replaySequence: 2
+        )
+
+        #expect(taken.isTaken)
```
Shahroz16: Could we keep the carried replay sequence in this assertion, for example `.taken(sequence: 2)`? `handleMovement` later publishes the returned sequence as applied, so minting a new one for a replay could let it outrank a genuinely newer queued movement while `isTaken` still passes. This is non-blocking for this head; the implementation preserves the sequence correctly.

Shahroz16 (approved):
Fresh movements are now stamped inside the gate while replays retain their original sequence. The focused coordinator suite passes 115/115 on the iPhone 17 simulator, and CI is green. This addresses my remaining ordering concern.

Tests/LocationGeofence/Geofence/Coordinator/GeofenceSyncCoordinatorTests.swift
mrehan27: Valid, done in e90bfa9b and e85b1daf.

All five replay call sites now assert `.taken(sequence: n)` instead of `isTaken`, and the helper is gone since nothing uses it.

Worth flagging what the first commit missed. Your test still did not discriminate: `replaySequence: 2` happened to equal the next number the allocator would mint, so minting and carrying produced the same answer. I only saw it by mutation — removing the carry left it green while three of the other four failed. Both affected tests now take their numbers from the allocator.

Mutation-checked after that: with the carry removed, all five fail and nothing else does. 684/684, lint 0. Tip e85b1daf.

Not re-requesting review since you have already approved and called this non-blocking — say the word if you would rather look again.

Shahroz16 (approved):
The latest two commits only strengthen the replay-sequence assertions. I reran the coordinator suite on the iPhone 17 simulator, 115/115 passed, and the adversarial diff review found no blocker. Still looks good.

Shahroz16 (approved):
I rechecked the restacked head and reran the coordinator suite on the iPhone 17 simulator, 120/120 passed. The movement-gate behavior still looks sound, and the current CI is green.

Sources/LocationGeofence/Coordinator/GeofenceSyncCoordinator.swift
```diff
@@ -131,53 +158,47 @@ final class GeofenceSyncCoordinatorImpl: GeofenceSyncCoordinator, @unchecked Sen
...
         switch await refreshAction(location: location, config: effectiveConfig) {
         case .remote:
-            return await performRemoteRefresh(
+            let remote = await performRemoteRefresh(
                 expectedUserId: userId,
                 anchor: location,
                 cachedConfig: cachedConfig,
                 anchorIsLiveFix: anchorIsLiveFix
             )
+            return MovementPassOutcome(result: remote, reCentred: remote.succeeded)
```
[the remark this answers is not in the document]
Shahroz16: Thanks, the latest commit now carries whether the monitor accepted the movement trigger through the refresh, wake, fallback, and restore paths. I reran the coordinator suite at this head, 124/124 passed.

Shahroz16 (approved):
Thanks, the re-centre result now follows whether the monitor accepted the trigger across each registration path, including the failed-fetch fallback. I reran the coordinator suite on the simulator, 124/124 passed, and CI is green. Looks good.

PR #1299, opened 2026-09-18
chore(geofence): reuse the entry's fix for the pass it starts

mrehan27:
The circle-entry re-arm added in #1297 starts a membership pass that can decide nothing.

Stack: #1296 → #1297 → #1298 → **this**

## The problem

`handleMovement` ends in `evaluatePolygonsAfterMovement` → `evaluateAllPolygons(requiresFreshFix: true)` → `resolveFix(requiringFresh: true)`, which only accepts a fix strictly newer than the last one the resolver delivered. After a circle entry that fix is the entry's own, delivered moments earlier.

When CoreLocation answers by echoing it, the request is refused and every polygon in the pass records `no_usable_fix`.

Measured on device, 09-17 capture, 15:10:12.978 → 15:10:13.005:

```
15:10:12.978  polygon.pass.started  why=movement n=8
15:10:12.996  fix.received          lat=31.36869 lon=74.17149
15:10:12.997  movement.fix.requested why=no_cached_fix     <- second pass, 27 ms later
15:10:13.004  fix.received          lat=31.36869 lon=74.17149   <- identical coordinates
15:10:13.005  polygon.undecided     id=21,18,16,27,17,26,25,20 why=no_usable_fix
```

A fix was delivered and the pass still decided nothing — all 8 polygons. #1297 creates this back-to-back pair on every covering-circle entry.

## The change

The entry's fix travels through `handleMovement` to the pass, which judges against it instead of asking again. Unchanged when no fix is supplied, which is every other caller.

`ResolvedFix` carries accuracy and timestamp, not just coordinates: membership judges against both.

A held fix past `movementFixMaxAge` is dropped and a request made instead. A movement deferred at the refresh gate replays with the fix it was recorded with, aged by however long the holder ran, and `PolygonMembershipDecision` refuses anything that old as `fix_too_old` — reusing one there loses the pass by the other route.

`held=` on `polygon.pass.started` records which branch ran (`none` / `reused` / `too_old`).

## Evidence

669 tests / 31 suites pass, plus the rest of the package.

Mutation, both halves:
- Ignore the held fix → `evaluateAllPolygons_givenTheCallersFix_…` fails, undecided and one request made.
- Drop the age guard → `evaluateAllPolygons_givenAHeldFixPastTheAgeCap_…` fails, refused as `fix_too_old`.

A control test asserts the broken behaviour so the fix cannot pass vacuously, and a coarse-fix test confirms a supplied fix is still judged on accuracy rather than waved through.

Simulator drive, iPhone 17 Pro / iOS 26, real backend, 12 fences:

```
os.callback.received  id=21 t=enter
contradiction.evaluated win=false            <- genuine crossing, not a launch replay
polygon.pass.started  why=os_transition n=1 pass=1 held=none
movement.radius.chosen rad=122 from=fix      <- the #1297 re-arm
polygon.pass.started  why=movement n=8 pass=2 held=reused
                                             <- 8 verdicts, no request
```

The matched control on 274e56c7 (same rig, reverse leg, no teleport) does **not** reproduce the failure: the second forced request returns a new fix and the pass decides. A simulator driving at 20 m/s never echoes, so the sim shows only the saved request. The lost-pass case is the device capture above.


Shahroz16 (approved):
The replay now carries the resolver’s already-acquired fix, so the deferred pass does not regress to stale storage. The simulator suite passes. This increment looks good; the gate-publication race in the parent PR still needs to be fixed there.

Shahroz16 (changes requested):
Could we avoid reusing a held fix right up to the full 30-second decision limit? `heldFixUse` checks its age once, but `resolvedOutcome` checks it again later for each polygon. A fix accepted at 29–30 seconds can cross the limit during the pass and be rejected as `fix_too_old` for every polygon, while the `.tooOld` branch would have requested a usable replacement. Leaving enough headroom for the pass, or falling back to a request when the fix is no longer usable before classification, would preserve the repair this PR is making. A clock-driven regression case with the held fix just below the cap would cover it.

The focused coordinator suite passes all 97 tests at this head. The branch also currently conflicts with its stack base, so I will do the final head check again after the restack.

mrehan27:
Valid, fixed in `e8e08c3b`.

You were right that the two checks disagree. `heldFixUse` accepts a fix inside `movementFixMaxAge`, and then `classifyMembership` re-read the clock for every polygon, so a fix taken just inside the cap crossed it mid-pass and every polygon from that point recorded `fix_too_old`.

I went with settling the age rather than adding headroom. The pass runs on one fix, so whether that fix is young enough to judge with is one question with one answer: `runPass` computes it once and threads it to the classification. Headroom would have needed a constant and still left the same drift, just narrower.

Not changed: the logged age on each verdict is still read at record time. That one is deliberately the live value and its doc already says so, since it exists to show how long a corroboration request took.

Test: `evaluate_givenTheFixAgedPastTheCapAfterThePassSettledIt_expectItStillDecides`. It drives `evaluate` with a fix already past the cap by the clock and a settled age of zero, so it only passes if the settled value is what gates. Driven that way on purpose — the drift itself is a race, and this pins the property without one. Mutation-checked: putting the recompute back fails that test and nothing else.

671/671, lint 0.

One thing to flag: this branch still needs restacking onto the merged feature tip, which I cannot push from my session. So the diff here is against the old base for now.

Shahroz16 (changes requested):
Thanks, settling the age for the whole pass fixes the per-polygon drift. There is still a smaller gap between heldFixUse and the second age read in runPass. I reproduced it with a 29.95-second held fix and a 100ms delay in the pass-start logger: the pass made no decision and requested no replacement. Could we carry the age accepted when selecting the held fix into the pass, or request another fix if it has expired before the pass starts?

mrehan27:
Restacked onto the merged stack since your last review — head is now `7b555d3c`, one commit. CI green (20 pass), no unresolved threads.

Two things in there beyond the fix you reviewed: `handleMovement` is re-implemented against the current gate rather than replayed, because the original predates the sequence/deferral rework and no longer applied; and `applyCachedRegistration` moves to its own file to keep the coordinator under the line cap, matching the existing +Gate / +Refresh / +OsRegistration split.

Sources/LocationGeofence/Polygon/PolygonMembershipResolver+Fix.swift
```diff
@@ -6,6 +6,64 @@ import Foundation
...
+        let age: TimeInterval
+    }
+
+    /// The fix a pass judges against — see `heldFixUse` for when the caller's is taken.
+    func passFix(heldFix: ResolvedFix?, decision: HeldFixDecision, requiringFresh: Bool) async -> PassFix? {
+        guard decision.use == .reused, let heldFix else {
+            return await requestedPassFix(requiringFresh: requiringFresh)
+        }
+        // The age from the decision, not a new reading: that is the whole point of carrying it.
+        return PassFix(location: heldFix.location, age: decision.age)
```
Shahroz16: Could we avoid reusing the held fix after corroboration has already produced newer contradictory evidence? I reproduced a circle entry where fix A is marginally inside and newer fix B is outside, so the entry pass correctly rejects the enter. The follow-up reuses A, rejects an echo of B against the resolver's updated freshness baseline, then commits an unconfirmed enter from A. Preserving the newer evidence or the earlier contradiction when choosing the held fix would keep the follow-up from reversing that decision.

Shahroz16 (changes requested):
Thanks, the earlier age-gap issue is fixed and the focused suite passes. One follow-up reuse path still reverses newer contradictory evidence, so I left that case inline.

Sources/LocationGeofence/Polygon/PolygonMembershipResolver+Fix.swift
mrehan27: Fixed in `107c72cf`. Reuse now requires that nothing newer has been delivered: if `fixResolver.latestFix` postdates the held fix, the pass requests instead of reusing, recorded as `held=superseded` so a capture can tell it from a reuse. Falling through keeps your fix B in play, and a pass that decides nothing is recovered by the next wake where a wrong enter is not. Mutation-checked — removing the check fails the new test.

Shahroz16 (approved):
The follow-up now uses the newer corroborating fix instead of reusing older evidence, while preserving valid held-fix reuse. I reran the focused suites and both the contradiction and age-bound regressions pass. Looks good.

Sources/LocationGeofence/Polygon/PolygonMembershipResolver+Fix.swift
Shahroz16: Thanks, confirmed the follow-up now selects the newer fix and the contradiction regression passes.

Sources/LocationGeofence/Polygon/PolygonMembershipResolver+Fix.swift
```diff
@@ -6,6 +6,80 @@ import Foundation
...
+    /// the newer fix in play, and deciding nothing is recoverable where deciding wrongly is not.
+    func heldFixUse(_ heldFix: ResolvedFix?) -> HeldFixDecision {
+        guard let heldFix else { return HeldFixDecision(use: .none, age: 0) }
+        let age = -heldFix.timestamp.timeIntervalSinceNow
+        guard age <= GeofenceConstants.movementFixMaxAge else {
+            return HeldFixDecision(use: .tooOld, age: age)
+        }
+        if let latest = fixResolver.latestFix?.timestamp, latest > heldFix.timestamp {
+            return HeldFixDecision(use: .superseded, age: age)
+        }
```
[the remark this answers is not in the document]
mrehan27: Correct, and replaced in `d5a0731b`. Refusing reuse was the over-correction — the agreed-corroboration branch then hit the echo refusal and lost the whole pass, which is what this path exists to avoid. The pass now judges on the newer fix instead of requesting, so no request is spent on either branch, and `held=newer` records it. Four tests, all four failing without the change.

PR #1300, opened 2026-09-18
chore(sample): measure the smallest region radius iOS promotes

mrehan27:
Sample-app diagnostic that measures the smallest region radius iOS will actually promote.

Independent of the #1296 → #1297 → #1298 → #1299 stack. Sample app only, no SDK change.

## Why

`PolygonWakeRadius.radius` floors the movement trigger at `polygonWakeMinRadius = 100`. Measured across two independent capture channels, that floor binds on roughly half of all sizings (28/60 and 45/87), and the clamped boundaries cluster at 50–75 m — so 50 m would unclamp 39 of 45, while 25 m adds only 4 more.

The floor exists because nobody has measured what iOS actually honours. This probe measures it instead of guessing.

## What it does

Registers concentric regions at 25 / 50 / 75 / 100 m around a persisted anchor and records what the OS reports for each. Off unless enabled, via the same three-state gate as `VisitProbe`: the `CIO_REGION_RADIUS_PROBE` environment variable, the `cio_region_radius_probe` launch argument, or the persisted flag.

The anchor is persisted deliberately. An OS relaunch re-entering `didFinishLaunchingWithOptions` would otherwise re-anchor the rings around wherever the device had reached, and every capture after the first would measure nothing.

The 100 m ring is the control: it is the radius we already ship, so a run where it stays silent is a broken capture, not a negative result.

## First capture failed, and why — new commit

The 09-19 capture produced 16 region callbacks and **no usable crossing**. Two separate artefacts, indistinguishable in the log as it was:

1. **Launch replay.** Four EXITs landed 63 ms after process start, reporting the device 1052 m from the anchor. Nobody newly exits a 25 m ring from a kilometre away — that is iOS handing a relaunched process its stored belief.
2. **Coalesced promotion.** Later, in a process already five hours old, all four rings ENTERed within 17 ms of each other at 22 m from the anchor. Not a replay, but not four measurements either: the OS noticed once, on arrival, and promoted everything at once. The 100 m control fired just as late, so nothing here separates "25 m works" from "the OS only looked once".

The latest commit stamps every crossing with `proc=` (seconds since process start), `dt=` (seconds since the previous probe callback) and `replay=`. Drop `replay=1`, then require `dt` to separate the rings. A real crossing is staggered — 100, then 75, then 50, then 25, seconds apart as the device moves.

Re-requesting review since this changes what the probe emits.

## Status

Armed on a device, all four radii accepted. No usable crossing yet — needs a drive with the new build.


Apps/APN-UIKit/APN UIKit/Diagnostics/RegionRadiusProbe.swift
```diff
@@ -0,0 +1,372 @@
...
+            level: .info
+        )
+    }
+
+    // MARK: - Reporting
+
+    private func report(_ region: CLRegion, transition: String) {
+        guard region.identifier.hasPrefix(Self.identifierPrefix) else { return }
+        let radius = (region as? CLCircularRegion)?.radius ?? -1
+        let fix = manager.location
```
Shahroz16, line 334: Could we resolve a current fix, or at least record and reject the location age, before using off= to judge whether a ring promoted on time? On a region cold wake, manager.location can still be the cached pre-crossing fix, so a late callback can be logged near the anchor and look like a valid small-radius result. Since the probe uses off= to distinguish works from eventually works, that would make the calibration result unreliable.

Shahroz16 (changes requested):
The probe is isolated to the sample app and the control-ring design is useful, but the value used to classify on-time versus late promotion can be stale on the cold-wake path being measured. That can turn a capture into a confident false result, so I would fix that before relying on this calibration.

Apps/APN-UIKit/APN UIKit/Diagnostics/RegionRadiusProbe.swift
mrehan27: Valid, fixed in 71191949.

The probe now records the fix age on both `off=` sites and marks the offset untrustworthy past 30 s, matching the SDK `movementFixMaxAge`. Rows with `offok=0` must be dropped before reading `off=`.

I recorded and rejected rather than resolving a current fix: a location request inside a region callback costs time and power on the exact path being measured, and a rejected sample is enough to keep the calibration honest. Say the word if you would rather it requested one.

Worth flagging since this is live: the build on the test device is updated with this fix, so the weekend capture will carry `age=` and `offok=`.

Shahroz16 (approved):
The probe now records fix age and marks stale offsets untrustworthy, so a cached cold-wake fix can no longer be read as valid calibration evidence. I also built the APN UIKit sample against the current head on the iPhone 17 simulator. This addresses my earlier concern.

Shahroz16 (approved):
The new process and callback timing fields make the replay and coalesced-promotion cases distinguishable, and the sample build is green at this head. One non-blocking follow-up I would add before another capture is a region-budget guard: the SDK can use all 20 Core Location slots, so if `monitoredRegions.count + 4` exceeds the limit, logging and skipping the probe would keep a budget rejection from looking like a radius rejection. The current capture confirmed all four rings were accepted, so this does not invalidate the evidence already collected.

PR #1301, opened 2026-09-19
fix(geofence): clear user state before stopping the OS monitor

mrehan27:
Teardown stopped the OS monitor and then cleared user-scoped state. Those are two separate awaits. A polygon pass resuming in that gap still read a populated `monitoredGeofenceIds`, passed the create guard in `recordPolygonMembership`, and could emit an enter for a fence being torn down.

Swapped both teardown sites — `reset` and `cleanupIfUserChanged` — so the clear runs first. A pass resuming after the clear now fails closed and is suppressed.

Ticket: https://linear.app/customerio/issue/MBL-2480

Tests: three added. Two pin the order at each site; both fail on the old stop-then-clear ordering and pass on the new one (verified by reverting the source change). The third pins what the ordering buys: once state is cleared, `recordPolygonMembership` returns `.suppressedUnmonitored`.

Full `LocationGeofenceTests`: 661 tests, 31 suites, green.

Stack: #1296 → #1297 → #1298 → #1299, #1300. This one is independent of that chain and branches off `feature/geofence-polygons` directly.


Shahroz16 (approved):
The clear-before-stop ordering closes the teardown window cleanly: once user-scoped state is cleared, a late polygon result is suppressed as unmonitored instead of emitting. I also ran the focused coordinator suite on an iPhone 17 simulator, all 96 tests passed.

PR #1302, opened 2026-09-19
chore(geofence): attribute every undecided polygon to its pass

mrehan27:
The `polygon.verdict` line carries `pass`. The `polygon.undecided` line did not.

## Why

Reading the 09-19 field capture, a pass that announced `n=4` logged three verdicts. That reads as a dropped polygon. It was not — the fourth had decided nothing, and its `polygon.undecided` record carried no `pass`, so it could not be tied back. Cost a false alarm and a detour before the numbers reconciled.

Three passes in one day were unreadable this way.

## What

`pass` added to `geofencePolygonUndecided`, with no default so every call site states its answer. The two `handleTransition` sites genuinely run outside any pass and write `pass=none` rather than omitting the key — absence should never have to be guessed at.

Threaded through `classifyMembership` and `corroborate`, which already run inside a pass.

## Tests

`GeofenceLogTailTests` requires the key. Mutation-checked: dropping the tail entry fails `everyRecord_expectMachineKeyAndReplayClassification`.

`LocationGeofenceTests`: 658 tests, 31 suites, green.

Independent of the #1296 → #1297 → #1298 → #1299 stack and of #1300 and #1301; branches off `feature/geofence-polygons` directly.


Shahroz16 (approved):
The pass identifier now stays attached to every undecided polygon result, while work outside a pass remains explicitly unattributed. The focused resolver and log-tail suites pass at this head.

Shahroz16 (approved):
The follow-up is formatting-only, and the focused polygon resolver suite passes all 67 tests at this head. Still looks good.

